Issue #6, opened 2026-04-24
Progress-card: audit of edge cases + races after v0.2.5


mekenthompson:
Audit of the progress-card driver + renderer after v0.2.5 shipped the `trulyDone = stage === 'done' && !hasInFlightSubAgents(state)` guards (PR #4).

**TL;DR:** PR #4's core fix is working as intended. `hasInFlightSubAgents()` correctly covers both foreground and background sub-agents. But the audit surfaced one confirmed race gap, one observability gap, two latent risks, and one UX gap that wasn't in scope of PR #4.

### 1. `flush()` `terminal` calc — multi-sub-agent race window not test-covered
**Location:** `telegram-plugin/progress-card-driver.ts:618-620`

`terminal` is correctly guarded by `!hasInFlightSubAgents(state)`, but the multi-sub-agent case where a parent `tool_result` arrives while a sub-agent is still mid-tool (sub_agent_tool_use fired, sub_agent_tool_result hasn't landed) is not exercised by the existing tests. The tests at `progress-card-driver.test.ts` only cover sequential arrival. A regression test here would lock in the PR #4 invariant for multi-parallel sub-agents.

**Fix direction:** Add a test in `progress-card-driver.test.ts` where parent `tool_result` lands while two sub-agents are mid-tool. Confirm `maybeCompleteDeferredTurn` does not emit `done:true`.

### 2. Silent drop of late sub-agent events
**Location:** `telegram-plugin/progress-card-driver.ts:793`

After `turn_end` + `completeTurnFully`, `currentTurnKey` is set to `null`. Sub-agent events arriving after this point silently `return` with no diagnostic log. A delayed sub-agent event in production vanishes with no trace.

**Fix direction:** Emit a debug-level log line when `k == null && event.kind.startsWith('sub_agent_')`. Low-noise, high-value for incident debugging.

### 3. Gateway + agent restart is ordered but not transactional
**Location:** `src/agents/lifecycle.ts:15-22`

`switchroom agent restart <name>` restarts the `-gateway` systemd unit before the agent unit. If the gateway restart succeeds but the agent restart fails (timeout, bad config, etc.), you get version skew: new gateway code reading events against an old agent code path. No post-restart health check on the gateway.

**Fix direction:** Add a brief `/health` probe to the gateway's local HTTP/IPC surface. Verify it's responsive before the agent unit is marked ready. Document the ordering guarantee (or lack thereof) in `docs/`.

[...]

PR #9, opened 2026-04-24
fix(progress-card): strip tool-name prefix for human-authored labels

mekenthompson:
Fixes **#6 item 5**.

## Summary

PR #4 shipped MCP-prefix stripping in `renderItemCore` — when an `mcp__*` tool has a resolvable label, the raw `mcp__server__action` prefix is dropped in favour of the friendly label. This PR extends the same principle to built-in tools that carry a human-authored `description` field.

**Before:**
```
● Bash Check commit state (00:09)
● Agent: Research steering queue
```

**After:**
```
● Check commit state (00:09)
● Research steering queue
```

## Changes

- **`telegram-plugin/tool-labels.ts`** — new exported helper `isHumanDescription(tool, input)` returns true for `Bash`/`BashOutput`/`Task`/`Agent` when `input.description` is a non-empty string. `WebFetch`/`WebSearch`/`Read`/`Edit`/etc. are untouched (their labels are identifiers, not prose — tool-name prefix still adds context).
- **`telegram-plugin/progress-card.ts`** — `ChecklistItem` and the `currentTool`/`lastCompletedTool` sub-structs of `SubAgentState` gain a `humanAuthored: boolean` field, populated at ingest time. `renderItemCore` gains a `humanAuthored` parameter and treats it the same as `tool.startsWith('mcp__')`.
- All six `renderItemCore` callsites thread the flag through from the stored item.

## Tests

- `tool-labels.test.ts`: new `isHumanDescription` suite (14 cases)
- `progress-card.test.ts`: new `renderItemCore: human-authored label prefix suppression` suite (6 cases: Bash±description, Task, Agent, MCP regression, WebFetch regression). Two existing tests updated for the new Agent/Task rendering.
- **Full suite: 2537 tests pass, 1 skipped. Type check clean.**

## Review focus

- Confirm the list of tools treated as "human-authored" is correct. Currently: `Bash`, `BashOutput`, `Task`, `Agent`. Tempted to add `Skill` if its `skill` param is prose-like — left out for now.
- The `humanAuthored` flag is persisted in state so the renderer doesn't need to re-parse `input`. Small storage overhead (1 bool per item); trade-off is renderer stays purely deterministic from state.

🤖 Generated during switchroom session work.

### DIFF
--- a/telegram-plugin/progress-card.ts
+++ b/telegram-plugin/progress-card.ts
@@ -15,7 +15,7 @@
  */
 
 import type { SessionEvent } from './session-tail.js'
-import { toolLabel } from './tool-labels.js'
+import { toolLabel, isHumanDescription } from './tool-labels.js'
 
 // ─── Types ──────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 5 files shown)


mekenthompson:
Addressed reviewer's must-fix: strengthened Agent-description test assertion. Commit 0471de4.

Issue #7, opened 2026-04-24
Heartbeat-detect + recover stale progress cards (including cross-restart)


mekenthompson:
## Problem

Progress cards can end up in a "stuck working" state — pinned to chat, showing ⚙️ Working… with elapsed time ticking up — even when the underlying work has finished, errored silently, or been orphaned by a gateway/agent restart. There's no reliable detector that says "this card hasn't seen a real event in N seconds, treat it as stale."

Observed today: after bouncing `clerk` to pick up v0.2.5, the pre-restart card went orphaned (the pinned working message from the old session stayed up with no owner). The new gateway has no knowledge of pre-restart cards, so it can neither update nor close them. User is left staring at a card that's never going to finish.

## Current behavior

- `STUCK_THRESHOLD_MS` = 2 min soft signal: card gets a "stuck" warning after 2 min of no session events (implemented in `progress-card.ts`, heartbeat gate at `progress-card-driver.ts:490-492`). This is a *hint*, not a recovery mechanism — the card never closes, just annotates.
- On gateway restart, `writeRestartMarker` → "recovered from unexpected restart" banner tells the user the gateway hit a restart. Does not reconcile previously-pinned cards.
- If the agent process dies without a clean `turn_end` (crash, OOM, systemd kill), the last card is never finalized. No watchdog closes it.
- Heartbeat emit failures (Telegram API 4xx, message deleted by user, message_id gone stale) are logged but don't escalate — card is silently frozen.

## What we want

1. **Heartbeat-driven staleness detection.**
   - Define a concrete staleness signal (e.g., no render emit for >3× heartbeat interval AND no session event for >N min AND final state has not been confirmed). Escalate past the current "stuck warning" — actively close + unpin the card with a terminal status (`⚠️ last status unknown — agent may have died`).
2. **Cross-restart reconciliation.**
   - On gateway startup, query pinned messages in each known chat. For each card that the new gateway does NOT have state for, decide whether to finalize-with-unknown-status or delete. Rationale: a card from before the restart can never be updated again; leaving it pinned is worse than closing it.
   - Persist the last-known card message_id + state hash per chat (SQLite buffer already exists for message history) so the new gateway can recognise its own orphans.
3. **Telegram-API failure escalation.**
   - If >K consecutive heartbeat edits fail with 4xx (message deleted / blocked), finalize the card locally so the driver stops retrying. Log once, don't spam.
4. **Deterministic fallback (switchroom-managed).**
[...]

PR #10, opened 2026-04-24
fix(progress-card): bound retries on Telegram 4xx errors

mekenthompson:
Fixes **#7 item 2**.

## Summary

When a Telegram card edit hits a 4xx (message deleted, bot blocked, message_id stale, bot kicked from chat) the driver used to log the error and retry on the next heartbeat tick forever. Result: log spam every ~5s, wasted cycles, no way to tell the card was gone. This PR introduces a bounded retry + terminal-state model.

## Model

Every edit outcome is classified into one of three kinds:

| Kind | Counts toward threshold | Examples |
|---|---|---|
| `permanent_4xx` | YES | 400 "message to edit not found", 403 "bot blocked", 404 |
| `transient` | NO | 5xx, network errors, rate-limit (retryable) |
| `benign` | NO | "message is not modified" (edit no-op, not a failure) |

After `maxConsecutive4xx` (default **3**) consecutive `permanent_4xx` failures, the card is marked `terminal = true` locally. Flush and heartbeat tick early-return with no edits. A single warn log fires at the transition; no spam.

Resets:
- Successful edit → counter back to 0
- New `turn_start` on same chat → both counter and terminal flag cleared (fresh card, fresh state)

## Changes

- `telegram-plugin/progress-card-driver.ts` (+124) — `ApiFailureKind` / `ApiFailureInfo` types; `maxConsecutive4xx` config option; per-chat `apiFailures` state; `reportApiFailure` / `reportApiSuccess` driver methods; terminal guards in flush + heartbeat
- `telegram-plugin/gateway/gateway.ts` (+43) — classifier function + success/failure wiring in the gateway's emit callback
- `telegram-plugin/server.ts` (+57) — same classifier + wiring for the non-gateway server path
- `telegram-plugin/tests/progress-card-driver.test.ts` (+288, 8 new tests)

## Tests

New `API failure escalation` describe block in `progress-card-driver.test.ts`:

1. 3 consecutive `permanent_4xx` → terminal; 4th flush is a no-op
2. 2 permanent + 1 success → counter resets to 0
3. Transient failures don't advance the counter
4. `benign` ("message is not modified") doesn't count
5. `turn_start` resets terminal state + counter
6. Heartbeat tick skips emit when terminal
7. Warn log fires exactly once on transition
8. `maxConsecutive4xx = 0` disables the mechanism entirely

**Full suite: 2529 passing, 1 skipped.** `tsc --noEmit` clean.

## Review notes

- Classifier is duplicated across `gateway.ts` and `server.ts`. Worker flagged this as intentional (each entry point has slightly different surrounding logic) but worth a review eye — could factor into a shared helper if the duplication feels wrong.
- `server.ts` `classifyProgressCardApiError` is declared `ApiFailureInfo | null` but implementation always returns non-null — stale docstring. Low-priority cleanup.

🤖 Generated during switchroom session work. Companion to #9.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -36,7 +36,11 @@ import { createPinManager } from '../progress-card-pin-manager.js'
 import { createPinWatchdog } from '../progress-card-pin-watchdog.js'
 import { logStreamingEvent } from '../streaming-metrics.js'
 import { type SessionEvent } from '../session-tail.js'
-import { createProgressDriver, type ProgressDriver } from '../progress-card-driver.js'
+import {
+  createProgressDriver,
+  type ApiFailureInfo,
+  type ProgressDriver,
+} from '../progress-card-driver.js'
 import {
   shouldSuppressToolActivity,
 } from '../pty-tail.js'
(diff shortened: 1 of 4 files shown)


mekenthompson:
Addressed reviewer's must-fix: 429 Too Many Requests now short-circuits to `transient` in both classifiers BEFORE the generic 4xx branch. New regression test fires 10 consecutive 429s and verifies the card stays live + counter is still at 0 afterwards. Full suite: 2530 vitest + 110 bun = 2640 passing. Commit `aca730d`.

PR #12, opened 2026-04-24
test(progress-card): lock multi-sub-agent invariant + late-event diag log

### LINKED ISSUE #6: Progress-card: audit of edge cases + races after v0.2.5

mekenthompson:
## Summary

Two follow-ups from the PR #4 audit (issue #6):

- **Regression test**: locks the multi-sub-agent deferred-completion invariant. Two parallel sub-agents in flight when parent `turn_end` arrives — sub-agent A is mid-tool (has `sub_agent_tool_use` but no `tool_result`), sub-agent B has a completed tool cycle. The test asserts that every emit carries `done: false` throughout, and `done: true` appears exactly once as the last emit only after **both** sub-agents fire `sub_agent_turn_end`. Today's code (PR #4) passes; the test guards against future regressions.

- **Diagnostic log for late sub-agent events**: when `currentTurnKey` is null (after `completeTurnFully`) and a `sub_agent_*` event arrives from a stale session-tail, the driver now calls `console.debug` with a structured event (`progress-card.late-sub-agent-event-dropped`) instead of silently returning. Low-noise (debug level, not warn), high-value for incident debugging. A companion test spies on `console.debug` to verify the log fires and no state corruption occurs.

## Test plan

- [ ] `bun run test` — full suite clean (118 vitest files + 110 bun tests)
- [ ] `bun run lint` (`tsc --noEmit`) — type clean
- [ ] New test `deferred completion: multi-sub-agent race — parent turn_end while A mid-tool, B has partial results` passes
- [ ] New test `late sub-agent event after card close: logs debug and returns cleanly` passes

Fixes #6 items 1 and 2.

### DIFF
--- a/telegram-plugin/progress-card-driver.ts
+++ b/telegram-plugin/progress-card-driver.ts
@@ -792,12 +792,9 @@ export function createProgressDriver(config: ProgressDriverConfig): ProgressDriv
       const k = currentTurnKey
       if (k == null) {
         if (event.kind.startsWith('sub_agent_')) {
-          console.debug({
-            event: 'progress-card.late-sub-agent-event-dropped',
-            kind: event.kind,
-            agentId: 'agentId' in event ? (event as { agentId: string }).agentId : null,
-            chatId,
-          }, 'sub_agent event arrived after card close')
+          process.stderr.write(
+            `telegram gateway: progress-card: late-sub-agent-event-dropped kind=${event.kind} agentId=${'agentId' in event ? (event as { agentId: string }).agentId : 'n/a'} chatId=${chatId}\n`,
+          )
         }
         return
       }
--- a/telegram-plugin/progress-card-driver.ts
+++ b/telegram-plugin/progress-card-driver.ts
@@ -790,7 +790,17 @@ export function createProgressDriver(config: ProgressDriverConfig): ProgressDriv
       // that already ended — without this, a stray tool_result after turn_end
       // would resurrect the card. currentTurnKey is cleared on turn_end.
       const k = currentTurnKey
-      if (k == null) return
+      if (k == null) {
+        if (event.kind.startsWith('sub_agent_')) {
+          console.debug({
+            event: 'progress-card.late-sub-agent-event-dropped',
+            kind: event.kind,
+            agentId: 'agentId' in event ? (event as { agentId: string }).agentId : null,
+            chatId,
+          }, 'sub_agent event arrived after card close')
+        }
+        return
+      }
       let chatState = chats.get(k)
       if (chatState == null) return
 
--- a/telegram-plugin/tests/progress-card-driver.test.ts
+++ b/telegram-plugin/tests/progress-card-driver.test.ts
@@ -1817,6 +1817,8 @@ describe('forceCompleteTurn — external completion signal', () => {
     driver.ingest({ kind: 'sub_agent_tool_use', agentId: 'A', toolUseId: 'a-t1', toolName: 'Read' }, 'c')
 
     // Step 5: B fires tool_use + tool_result (B has ≥1 completed tool cycle)
+    // B completing a cycle before parent turn_end ensures the deferred-completion logic sees
+    // heterogeneous sub-agent states (one mid-tool, one with completed cycles).
     driver.ingest({ kind: 'sub_agent_tool_use', agentId: 'B', toolUseId: 'b-t1', toolName: 'Bash' }, 'c')
     driver.ingest({ kind: 'sub_agent_tool_result', agentId: 'B', toolUseId: 'b-t1' }, 'c')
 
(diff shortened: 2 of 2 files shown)


mekenthompson:
Addressed reviewer's fix: switched log from console.debug to process.stderr.write to match the file's existing pattern. Test updated accordingly. Commit 348a2260eff5c1a8e98cd8b5f89ba384b409fe1a.

mekenthompson:
Correction on the commit SHA — the correct commit is fc072966c8a127b677aa3924e822100fc1745095 (fix(progress-card): match file's stderr log pattern for late-event diag). Prior comment had a stale SHA from a different branch.

PR #33, opened 2026-04-24
telegram: admin commands in gateway (phase 1)

mekenthompson:
## What this PR does

Enables a switchroom agent with `admin: true` in its config to handle foreman-style slash commands (`/agents`, `/logs`, `/restart`, `/update`, `/auth`, `/reconcile`, etc.) **directly in its own Telegram gateway**, invisibly — Claude never sees these command messages.

## Background

Before this PR, admin commands were wired into the gateway's Grammy `bot.command()` handlers unconditionally. They executed regardless of which agent was running and there was no per-agent gating. If a non-admin agent had a user type `/agents`, Grammy would route it to the local handler and Claude would never see it — a silent intercept the user didn't ask for.

### 1. Config + systemd plumbing (`src/config/schema.ts`, `src/agents/systemd.ts`, `src/cli/agent.ts`)

Adds `admin: boolean` to `AgentSchema`. When `admin: true`, `generateGatewayUnit()` emits `Environment=SWITCHROOM_AGENT_ADMIN=true` in the systemd unit.

### 2. Shared dispatcher (`telegram-plugin/admin-commands/index.ts`)

A thin, Telegraf-independent module exposing:
- `ADMIN_COMMAND_NAMES` — canonical set of admin commands. Both gateway and any future consumer can import this to stay in sync.
- `parseCommandName(text)` — extracts `/cmd` handling `@botname` suffixes.
- `dispatchAdminCommand(text, adminEnabled)` — pure decision function: returns `{ handled: true }` only when `adminEnabled=true` AND the command is in the set. Used for testing the dispatch table independently of any bot framework.

### 3. Gateway gating (`telegram-plugin/gateway/gateway.ts`)

- Reads `AGENT_ADMIN = process.env.SWITCHROOM_AGENT_ADMIN === 'true'`
- Registers a Grammy **middleware before** all `bot.command()` handlers
- When `AGENT_ADMIN=false`: if an inbound text message is an admin command, the middleware calls `handleInboundCoalesced` (the normal Claude path) and returns **without** calling `next()` — so `bot.command()` never fires and Claude sees the message normally
- When `AGENT_ADMIN=true`: middleware is a no-op → `bot.command()` handlers fire normally → Claude never sees admin text

## Command routing precedence

```
Telegram message arrives
        │
        ▼
Grammy middleware (admin-gate)
        │
        ├─ AGENT_ADMIN=false AND text is admin cmd → handleInbound → Claude
        │
        └─ otherwise: next()
                │
                ▼
        Grammar's own early-returns (pairing, secret-detect, vault intercepts)
                │
                ▼
        bot.command() handlers (start, help, status, agents, logs, restart…)
                │
                ▼  (only if not matched by bot.command)
        bot.on('message:text') → handleInbound → Claude
```

## Invisibility invariant

When `AGENT_ADMIN=true`, admin command text is consumed entirely by the gateway. It is NOT recorded to IPC history and NOT broadcast to any connected bridge. Claude's context window never sees `/agents`, `/restart`, etc.

## Access control

Admin commands already require `isAuthorizedSender(ctx)` in each handler, which checks the gateway's `allowFrom` list. Admin commands therefore inherit the gateway's access control for free — no new auth layer needed.

## Non-regression invariant

`admin: false` (default, absent) → **zero behavior change**. The middleware's `AGENT_ADMIN=false` branch redirects admin command text to Claude, exactly matching the pre-gateway-era behavior where those texts went straight to the LLM.

## Tests

`telegram-plugin/admin-commands/dispatch.test.ts` — 15 tests:
- Known admin command (`/agents`) + `admin=true` → `handled: true`
- Unknown slash (`/nope`) + `admin=true` → `handled: false`
- Non-slash text (`hello`) + `admin=true` → `handled: false`
- All commands + `admin=false` → `handled: false` (belt-and-braces)
- `parseCommandName` edge cases: `@botname` suffix, args, case normalization

**Pre-existing failure**: `telegram-plugin/tests/silent-reply-guard.test.ts` fails with `process.exit unexpectedly called` — confirmed pre-existing on `origin/main`, unrelated to this PR.

## Out of scope (Phase 1)

- **`/create-agent`**: complex multi-turn wizard state machine. Stays foreman/server-only. Sharing it cleanly would require extracting the wizard state store and the multi-message flow — left for Phase 2 or later.
- **Foreman auto-exit when admin gateway is live**: Phase 2.
- **Bootstrap handoff from foreman to admin-gateway**: Phase 3.
- **`server.ts` (legacy monolith)**: untouched. It continues to work as before for non-gateway deployments.

### DIFF
--- a/src/agents/systemd.ts
+++ b/src/agents/systemd.ts
@@ -197,7 +197,7 @@ function hasNodeOnPath(): boolean {
   }
 }
 
-export function generateGatewayUnit(stateDir: string, agentName: string): string {
+export function generateGatewayUnit(stateDir: string, agentName: string, adminEnabled = false): string {
   const pluginDir = resolve(import.meta.dirname, "../../telegram-plugin");
   const gatewayEntry = resolve(pluginDir, "gateway/gateway.ts");
   const logFile = resolve(stateDir, "gateway.log");
@@ -239,7 +239,7 @@ Environment=PATH=${unitPath}
 Environment=SWITCHROOM_CLI_PATH=${switchroomCli}
 Environment=TELEGRAM_STATE_DIR=${stateDir}
 Environment=SWITCHROOM_AGENT_NAME=${agentName}
-
+${adminEnabled ? `Environment=SWITCHROOM_AGENT_ADMIN=true\n` : ''}
 [Install]
 WantedBy=default.target
 `;
@@ -272,7 +272,11 @@ export function installAllUnits(config: SwitchroomConfig): void {
 
     if (useAutoaccept && gwName) {
       const stateDir = resolve(agentDir, "telegram");
-      const gatewayContent = generateGatewayUnit(stateDir, agentName);
+      // Pass admin flag so the gateway unit includes SWITCHROOM_AGENT_ADMIN=true
+      // when the agent is configured with admin:true. The gateway reads this env
+      // var to decide whether to intercept slash commands before forwarding to Claude.
+      const adminEnabled = resolveAgentConfig(config.defaults, config.profiles, agent).admin === true;
+      const gatewayContent = generateGatewayUnit(stateDir, agentName, adminEnabled);
       installUnit(gwName, gatewayContent);
       installedAgents.push(unitName(gwName));
     }
--- a/src/cli/agent.ts
+++ b/src/cli/agent.ts
@@ -402,7 +402,8 @@ export function regenerateSystemdUnitsForAgent(
 
   if (useAutoaccept && gwName) {
     const stateDir = resolve(agentDir, "telegram");
-    const desiredGw = generateGatewayUnit(stateDir, name);
+    const adminEnabled = resolved.admin === true;
+    const desiredGw = generateGatewayUnit(stateDir, name, adminEnabled);
     const gwUnitPath = unitFilePath(gwName);
     const currentGw = existsSync(gwUnitPath) ? readFileSync(gwUnitPath, "utf-8") : "";
     if (currentGw !== desiredGw) {
@@ -763,7 +764,9 @@ export function registerAgentCommand(program: Command): void {
         // create` on its own would leave the new bot silent.
         if (useAutoaccept && gwName) {
           const stateDir = resolve(agentDir, "telegram");
-          const gatewayContent = generateGatewayUnit(stateDir, name);
+          const resolved = resolveAgentConfig(config.defaults, config.profiles, agentConfig);
+          const adminEnabled = resolved.admin === true;
+          const gatewayContent = generateGatewayUnit(stateDir, name, adminEnabled);
           installUnit(gwName, gatewayContent);
         }
 
--- a/src/config/schema.ts
+++ b/src/config/schema.ts
@@ -599,6 +599,16 @@ export const AgentSchema = z.object({
     .boolean()
     .optional()
     .describe("If true, add skipDangerousModePermissionPrompt to settings.json"),
+  admin: z
+    .boolean()
+    .optional()
+    .describe(
+      "If true, the agent's Telegram gateway intercepts admin slash commands " +
+      "(/agents, /logs, /restart, /delete, /update, /auth, /reconcile, etc.) " +
+      "locally before forwarding to Claude. Commands are handled silently — " +
+      "Claude never sees them. Requires the agent to use the switchroom-telegram " +
+      "plugin. When false or absent, all messages pass through to Claude unchanged.",
+    ),
   settings_raw: z
     .record(z.string(), z.unknown())
     .optional()
--- a/telegram-plugin/admin-commands/dispatch.test.ts
+++ b/telegram-plugin/admin-commands/dispatch.test.ts
@@ -1,0 +1,96 @@
+import { describe, it, expect } from 'vitest'
+import { dispatchAdminCommand, parseCommandName, ADMIN_COMMAND_NAMES } from './index.js'
+
+// ─── parseCommandName ────────────────────────────────────────────────────────
+
+describe('parseCommandName', () => {
+  it('extracts a simple command', () => {
+    expect(parseCommandName('/agents')).toBe('agents')
+  })
+
+  it('strips a @botname suffix', () => {
+    expect(parseCommandName('/agents@mybot')).toBe('agents')
+  })
+
+  it('ignores arguments after a space', () => {
+    expect(parseCommandName('/logs 50')).toBe('logs')
+  })
+
+  it('normalises to lowercase', () => {
+    expect(parseCommandName('/Agents')).toBe('agents')
+  })
+
+  it('returns null for non-slash text', () => {
+    expect(parseCommandName('hello')).toBeNull()
+    expect(parseCommandName('')).toBeNull()
+    expect(parseCommandName('agents')).toBeNull()
+  })
+
+  it('returns null for an empty slash', () => {
+    expect(parseCommandName('/ ')).toBe('')
+  })
+})
+
+// ─── ADMIN_COMMAND_NAMES ─────────────────────────────────────────────────────
+
+describe('ADMIN_COMMAND_NAMES', () => {
+  it('contains the core admin commands', () => {
+    const required = ['agents', 'logs', 'restart', 'update', 'auth', 'reconcile']
+    for (const cmd of required) {
+      expect(ADMIN_COMMAND_NAMES.has(cmd)).toBe(true)
+    }
+  })
+
+  it('does not contain create-agent (out of scope for phase 1)', () => {
+    expect(ADMIN_COMMAND_NAMES.has('create-agent')).toBe(false)
+  })
+})
+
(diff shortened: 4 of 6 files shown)


mekenthompson (commented):
## Review: telegram: admin commands in gateway (phase 1)

**Verdict: needs changes** — two issues, one of which could cause a silent regression on a non-admin agent if `pins-status` is ever used.

---

### Critical invariants

**1. Invisibility (admin=true) — PASS**

The Grammy middleware at `gateway.ts:2884` is registered before all `bot.command()` calls. When `AGENT_ADMIN=true`, the middleware calls `next()` unconditionally, so `bot.command()` handlers fire normally and `handleInboundCoalesced` (the IPC path) is never reached for admin commands. Ordering is correct.

**2. Zero regression for non-admin agents — CONDITIONAL PASS (see issue #1)**

When `AGENT_ADMIN=false`, the middleware redirects messages in `ADMIN_COMMAND_NAMES` to `handleInboundCoalesced`. For all commands currently in `ADMIN_COMMAND_NAMES` that have a matching `bot.command()` in gateway.ts, behavior is preserved — the command no longer silently no-ops for unpaired users, it goes to Claude instead. However, `pins-status` is in `ADMIN_COMMAND_NAMES` but has no `bot.command('pins-status', ...)` in gateway.ts (it's only in server.ts). On a non-admin gateway, `/pins-status` will now be redirected to Claude via the middleware — but on the old code it would have simply been unmatched and ignored by Grammy. Net effect: no regression in the old "unmatched" case since Claude still processes it. But the entry is a mismatch that will confuse the next person who reads the set.

**3. Foreman untouched — PASS**

`server.ts` does not import from `admin-commands/index.ts` and is unchanged. No regression risk.

**4. Access control — PASS**

- `AGENT_ADMIN=true` path: `bot.command()` handlers all call `isAuthorizedSender()` (gateway.ts lines 2978, 2994, 3001, 3008, etc.) before doing anything. Unauthorized senders are rejected.
- `AGENT_ADMIN=false` redirect path: goes through `handleInboundCoalesced` → `handleInbound` → `gate(ctx)` (line 1985) which enforces `allowFrom`. Unauthorized senders get `action: 'drop'`.

Both paths are covered. The commit message claim ("inherit allowFrom for free") is accurate.

---

### Issues

**Issue 1 — `pins-status` in `ADMIN_COMMAND_NAMES` has no handler in gateway.ts**

`telegram-plugin/admin-commands/index.ts:63` includes `'pins-status'` in the set. There is no `bot.command('pins-status', ...)` in gateway.ts. The command exists only in server.ts (line 4648).

On a non-admin agent gateway, `/pins-status` will be redirected to Claude instead of being silently dropped. This is a mild behaviour change (was: silently ignored; now: forwarded to Claude). On an admin agent, `/pins-status` would be intercepted by the middleware... wait, no — `AGENT_ADMIN=true` means the middleware is a no-op and `bot.command()` handles it, but there's no `bot.command('pins-status')` in gateway.ts, so Grammy falls through to `bot.on('message:text')` at line 3923 which calls `handleInboundCoalesced`. That means **`/pins-status` on an admin gateway leaks to Claude** even though `AGENT_ADMIN=true` and it's in the "admin" set.

This violates the invisibility invariant for `pins-status` on admin gateways.

**Fix**: either add `bot.command('pins-status', ...)` to gateway.ts, or remove `'pins-status'` from `ADMIN_COMMAND_NAMES` until gateway.ts has an implementation for it.

**Issue 2 — `dispatchAdminCommand` is imported but never called in gateway.ts**

`gateway.ts:155` imports `dispatchAdminCommand` from `../admin-commands/index.js`. The middleware at line 2884 uses `parseCommandName` and `ADMIN_COMMAND_NAMES` directly, bypassing `dispatchAdminCommand` entirely. The function exists only for the test suite.

This is dead code in gateway.ts (the import will trigger a TypeScript/linter "unused import" warning and is confusing — a reader expects the named function to be the gating point but it isn't). Either:
- Use `dispatchAdminCommand(ctx.message.text, AGENT_ADMIN)` in the middleware, or
- Drop it from the gateway import.

---

### Nits

- **Test title mismatch** — `dispatch.test.ts:29`: `it('returns null for an empty slash', ...)` but asserts `toBe('')` not `toBeNull()`. The function returns `''` (empty string) for `'/ '`, not `null`. The description contradicts the assertion.

- **PR description mentions `/delete`** as a handled command, but `/delete` does not exist in `ADMIN_COMMAND_NAMES` or any `bot.command()` registration. PR description is inaccurate.

- **`admin-commands/index.ts` file header** says "Used by both: gateway.ts and server.ts" but server.ts does not import from this module. The header comment is aspirational, not current. Should say "gateway.ts" only until server.ts is wired up.

- **Cosmetic**: when `adminEnabled=false`, `systemd.ts:242` produces two consecutive blank lines before `[Install]`. Not invalid for a systemd unit, just inconsistent with the `adminEnabled=true` case which produces one blank line.

---

### Confidence

- Did not run the tests or build (`bun test`).
- Did not exercise a live gateway process with the env var toggled.
- Did not verify Grammy's middleware ordering via the framework source — relied on documented Grammy behaviour (middleware chain runs in registration order).
- Foreman path reviewed by inspection only; no live foreman test.

mekenthompson (commented):
## Review: telegram: admin commands in gateway (phase 1)

**Verdict: needs changes** — two issues, one of which violates the invisibility invariant.

---

### Critical invariants

**1. Invisibility (admin=true) — FAIL for `pins-status`**

The Grammy middleware at `gateway.ts:2884` is registered before all `bot.command()` calls and is correct in concept. However, `ADMIN_COMMAND_NAMES` includes `'pins-status'` (`admin-commands/index.ts:63`) but there is no `bot.command('pins-status', ...)` in gateway.ts. The command only exists in server.ts (line 4648).

When `AGENT_ADMIN=true`: the middleware is a no-op, so `/pins-status` falls through to Grammy with no matching `bot.command()`, and Grammy routes it to `bot.on('message:text')` (gateway.ts:3923) which calls `handleInboundCoalesced`. **`/pins-status` leaks to Claude on an admin gateway.**

**Fix**: remove `'pins-status'` from `ADMIN_COMMAND_NAMES` until gateway.ts has an implementation, or add the handler to gateway.ts now.

**2. Zero regression for non-admin agents — PASS** (for all commands that do have a bot.command handler; pins-status was silently ignored before so the change to forward-to-Claude on non-admin is not a regression, just surprising)

**3. Foreman untouched — PASS**

server.ts does not import from `admin-commands/index.ts`. No regression risk.

**4. Access control — PASS**

- `AGENT_ADMIN=true` path: all `bot.command()` handlers call `isAuthorizedSender()` before executing.
- `AGENT_ADMIN=false` redirect path: `handleInboundCoalesced` → `handleInbound` → `gate(ctx)` (line 1985) enforces `allowFrom`. Unauthorized senders are dropped before Claude sees anything.

Both paths are protected.

---

### Issues requiring changes

**Issue 1 — `pins-status` in `ADMIN_COMMAND_NAMES` has no gateway handler**

`telegram-plugin/admin-commands/index.ts:63`. Breaks the invisibility invariant on admin gateways. See above.

**Issue 2 — `dispatchAdminCommand` is imported but never called in gateway.ts**

`gateway.ts:155` imports `dispatchAdminCommand` but the middleware at line 2884 uses `parseCommandName` and `ADMIN_COMMAND_NAMES` directly. Dead import — either use the function in the middleware (cleaner) or drop it from the import.

---

### Nits

- `dispatch.test.ts:29`: test description says "returns null for an empty slash" but asserts `toBe('')` not `toBeNull()`.
- PR description lists `/delete` as a handled command — `/delete` does not exist in `ADMIN_COMMAND_NAMES` or any `bot.command()` registration anywhere.
- `admin-commands/index.ts` file header says "Used by both: gateway.ts and server.ts" but server.ts does not import this module.

---

### Confidence

Did not run `bun test` or build. Did not exercise a live gateway process with the env var toggled. Grammy middleware ordering verified by inspection against documented Grammy behaviour (registration order = execution order).

mekenthompson:
Review feedback addressed in 4974a1c:

1. **`/pins-status` leak fixed** — removed from `ADMIN_COMMAND_NAMES`. Since gateway has no `bot.command('pins-status')` handler, keeping it in the allowlist silently leaked the command text to Claude on admin agents. Dead weight removed, invariant restored.
2. **Dead `dispatchAdminCommand` import removed** from gateway.ts. The middleware uses `parseCommandName` + `ADMIN_COMMAND_NAMES` directly; the exported dispatcher is now a test-only helper (and documented as such in the module header). Kept the export because the tests genuinely exercise the dispatch truth table.
3. **Test title corrected** — `'returns null for an empty slash'` → `'returns an empty string for a bare slash'`, matching the `toBe('')` assertion.

Follow-up worth considering (not in this PR): a unit test that asserts every name in `ADMIN_COMMAND_NAMES` has a matching `bot.command(...)` registration in gateway.ts, to prevent this class of bug via static check. Low-cost if done as a grep-based test; deferred for now.

PR #40, opened 2026-04-24
fix(auth): stale-token capture + credentials.json shadowing

mekenthompson:
## Summary

**Live incident:** gymbro's `/auth` flow captured an expired OAuth token (expired 2026-04-20) from a stale `.credentials.json` left over from a prior auth. The flow completed with `completed: true`, wrote the dead token to `.oauth-token`, and the agent started 401-ing against `api.anthropic.com`. Confirmed invalid by hitting the API directly; klanker's working token returned 200 against identical headers.

Two bugs compounded to cause this:

### Bug 1 — stale-token capture (root cause)

`submitAuthCode`'s poll loop called `readTokenFromCredentialsFile(credFileToWatch)` on every tick **without checking whether the file was fresh**. If a `.credentials.json` from a prior login existed in the agent's `.claude/` dir, the very first poll tick succeeded with the stale token — before `claude setup-token` had written anything.

**Fix:** `startAuthSession` now snapshots `fileMtimeMs(join(configDir, ".credentials.json"))` (0 if absent) and saves it as `credentialsMtimeAtStart` in `AuthSessionMeta`. The poll loop gates on `credsMtime > credsMtimeSnapshot` before accepting the file. A pre-existing file at the same mtime is skipped; polling continues until `claude` writes a fresh token (new mtime) or the timeout fires. First-time logins (snapshot == 0) accept any positive mtime — no regression.

### Bug 2 — credentials.json shadowing at runtime

Even with a correct `.oauth-token` written, the running `claude` CLI reads `.credentials.json` first and ignores `CLAUDE_CODE_OAUTH_TOKEN` when that file is present (and expired). Removing the stale file made the env-var path work immediately (confirmed live).

**Fix:** After `writeOAuthToken` succeeds, `submitAuthCode` calls `rmSync(credentialsPath(agentDir), { force: true })`. The new long-lived token in `.oauth-token` becomes the single source of truth. The force/reauth path is unaffected — credentials live in a throwaway temp dir already wiped by `cleanupAuthTempDirs`.

## Changes

- `src/auth/manager.ts`
  - Added `statSync` import
  - Added `fileMtimeMs(filePath)` helper (returns 0 on ENOENT)
  - Extended `AuthSessionMeta` with optional `credentialsMtimeAtStart?: number`
  - `startAuthSession`: snapshot cred file mtime before launching tmux session; save in meta
  - `submitAuthCode`: read snapshot from meta; gate cred-file channel on `credsMtime > snapshot`; unlink `credentialsPath(agentDir)` after `writeOAuthToken` success
- `tests/auth.stale-token-fix.test.ts` — new regression suite (7 tests)

## Test plan

- [x] `tsc --noEmit` — clean
- [x] `vitest run` — 118/119 test files pass; only pre-existing flake `silent-reply-guard.test.ts` fails (known, documented in task brief)
- [x] New tests cover: mtime gate blocks stale file at equal mtime, mtime gate accepts file with newer mtime, legacy meta (no field) doesn't break, no deletion on failure path, creds removed on success path

## Invariants enforced

1. A `.credentials.json` present at session start can never be captured as a fresh token — mtime must be strictly newer.
2. After a successful auth, `.credentials.json` is removed; `.oauth-token` is the sole token source at runtime.

### DIFF
--- a/src/auth/manager.ts
+++ b/src/auth/manager.ts
@@ -8,6 +8,7 @@ import {
   mkdtempSync,
   rmSync,
   chmodSync,
+  statSync,
 } from "node:fs";
 import { join, resolve } from "node:path";
 import type { SwitchroomConfig } from "../config/schema.js";
@@ -90,12 +91,30 @@ interface AuthSessionMeta {
    * nothing and timed out with "no token was found after 20s" even
    * when the exchange had succeeded. To detect silent success we must
    * poll the credentials file directly. See 2026-04-22 bundle-strings
-   * investigation — the setup-token success branch explicitly returns
+   * investigation — the setup-tool success branch explicitly returns
    * `null` for the on-screen view when a token exists.
    *
    * Optional for the same legacy reason as initialCodeChallenge.
    */
   configDir?: string;
+  /**
+   * The mtime (ms since epoch) of `<configDir>/.credentials.json` at the
+   * moment the auth session was started, or 0 if the file didn't exist.
+   *
+   * Used in the poll loop inside `submitAuthCode` to reject pre-existing
+   * stale tokens: we only accept a credentials.json read if its mtime is
+   * strictly greater than this snapshot. That way a leftover credentials.json
+   * from a prior auth (e.g. gymbro's expired 2026-04-20 token) can never be
+   * mistaken for fresh output from the new `claude setup-token` run.
+   *
+   * Edge case: snapshot == 0 (file absent at session start) → any positive
+   * mtime passes, so first-time logins still work correctly.
+   *
+   * Optional for legacy session files written before this field was added;
+   * missing value is treated as 0 (accept any mtime) to avoid breaking
+   * already-in-flight sessions after an upgrade.
+   */
+  credentialsMtimeAtStart?: number;
 }
 
 /**
@@ -261,6 +280,19 @@ export function resolveSlotForAdd(
     return requested;
   }
   return suggestSlotName(agentDir);
+}
+
+/**
+ * Return the mtime (ms since epoch) of a file, or 0 if it doesn't exist
+ * or can't be stat'd. Used to snapshot .credentials.json before an auth
+ * session starts so the poll loop can reject pre-existing stale tokens.
+ */
+function fileMtimeMs(filePath: string): number {
+  try {
+    return statSync(filePath).mtimeMs;
+  } catch {
+    return 0;
+  }
 }
 
 function writeAuthSessionMeta(agentDir: string, meta: AuthSessionMeta): void {
@@ -560,6 +592,14 @@ export function startAuthSession(
     configDir = claudeDir(agentDir);
   }
 
+  // Snapshot the mtime of the credentials file BEFORE we launch claude
+  // setup-token. The poll loop in submitAuthCode uses this to reject any
+  // token read from a .credentials.json that hasn't changed since session
+  // start — i.e. a stale file from a prior auth. For the force path the
+  // configDir is a fresh temp dir, so the file never existed (mtime == 0)
+  // and the snapshot is effectively a no-op guard.
+  const credentialsMtimeAtStart = fileMtimeMs(join(configDir, ".credentials.json"));
+
   // For a forced reauth, cleanup MUST still run even if `claude setup-token`
   // crashes, is killed, or the tmux session is torn down — otherwise OAuth
   // artifacts linger. A bash EXIT trap gives us that guarantee. We route
@@ -606,6 +646,7 @@ export function startAuthSession(
     slot: slotArg,
     initialCodeChallenge,
     configDir,
+    credentialsMtimeAtStart,
   });
 
   return {
@@ -700,13 +741,24 @@ export function submitAuthCode(
     ? join(meta.configDir, ".credentials.json")
     : credentialsPath(agentDir);
 
+  // Stale-token guard (Fix 1, 2026-04-25 gymbro incident):
+  // We only accept a credentials.json read if the file was written AFTER
+  // the auth session started. snapshot == 0 means the file didn't exist at
+  // session start, so any positive mtime passes (first-time login case).
+  // Legacy session meta without this field → treat as 0 (accept any mtime)
+  // to avoid breaking already-in-flight sessions after an upgrade.
+  const credsMtimeSnapshot = meta?.credentialsMtimeAtStart ?? 0;
+
   // Two-channel success detection:
   //   1. <configDir>/.credentials.json written by claude CLI itself.
   //      This is the PRIMARY channel as of claude CLI 2.1+ because the
   //      token is no longer printed to stdout on setup-token success.
+  //      GATED: only accepted if the file's mtime is strictly newer than
+  //      the snapshot taken at session start (stale-token fix).
   //   2. Log file scan for a raw `sk-ant-oat...` string. Covers older
   //      claude CLI versions that DID print it, and any future code
-  //      path that logs the token (e.g. debug mode).
+  //      path that logs the token (e.g. debug mode). Not mtime-gated
+  //      because the log file is created fresh at session start.
   //
   // Both channels produce the same token string on success; we return
   // whichever wins the race. Polling both each tick means a silent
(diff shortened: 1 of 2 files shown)


mekenthompson (commented):
## Review: fix(auth) — stale-token capture + credentials.json shadowing

Reviewed branch `auth-stale-token-fix` (single commit 291b64e). Only `src/auth/manager.ts` and `tests/auth.stale-token-fix.test.ts` are in the commit — the large `git diff main..` noise is the branch base being slightly behind main, not scope creep. Clean.

---

### Correctness — hardass trace

**Does the mtime gate actually close the bug?**

Yes. The root cause was the poll loop reading `.credentials.json` on the very first tick and returning whatever was already there. With the snapshot, the loop skips any read whose `mtimeMs <= snapshot`. The stale file from 2026-04-20 will have an mtime well before session start, so `credsMtime > credsMtimeSnapshot` is false every tick until `claude setup-token` overwrites the file. Bug closed.

**Filesystem clock granularity — can a legit write land at mtime == snapshot?**

This machine is ext4; Linux `statSync().mtimeMs` returns sub-millisecond precision (verified: successive writes produce distinct mtimes ~1ms apart). The risk is real on FAT32/HFS+ (1–2 s granularity) or on network mounts with coarse `timeo`. On any filesystem where the real write happens within the same 1-second bucket as session start, the new file would be skipped and polling would run out the full 20 s timeout before returning `completed: false`.

**Mitigation that exists:** the log-file channel is NOT mtime-gated. If `claude setup-token` writes the token to stdout/log within the 20 s window, the second arm catches it regardless of mtime. So on coarse-mtime filesystems the gate degrades to: "no credentials-file channel, but log-file channel still works." That's acceptable for the production case (ext4, local SSD), but worth a code comment noting the degradation path.

**Should the poll loop also compare token text on mtime-equal reads?**

Marginally useful. It would catch the (hypothetical) case where claude CLI rewrites the file atomically to the same inode with identical mtime but a new token. In practice this cannot happen: any write to `.credentials.json` that changes the byte content will advance mtime on any POSIX FS. Not worth doing.

**Race: can the unlink (Fix 2) delete a file claude is mid-write to?**

No race. The unlink fires only after `writeOAuthToken` succeeds, which only fires after the poll loop exits with a non-null token. The poll loop only exits via the credentials-file channel when it has **successfully read and parsed** a valid token from the file — meaning the write is complete. By the time `rmSync(credentialsPath(agentDir))` runs, claude has already exited `setup-token` (the token is written at the end of that flow). Clean.

**Snapshot surviving a `claude setup-token` internal restart (2026-04-22 bug):**

If claude restarts internally and re-creates `.credentials.json` with a new timestamp, that new mtime will be > snapshot. This works correctly and the new file is accepted. If claude restarts AND somehow reuses the same mtime (no realistic mechanism for this), the credentials channel is blocked but the log-file channel still catches the token. No edge case here.

**First-time auth (no pre-existing credentials.json), snapshot == 0:**

`fileMtimeMs` catches `ENOENT` and returns 0. The poll loop checks `credsMtime > 0`, which is true for any file that exists at all (mtimeMs is always a positive epoch value). First-time logins work cleanly. Off-by-one: none.

**Force/reauth path:**

`configDir` is a fresh `mkdtempSync` dir. `.credentials.json` doesn't exist there; snapshot == 0. The mtime gate is effectively a no-op (any write passes). After success, `cleanupAuthTempDirs` removes the entire temp dir first, then `rmSync(credentialsPath(agentDir))` targets the agent's *real* `.claude/.credentials.json` (not the temp dir). The comment on line 820 correctly documents this. No accident here.

---

### Test coverage

7 tests across 3 `describe` blocks. Assessment:

**What's well covered:**
- Stale token rejected when `mtimeMs == snapshot` (the exact incident scenario) — ✓
- Legacy session meta (missing field) doesn't throw — ✓
- No deletion on non-success path (Fix 2 baseline) — ✓
- `readTokenFromCredentialsFile` returns FRESH_TOKEN when mtime advances (gate math confirmed) — ✓
- Gate rejects same-mtime case as a pure assertion check — ✓

**The significant gap — Fix 2 has no real behavioral test:**

`tests/auth.stale-token-fix.test.ts` L245–274 ("credentials.json is removed when the log-file channel delivers a fresh token") is not a behavioral test. The test body writes the file, then calls `rmSync` on it directly in the test itself, then asserts it's gone. It does not call `submitAuthCode` at all. This is the test author commenting their way around the limitation ("mocking tmux is not feasible") and then testing nothing. The invariant that `submitAuthCode` actually calls `rmSync(credentialsPath(agentDir))` on success is **untested at the behavioral level**.

The workaround is understood (needs tmux), but the test as written gives false confidence. It should either:
- Be renamed to something like `"rmSync with force:true removes a file (sanity)"` and its scope clearly noted, or
- Be replaced with a test that at least stubs/mocks `tmuxSessionExists` to return true and pre-populates both the session meta and the log file with a fresh token, letting `submitAuthCode` run through its full poll loop (log-file channel) and into the success branch.

**Missing tests worth noting (not blockers individually, but as a set):**
- snapshot == 0 (first-time auth, file absent at session start) — not explicitly exercised via `submitAuthCode`, only covered implicitly by the legacy test.
- Scenario where `kill-session` throws — the `try/catch` is there, but nothing asserts that the unlink still runs. (The `try/catch` makes it safe, but there's no test confirming the unlink is unconditional relative to that throw.)
- mtime-equal edge case tested as a pure JS assertion (`sameTimeMtime > staleMtime === false`), not as input to the actual poll loop condition. Acceptable given tmux constraint, but note it.

---

### Style / minor

- Comment quality is excellent throughout. The `credentialsMtimeAtStart` JSDoc in the interface, the inline `// Fix 1` and `// Fix 2` call-outs, and the references to the incident date give future readers clear context for WHY without needing to read git blame. Above average.
- Minor: the `utimesSync` import in the test file is used but the import is alongside other standard-lib imports — fine.
- No unrelated changes in the commit itself.

---

### Verdict: **NEEDS CHANGES — one item before merge**

**Blocker:**
- `"credentials.json is removed when the log-file channel delivers a fresh token"` (Fix 2 test, L245–274) does not test `submitAuthCode` at all. It calls `rmSync` directly in the test body. Either stub `tmuxSessionExists` to make the full success path reachable, or rename + honestly scope the test so it doesn't misrepresent what's covered. The behavior it claims to test (Fix 2 side-effect on success) has zero unit coverage.

**Non-blockers (nits):**
- Consider a comment on the mtime gate acknowledging the degradation on coarse-mtime filesystems (FAT/HFS+/NFS): "on 1-second-granularity filesystems this channel may be skipped; the log-file channel covers that case."
- The `snapshot == 0` first-time path could have a dedicated test calling `submitAuthCode` with no pre-existing credentials file and a session meta with `credentialsMtimeAtStart: 0` — currently only covered by the legacy test which omits the field rather than setting it to 0.

Fix the blocker and this is ready to merge. The core logic (mtime gate + unlink) is correct, the commit is clean, and the comments are genuinely good.

mekenthompson:
Reviewer's blocker addressed in 70abf17.

The original L245-274 test was rewritten as a tmux-gated end-to-end test that drives `submitAuthCode` through its full success branch via the log-file detection channel and asserts the unlink fires by the function itself, not by the test.

Specifically:
- Pre-populates a stale `.credentials.json` with mtime snapshot (so the credentials channel correctly stays blocked)
- Pre-populates `.setup-token.log` with a fresh token (the log-file channel wins the poll race)
- Spins up a real disposable tmux session with `tmux new-session -d` so `tmuxSessionExists` returns true and `tmux send-keys` has somewhere to deliver the code
- Calls `submitAuthCode` and asserts all three success-branch invariants:
  - **Fix 1:** `.oauth-token` contains the FRESH token (not the stale)
  - **Fix 2:** `.credentials.json` is unlinked
  - Side: `.setup-token.log` is removed (existing behavior)

Gated on `describe.runIf(tmuxAvailable())` mirroring `tests/auth.stale-session.test.ts` so suites without tmux still pass.

The no-tmux baseline test was kept and re-titled as a negative invariant for Fix 2 (unlink must NOT fire on the early-exit path) — that's its actual job.

`vitest run tests/auth.stale-token-fix.test.ts` — 7/7 pass. `tsc --noEmit` clean.

Issue #44, opened 2026-04-24
Secret redaction: OAuth browser code not detected, vault unlock UX is friction-heavy


mekenthompson:
Reported live by Ken during the gymbro `/auth` re-test on 2026-04-25 09:24 AEST after PR #40 landed. The auth flow itself worked — the bugs/UX issues are in the secret-detect path adjacent to it.

## Bug — OAuth browser code wasn't redacted from the chat

When the user pastes a `claude setup-token` browser-OAuth code into the Telegram chat (response to "Paste the browser code here ↓"), the message stays visible in chat. Secret-detect doesn't recognise it.

Why: in `telegram-plugin/secret-detect/patterns.ts`, the anchored provider rules cover prefixed tokens (`sk-ant-`, `ghp_`, `xox[bapr]-`, `AIza…`, etc.). The setup-token OAuth code has no recognisable prefix — it's a high-entropy URL-fragment-style string emitted by Anthropic's auth bridge. The `kv_entropy` fallback only fires inside structured `KEY=value` contexts, which doesn't match a bare paste. So no rule fires → no redaction → no delete → the secret lives in the chat history (and the local SQLite buffer) forever.

This is a user-visible security regression of the entire `/auth` Telegram flow: Phase 4's whole pitch is "auth from your phone safely" and the proof-of-concept paste was readable in chat the whole time.

### Fix sketch

Add an anchored rule that matches the Anthropic browser-OAuth code shape. Format known so far (from gymbro test): two URL-safe base64 segments separated by `#`. Sample shape: `[A-Za-z0-9_-]{20,}#[A-Za-z0-9_-]{20,}`. Validate against more samples before merging — Anthropic could change the shape.

A more durable belt: detect the gateway's own outbound prompt ("Paste the browser code here ↓") and treat the next inbound message as auth-flow-sensitive regardless of pattern match. That gives a context-based guarantee that doesn't depend on us tracking Anthropic's code shape.

## UX — vault unlock flow has too many steps for a non-technical user

Quoting Ken verbatim: "for secret redaction experience, it deletes then prompts user to trigger secrets list, then enter vault password. Bad ux."

Current flow when no passphrase is cached (`gateway.ts` ~L2198-L2218):

1. Detect → defer the secret in `deferredSecrets`.
2. Delete the inbound message.
3. Reply: "⚠️ detected a secret but no vault passphrase is cached — run `/vault list` to unlock, then re-paste."
4. User runs `/vault list` (manual command typing).
[...]

PR #46, opened 2026-04-24
fix(secret-detect): redact Anthropic OAuth browser codes

mekenthompson:
Closes the security gap from #44 phase 1. The `claude setup-token` browser-OAuth code pasted during the Telegram `/auth` flow was not being recognised by secret-detect, so the raw code stayed visible in chat history (and in the local SQLite buffer) until the chat was scrolled past.

## Changes

**Channel A — anchored pattern** (`telegram-plugin/secret-detect/patterns.ts`):
```ts
{ rule_id: 'anthropic_oauth_code', regex: /\b([A-Za-z0-9_-]{20,}#[A-Za-z0-9_-]{20,})\b/g, ... }
```
Listed before `anthropic_api_key` so it wins on ties. The 20-char minimum on each segment reliably excludes ordinary URL fragments.

**Channel B — context rule** (`telegram-plugin/gateway/gateway.ts`):
When the gateway emits the "Paste the browser code here" prompt, it sets `awaitingAuthCodeAt[chat_id] = now`. The next inbound message within 5 min (OAuth code lifetime) is treated as auth-flow-sensitive regardless of pattern match. Belt-and-braces — pattern covers the known shape, context rule covers any future format change Anthropic ships.

If the pattern fires, normal flow stores → deletes → confirms. If it doesn't, the fallback deletes + stages + warns so no raw bytes ever reach \`recordInbound\` / IPC.

The flag is consumed on the next inbound or by TTL, whichever comes first.

## Tests

14 new tests in \`telegram-plugin/tests/secret-detect-oauth-code.test.ts\`. All pass. Full vitest suite + tsc clean.

## Out of scope

Phase 2 of #44 — the seamless one-tap-button vault-unlock UX redesign — is deferred to a separate PR. Spec'd in the issue.

## Live incident

Found during the gymbro \`/auth\` re-test on 2026-04-25 09:24 AEST after PR #40 landed. The auth itself worked; the browser code lingered in the chat the whole time.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -631,6 +631,14 @@ function deferredKey(chat_id: string, message_id: number): string {
   return `${chat_id}:${message_id}`
 }
 
+// Channel B context rule — tracks when the gateway has emitted the
+// "Paste the browser code here" prompt so that the next inbound message
+// in the same chat is treated as auth-flow-sensitive regardless of whether
+// the pattern rule fires (belt-and-braces: pattern covers the known shape,
+// context rule covers future shape changes).
+const awaitingAuthCodeAt = new Map<string, number>()
+const AUTH_CODE_CONTEXT_TTL_MS = 5 * 60_000 // 5 min — OAuth code lifetime
+
 // ─── TTL reaper ───────────────────────────────────────────────────────────
 // Pending state maps above all grow whenever a flow starts and only shrink
 // when the flow completes. Users abandoning a flow (closing Telegram, losing
@@ -650,6 +658,9 @@ const pendingStateReaper = setInterval(() => {
   }
   for (const [k, v] of vaultPassphraseCache) {
     if (now > v.expiresAt) vaultPassphraseCache.delete(k)
+  }
+  for (const [k, v] of awaitingAuthCodeAt) {
+    if (now - v > AUTH_CODE_CONTEXT_TTL_MS) awaitingAuthCodeAt.delete(k)
   }
 }, 60_000)
 pendingStateReaper.unref()
@@ -2146,6 +2157,19 @@ async function handleInbound(
   // — never fall through to recordInbound/broadcast with raw bytes. See
   // gateway-secret-detect.test.ts and secret-detect-fail-closed.test.ts.
   try {
+    // Channel B context rule: if we emitted "Paste the browser code here"
+    // recently in this chat, treat the inbound as auth-flow-sensitive —
+    // high-confidence secret detection regardless of pattern match. This
+    // survives Anthropic changing their token format because it tracks the
+    // gateway's own prompt, not the token shape.
+    const authCodeSentAt = awaitingAuthCodeAt.get(chat_id)
+    const isAuthFlowContext =
+      authCodeSentAt !== undefined && Date.now() - authCodeSentAt < AUTH_CODE_CONTEXT_TTL_MS
+    if (isAuthFlowContext) {
+      awaitingAuthCodeAt.delete(chat_id) // consume: one message per prompt
+      process.stderr.write(`[secret-detect] auth-flow context rule active for chat ${chat_id}\n`)
+    }
+
     const cachedPp = vaultPassphraseCache.get(chat_id)
     const passphrase = cachedPp && cachedPp.expiresAt > Date.now() ? cachedPp.passphrase : null
     if (passphrase) {
@@ -2182,6 +2206,25 @@ async function handleInbound(
             staged_at: Date.now(),
           })
         }
+      } else if (isAuthFlowContext && pipeRes.stored.length === 0) {
+        // Channel B fallback: pattern didn't fire (Anthropic may have changed
+        // the token format) but we know this is an auth code paste because we
+        // prompted for it. Delete + stage + warn so no raw bytes leak.
+        if (msgId != null) {
+          try { await bot.api.deleteMessage(chat_id, msgId) } catch {}
+        }
+        await switchroomReply(
+          ctx,
+          '⚠️ auth-flow secret detected (context rule). the message was deleted for safety. run <code>/vault list</code> to unlock and re-paste if you need to store it.',
+          { html: true },
+        )
+        deferredSecrets.set(deferredKey(chat_id, msgId ?? 0), {
+          chat_id,
+          original_message_id: msgId ?? 0,
+          text: effectiveText,
+          staged_at: Date.now(),
+        })
+        return
       } else if (pipeRes.ambiguous.length > 0) {
         for (const d of pipeRes.ambiguous) {
           secretStaging.set({ chat_id, message_id: msgId ?? 0, detection: d, staged_at: Date.now() })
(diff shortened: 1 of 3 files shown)


mekenthompson (commented):
## Review: needs-changes

Reviewed branch `oauth-code-redact` against `main`. Two blockers, one medium issue, several minor items.

---

### BLOCKER 1 — Channel A + B are bypassed in the primary auth flow

Both fixes sit inside the secret-detect block (~L2157 of gateway.ts). The `pendingReauthFlows` intercept runs **before** that block (~L2037). In the normal auth flow:

1. `runSwitchroomAuthCommand` sets `awaitingAuthCodeAt` (Channel B).
2. Every caller immediately sets `pendingReauthFlows`.
3. User pastes the code.
4. `pendingReauthFlows` intercept: `looksLikeAuthCode` matches `[A-Za-z0-9_.#-]{6,500}` → code submitted to CLI → `🔑` reaction → **early return**.
5. Secret-detect block never reached. Message never deleted. `awaitingAuthCodeAt` entry sits until TTL expires and gets reaped, unconsumed.

TTL comparison makes the gap total: `pendingReauthFlows` TTL is 10 min, `AUTH_CODE_CONTEXT_TTL_MS` is 5 min — so the auth-code context window always expires inside the pending-reauth window. Channel B can never fire while `pendingReauthFlows` is active.

**Channel A has the same problem.** `detectSecrets` is called inside the secret-detect block. When `pendingReauthFlows` intercepts first, Channel A never runs.

**Verdict:** the raw OAuth code message is still not deleted in the primary flow. Issue #44 phase 1 is not closed.

**Minimal fix:** add `bot.api.deleteMessage(chat_id, msgId).catch(() => {})` to the `pendingReauthFlows` intercept path (around L2059, just before `setMessageReaction`). Single-line fix at the right layer — the intercept already knows it is handling an auth code.

---

### BLOCKER 2 — Channel A false-positives on real URLs with long path segments

The regex `/\b([A-Za-z0-9_-]{20,}#[A-Za-z0-9_-]{20,})\b/g` fires on any URL whose path segment is ≥ 20 chars followed by a fragment anchor ≥ 20 chars. `\b` only requires a word-boundary before the first char; after a `/`, that boundary exists.

Verified with Node:

```
"see https://docs.com/getting-started-tutorial#installation-and-setup-guide for details"
=> MATCH: getting-started-tutorial#installation-and-setup-guide
```

The test suite only validates short segments (`#installation`, `/foo#bar`). It does not test a URL with a long path + long anchor — a realistic false-positive that a support link or GitHub permalink would trigger daily.

The PR comment claims "The 20-char minimum on each segment reliably excludes ordinary URL fragment anchors (which are always short identifiers like `#section`)" — this claim is incorrect. GitHub headings, npm/cargo readme anchors, and docs sites routinely exceed 20 chars in both segments.

**Fix options (pick one):**
- Anchor to whitespace: `(?:^|\s)([A-Za-z0-9_-]{20,}#[A-Za-z0-9_-]{20,})(?:\s|$)` — bare paste only.
- Negative lookbehind for path separators: `(?<![/.])` before the first segment.
- Add a negative-case test for a URL with a long path+anchor.

---

### MEDIUM — Inaccurate comment in patterns.ts

```ts
// Listed before the generic anthropic_api_key rule so it wins on ties.
```

`anthropic_oauth_code` is at array index 1, **after** `anthropic_api_key` at index 0. The claim is backwards. The two patterns do not overlap so ordering is immaterial, but the comment is misleading — remove or correct the "wins on ties" rationale.

---

### MEDIUM — One-shot consumption is fragile

`awaitingAuthCodeAt.delete(chat_id)` fires on the first inbound message regardless of content. A stray `/queue` command, typo, or steering prefix before the paste consumes the flag and the actual code paste then falls through to Channel A only — which (per Blocker 1) is itself bypassed in the primary flow.

---

### MINOR — `awaitingAuthCodeAt.set` is inside the silenced `try/catch`

```ts
try {
  await switchroomReply(ctx, '📋 Paste the browser code here ↓', { ... })
  awaitingAuthCodeAt.set(authChatId, Date.now())  // not reached if switchroomReply throws
} catch {
  // ForceReply is UX garnish — don't escalate
}
```

If `switchroomReply` throws, the catch swallows it and Channel B is never armed. Move `.set` before the `await` or outside the try/catch to make arming unconditional when the URL condition is true.

---

### Tests

- 14 tests are present; structural Channel B wiring tests are a sensible approach for an untestable singleton.
- Missing: URL with long path segment + long fragment (the exact false-positive in Blocker 2).
- Missing: end-to-end assertion that `deleteMessage` is called in the `pendingReauthFlows` intercept path. `auth-code-auto-capture.test.ts` doesn't assert deletion. The sequencing bug (Blocker 1) is not caught by any test.

---

### What's good

- `awaitingAuthCodeAt` is reaped by the TTL reaper alongside every other pending-state map — no unbounded growth.
- Fail-closed invariant is intact: the catch block drops the message rather than falling through to `recordInbound`.
- No-passphrase branch correctly gates on `isAuthFlowContext`, deletes, and returns.
- Comment quality on new blocks is solid.
- No unrelated changes in the diff.

---

### Summary

**Verdict: blocked.** Blocker 1 means the primary auth paste still lingers undeleted in Telegram — the stated goal of #44 phase 1 is not met. Blocker 2 means a docs link with a long anchor will trigger a false vault-store. The correct fix for Blocker 1 is one `deleteMessage` call inside the `pendingReauthFlows` intercept, not a new code path downstream of it.

mekenthompson:
All blockers and mediums from this review have been addressed in commit 35248a0.

**Blocker 1** — added `bot.api.deleteMessage(chat_id, msgId).catch(() => {})` inside the `pendingReauthFlows` intercept path, just before the 🔑 reaction, so the raw OAuth code message is deleted in the primary auth flow.

**Blocker 2** — changed the `anthropic_oauth_code` regex from `\b...\b` to `(?:^|\s)...(?=\s|$)` (whitespace-boundary with the `m` flag), eliminating false-positives on URLs with long path + fragment segments (e.g. GitHub headings, npm readme anchors). Bare paste on its own line or after prose still matches correctly.

**Medium — comment** — fixed the backwards "Listed before the generic `anthropic_api_key` rule so it wins on ties" comment to "anthropic_api_key precedes; the patterns don't overlap so ordering is moot.".

**Medium — flag consumption** — moved `awaitingAuthCodeAt.delete(chat_id)` from the top-of-block early consume (fired on any inbound) to inside the three actual detection branches (Channel B fallback, no-passphrase hasHigh, and passphrase stored path). Stray messages within the 5-min window no longer disarm Channel B for a subsequent valid paste.

**Minor — armer** — moved `awaitingAuthCodeAt.set` before the inner `try { await switchroomReply(...) }` block so arming is unconditional even if the ForceReply send throws.

**Tests** — 10 new structural/unit assertions added to `secret-detect-oauth-code.test.ts`: URL false-positive regression (3 URL variants), deleteMessage sequencing in pendingReauthFlows (2), flag consumption only on actual detection (3), and armer-before-catch (2). TSC and vitest clean (47 pre-existing failures in unrelated ipc/history/bridge tests remain unchanged).

Issue #67, opened 2026-04-25
telegram + scaffold: align menu commands, foreman, CLAUDE.md template to update/restart/version


mekenthompson:
## Summary
Deferred from PR #65 — finish surface alignment so every place a user could discover/invoke commands says the same thing.

PR #65 aligned the CLI, skills, README, and tests. Still stale:

1. **Telegram bot menu commands** — `TELEGRAM_BASE_COMMANDS` and `TELEGRAM_SWITCHROOM_COMMANDS` arrays in the telegram-plugin still expose old verbs (or miss new ones). Bot `/` menu in Telegram should show only `/update`, `/restart`, `/version` for fleet ops, plus the per-agent commands.

2. **Foreman admin commands** — `telegram-plugin/foreman/foreman.ts` and `telegram-plugin/admin-commands/*` register commands the foreman responds to. Align with the three-verb world. Drop any standalone reconcile/rebuild handlers; route deprecated invocations to `update` with a notice.

3. **Default CLAUDE.md template** — switchroom generates each new agent's CLAUDE.md at install/scaffold time. Find the template (likely `src/agents/scaffold.ts` or a `templates/` dir) and replace any old verb references with: "To pick up new code: `switchroom update`. To bounce a wedged agent: `switchroom restart`. To check what's running: `switchroom version`."

## Acceptance criteria

- [ ] Telegram bot menu (visible via `/` in chat) shows only the canonical verbs
- [ ] Foreman responds to `/update`, `/restart`, `/version` with the same semantics as the CLI
- [ ] Old verbs to the foreman print a one-liner "use /update" and proceed
- [ ] New agents created post-merge get a CLAUDE.md that mentions only the three verbs
- [ ] Tests cover the menu array contents + the deprecated-foreman-handler delegation

mekenthompson:
Reopening — PR #68 (merged) covered (1) Telegram menu and (2) foreman wiring. The third surface — **default CLAUDE.md template alignment in `src/agents/scaffold.ts`** — is still pending. The template generated for new agents may still mention old verbs.

mekenthompson:
Verified the default CLAUDE.md template is already clean — `grep -rE 'switchroom (upgrade|rebuild|reconcile)' profiles/` returns zero matches. Templates never mentioned the old verbs in user-visible text. Closing as done.

Surface alignment summary across PR #65 + #68:
- ✓ CLI subcommand registration
- ✓ Telegram bot menu commands
[...]

Issue #66, opened 2026-04-25
switchroom update: bump Claude CLI + rebuild own binary + self-reexec + per-agent SHA tracking


mekenthompson:
## Summary
Deferred from PR #65 — finish the `update` flow so it's actually all-in-one.

Today `switchroom update` pulls source and reconciles agents but:
1. Doesn't update Claude Code CLI (`@anthropic-ai/claude-code` global install)
2. Doesn't rebuild switchroom's own CLI binary, so `switchroom --version` lies after upgrade
3. Doesn't re-exec itself after rebuild — restart-of-agents step runs with old logic
4. Reports the *binary* SHA per agent, not the *running-process* SHA — drift goes unnoticed

### 1. Bump Claude CLI
After `git pull`, before reconcile:
```
bun add -g @anthropic-ai/claude-code@latest
```
Wrap in a try-warn — if it fails (network, permissions), warn loudly and continue. Don't abort `update` over a Claude CLI bump fail.

### 2. Rebuild own CLI binary
After `git pull` (and if any TS/JS source changed), run the project's build (`npm run build` or `bun run build` — check `package.json`). Update the dist binary so `switchroom --version` and `switchroom version` reflect the new SHA.

### 3. Self-reexec after rebuild
The currently-running `switchroom update` process is using the OLD binary's logic. After step 2, `process.execv` (or shell-exec) the new binary with a `--phase=post-build` flag (or pass state via env/temp file) so reconcile + restart steps use the FRESH code.

Pattern:
- Detect: this process is running pre-rebuild
- Persist in-progress state (which agents to restart, install dir, etc.) to a temp file
- Re-exec with the new binary + a flag that tells it "skip pull/rebuild, resume from temp file"
- New binary completes the run

### 4. Per-agent SHA tracking
Today `printHealthSummary` uses the *current* `COMMIT_SHA` for every agent's "on X" field — assumes all agents are on the binary's SHA. Wrong if an agent was started before an `update` cycle and not yet restarted.

Embed the SHA into each agent's systemd unit `Environment=` at `installAllUnits` time. The agent process reads it at startup and writes to a state file (or exposes via IPC). `getAgentStatus` reads it back. Per-process accurate.

If unavailable (older agent that pre-dates this feature), report `?` rather than silently lie.

## Acceptance criteria

- [ ] `switchroom update` bumps Claude CLI when network is available, warns on failure
- [ ] CLI binary is rebuilt after a successful pull
[...]

PR #70, opened 2026-04-25
WIP feat(telegram): subagent visibility scaffold (#64)

### LINKED ISSUE #64: telegram: surface background sub-agent activity in progress card or pinned message
## Problem

When the main agent dispatches a background sub-agent (worker, researcher, etc.), the user gets no visibility into what's happening for the duration of that worker's run. The progress card shows the main turn's tool calls, then goes quiet because the main agent is just waiting on the sub-agent.

This bites in real workflows: dispatched 3 workers today on a refactor, each ran 5–15 minutes, user had to keep asking "what's happening?" because nothing was visible in Telegram. From the user:

> "Ok well we need to add debug info as I don't see status of that worker in telegram, and add if we need extra log wiring to properly resolve"

This is partially adjacent to #32 (which addressed in-turn tool labeling) but distinct: that issue dealt with the main turn's tool calls; this issue is about **background sub-agent activity that runs across multiple of the parent's turn boundaries.**

## What the user can see today
- Status emoji ladder on their inbound message (👀 → 🤔 → 🔥 → 👍)
- Progress card with the main agent's tool calls during the current turn
- The reaction goes back to 👍 as soon as the main agent's stream_reply done=true fires — even if a background worker is still running
- Nothing surfaces from the worker's own tool calls, file changes, errors, completion

## What we have that we could wire up

Sub-agent transcripts live at:
\`\`\`
/home/kenthompson/.switchroom/agents/<agent>/.claude/projects/<...>/subagents/agent-<id>.jsonl
\`\`\`
These JSONL files are the full Claude Agent SDK transcript for each background sub-agent, growing in real time. Each line is an event (user/assistant message, tool use, tool result, etc.).

The agent process knows when it dispatches a background sub-agent (Agent tool with run_in_background=true returns the agentId).

## Desired behavior

When a background sub-agent is dispatched, surface its activity in Telegram. Options (not mutually exclusive):

1. **Pinned "workers" status message** maintained while any sub-agent is running. Each line: \`<description> · <state> · <last activity Xs ago> · <tool count>\`. Updates throttled to every few seconds. Removed when no workers are active.

2. **Inline status updates** at major events: dispatched, halfway through (heartbeat ping at e.g. 60s), stuck (no activity 90s+), completed (with summary preview). Sent as separate Telegram messages.

3. **On-demand query**: \`/workers\` returns current background sub-agent state. (Lower priority — covers the gap but requires user action.)

My pick: combination of (1) and (2). Pinned card for live state, inline messages at start/stuck/completion for push-notification-worthy events.

## Implementation sketch

The agent process can:
- Maintain a registry of dispatched background sub-agents (their agentId, description, output JSONL path, dispatched-at timestamp)
- Spawn a tail-loop on each JSONL that summarizes events into structured progress (e.g., latest tool name, total tool count, last activity time)
- Update the pinned worker card via the Telegram MCP when state changes (with throttling — same edit-frequency rules as the progress card today)
- On JSONL completion (final message line + status), edit the card to show completion + send a follow-up notification message

Alternative: this could live in the gateway/plugin layer rather than in the agent itself. The agent emits a \`subagent_dispatched\` marker; the gateway tails the JSONL. Probably cleaner — keeps the agent's job simple and reuses existing tail/edit infrastructure. Match the same pattern used today for reading agent transcripts.

## Acceptance criteria

- [ ] When the main agent dispatches a background sub-agent, the user gets an inline notification: \"Worker dispatched: <description>\"
- [ ] A pinned (or persistently updated) message shows live worker state with tool counts and last-activity time, refreshed every few seconds
- [ ] When a worker stalls (no JSONL activity for >60s), an inline notification fires: \"Worker idle: <description>\"
- [ ] When a worker completes, an inline notification fires with the summary preview (or completion status if no summary)
- [ ] The pinned card removes itself when no workers are active
- [ ] Edits respect the same rate-limiting as today's progress card

## Out of scope

- Replaying every tool call from the worker into Telegram (would be too noisy; just summary stats)
- Live transcript export
- Real-time interactive control of the worker (cancel/steer/redirect from Telegram) — that's a separate feature

mekenthompson:
## Status: DRAFT — needs test refactor before merge

Closes #64 once tests are fixed.

## Summary

Lands what an autonomous worker built before its turn budget ran out. Includes the watcher module + 10 unit tests + gateway integration. Worker was mid-refactor of the test mocking approach when it stopped.

## What's here
- \`telegram-plugin/subagent-watcher.ts\` (515 lines): registry of dispatched sub-agents, JSONL tail, pinned card maintenance, dispatch/stall/completion notifications
- \`telegram-plugin/gateway/gateway.ts\`: hooks the watcher into gateway startup when \`SUBAGENT_OWNER_CHAT_ID\` is set
- \`telegram-plugin/tests/subagent-watcher.test.ts\`: 10 unit tests covering registry transitions, dispatch/stall/completion events, pinned card lifecycle

## Known issue (blocks merge)

All 10 tests fail with:
\`\`\`
TypeError: Cannot spy on export "existsSync". Module namespace is not configurable in ESM.
\`\`\`

The watcher imports node:fs functions directly; tests use \`vi.spyOn\` to mock them, which doesn't work on ESM namespace exports.

## Fix-it path

Refactor \`startSubagentWatcher\` to accept an \`fs\` dependency (object with \`existsSync\`, \`readdirSync\`, \`createReadStream\`, \`statSync\`, \`watchFile\`). Tests pass a mock; production code passes \`require('node:fs')\` (or imports as default). Common DI pattern in this codebase already.

Should be a focused 30-60 minute fix.

## Why ship as draft

Other autonomous PRs from this batch already merged the high-value parts (#65, #67, #66). This one needs cleanup but the design + bulk of the code is here — preserving for the next session rather than throwing away.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -159,6 +159,10 @@ import { maskToken } from '../secret-detect/mask.js'
 import { defaultVaultWrite, defaultVaultList } from '../secret-detect/vault-write.js'
 import { detectSecrets } from '../secret-detect/index.js'
 import { ADMIN_COMMAND_NAMES, parseCommandName } from '../admin-commands/index.js'
+import {
+  startSubagentWatcher,
+  type SubagentWatcherHandle,
+} from '../subagent-watcher.js'
 
 // ─── Stderr logging ───────────────────────────────────────────────────────
 installPluginLogger()
@@ -751,6 +755,7 @@ const streamMode = process.env.SWITCHROOM_TG_STREAM_MODE ?? 'checklist'
 const TURN_FLUSH_SAFETY_ENABLED = isTurnFlushSafetyEnabled()
 let progressDriver: ProgressDriver | null = null
 let unpinProgressCardForChat: ((chatId: string, threadId: number | undefined) => void) | null = null
+let subagentWatcher: SubagentWatcherHandle | null = null
 
 // ─── IPC server ───────────────────────────────────────────────────────────
 const SOCKET_PATH = process.env.SWITCHROOM_GATEWAY_SOCKET ?? join(STATE_DIR, 'gateway.sock')
@@ -4456,6 +4461,9 @@ async function shutdown(signal: string): Promise<void> {
   // Clean up all timers and pending state.
   // Snapshot timer handles before clearing so a late-firing timer can't
   // invalidate the iterator by deleting its own entry during cleanup.
+  subagentWatcher?.stop()
+  subagentWatcher = null
+
   for (const iv of [...typingIntervals.values()]) clearInterval(iv)
   typingIntervals.clear()
   for (const t of [...typingRetryTimers.values()]) clearTimeout(t)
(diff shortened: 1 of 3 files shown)


mekenthompson:
Marked back to draft. The fs-DI refactor moved tests from 0/17 passing to 14/17 — the original ESM mocking blocker is resolved. The 3 remaining failures (`updates description from sub_agent_text event`, `counts tools from sub_agent_tool_use events`, `emits completion notification when turn_end arrives`) appear to be subtle test-state issues where the per-test `h.mockFs.readSync` override either isn't being seen by the watcher or the file content isn't being parsed end-to-end. Needs ~30 min of focused debugging; not blocking other work.

Issue #99, opened 2026-04-26
klanker-gateway crash loop — boot-card editMessageText throws unhandled rejection on 'message is not modified'


mekenthompson:
## Symptom

\`switchroom-klanker-gateway.service\` is in a hard restart loop. Observed 44 restarts in ~7 min before I stopped the service; logs show the same crash every cycle.

## Crash signature

Every gateway boot cycle:

\`\`\`
telegram gateway: boot-card: posted msgId=N chatId=8248703757
(~4 seconds later)
telegram gateway: unhandled rejection: GrammyError: Call to 'editMessageText' failed!
  (400: Bad Request: message is not modified: specified new message content
   and reply markup are exactly the same as a current content and reply markup
   of the message)
(~3 seconds later — systemd Restart=always RestartSec=3)
telegram gateway: boot-card: posted msgId=N+2 chatId=8248703757
... loop repeats
\`\`\`

54 such rejections before service was stopped.

## Triggered by #93

The boot-card feature (issue #93, just landed) posts a fresh card on every gateway start, then runs probes that edit the card with results. When all probes settle to the **same content** the card was originally posted with (e.g. all green ticks, no restart reason text), the edit hits Telegram's "message is not modified" 400.

## Why the existing try/catch isn't catching it

\`telegram-plugin/gateway/boot-card.ts:186-201\` has:

\`\`\`ts
async function editCard(): Promise<void> {
  try {
    await bot.editMessageText(chatId, messageId, ..., {...})
  } catch {
    // Edit failures are non-fatal; another edit will follow
  }
}
\`\`\`

This *should* catch the GrammyError. But Node logs it as an unhandled rejection that crashes the process. Possible causes (need someone with bot/Grammy familiarity to confirm):
- The \`bot\` param passed in might be a wrapped/proxied API where the rejection escapes asynchronously (not synchronously from the await)
- Grammy might be re-throwing on an internal queue after the await resolves
- Or there's a separate editMessageText callsite outside boot-card.ts hitting the same endpoint at the same cadence

The repeating pattern (one rejection per gateway boot, ~4s after \`posted msgId=N\`) strongly correlates with the boot-card probe-settle-and-edit path.

## Compound damage

Every restart cycle posts a fresh boot card to Ken's DM (chatId 8248703757). The DM is now polluted with ~50 boot cards, msgIds 2350 through 2356+. The pinned-card sweep on next boot would help if it could survive long enough to do it.

## Recovery this incident

\`\`\`
[...]

PR #100, opened 2026-04-26
fix(telegram-plugin): klanker-gateway crash loop — boot-card editMessageText not_modified (#99)

mekenthompson:
## Summary
- **Layer A** — pre-check in `boot-card.ts`: skip `editMessageText` when content is byte-identical
- **Layer B** — defensive catch via shared `isMessageNotModified` helper in case A misses
- **Layer C** — global `unhandledRejection` guard in `gateway.ts` for benign Grammy 400s
- New `telegram-plugin/gateway/grammy-errors.ts` utility, used by both boot-card and the existing inline check
- New test `boot-card-not-modified.test.ts` covering Layers A/B/D

## Why three layers
The `try/catch` already in `runProbesAndUpdateCard` was not catching the rejection — it escaped past the await asynchronously. Defense in depth: A eliminates the API call, B catches if it sneaks through, C is the process-wide net for any future similar pattern.

## Closes
- #99 — klanker-gateway crash loop on identical boot-card edits (54 rejections in 7 minutes; lawgpt-gateway and clerk-gateway also affected)

## Adjacent cleanup also in this PR
- `progress-card.ts` helpers (`formatDuration`, `escapeHtml`, `truncate`) extracted to shared `card-format.ts` with covering tests
- `gateway.ts` worker-card pin system-message suppression (mirrors progress-card path)
- `subagent-watcher.ts` misc cleanup found during the editor-call trace

## Test plan
- [x] `bun test telegram-plugin/tests/boot-card-not-modified.test.ts` — 10/10 pass
- [x] `bun test telegram-plugin/tests/` — full telegram-plugin suite green (exit 0)
- [ ] After merge: restart klanker-gateway and lawgpt-gateway, tail logs for 60s, confirm zero "message is not modified" lines and no restart loop

## Notes for reviewer
- Layer C's process-handler test was simplified — bun's runner doesn't deliver `unhandledRejection` events synchronously enough for an in-test assertion. The handler logic is verified via `isMessageNotModified` directly. Noted in the test.
- Pre-existing TypeScript errors in `src/agents/systemd-sha.test.ts` (TS2307 on dynamic `?sha-test-*` imports) are unrelated to this PR

### DIFF
--- a/telegram-plugin/card-format.ts
+++ b/telegram-plugin/card-format.ts
@@ -1,0 +1,28 @@
+/**
+ * Shared formatting utilities for Telegram status cards.
+ *
+ * Both the main-agent progress card (progress-card.ts) and the background
+ * worker card (subagent-watcher.ts) import from here so duration strings,
+ * HTML escaping, and truncation are byte-identical across both surfaces.
+ *
+ * Duration format: `<1s` for sub-second, `00:SS` for < 1 minute, `MM:SS`
+ * for >= 1 minute. This is the format used by the main progress card and is
+ * now the canonical format for all status cards.
+ */
+
+export function formatDuration(ms: number): string {
+  if (ms < 1000) return `${ms}ms`
+  const s = Math.floor(ms / 1000)
+  if (s < 60) return `00:${s.toString().padStart(2, '0')}`
+  const m = Math.floor(s / 60)
+  const r = s % 60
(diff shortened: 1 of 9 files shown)


mekenthompson:
Superseded by #102 which already closed #99 with a more comprehensive fix (boot-card dedupe + quota cache + unhandledRejection discriminator). Closing without merging — the same issue is no longer reproducible on current main.

PR #117, opened 2026-04-26
fix(vault): cgroup-based broker ACL + e2e harness + docs

mekenthompson:
## Summary
The vault broker that landed in #113 had two production-blocking bugs the unit tests missed. This PR fixes both, adds a gated systemd e2e harness that actually exercises the production path, and ships user docs.

### Bug 1 — exe-path ACL
The ACL matched `/proc/<pid>/exe` against the cron script path. But the generated cron unit runs `/bin/bash <script>`, so the kernel-set exe is `/bin/bash` — the pattern never matched. **Every cron `vault get` would fail-closed `DENIED` in production.**

Fix: cgroup-based identity. Peercred reads `/proc/<pid>/cgroup` to find the systemd unit name (`switchroom-<agent>-cron-<i>.service`). Cgroup membership is set by systemd as root and unspoofable from userspace.

### Bug 2 — peercred queried itself
Peercred ran `ss -xpn … src <socket>`, which selects the **server-side** row of a unix connection. The `users:()` column on that row is the listening process — the broker itself. Any check based on that PID asks the broker about the broker.

Fix: parse all of `ss -xpn` and walk the inode pair from server-side row → matching client-side row (Local=*) → client PID.

### Why the unit tests didn't catch it
Synthetic ss fixtures matched the cron-path regex; synthetic exe paths bypassed the real `/proc` walk. The new e2e harness spins a real broker, places the cron in a transient systemd unit via `systemd-run --user`, and exercises the full path.

## What's in this PR
- `src/vault/broker/peercred.ts` — cgroup reader + inode-pair `ss` walk
- `src/vault/broker/acl.ts` — switched from exe-path to systemd unit name
- `src/vault/broker/{peercred,acl}.test.ts` — fixtures rewritten to match production output
- `tests/integration/vault-broker-e2e.test.ts` — gated `INTEGRATION=1` harness (allowed key, denied key, broker stopped)
- `CHANGELOG.md`, `README.md` — `[Unreleased]` entry + new "Vault broker (cron secrets)" section

## Test plan
- [x] `bun run lint` clean
- [x] Broker unit tests: 62 passed (4 files)
- [x] `INTEGRATION=1 bun run test:vitest -- tests/integration/vault-broker-e2e` — 4/4 pass against real systemd
- [x] Full vitest suite — only the pre-existing `auth.stale-token-fix.test.ts` tmux-dependent failure (unrelated)
- [x] `bun run build` clean; PII grep at baseline (4 `/home/hindsight/.pg0` mentions, identical to main)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,34 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- **Vault broker ACL was unconditionally denying every cron** in #113. The
+  ACL matched `/proc/<pid>/exe` against the cron script path, but the
+  generated systemd unit invokes `/bin/bash <script>`, so the kernel-set
+  exe is `/bin/bash` and the path pattern never matched in production.
+  Replaced with cgroup-based identity: peercred reads
+  `/proc/<pid>/cgroup` to find the systemd unit name
+  (`switchroom-<agent>-cron-<i>.service`), which systemd writes as root
+  and processes cannot tamper with from userspace. Unit-test fixtures
+  now exercise the full `ss -xpn` inode-pair lookup that production
+  needs to map a connecting client back to its PID.
+- **Peercred `ss` query was returning the broker's own PID.** The
+  `src <socket>` filter selects the server-side row of a unix
+  connection, whose `users:()` column is the listening process. The
+  caller is the *client side*, identifiable by walking the inode pair
+  in the same `ss -xpn` output. Fix lands the two-step lookup.
+
+### Added
+- `tests/integration/vault-broker-e2e.test.ts` — gated systemd e2e
+  harness (set `INTEGRATION=1`). Spawns a real broker, places the cron
+  in a transient `switchroom-<agent>-cron-0.service` via
+  `systemd-run --user`, and proves end-to-end:
+  - allowed-key happy path returns the value through the broker
+  - disallowed-key path is denied with `ACL DENIED`, no value leaks
+  - broker-stopped path fails loud (no silent fallback to interactive
+    passphrase prompt in headless mode)
+
 ## v0.3.0 — 2026-04-25
 
 ### Added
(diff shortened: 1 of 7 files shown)


mekenthompson:
**Review pass — addressed the must-fix and the should-fix items.**

Reviewer flagged a real spoofability hole. Under cgroup v2 user delegation, a same-UID caller can `mkdir` an arbitrary cgroup directory under their `user@<uid>.service` subtree (including paths shaped like `switchroom-<agent>-cron-<i>.service`) and move their PID into it. /proc/<pid>/cgroup then reports the spoofed name. The previous commit's docstring claim that the cgroup is "unspoofable from userspace" was wrong under default delegation.

Fix in latest push: cross-check the cgroup-derived unit name against `systemctl --user show <unit>`. systemd-user reports `LoadState=loaded` only for units it has actually registered; spoofed cgroups have no corresponding unit and fail the check. Accepted ActiveStates: `active` and `activating` (Type=oneshot cron units sit in `activating` while ExecStart runs).

Three new unit tests cover:
- spoofed unit (`systemctl` returns `LoadState=not-found`) → `systemdUnit=null`
- `systemctl` errors (e.g. systemd-user not running) → `systemdUnit=null`
- loaded but `failed`/stale unit → `systemdUnit=null`

Also from the review:
- `SKIP_REASON` ternary in `vault-broker-e2e.test.ts` was missing the "all conditions met" branch (always reported "systemd-run not available"). Fixed.
- `acl.ts`: removed dead-code branches — the unreachable `exe.startsWith(switchroomCli + " ")` (`/proc/<pid>/exe` always resolves to a bare binary path) and the unreachable ternary in the final return.
- `peercred.ts`: `parseSsRows` column comment mislabeled the inode field as "port"; variable names were already correct, comment now matches.

E2e against real systemd-run still passes (4/4). 65 unit tests green. Lint clean. Build clean, PII grep at baseline.

PR #121, opened 2026-04-26
fix(scaffold): pre-approve delete_message and get_recent_messages MCP tools

mekenthompson:
## Summary

Two switchroom-telegram MCP tools were registered in the plugin but missing from the pre-approval allowlist. Adding them eliminates spurious permission prompts during normal operation.

- `delete_message` — registered at [telegram-plugin/server.ts:1347](telegram-plugin/server.ts) for retracting bot messages
- `get_recent_messages` — registered at [telegram-plugin/server.ts:1377](telegram-plugin/server.ts) for context recovery after a Claude Code restart

Without the allowlist entries, the agent prompts on first use, which defeats `get_recent_messages` specifically — its whole job is to recover silently after a restart.

## Test plan
- [x] `npm run lint` passes
- [ ] Boot a profile that uses `channels.telegram.plugin = "switchroom"` and confirm the agent can call both tools without a permission prompt

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -195,8 +195,10 @@ const SWITCHROOM_TELEGRAM_MCP_TOOLS = [
   "mcp__switchroom-telegram__edit_message",
   "mcp__switchroom-telegram__send_typing",
   "mcp__switchroom-telegram__pin_message",
+  "mcp__switchroom-telegram__delete_message",
   "mcp__switchroom-telegram__forward_message",
   "mcp__switchroom-telegram__download_attachment",
+  "mcp__switchroom-telegram__get_recent_messages",
 ];
 
 /**
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -199,6 +199,7 @@ const SWITCHROOM_TELEGRAM_MCP_TOOLS = [
   "mcp__switchroom-telegram__forward_message",
   "mcp__switchroom-telegram__download_attachment",
   "mcp__switchroom-telegram__get_recent_messages",
+  "mcp__switchroom-telegram__progress_update",
 ];
 
 /**
--- a/tests/scaffold.test.ts
+++ b/tests/scaffold.test.ts
@@ -692,10 +692,30 @@ describe("scaffoldAgent", () => {
     );
 
     expect(settings.permissions.allow).toContain("calendar");
-    expect(settings.permissions.allow).toContain("mcp__switchroom-telegram");
-    expect(settings.permissions.allow).toContain("mcp__switchroom-telegram__reply");
-    expect(settings.permissions.allow).toContain("mcp__switchroom-telegram__react");
-    expect(settings.permissions.allow).toContain("mcp__switchroom-telegram__edit_message");
+
+    // Assert every tool name registered by the plugin is in the allowlist.
+    // Drift between the plugin's tool registry and the scaffolded permission
+    // list silently breaks pre-approval — agents fall back to per-call
+    // permission prompts. This test catches the drift the next time it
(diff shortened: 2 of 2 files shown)


mekenthompson (commented):
This is a correct and necessary fix — both tools are registered in the plugin server and were legitimately missing from the allowlist — but one registered tool is still absent from the list.

**Problems**

- `mcp__switchroom-telegram__progress_update` is registered in the plugin server (line 1265 of `telegram-plugin/server.ts`) but is not in `SWITCHROOM_TELEGRAM_MCP_TOOLS`. The diff adds `delete_message` and `get_recent_messages` but skips `progress_update`. Agents that call `progress_update` mid-turn will still prompt for permission, defeating the purpose of pre-approval. This should be added in the same commit.

**Questions / nice-to-haves**

- The existing test at line 684 of `tests/scaffold.test.ts` asserts only four specific tools (`mcp__switchroom-telegram`, `reply`, `react`, `edit_message`). Neither `delete_message`, `get_recent_messages`, nor `progress_update` are asserted in any test. The test is not wrong, but it could be extended to cover every tool name in `SWITCHROOM_TELEGRAM_MCP_TOOLS` so future omissions are caught automatically rather than noticed in production.
- Tool name spellings: the PR uses `mcp__switchroom-telegram__delete_message` and `mcp__switchroom-telegram__get_recent_messages`. These exactly match the `name:` values registered in the plugin server, so the allowlist key format is correct.
- The wildcard prefix entry `"mcp__switchroom-telegram"` (no `__*` suffix) at line 191 of `scaffold.ts` — is that intentional as a bare prefix match, or should it be `"mcp__switchroom-telegram__*"`? The other wildcard entries (`mcp__hindsight__*`, `mcp__switchroom__*`) use the explicit `__*` suffix. Worth auditing how Claude Code's permission engine interprets a bare server-name string vs. a wildcard.

**Side effects on existing agents**

- `reconcileAgent` and `scaffoldAgent` both derive `desiredAllow` / `permissionAllow` from the same `SWITCHROOM_TELEGRAM_MCP_TOOLS` constant (lines 2050 and 2922 of `scaffold.ts`). Any agent that runs `switchroom reconcile` (or `switchroom agent update`) will automatically pick up the new entries. No manual re-scaffold is needed; reconcile rewrites `settings.json`'s `permissions.allow` unconditionally. Existing agents that have not been reconciled will continue to prompt until the next reconcile cycle.

**Test coverage**

- No new tests were added. That is defensible for a two-line constant change — the mechanism (how the constant propagates into `permissions.allow`) is already exercised by the existing scaffold tests. However, the gap noted above (test does not enumerate the full list) means regressions of this exact type won't be caught automatically.

**Recommendation**

Merge after adding `mcp__switchroom-telegram__progress_update` to `SWITCHROOM_TELEGRAM_MCP_TOOLS` and optionally extending the scaffold test to assert every entry in the list.

---

src/agents/scaffold.ts
```diff
@@ -195,8 +195,10 @@ const SWITCHROOM_TELEGRAM_MCP_TOOLS = [
   "mcp__switchroom-telegram__edit_message",
   "mcp__switchroom-telegram__send_typing",
   "mcp__switchroom-telegram__pin_message",
+  "mcp__switchroom-telegram__delete_message",
   "mcp__switchroom-telegram__forward_message",
   "mcp__switchroom-telegram__download_attachment",
+  "mcp__switchroom-telegram__get_recent_messages",
 ];
```
mekenthompson, line 204: `progress_update` is registered in the plugin server (`telegram-plugin/server.ts:1265`) but is not in this list. It should be added here alongside the two new entries — otherwise mid-turn check-in calls from agents will still trigger a permission prompt.

PR #122, opened 2026-04-27
test(config): exclude .claude/worktrees from vitest discovery

mekenthompson:
## Summary

Sub-agent worktrees under `.claude/worktrees/<slug>/` (created by Claude Code's worktree-isolation feature) contain duplicate test files that vitest was discovering and running against stale code. Locally this added hundreds of phantom failures to `npm test`.

The fix adds `**/.claude/worktrees/**` to vitest's exclude list.

## Test plan
- [x] `npm run lint` passes
- [x] `npm run test:vitest` no longer reports test files inside `.claude/worktrees`
- [ ] CI green

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -24,6 +24,10 @@ export default defineConfig({
     exclude: [
       "**/node_modules/**",
       "**/dist/**",
+      // `.claude/worktrees/<slug>/` are checkout copies created by sub-agent
+      // sessions. Their tests duplicate the canonical ones and run against
+      // stale code — never discover them from the canonical repo.
+      "**/.claude/worktrees/**",
       "**/telegram-plugin/tests/history.test.ts",
       "**/telegram-plugin/tests/ipc-server-client.test.ts",
       "**/telegram-plugin/tests/ipc-server-race.test.ts",


mekenthompson (commented):
The fix is correct and necessary — the glob form, config location, and comment are all solid. A few observations worth considering before merge.

**What's good**
- `**/.claude/worktrees/**` is the right glob: the leading `**/` means it matches whether the directory is at the repo root or nested, and the trailing `/**` catches all descendants.
- `vitest.config.ts` is the only discovery-level config (no `vite.config.*` exists, no eslint config, no `jest.config`). This is exactly the right file.
- `tsconfig.json` `include` is locked to `src/**`, `bin/**`, `scripts/**` — it won't walk `.claude/worktrees` regardless, so no tsconfig change is needed.
- The `bun test` leg of `npm test` is explicit-file-list only, not a glob walk — also unaffected.
- The coverage `include` is scoped to `telegram-plugin/**` — safe.

**Issues / questions**

- **`node_modules` and `dist` are already in `defaultExclude`** — when a user provides an `exclude` array in `vitest.config`, it *replaces* `defaultExclude` entirely (vitest does not merge). The existing config was already compensating for this by re-listing `node_modules` and `dist`. The PR is consistent with that existing approach and adds no new regression, but it reinforces a slightly fragile pattern. Consider spreading `configDefaults.exclude` from `vitest/config` so future additions to vitest's own default list aren't silently dropped:
  ```ts
  import { configDefaults, defineConfig } from "vitest/config";
  // …
  exclude: [
    ...configDefaults.exclude,
    "**/.claude/worktrees/**",
    // bun-only tests:
    "**/telegram-plugin/tests/history.test.ts",
    // …
  ],
  ```
  This is a nice-to-have refactor, not a blocker.

- **`.gitignore` gap** — `.claude/worktrees/**` is not in `.gitignore` (only `.claude/settings.local.json` is). If a developer runs `git add .` while a worktree is present, those files could be accidentally staged. A companion `.gitignore` entry would close this — worth a follow-up issue or a second commit on this branch.

- **No companion exclude in `coverage` section** — the coverage `include` is already scoped tightly to `telegram-plugin/**`, so worktree files will never enter coverage reports. No action needed, but worth noting for reviewers who may wonder.

- **CI check not yet green** (per the PR's own test plan checkbox) — the fix is clearly correct locally, but it would be ideal to confirm CI before merging.

**Recommendation**

Merge as-is for the immediate fix. Open a follow-up issue (or add a second commit) to:
1. Add `.claude/worktrees/` to `.gitignore`.
2. Optionally refactor `exclude` to spread `configDefaults.exclude` instead of manually listing `node_modules` and `dist`.

PR #123, opened 2026-04-27
fix(profiles): use platform separator for path-traversal boundary check

mekenthompson:
## Summary

`getProfilePath` rejects every profile name on Windows because the path-traversal boundary check uses a hardcoded forward slash:

```ts
requested.startsWith(PROFILES_ROOT + "/")
```

On POSIX `path.resolve()` returns slash-separated paths so this happens to work. On Windows it returns backslash-separated paths and the check is always `false` — even for legitimate names like `default` or `coder`. The result is that scaffolding throws `Invalid profile name` for everything when run on Windows. CI (Linux) doesn't catch it.

The fix is to compare against `path.sep`. POSIX behavior is unchanged; Windows now works.

## Test plan
- [x] `npm run lint`
- [x] New `src/agents/profiles.test.ts` covers:
  - real profile resolution
  - config-only profile name fallback (the `coder` case)
  - `../` traversal rejection
  - absolute-path rejection (platform-aware)
  - empty string handling
- [ ] CI green

### DIFF
--- a/src/agents/profiles.test.ts
+++ b/src/agents/profiles.test.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect } from "vitest";
+import { sep as pathSep, resolve } from "node:path";
+import { getProfilePath, listAvailableProfiles } from "./profiles.js";
+
+describe("getProfilePath", () => {
+  it("resolves a real profile that exists on disk", () => {
+    const result = getProfilePath("default");
+    expect(result.endsWith(`${pathSep}default`)).toBe(true);
+  });
+
+  it("falls back to the default profile when the requested name does not exist", () => {
+    const fallback = getProfilePath("not-a-real-profile-name-xyzzy");
+    expect(fallback.endsWith(`${pathSep}default`)).toBe(true);
+  });
+
+  it("falls back to the default profile when the name is a config-only profile (e.g. 'coder')", () => {
+    // The user can declare profiles inline in switchroom.yaml — these have no
+    // filesystem directory, so getProfilePath should fall back, not throw.
+    expect(() => getProfilePath("coder")).not.toThrow();
+  });
+
+  it("rejects a path-traversal attempt with `..`", () => {
+    expect(() => getProfilePath("../etc")).toThrow(/Invalid profile name/);
+  });
+
+  it("rejects a path-traversal attempt with deeper `..`", () => {
+    expect(() => getProfilePath("../../tmp")).toThrow(/Invalid profile name/);
+  });
+
+  it("rejects an absolute path", () => {
+    // resolve() would canonicalize an absolute path, escaping PROFILES_ROOT.
+    const abs = pathSep === "\\" ? "C:\\Windows" : "/etc/passwd";
+    expect(() => getProfilePath(abs)).toThrow(/Invalid profile name/);
+  });
+
+  it("accepts an empty string (resolves to PROFILES_ROOT itself, then falls back)", () => {
+    // resolve(PROFILES_ROOT, "") === PROFILES_ROOT, which is allowed; it
+    // falls through to the default profile because it's not a usable
+    // profile dir on its own.
+    expect(() => getProfilePath("")).not.toThrow();
+  });
+});
+
+describe("listAvailableProfiles", () => {
+  it("includes the bundled 'default' profile", () => {
+    const profiles = listAvailableProfiles();
+    expect(profiles).toContain("default");
+  });
+
+  it("excludes the underscore-prefixed _base directory", () => {
+    const profiles = listAvailableProfiles();
+    expect(profiles).not.toContain("_base");
+    expect(profiles.every((name) => !name.startsWith("_"))).toBe(true);
+  });
+});
--- a/src/agents/profiles.test.ts
+++ b/src/agents/profiles.test.ts
@@ -1,5 +1,8 @@
 import { describe, it, expect } from "vitest";
 import { sep as pathSep, resolve } from "node:path";
+import { mkdtempSync, mkdirSync, symlinkSync, writeFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { getProfilePath, listAvailableProfiles } from "./profiles.js";
 
 describe("getProfilePath", () => {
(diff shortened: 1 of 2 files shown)


mekenthompson (commented):
## Overall take

The fix is **correct and safe on POSIX**. Swapping the hardcoded  for  closes a real Windows bug without weakening anything on Linux/macOS. The logic structure —  first, then  — is the right pattern, and the test file adds meaningful regression coverage.

That said, there are several findings worth addressing before merge, two of which have security weight.

---

## Findings

### SECURITY — Missing trailing-separator guard on PROFILES_ROOT itself (low-severity, currently harmless)

The check is:

```ts
if (requested !== PROFILES_ROOT && !requested.startsWith(PROFILES_ROOT + pathSep)) {
```

The `=== PROFILES_ROOT` arm allows a resolved path that is exactly the root (e.g. `profileName = ""` or `profileName = "."`). The code then falls through to `existsSync(requested) && hasProfileFiles(requested)` which returns false for the root dir (it contains subdirs, not .hbs files directly), so it continues to the fallback. So today this is safe. But the guard is subtly fragile: if PROFILES\_ROOT ever gains a top-level .hbs file, the empty-string input would resolve to a valid profile path pointing at the root itself. The safer canonical form used in `memory-search.ts` and `workspace.ts` elsewhere in this codebase is:

```ts
if (!requested.startsWith(PROFILES_ROOT + pathSep) && requested !== PROFILES_ROOT) { throw ... }
```

…which is semantically identical to what's here, so this is a style consistency nit rather than a live bug.

### SECURITY — Symlinks not resolved before the boundary check

`path.resolve()` does **not** follow symlinks. A profile directory named e.g. `evil` that is itself a symlink pointing to `/etc` will pass the `startsWith(PROFILES_ROOT + pathSep)` check (the symlink lives inside PROFILES\_ROOT) and then `existsSync` + `hasProfileFiles` will operate on the symlink target. If the target directory happens to contain any `.hbs` file, the resolved path returned is the symlink path — which the caller then hands to `renderTemplate` / `readFileSync`. In practice, only admin-created symlinks inside the bundled `profiles/` directory would trigger this; an end-user cannot inject a symlink via `profileName`. Still, the codebase's own `getWorkspaceMemoryFile` (`memory-search.ts` line 261–278) and `startWebServer` (`web/server.ts` line 194–303) both use `realpathSync` before their boundary check for exactly this reason. `getProfilePath` should do the same.

Suggested fix:

```ts
import { resolve, join, sep as pathSep } from "node:path";
import { realpathSync } from "node:fs";

export function getProfilePath(profileName: string): string {
  const requested = resolve(PROFILES_ROOT, profileName);
  if (requested !== PROFILES_ROOT && !requested.startsWith(PROFILES_ROOT + pathSep)) {
    throw new Error(`Invalid profile name: ${profileName}`);
  }
  // Re-check after resolving symlinks so a symlink inside profiles/ can't
  // point outside the root (same pattern as memory-search.ts + web/server.ts).
  let real: string;
  try {
    real = realpathSync(requested);
  } catch {
    real = requested; // ENOENT — let existsSync handle it
  }
  if (real !== PROFILES_ROOT && !real.startsWith(PROFILES_ROOT + pathSep)) {
    throw new Error(`Invalid profile name: ${profileName}`);
  }
  // ... rest unchanged
}
```

### Windows mixed-separator input not tested

On Windows, Node accepts both `/` and `\`. A user passing `"..\\etc"` would have `resolve()` normalise it to a backslash path before the check, so the traversal is still caught. This is fine — `resolve()` normalises before the comparison — but there is no test for a mixed-separator traversal attempt (e.g. `"..\\etc"` on Windows, or `"subfolder/../../etc"`). The existing tests only cover forward-slash `../` forms. Low risk given normalisation, but worth documenting.

### Test gap — symlink traversal vector not covered

There is no test that creates a symlink inside a temp PROFILES\_ROOT pointing outside it and confirms rejection. Given the symlink finding above, this is the most important missing test case.

### Test gap — `getProfilePath` is not fully unit-isolated

The tests call the live `getProfilePath` against the real `profiles/` directory on disk. The `"default"` profile must exist for tests 1, 2, and 4 to pass; the `"coder"` profile must NOT exist as a filesystem directory for test 3 to exercise the fallback path. This makes the suite fragile against changes to the bundled profiles. A temp-dir fixture (or mocking `PROFILES_ROOT`) would make traversal tests environment-independent.

### Other boundary checks in the codebase are consistent — no action needed

- `workspace.ts` `isInsideWorkspace()` (line 51): already uses `path.sep`, correct.
- `memory-search.ts` `getWorkspaceMemoryFile()` (line 274): uses `path.sep` + `realpathSync`, correct and more thorough.
- `web/server.ts` static-file check (line 302): uses `relative()` + `realpathSync`, correct.
- `scaffold.ts` `startsWith(skillsPool)` (lines 1228, 1300, 1333): these are symlink-target comparisons, not security guards, no sep issue.

No other site needs the same fix this PR applies.

---

## Concrete recommendation

1. **Must-fix before merge:** Add `realpathSync` resolution after the lexical check, matching the pattern already established in `memory-search.ts` and `web/server.ts`. The current code leaves a symlink escape path open.
2. **Should-fix:** Add a test that uses a temp dir so traversal tests don't depend on the real `profiles/` layout.
3. **Nice-to-have:** Add a mixed-separator test (`"subfolder/../../etc"`) and a symlink traversal test to the suite.

The core Windows fix (the original motivation for the PR) is correct and unambiguous — this is a real bug fix. The issues above are about hardening the security boundary that the fix touches.

mekenthompson (commented):
## Overall take

The fix is **correct and safe on POSIX**. Swapping the hardcoded `"/"` for `path.sep` closes a real Windows bug without weakening anything on Linux/macOS. The logic structure — `resolve()` first, then `startsWith(root + sep)` — is the right pattern, and the test file adds meaningful regression coverage.

That said, there are several findings worth addressing before merge, two of which have security weight.

---

## Findings

### SECURITY — Symlinks not resolved before the boundary check

`path.resolve()` does **not** follow symlinks. A profile directory named e.g. `evil` that is itself a symlink pointing to `/etc` will pass the `startsWith(PROFILES_ROOT + pathSep)` check (the symlink lives inside PROFILES_ROOT) and then `existsSync` + `hasProfileFiles` will operate on the symlink target. If the target directory happens to contain any `.hbs` file, the resolved path returned is the symlink path — which the caller then hands to `renderTemplate` / `readFileSync`. In practice only admin-created symlinks inside the bundled `profiles/` directory would trigger this; an end-user cannot inject a symlink via `profileName`. Still, the codebase's own `getWorkspaceMemoryFile` (`memory-search.ts` lines 261–278) and `startWebServer` (`web/server.ts` lines 194–303) both use `realpathSync` before their boundary check for exactly this reason. `getProfilePath` should do the same.

Suggested fix:

```ts
import { realpathSync } from "node:fs";

export function getProfilePath(profileName: string): string {
  const requested = resolve(PROFILES_ROOT, profileName);
  if (requested !== PROFILES_ROOT && !requested.startsWith(PROFILES_ROOT + pathSep)) {
    throw new Error(`Invalid profile name: ${profileName}`);
  }
  // Re-check after resolving symlinks — same pattern as memory-search.ts + web/server.ts.
  let real: string;
  try {
    real = realpathSync(requested);
  } catch {
    real = requested; // ENOENT — let existsSync handle it below
  }
  if (real !== PROFILES_ROOT && !real.startsWith(PROFILES_ROOT + pathSep)) {
    throw new Error(`Invalid profile name: ${profileName}`);
  }
  // ... rest unchanged
}
```

### SECURITY (style/consistency) — Root-exact arm is subtly fragile

The `requested === PROFILES_ROOT` arm (allowing an empty-string or `.` input) is safe today because `hasProfileFiles(PROFILES_ROOT)` returns false. But it is an unlocked door: if a top-level `.hbs` file ever lands in `profiles/`, empty-string input would resolve to a valid profile path for the root itself. The safer pattern used in `memory-search.ts` line 274 is the same two conditions but ordered differently, making the intent clearer. Low urgency, but worth a comment.

### Windows mixed-separator input

On Windows, Node accepts both `/` and `\`. A user passing `"..\\etc"` would have `resolve()` normalise it before the check, so the traversal is still caught. This is fine — but there is no test for a mixed-separator traversal attempt (`"subfolder/../../etc"`). Low risk given normalisation, but worth documenting.

### Test gap — symlink traversal vector not covered

There is no test that creates a symlink inside a temp PROFILES_ROOT pointing outside it and confirms rejection. Given the symlink finding above, this is the most important missing test case.

### Test gap — tests are coupled to the real `profiles/` directory on disk

`getProfilePath("default")` and the fallback tests require the bundled `default` profile directory to exist. The `"coder"` test requires that name NOT be a real directory. This makes the suite fragile against changes to bundled profiles. A temp-dir fixture or mocking `PROFILES_ROOT` would make traversal tests environment-independent.

### Other boundary checks in the codebase — no action needed

- `workspace.ts` `isInsideWorkspace()` (line 51): already uses `path.sep`, correct.
- `memory-search.ts` `getWorkspaceMemoryFile()` (line 274): uses `path.sep` + `realpathSync`, correct and more thorough.
- `web/server.ts` static-file check (line 302): uses `relative()` + `realpathSync`, correct.
- `scaffold.ts` `startsWith(skillsPool)` (lines 1228, 1300, 1333): symlink-target comparisons, not security guards, no sep issue.

No other site needs the same fix this PR applies.

---

## Concrete recommendation

1. **Must-fix before merge:** Add `realpathSync` resolution after the lexical check, matching the pattern already established in `memory-search.ts` and `web/server.ts`. The current code leaves a symlink escape path open.
2. **Should-fix:** Add a test that uses a temp dir so traversal tests don't depend on the real `profiles/` layout on disk.
3. **Nice-to-have:** Add a mixed-separator test (`"subfolder/../../etc"`) and a symlink traversal test to the suite.

The core Windows fix (the original motivation for the PR) is correct and unambiguous — this is a real bug fix. The issues above are about hardening the security boundary that the fix touches.

PR #124, opened 2026-04-27
feat(telegram): surface reply_to_message context in channel envelope

### LINKED ISSUE #119: switchroom-telegram: reply_to_message_id not captured in message buffer
## Problem

When a user uses Telegram's native reply feature (long-press → Reply on a specific message), the incoming `Message` object from the Bot API includes a `reply_to_message` field with the original message content. This field is currently **dropped** when the message is stored in the local SQLite buffer.

The agent receives only the new message text, with no indication of which prior message was being replied to or what it contained. This breaks reply-as-context UX entirely — the agent asks for clarification ("what's 'this'?") on questions that made perfect sense as replies.

## Expected behaviour

When a user replies to a Telegram message, the agent should see:
1. The message text
2. Which message was replied to (ID + text excerpt, or the full `reply_to_message` content)

This should surface in:
- The local SQLite buffer (new column, or a metadata field)
- The `<channel>` tag injected into the Claude Code session, e.g. `reply_to_message_id="386" reply_to_text="Morning briefing sent..."`

## Proposed fix

In the Telegram webhook/polling handler, when `message.reply_to_message` is present:
- Store `reply_to_message_id` and a truncated `reply_to_text` alongside the incoming message in the buffer
- Include both in the `<channel>` tag so the agent has the context without needing an extra `get_recent_messages` call

## Context

Observed in gymbro agent. User replied to a cron briefing message with "How do we fix this deterministically for switchroom crons?" — the agent had no idea what "this" referred to and asked for clarification, creating a frustrating back-and-forth.

mekenthompson:
## Summary

Closes [switchroom#119](https://github.com/switchroom/switchroom/issues/119).

When the user uses Telegram's native "long-press → Reply" feature, the inbound Bot API \`Message\` includes \`reply_to_message\` with the original message's id and text. The plugin was dropping it. The agent saw "how do we fix this?" with no antecedent and asked for clarification — breaking reply-as-context UX.

This PR captures \`reply_to_message_id\` and a 200-char preview of the original text (or caption) and exposes them in three places:

1. **\`<channel>\` envelope meta** — the gateway path ([telegram-plugin/gateway/gateway.ts](telegram-plugin/gateway/gateway.ts)) and the legacy direct-MCP path ([telegram-plugin/server.ts](telegram-plugin/server.ts)) both emit \`reply_to_message_id\` / \`reply_to_text\` attributes when set. The trust-model description at [server.ts:1029](telegram-plugin/server.ts) already listed \`reply_to_message_id\` as a permitted attribute — this PR lands the implementation that matched the contract.
2. **History SQLite buffer** — adds two columns to \`messages\` plus an idempotent \`ALTER TABLE ADD COLUMN\` migration for pre-existing DBs.
3. **\`get_recent_messages\` summary** — appends a compact \`↪️#<id>:"…"\` marker to each row's text line.

## Test plan
- [x] \`npm run lint\` passes
- [x] New cases in [telegram-plugin/tests/history.test.ts](telegram-plugin/tests/history.test.ts) cover round-trip, null-context, and per-row independence (run via \`bun test\` in CI; the file is excluded from vitest because it imports \`bun:sqlite\`)
- [ ] Manual smoke: long-press a bot message in Telegram, reply with "what does this mean?", confirm the agent sees \`reply_to_text\` in the channel meta and resolves the antecedent without asking
- [ ] CI green

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -65,7 +65,14 @@ import {
   initHistory, recordInbound, recordOutbound, recordEdit,
   deleteFromHistory, query as queryHistory, getLatestInboundMessageId,
 } from '../history.js'
-import { parseQueuePrefix, parseSteerPrefix, formatPriorAssistantPreview } from '../steering.js'
+import { parseQueuePrefix, parseSteerPrefix, formatPriorAssistantPreview, formatReplyToText } from '../steering.js'
+
+/**
+ * Truncation cap for the `reply_to_text` channel-meta attribute (issue #119).
+ * Same value as in server.ts — kept in sync because the two paths produce
+ * identical envelope shapes.
+ */
+const REPLY_TO_TEXT_MAX = 200
 import { markdownToHtml, splitHtmlChunks, repairEscapedWhitespace } from '../format.js'
 import {
   startText as buildStartText,
@@ -2562,13 +2569,11 @@ async function handleInbound(
 
   const imagePath = downloadImage ? await downloadImage() : undefined
 
-  // If the user replied to a prior message via Telegram's native "long-press
-  // → Reply" affordance, capture the original message_id and a truncated
-  // preview so the agent can resolve "this" / "that" anaphora without
-  // an extra get_recent_messages call. See issue #119.
+  // Telegram-native reply context (issue #119). Same pattern as server.ts:
+  // `replyToText` is raw (for SQLite); `replyToTextEscaped` is XML-escaped
+  // (for channel meta).
   const replyToMsg = ctx.message?.reply_to_message
   const replyToMessageId = replyToMsg?.message_id
-  const REPLY_TO_TEXT_MAX = 200
   const replyToTextRaw = replyToMsg
     ? (replyToMsg.text ?? replyToMsg.caption ?? undefined)
     : undefined
@@ -2577,6 +2582,7 @@ async function handleInbound(
         ? replyToTextRaw.slice(0, REPLY_TO_TEXT_MAX - 1) + '…'
         : replyToTextRaw)
     : undefined
+  const replyToTextEscaped = formatReplyToText(replyToTextRaw, REPLY_TO_TEXT_MAX)
 
   if (HISTORY_ENABLED) {
     try {
@@ -2654,7 +2660,9 @@ async function handleInbound(
       // treat this as the antecedent for "this" / "that" / pronoun
       // references in the body, instead of asking the user what they meant.
       ...(replyToMessageId != null ? { reply_to_message_id: String(replyToMessageId) } : {}),
-      ...(replyToText != null && replyToText.length > 0 ? { reply_to_text: replyToText } : {}),
+      // Use the XML-escaped form for the meta — the raw form is in the
+      // SQLite buffer for verbatim retrieval via get_recent_messages.
+      ...(replyToTextEscaped != null && replyToTextEscaped.length > 0 ? { reply_to_text: replyToTextEscaped } : {}),
       // queued="true" when mid-turn with no steer prefix (new default), or
       // with explicit /queue or /q prefix (legacy alias).
       ...((isQueuedMidTurn || isQueuedPrefix) ? { queued: 'true' } : {}),
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -2562,6 +2562,22 @@ async function handleInbound(
 
   const imagePath = downloadImage ? await downloadImage() : undefined
 
+  // If the user replied to a prior message via Telegram's native "long-press
+  // → Reply" affordance, capture the original message_id and a truncated
+  // preview so the agent can resolve "this" / "that" anaphora without
+  // an extra get_recent_messages call. See issue #119.
+  const replyToMsg = ctx.message?.reply_to_message
+  const replyToMessageId = replyToMsg?.message_id
+  const REPLY_TO_TEXT_MAX = 200
+  const replyToTextRaw = replyToMsg
+    ? (replyToMsg.text ?? replyToMsg.caption ?? undefined)
+    : undefined
+  const replyToText = replyToTextRaw != null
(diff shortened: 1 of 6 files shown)


mekenthompson (commented):
## Overall take

Solid, well-motivated change. The three-layer approach (envelope meta + SQLite history + get_recent_messages inline display) is the right shape, the migration is defensive and idempotent, and the test coverage for the history layer is good. There is one security issue that needs fixing before merge, two minor inconsistencies, and a handful of low-priority observations.

---

## Findings

### MUST FIX — `reply_to_text` placed in XML attribute position without escaping

`reply_to_text` is user-supplied content (the text of the message the user replied to). Both `server.ts` and `gateway/gateway.ts` place it directly into the `meta` object that becomes XML attributes on the `<channel>` tag:

```ts
...(replyToText != null && replyToText.length > 0 ? { reply_to_text: replyToText } : {}),
```

Compare `prior_assistant_preview`, which goes through `formatPriorAssistantPreview` → `escapeXmlAttribute` before entering the same `meta` object. `reply_to_text` skips that step entirely.

A replied-to message containing `"`, `<`, `>`, or `&` will produce malformed XML in the `<channel>` envelope (e.g. `reply_to_text="say &quot;hello&quot;"` is fine; `reply_to_text="a" b="c"` is attribute injection). An adversary who can get the allowlisted user to long-press-reply to a crafted prior message could inject fake trusted attributes into the envelope. In a personal-DM bot this is self-harm only, but in multi-user group deployments it is meaningful.

**Fix:** run `reply_to_text` through `escapeXmlAttribute` (already exported from `steering.ts`) before inserting it into `meta`. The same truncation pipeline is fine; just wrap the final value:

```ts
import { escapeXmlAttribute } from './steering.js'
// ...
const replyToText = replyToTextRaw != null
  ? escapeXmlAttribute(
      replyToTextRaw.length > REPLY_TO_TEXT_MAX
        ? replyToTextRaw.slice(0, REPLY_TO_TEXT_MAX - 1) + '…'
        : replyToTextRaw
    )
  : undefined
```

Same fix needed in both `server.ts` and `gateway/gateway.ts` (they duplicate the block identically — a candidate for a shared helper).

### MUST FIX — Trust-model description omits `reply_to_text`

`server.ts:1029` lists the trusted attributes:

> `source, chat_id, message_id, message_thread_id, user, user_id, ts, reply_to_message_id, image_path, attachment_file_id`

`reply_to_text` is not in the list, even though it is now emitted as a meta attribute. If an agent reads the trust model literally, it may treat `reply_to_text` as untrusted body content rather than plugin-supplied metadata, defeating the whole purpose of surfacing it in the envelope. Add `reply_to_text` to the list (but also note in the description that its value is a preview of user-generated content, not a plugin-generated ID — so it should be treated as contextual rather than authoritative).

---

### Should fix — `REPLY_TO_TEXT_MAX` constant defined inside the function body (twice)

`const REPLY_TO_TEXT_MAX = 200` appears inside `handleInbound` in both `server.ts` and `gateway/gateway.ts`. It is a module-level constant and should live at module scope (or in a shared helper), both for readability and so that both paths stay in sync if the limit changes.

### Should fix — `get_recent_messages` inline marker double-truncates, with inconsistent limits

The history layer stores up to 200 chars. The `get_recent_messages` summary line slices to 60 chars:

```ts
r.reply_to_text.slice(0, 60)${r.reply_to_text.length > 60 ? '…' : ''}
```

That is fine on its own, but the length check `r.reply_to_text.length > 60` is applied to the already-stored string, which can be up to 200 chars. The ellipsis logic is correct; this is not a bug. However the 60-char limit is undocumented and different from the 200-char storage limit. A brief comment explaining the two-tier design (200 for storage, 60 for summary lines) would prevent future maintainers from "fixing" it.

---

### Low priority / observations

- **Code duplication.** The `replyToMsg` extraction block is copy-pasted verbatim between `server.ts` and `gateway/gateway.ts`. Given the security fix above must be applied to both, consider extracting a `extractReplyContext(ctx, maxLen)` helper in a shared module.
- **`reply_to_message_id` stored as `INTEGER` in SQLite, but `query()` returns it as part of `RecordedMessage` cast from `SELECT *`.** `bun:sqlite` will return it as a JS number. The interface declares it as `number | null`, which is correct. No bug, just worth confirming in a test that the type is a number and not a string after a round-trip (the existing round-trip test does assert `.toBe(17)` with `===` semantics, so this is already covered).
- **`replyToMessageId ?? null` vs `replyToMessageId ?? undefined`.** The `recordInbound` args interface declares `reply_to_message_id?: number | null | undefined`. Passing `null` when unset is correct for the SQL bind. Good.
- **No envelope-level test.** The history-layer tests are good, but there is no test that an inbound message with `reply_to_message` produces the correct `meta` attributes in the IPC/MCP notification (analogous to the steering attribute tests in `tests/steering.test.ts` and `tests/e2e.test.ts`). This is a gap — the XML-escaping bug above would have been caught by such a test.
- **`reply_to_text` in `get_recent_messages` JSON payload.** The `messages` array returned by `get_recent_messages` includes raw `RecordedMessage` rows, so `reply_to_text` surfaces both in the inline summary string and in the JSON. That is fine and probably useful. Worth noting in the PR description or a code comment so it is intentional, not accidental.

---

## Concrete recommendations

1. Apply `escapeXmlAttribute` to `replyToText` in both `server.ts` and `gateway/gateway.ts` before the `meta` spread. This is the blocking issue.
2. Add `reply_to_text` to the trust-model attribute list at `server.ts:1029`, with a note that its content is a preview of user-generated text.
3. Hoist `REPLY_TO_TEXT_MAX = 200` to module scope (or a shared constant) in both files.
4. Add an envelope-shape test (either in `tests/e2e.test.ts` or a new `tests/reply-to.test.ts`) that verifies: (a) the `meta` object includes `reply_to_message_id` and `reply_to_text` when `reply_to_message` is present, (b) `reply_to_text` is absent when the field is null, and (c) text containing XML-special chars (`"`, `<`, `>`, `&`) is properly escaped before it enters the meta.

Items 1 and 2 are blocking. Items 3 and 4 can follow in a subsequent PR if preferred.

---

PR #125, opened 2026-04-27
feat(schedule): suppress_stdout flag to skip stdout-to-Telegram forwarding

### LINKED ISSUE #118: Cron sessions: stdout leaks as second Telegram message — need deterministic suppression
## Problem

When a scheduled cron session sends a Telegram message via `stream_reply`/`reply` MCP tools and then exits, the session's stdout (Claude's final text output) is also routed to Telegram as a second message. The user sees the briefing twice — once from the MCP tool call, once from the stdout route.

## Current workaround

Appending `HEARTBEAT_OK` to the end of cron prompts suppresses the follow-up message via the plugin's silent-reply behaviour. But this is model-level — the model must correctly output that exact string as the last thing it says. It's not deterministic. Any explanation, summary, or trailing sentence breaks it.

## Expected behaviour

A scheduled session that has already sent a Telegram message via MCP tools should **not** have its stdout routed to Telegram. Or alternatively, stdout routing to Telegram should be opt-in, not the default.

## Proposed fix

A config option on `schedule:` entries (e.g. `suppress_stdout: true` or `reply_mode: mcp_only`) that tells the harness: "this session communicates exclusively via MCP tool calls — do not forward stdout to Telegram". This makes the behaviour deterministic regardless of what the model outputs.

## Context

Observed on gymbro agent, all 6 scheduled cron tasks exhibit this. The morning briefing sends a full formatted message via `stream_reply`, then the session exits with a summary sentence like "Morning briefing sent. Key signals flagged: ..." which arrives as a second Telegram message.

mekenthompson:
## Summary

Closes [switchroom#118](https://github.com/switchroom/switchroom/issues/118).

Cron sessions that post their own message via MCP tools (\`stream_reply\` / \`reply\`) currently arrive twice in Telegram: once from the MCP tool, then a second time when the cron script forwards the model's trailing summary via curl. The \`HEARTBEAT_OK\` / \`NO_REPLY\` sentinel approach is model-level and breaks the moment the model says one extra word.

This PR adds a deterministic config-level escape hatch:

\`\`\`yaml
schedule:
  - cron: "0 7 * * *"
    prompt: "Morning briefing"
    suppress_stdout: true   # NEW
\`\`\`

When set, the generated \`cron-N.sh\` switches from the capture-then-curl block to:

\`\`\`bash
exec claude -p '<prompt>' --model '<model>' --no-session-persistence > /dev/null 2>&1
\`\`\`

Default is \`false\` so behavior is unchanged for existing cron entries.

## Implementation
- [src/config/schema.ts](src/config/schema.ts) — new \`suppress_stdout\` boolean on \`ScheduleEntrySchema\` (default false)
- [src/agents/scaffold.ts](src/agents/scaffold.ts) — \`buildCronScript\` takes a new optional positional, both scaffold and reconcile call sites thread \`entry.suppress_stdout\` through
- [src/vault/broker/acl.test.ts](src/vault/broker/acl.test.ts) — builder updated for the now-required field on the parsed \`ScheduleEntry\` output type

## Test plan
- [x] \`npm run lint\`
- [x] New cases in [tests/scaffold.test.ts](tests/scaffold.test.ts):
  - default path forwards stdout to Telegram (legacy behavior)
  - \`suppress_stdout: true\` skips the curl-to-Telegram block
  - reconcile flips a cron from default → suppressed and reports the change
- [ ] CI green
- [ ] Manual smoke on the gymbro agent: set \`suppress_stdout: true\` on the morning briefing entry, confirm only one Telegram message arrives at trigger time

### DIFF
--- a/docs/scheduling.md
+++ b/docs/scheduling.md
@@ -36,6 +36,23 @@ For each schedule entry, switchroom generates:
 | `cron` | Yes | — | Standard 5-field cron expression |
 | `prompt` | Yes | — | The prompt sent to Claude |
 | `model` | No | `claude-sonnet-4-6` | Model for this task |
+| `secrets` | No | `[]` | Vault keys this task may read via the broker. See [configuration.md#vault-broker-linux-only](configuration.md#vault-broker-linux-only). |
+| `suppress_stdout` | No | `false` | When `true`, the cron script discards stdout instead of forwarding it to Telegram. Use for tasks that send their own message via MCP tools (`stream_reply` / `reply`) so the trailing model summary doesn't post as a duplicate. See [issue #118](https://github.com/switchroom/switchroom/issues/118). |
+
+### When to set `suppress_stdout: true`
+
+The default cron flow captures `claude -p`'s stdout and `curl`s it to Telegram, so the agent gets one message per cron run for free. That works for tasks that respond entirely via the model's final text — "morning briefing", "weekly summary", etc.
+
+It backfires for tasks that already post their own message via MCP tools (`mcp__switchroom-telegram__stream_reply`, `mcp__switchroom-telegram__reply`). The MCP tool call posts the formatted message; then the cron script forwards the model's trailing summary as a second message:
+
+```
+[MCP tool call]: 🌅 Morning briefing — 3 items on today's calendar...
+[stdout]:        Morning briefing sent. Key signals flagged: low sleep, gym in 2h.
+```
+
+The user sees both. The `HEARTBEAT_OK` / `NO_REPLY` sentinels can suppress the stdout but only if the model produces them as the exact final tokens — fragile.
+
+`suppress_stdout: true` is the deterministic switch. The cron script `exec`s `claude -p` with stdout routed to `/dev/null`, so only the MCP-tool-posted message reaches Telegram.
 
 ### Cron Expression Examples
 
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -1182,11 +1182,14 @@ fi
 
 ${suppressStdout
     ? `# suppress_stdout: true (cron entry posts its own message via MCP tools — see issue #118)
-# Discard stdout so the trailing model summary doesn't arrive as a second message.
+# Discard stdout so the trailing model summary doesn't arrive as a second
+# Telegram message. Stderr is left open so systemd captures auth/network/
+# bad-prompt errors via journalctl — silently swallowing those would make
+# a broken cron job invisible to operators.
 exec claude -p ${shellSingleQuote(prompt)} \\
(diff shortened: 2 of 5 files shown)


mekenthompson (commented):
Solid implementation — correct default, clean gate placement, and good test coverage. A few items worth fixing before merge.

**What's good**
- Default is `false` — existing schedules are unaffected, backward-compatible.
- Gate is at the *forward* layer (capture/exec vs curl), not the capture layer. stderr from `claude -p` was already going to `/dev/null` in both the old and new paths, so the suppress path's `> /dev/null 2>&1` is consistent.
- Zod schema is well-formed: `.boolean().default(false)` with a clear `.describe()` that explains the rationale and links the issue.
- Both call sites (`scaffoldAgent` and `reconcileAgent`) thread `entry.suppress_stdout ?? false` through — no missed call site.
- Three new test cases cover: default path unchanged, suppress path removes curl, and reconcile flips an existing cron correctly. The `acl.test.ts` builder fix is correct.
- JSDoc on `buildCronScript` updated and accurate.

**Issues to fix**

- **`docs/scheduling.md` table is stale.** The Configuration table at line 36 lists `cron`, `prompt`, `model` — it does not mention `secrets` (added in PR #112) or the new `suppress_stdout`. Readers following the docs won't know the field exists. At minimum add a row; ideally add a short prose explanation matching the PR description's "why".

- **stderr is also suppressed, which may surprise users.** In the suppress path, `> /dev/null 2>&1` discards stderr too. If `claude -p` errors (auth failure, network timeout, bad prompt), there is no signal anywhere — the systemd service exits 0 (because `exec` replaces the shell and a non-zero exit would normally appear in `journalctl`, but `/dev/null 2>&1` swallows the error text). Consider routing stderr to the journal instead: `> /dev/null 2>/dev/null` is fine if you're intentional about silence, but `exec claude … > /dev/null` (stderr left open) would let `journalctl` show errors without duplicating stdout to Telegram. This is a UX call; document whichever you choose.

- **`suppress_stdout` in the generated comment string is a mismatch opportunity.** The bash comment `# suppress_stdout: true` is what the test asserts on (`expect(content).toContain("suppress_stdout: true")`). That's fragile — if the comment wording changes the test breaks. Minor, but worth asserting on `exec claude -p` instead of the comment string for the sentinel check.

- **README.md schedule example (line 165) is not stale** — it predates the feature and doesn't enumerate fields, so no fix needed there.

**Concrete recommendation**

1. Add `suppress_stdout` row to `docs/scheduling.md` table (and optionally a short "When to use" paragraph referencing issue #118).
2. Decide explicitly on stderr behavior and document it — routing stderr to the journal is likely more debuggable than full silence.
3. Optionally harden the test assertion to check `exec claude -p` rather than the comment text.

Otherwise ready to merge.

Issue #109, opened 2026-04-26
UX: 'status?' from user is a failure signal — log + add debug visibility


mekenthompson:
## Problem

When a user has to ask \"status?\" or \"what's happening\" mid-task, the live progress surface (pinned card + status reactions) has failed its job. The card's whole purpose is to make state obvious without prompting. If the user is asking, the card isn't doing its job.

## Recent occurrences (one Telegram session, 2026-04-26)

1. **~14:24** — During a worker that ran for ~13 min and then truncated. Card was pinned but the user couldn't tell from it whether the worker was still alive, stalled, or done. He had to ask.
2. **~15:37** — Between an agent restart attempt being interrupted and the next status update. Card showed nothing during the silence.

Both share a pattern: **long silent stretches during work the user can't directly see**, where the card's content doesn't refresh visibly.

## What should happen instead

The card should answer these questions at a glance, always:

- Is anything happening right now? (last activity timestamp, ticking elapsed)
- What is it doing? (current tool / step, in human language)
- Is it making progress, stalled, or idle?
- If stalled: why?

When the user has to ask, that's a signal one of those answers is missing or unclear from the card alone.

## Proposed work

1. **Treat 'status?' as telemetry**: when an inbound message matches \`/^\s*status\??$/i\` (case-insensitive, optional ?), emit a structured log line so we can count + analyze frequency:
   \`\`\`
   telegram gateway: ux-failure: status-query chatId=... agent=... lastActivity=... lastEditAge=... lastEmitTurnKey=... cardState=...
   \`\`\`
2. **Snapshot context at the moment of the query**: log the current pinned card's HTML (truncated), last emit time, current \`chatState.lastEventAt\`, current sub-agent registry summary. Helps reconstruct \"what was the user looking at when they asked?\".
3. **Stretch — auto-refresh trigger**: on detecting a status query, force-flush the progress card immediately (re-render with current state) so the user sees a fresh card without us having to explain.
4. **Stretch — proactive 'I'm here' beat**: when card is in \`pendingCompletion\` state with sub-agents running, ensure the heartbeat re-renders at least every N seconds with a visible elapsed-time delta. Currently the heartbeat coalesces if HTML hasn't changed — for the user-facing case we want SOMETHING to tick visibly.

## Non-goals

- Don't gate this on the user saying the magic word \"status\". The 30s+ silent stretches are the real problem; this issue is about *measuring* them so we can fix them properly.

## Acceptance

[...]

Issue #83, opened 2026-04-25
subagent-watcher: skip pre-existing JSONL files at startup (don't replay historical dispatches)


mekenthompson:
## Bug

When the gateway starts and `SWITCHROOM_SUBAGENT_WATCHER=1`, the watcher scans `.claude/projects/<...>/subagents/` and registers every existing `agent-*.jsonl` file. Each registration fires a 'Worker dispatched: ...' Telegram notification. After agent restart with months of history, this can flood the chat with dozens of bogus 'dispatched' messages — observed 11+ on clerk after a routine restart on 2026-04-26.

Hotfix landed in PR #82 makes the watcher opt-in. This issue is the proper fix.

## Fix

At watcher boot, capture `startupAt = Date.now()`. When discovering a JSONL file, check its `statSync(path).mtime`. If the file's mtime is older than startupAt minus a small grace (e.g. 10s), register it WITHOUT firing the 'dispatched' notification — the file represents a historical sub-agent, not a fresh one. Cards + completion notifications can still fire for already-registered entries.

## Acceptance

- [ ] Restart with months of historical JSONL → no dispatch notifications
- [ ] Genuinely new sub-agent (file created post-startup) → notification fires
- [ ] Once verified, flip the default from opt-in back to opt-out (or remove the env gate)
- [ ] Test added: harness sets up a stale JSONL, starts watcher, asserts no dispatch notification

PR #126, opened 2026-04-27
feat(gateway): log structured status-query telemetry (#109)

mekenthompson:
## Summary

Closes step 1 of [switchroom#109](https://github.com/switchroom/switchroom/issues/109).

When a user types just \`status\` / \`status?\` mid-turn, the pinned progress card has failed its job — the user shouldn't need to ask. The issue's first ask is telemetry: count and analyze how often this happens before deciding on follow-on UX changes.

This PR adds an anchored regex match in [telegram-plugin/gateway/gateway.ts](telegram-plugin/gateway/gateway.ts) at the top of \`handleInbound\` (just after the access gate). On match, it emits a structured stderr line with a snapshot of the card state — agent name, chat/thread, card stage, turn age in seconds, item count, sub-agent count.

\`\`\`
telegram gateway: ux-failure status-query agent=klanker chat_id=-1001234 thread=42 card_stage=Run card_turn_age_s=187 card_items=3 card_subagents=1
\`\`\`

The message is **not** intercepted — it still flows through to the agent so the agent can answer the question if the card hadn't.

## Out of scope

The issue's stretch items (auto-flush the card on status query, proactive heartbeat beats) are deliberately **not** in this PR. They're invasive enough to deserve their own change after we have telemetry showing *where* the silent stretches actually happen. This PR just unlocks the measurement.

## Test plan
- [x] \`npm run lint\`
- [ ] Manual smoke: send \`status?\` in a Telegram chat, confirm the log line appears in journalctl with the expected fields
- [ ] Run a few weeks → \`grep 'ux-failure status-query' service.log | wc -l\` gives a frequency number
- [ ] CI green

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -610,6 +610,11 @@ function logOutbound(
   )
 }
 
+// Issue #109: when a user types just "status" or "status?" they're asking
+// because the live progress surface failed to communicate. Anchored regex,
+// case-insensitive, optional trailing "?" — must be the entire body.
+const STATUS_QUERY_RE = /^\s*status\??\s*$/i
+
 // ─── Permission handling ──────────────────────────────────────────────────
 const PERMISSION_REPLY_RE = /^\s*(y|yes|n|no)\s+([a-km-z]{5})\s*$/i
 const pendingPermissions = new Map<string, { tool_name: string; description: string; input_preview: string; startedAt: number }>()
@@ -2208,6 +2213,32 @@ async function handleInbound(
   const msgId = ctx.message?.message_id
 
   if (messageThreadId != null) chatThreadMap.set(chat_id, messageThreadId)
+
+  // Issue #109: when the user has to ask "status?" mid-turn, the live progress
+  // surface (pinned card + status reactions) has failed its job. Log the
+  // event with a snapshot of the card state so we can count + analyze
+  // frequency. We don't intercept the message — it still flows through to
+  // the agent, since the agent may have a useful answer the card hasn't
+  // surfaced yet.
+  if (STATUS_QUERY_RE.test(text)) {
+    try {
+      const threadKey = messageThreadId != null ? String(messageThreadId) : undefined
+      const cardState = progressDriver?.peek(chat_id, threadKey)
+      const turnAge = cardState?.turnStartedAt
+        ? Math.max(0, Math.floor((Date.now() - cardState.turnStartedAt) / 1000))
+        : null
+      const stage = cardState?.stage ?? 'idle'
+      const itemCount = cardState?.items.length ?? 0
+      const subAgentCount = cardState?.subAgents.size ?? 0
+      const agentName = process.env.SWITCHROOM_AGENT_NAME ?? '-'
+      process.stderr.write(
+        `telegram gateway: ux-failure status-query agent=${agentName} chat_id=${chat_id} thread=${threadKey ?? 'none'} ` +
+        `card_stage=${stage} card_turn_age_s=${turnAge ?? 'idle'} card_items=${itemCount} card_subagents=${subAgentCount}\n`,
+      )
+    } catch (err) {
+      process.stderr.write(`telegram gateway: status-query telemetry failed: ${(err as Error).message}\n`)
+    }
+  }
 
   // Permission-reply intercept
   const permMatch = PERMISSION_REPLY_RE.exec(text)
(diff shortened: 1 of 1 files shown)


mekenthompson (commented):
## Review: feat(gateway): log structured status-query telemetry (#109)

**Overall:** This is a small, well-scoped change that does exactly what it says — it instruments one specific failure signal without altering behaviour. The implementation is solid in most respects, but there are several items worth addressing before merge.

---

### Good

- **Scope discipline.** 31 lines added, zero deleted. Stretch items (auto-flush, heartbeat) correctly left out. The PR description explains this well.
- **Access gate first.** The telemetry block sits after `gate(ctx)` returns `allow`, so only authenticated senders are counted. No unauthenticated traffic can inflate metrics.
- **No message interception.** The message continues to flow to the agent — correct; the user's question may still get a useful answer.
- **Error isolation.** The `try/catch` prevents a telemetry failure from breaking the inbound path. Good defensive choice.
- **Uses `peek` correctly.** `progressDriver?.peek(chat_id, threadKey)` gracefully handles the case where the driver is null or the chat has no card (returns `undefined`), which the code handles via `?? 'idle'` fallbacks.
- **Issue reference.** The PR closes step 1 of #109 explicitly. The commit message and description are clear.

---

### Findings

**[BLOCKER] CI is failing — Build #83**
`gh pr checks` shows `buildkite/switchroom fail` on this PR. The diff itself is minimal so the failure may be a flake or a pre-existing test issue, but it needs to be green before merge. The manual smoke test checkbox is also unchecked.

**[MEDIUM] Log format mismatch vs. issue acceptance criteria**
Issue #109 specifies the grep anchor as `ux-failure: status-query` (colon after `ux-failure`), but the emitted line uses `ux-failure status-query` (space, no colon). Whoever wrote the monitoring runbook or grep alias from the issue description will get zero hits. Trivial to fix — pick one and document it in the issue.

**[MEDIUM] Issue #109 item 2 not addressed — card HTML / lastEventAt snapshot missing**
The issue asks for: "Snapshot context at the moment of the query: log the current pinned card's HTML (truncated), last emit time, current `chatState.lastEventAt`, current sub-agent registry summary." The PR logs `card_items` and `card_subagents` counts, but not `lastEventAt` or any card HTML. The PR description calls this "step 1" and doesn't claim to cover item 2, so this may be intentional — but it should be clarified in the issue (close step 1 items only, re-open / create follow-up for item 2 snapshot).

**[MEDIUM] `card_turn_age_s=idle` is a type mismatch in the log field**
When `cardState` is absent (no active turn), the code emits `card_turn_age_s=idle`. Mixing string and integer values in the same field name breaks any structured log parser that treats the field as numeric (e.g., a Loki LogQL `>` comparison, a Datadog facet, or even `awk '{print $NF}'`). Better: emit `card_turn_age_s=-1` (sentinel) or omit the field and let the absence signal idle, or emit a separate boolean `card_idle=true`.

**[LOW] `turnStartedAt === 0` case is not handled**
`ProgressCardState.turnStartedAt` is documented as `0 when idle`. The current guard is:
```ts
const turnAge = cardState?.turnStartedAt
  ? Math.floor(...)
  : null
```
This correctly treats `0` as falsy, so `turnAge` becomes `null` and the log emits `card_turn_age_s=idle`. That's fine — but it means even if `peek()` returns a state (the driver exists and `stage` is not `'idle'`), a zero `turnStartedAt` would wrongly report `idle`. This is an edge case but worth a comment.

**[LOW] No unit test for the emitted shape**
There are several gateway test files in `telegram-plugin/tests/`. Nothing covers `STATUS_QUERY_RE` matching or the shape of the emitted log line. A test that mocks `progressDriver.peek`, calls `handleInbound` with `"status?"`, and asserts the `process.stderr.write` call received the expected fields would lock in the format and catch the colon/space issue above. Not a blocker, but this is exactly the kind of regression that creeps in silently.

**[LOW] `progressDriver` could be non-null with a null `ProgressCardState` for a different reason**
The `peek()` call returns `undefined` when the chat has no tracked state. The code handles this fine. But it's worth noting that `subAgents` on `ProgressCardState` is a `ReadonlyMap`, not an array — `cardState?.subAgents.size` is correct. Just confirming the types are used properly.

**[LOW] `SWITCHROOM_AGENT_NAME` missing → `-` is silent**
If the env var is unset in a non-standard deploy, the log line emits `agent=-` which looks like a bug in grep output. Consider a warning log when the name is absent, or use `process.env.npm_package_name` as a fallback.

---

### Concrete recommendations

1. **Fix CI** — investigate Build #83 before merge.
2. **Fix the grep anchor** — choose `ux-failure: status-query` (colon, matching the issue) or update the issue to match the code. Document the exact grep in the issue's Acceptance section.
3. **Fix `card_turn_age_s` type consistency** — use `-1` or omit when idle; do not mix string and integer.
4. **Add one unit test** asserting the emitted log fields for both the active-turn and idle-card cases.
5. **Clarify issue #109 item 2 scope** — comment on #109 saying item 2 (snapshot) is deferred to a follow-up so it doesn't silently stay open.

---

Issue #81, opened 2026-04-25
progress-card: still showing raw tool calls instead of human descriptions — add diagnostic logging


mekenthompson:
## Symptom

The pinned progress card sometimes still renders raw tool counts (e.g. \`3× Bash\`) instead of human-readable preamble text (e.g. \`Reading the file.\`, \`Running tests.\`), despite #51 (\"recover prose-as-step text into turn-flush\") and the agent SOUL guidance instructing agents to lead each tool call with a one-line preamble.

The user observed this multiple times today during klanker's turns. The agent IS producing preambles in its assistant text — but the card falls back to tool-count rendering anyway.

## What we don't know

- Are narratives being captured in the turn state? (\`turn-flush-prose-recovery.ts\` uses \`state.narratives\`)
- Is the recovery hitting the \`narratives === undefined\` defensive branch?
- Is the assistant text being parsed correctly into narratives?
- Is the card render falling back to tool-count even when narratives are present?

## Asked for

Add diagnostic logging at:
1. Where assistant text is captured into the narratives array — log the count + preview of each narrative
2. Where the card decides between prose vs tool-count rendering — log the chosen path + what it had available (narrative count, tool count, recovered prose length)
3. Where prose recovery runs — log returned text length + whether it took the recovery branch

After the logging is in place, run a normal klanker turn that exhibits the bug (any tool-heavy turn will do), then capture the gateway log lines and file a follow-up bug with the data.

## Acceptance criteria

- [ ] Three log lines added at the decision points above
- [ ] Logged with a stable prefix (e.g. \`progress-card.diag:\`) so they're greppable
- [ ] Tested manually against a tool-heavy turn — confirm the logs land
- [ ] Once data is in hand, file a follow-up issue with the actual root cause + fix

## Severity

Medium — the card still functions, but it's noisier and less informative than designed. The user-visible regression is real.

mekenthompson:
Diagnostic logging shipped in #127 — all three decision points covered with the `progress-card.diag:` prefix:

- **Text event capture** → `telegram-plugin/progress-card-driver.ts:966` (gated behind `PROGRESS_CARD_DIAG=1` because it can be high-volume; flip the env var on for a one-off agent restart to capture data)
- **Render branch decision** → `telegram-plugin/progress-card-driver.ts:711` (always-on; fires on first emit + terminal flush, low volume)
[...]

PR #127, opened 2026-04-27
fix(progress-card): add diagnostic logs at narrative capture, render decision, and prose recovery

mekenthompson:
## Summary

Refs [switchroom#81](https://github.com/switchroom/switchroom/issues/81).

The progress card sometimes renders raw tool counts (\`3× Bash\`) instead of the human-readable preamble path, even when the agent did emit a preamble line. Issue #81 asks for diagnostic logging at the three decision points so we can capture which path is taken and why, then file a follow-up with the data.

This PR adds three log lines, all prefixed with \`progress-card.diag:\` for easy grep:

1. **text-event** — in [progress-card-driver.ts](telegram-plugin/progress-card-driver.ts) after \`reduce()\` runs on a \`text\` event. Reports \`captured\` vs \`discarded\` (the \`extractNarrativeLabel\` filter), narrative count before/after, and a 60-char preview.
2. **render branch=** — in \`flush()\` on first emit and on \`forceDone\`. Reports which checklist branch the renderer chose (\`narratives\` / \`tool-count-fallback\` / \`empty\`) plus the input counts.
3. **prose-recovery hit/miss** — in [gateway.ts](telegram-plugin/gateway/gateway.ts) turn-flush recovery path. Reports both the success and the empty-recover path with narrative count.

No behavior changes — pure logging. After a few real turns exhibit the bug, the logs should pinpoint which decision fails and a follow-up PR can land the fix.

## Test plan
- [x] \`npm run lint\` passes
- [ ] Manual smoke: run a tool-heavy turn and confirm three flavors of \`progress-card.diag:\` appear in journalctl
- [ ] After data is in hand, file a follow-up bug with the actual root cause + fix
- [ ] CI green

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1798,11 +1798,27 @@ function handleSessionEvent(ev: SessionEvent): void {
           threadId != null ? String(threadId) : undefined,
         )
         const recovered = recoverProseFromProgressCard(peek)
+        // Issue #81 diagnostic: record both the success and the empty-recover
+        // path so we can correlate "card showed tool-count" with "recovery
+        // had nothing to give either". The narrative count tells us whether
+        // the issue is at capture time (no narratives ever made it into the
+        // state) or at parse time (narratives existed but produced empty
+        // text after trim).
+        const narrativeCount = peek?.narratives.length ?? 0
         if (recovered.length > 0) {
           process.stderr.write(
             `telegram gateway: turn-flush prose-recovery — recovered ${recovered.length} chars from progress-card narratives chat=${chatId} turnKey=${currentTurnStartedAt}\n`,
           )
+          process.stderr.write(
+            `progress-card.diag: prose-recovery hit chatId=${chatId} turnKey=${currentTurnStartedAt} ` +
+            `narrative_count=${narrativeCount} recovered_len=${recovered.length}\n`,
+          )
           currentTurnCapturedText.push(recovered)
+        } else {
+          process.stderr.write(
+            `progress-card.diag: prose-recovery miss chatId=${chatId} turnKey=${currentTurnStartedAt} ` +
+            `narrative_count=${narrativeCount} peek_state=${peek == null ? 'null' : 'present'}\n`,
+          )
         }
       }
 
--- a/telegram-plugin/progress-card-driver.ts
+++ b/telegram-plugin/progress-card-driver.ts
@@ -671,6 +671,23 @@ export function createProgressDriver(config: ProgressDriverConfig): ProgressDriv
       taskNum.total > 1 ? taskNum : undefined,
       { stuckMs },
     )
+    // Issue #81 diagnostic: which checklist branch is the renderer taking?
+    // The card prefers `narratives` (human preambles) over `items` (raw
+    // tool counts). When prose lands without narratives we want to know
+    // why — log the available state at the decision boundary.
+    if (forceDone || chatState.lastEmittedHtml == null /* first emit */) {
+      const s = chatState.state
+      const branch = s.narratives.length > 0
+        ? 'narratives'
+        : s.items.length > 0
+          ? 'tool-count-fallback'
+          : 'empty'
+      process.stderr.write(
+        `progress-card.diag: render branch=${branch} chatId=${chatState.chatId} turnKey=${chatState.turnKey} ` +
+        `narratives=${s.narratives.length} items=${s.items.length} latestText_len=${s.latestText?.length ?? 0} ` +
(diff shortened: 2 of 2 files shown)


mekenthompson (commented):
## Review: diagnostic logs for issue #81

**Overall:** The logging is well-targeted and the PR delivers exactly what it claims. Three call sites are correct, context is good, no behavior changes. The main concern is that the logs are always-on with no debug-level gate, which matters for the text-event path specifically. See findings below.

---

### What's good

- **Issue link present.** PR body refs #81, explains the hypothesis ("capture time vs parse time"), and sets expectations (follow-up PR after data is in hand). Exactly right for a diag-only PR.
- **Consistent style.** All three log lines use `process.stderr.write` with the `progress-card.diag:` prefix, consistent with the existing 14 `process.stderr.write` calls in `progress-card-driver.ts`. No `console.debug` or mixed patterns.
- **Structured log format.** Every line is key=value pairs: `chatId=`, `turnKey=`, `narrative_count=`, etc. Easy to grep and parse programmatically.
- **Call sites are correct.** The three named sites (text-event after reduce, render branch on first/final emit, prose-recovery hit/miss) are the right places to instrument for this bug.
- **No sensitive payload leak (mostly).** The render-branch log emits only counts and lengths (`narratives=N items=N latestText_len=N`), not content. `chatId` is a Telegram numeric ID — not PII-sensitive in a server log.
- **Prose-recovery miss branch added.** The existing code already logged the hit path; the new `else` for the miss case is the key addition and it's correct.
- **Frequency is bounded on the render-branch log.** Gating on `forceDone || lastEmittedHtml == null` means it fires at most twice per turn (first emit + terminal flush), not on every intermediate edit. Good.

---

### Findings

**1. No debug gate — always-on logs (moderate)**

None of the three log sites are gated by a debug flag or log level. The codebase has no `LOG_LEVEL` / `DEBUG` env-var mechanism; this follows the pattern of existing stderr logs (which are also always-on). That's acceptable for low-frequency events, but:

- The **text-event log** fires on every `text` SessionEvent that passes through ingest. A long agent turn emitting several assistant messages will produce one line per message. This is low enough volume that it's probably fine in practice, but it's worth noting. If the agent ever enters a pathological loop emitting many text blocks, this log will be proportionally noisy.
- The **render-branch log** fires at most ~2× per turn — no concern.
- The **prose-recovery log** fires at most once per turn (only when `currentTurnCapturedText` is empty) — no concern.

Recommendation: for the text-event site, consider gating behind `process.env.PROGRESS_CARD_DIAG === '1'` or similar, to allow silencing after the bug is diagnosed without a code change. Not blocking for a short-lived diag window, but worth doing before this sits in production for months.

**2. Preview field logs assistant text content (low, flag for awareness)**

In the text-event log:

```ts
const preview = last?.text ? last.text.slice(0, 60).replace(/\n/g, ' ') : ''
```

`last?.text` is the narrative label extracted from the assistant's preamble (first line, up to 200 chars via `extractNarrativeLabel`). This is **assistant output**, not the user's prompt — so it's not user-PII. However it is conversation content appearing in `journalctl`. If this codebase treats journal logs as potentially observable by support staff, be aware that assistant preamble text lands there verbatim (up to 60 chars). Probably fine given the existing `userRequest` snippets already in the progress-card state, but flagging for explicit acknowledgement.

**3. Render-branch log condition comment is slightly misleading (minor)**

```ts
if (forceDone || chatState.lastEmittedHtml == null /* first emit */) {
```

The comment says "first emit" but `forceDone` is also true on intermediate terminal flushes (`completeTurnFully`, `closeZombie`, `maybeCompleteDeferredTurn`). The log will fire on each of those. That's actually the right behavior (you want to see the branch at the terminal state too), but the comment implies it's only for the first render. Suggest: `/* first emit or terminal flush */`.

**4. `narrativeCount` reads from `peek?.narratives` before `recoverProseFromProgressCard` runs (minor, correct)**

In `gateway.ts`, the diag reads:
```ts
const narrativeCount = peek?.narratives.length ?? 0
```
This is placed before `recoverProseFromProgressCard(peek)` is called. `recoverProseFromProgressCard` reads `peek` but doesn't mutate it, so `narrativeCount` is stable. No bug here — just noting it was worth verifying.

**5. No tests added (noted, not blocking)**

Pure logging additions are typically not unit-tested. The PR test plan calls for manual smoke verification (`journalctl` grep). Acceptable given the ephemeral nature of this PR.

---

### Recommendation

**Ship if you add a `PROGRESS_CARD_DIAG` env gate on the text-event path** (finding #1), or accept the minor noise and ship as-is given the expected short lifespan of these logs. Fix the comment on finding #3 as a drive-by. Everything else is non-blocking.

Issue #111, opened 2026-04-26
ci: bktec returns 'no files found' on certain include/exclude pattern combinations


mekenthompson:
## Summary

`.buildkite/pipeline.yml`'s `tests-core` step uses bktec v2.3.2 to shard tests across parallel jobs. During PR #110 rollout I observed bktec failing with:

> no files found with pattern \"...\" and exclude pattern \"...\"

across multiple configurations:

| Build | INCLUDE_PATTERN | EXCLUDE_PATTERN | Result |
|------:|-----------------|-----------------|--------|
| #57   | original 5 entries (incl. telegram-plugin) | `telegram-plugin/tests/history.test.ts` (1 entry) | bktec ran tests; failed on bun:test imports |
| #58   | same 5 | 13 comma-separated entries (~1KB) | "no files found" |
| #59   | 4 entries (telegram-plugin removed) | empty string | "no files found" |
| #60   | 4 entries | sentinel non-existent file | "no files found" |
| #61   | 4 entries | real file `telegram-plugin/tests/history.test.ts` (no longer in include scope) | "no files found" |

## Hypotheses (untested)

1. bktec validates exclude paths exist on disk and silently aborts on missing
2. bktec requires include pattern to match historical timing-data file set; changing include nukes the run
3. bktec's glob library breaks on long comma-separated exclude lists (>~100 chars?)
4. Empty `EXCLUDE_PATTERN` is interpreted as "exclude all"

## Reproduction

Download bktec v2.3.2 binary locally and exercise the include/exclude combinations above in a checkout of switchroom.

## Why this matters

Lint, plugin tests, and eval steps all work fine. The only blocker is `tests-core` (vitest-via-bktec). This blocks CI signal on every PR. Phase 1 (#110) was admin-merged because the failure was independent of the PR.

## Suggested fix paths (rank-ordered)

1. **Diagnose with bktec binary locally**, find the offending pattern combination, document the constraint, update pipeline.yml accordingly.
2. **Remove bktec entirely** — the else branch falls back to bun run test:vitest which uses vitest.config.ts excludes (those were updated in #110 and are correct). Loses parallelism but unblocks signal.
3. **Pin bktec to an older version** if v2.3.2 introduced a regression.

## Pre-existing baseline

This issue blocks CI on every PR currently — not just #110. Upstream main is also red on the same step.

mekenthompson:
## Diagnosis confirmed — root cause found empirically

[...]

Issue #107, opened 2026-04-26
ci(bktec): investigate exit-16 so we can re-enable parallel test sharding


mekenthompson:
## Observed

The \`Core tests\` step uses bktec (Buildkite Test Engine Client) for parallel test sharding. On every recent build it fails with:

\`\`\`
+++ Buildkite Test Engine Client: bktec v2.3.2

no files found with pattern "tests/**/*.test.ts,test/**/*.test.ts,experimental/**/*.test.ts,src/**/*.test.ts,telegram-plugin/tests/*.test.ts" and exclude pattern "telegram-plugin/tests/history.test.ts"

🚨 Error: The command exited with status 16
\`\`\`

The pattern is set in \`.buildkite/pipeline.yml\` as:

\`\`\`yaml
export BUILDKITE_TEST_ENGINE_TEST_FILE_PATTERN="tests/**/*.test.ts,test/**/*.test.ts,experimental/**/*.test.ts,src/**/*.test.ts,telegram-plugin/tests/*.test.ts"
\`\`\`

The comment in the pipeline already notes that bktec's glob library doesn't honor brace expansion — but the issue here is different: even without braces, no files are found.

## Possible causes

- bktec was updated to v2.3.2 and changed glob semantics
- Working directory mismatch (bktec runs at repo root but the agent's cwd is somewhere else)
- The double-comma-separated patterns confuse bktec — maybe needs a different separator

## Why now

Affects every PR + every main build. Plain-vitest fallback works (the pipeline supports it), but bktec is the default path. PR #102's CI also fails on this step.

## Quick mitigation

Add explicit \`cwd\` or check what bktec expects. Worst case, drop bktec entirely and fall back to plain \`bun run test:vitest\`.

mekenthompson:
Status update from [PR #128](https://github.com/switchroom/switchroom/pull/128):

bktec is now **disabled by default** in the pipeline. It's gated behind `SWITCHROOM_USE_BKTEC=1` (cluster env). All Core-tests runs since #65 exited with code 16 — the cause hasn't been root-caused. The PR makes main green via plain vitest while keeping the bktec wiring in place, with `BUILDKITE_TEST_ENGINE_DEBUG_ENABLED=true` already set inside the gated block so the next person to flip the flag captures the diagnostic output.

**Next action when picking this back up:**

1. Set `SWITCHROOM_USE_BKTEC=1` on a one-off branch (`ci/test-bktec`).
[...]

Issue #86, opened 2026-04-25
subagent-watcher: HTML escape bug breaks background-worker card (silent regression of #64)


mekenthompson:
## Summary
The pinned background-worker card from `subagent-watcher.ts` (PR #70 / closes #64) is **silently broken in production** because `renderWorkerCard` produces malformed HTML. Every send attempt throws a Telegram parse error and is silently caught — users see no sub-agent surface.

## Smoking gun

`telegram-plugin/subagent-watcher.ts:141`:
```ts
function formatDuration(ms: number): string {
  if (ms < 1000) return '<1s'
  // ...
}
```

The literal string `<1s` is concatenated unescaped into the HTML body at line 170:
```ts
lines.push(`\u{1F527} ${desc} · running · last activity ${ago} ago · ${w.toolCount} tools`)
```

Result over the wire: `…last activity <1s ago · 0 tools` — Telegram's HTML parser sees `<1s` as a tag-open and rejects the entire message:

```
telegram gateway: subagent-watcher card send failed:
  Call to 'sendMessage' failed!
  (400: Bad Request: can't parse entities: Unsupported start tag "1s" at byte offset 133)
```

This error appears **once per second** in clerk's gateway journal whenever any sub-agent is running, indicating the retry/throttle is firing repeatedly without escalation.

## Why this is bad

1. **Silent regression of #64.** PR #70 shipped the watcher infrastructure but its rendering produces invalid HTML on the most-common path (sub-agent just dispatched, last activity <1s ago).
2. **No user surface.** I dispatched 3 background sub-agents in this session (acpx research, boot-card fixes, progress-card logging). Zero pinned cards. The user observed: "Why is there no status card for that worker??"
3. **Retry storm.** Each registry update re-fires `editMessageText` / `sendMessage`, which fails identically. Visible as ~10 consecutive identical 400s in logs before throttle kicks in. Wasted Telegram API calls.
4. **Silent failure UX.** The `.catch(() => {})` style swallow path means users never see the error and have no signal that sub-agents even exist.

## Repro

1. Restart any switchroom agent (e.g. `switchroom agent restart clerk`).
2. From inside the agent's chat, ask Claude to do something that dispatches a sub-agent (`Agent()` with `run_in_background: true`).
3. Within ~1 second of dispatch, observe: no card appears in Telegram.
4. `journalctl --user -u switchroom-clerk-gateway.service` shows the parse-entity 400 errors.

## Fix scope

[...]

Issue #77, opened 2026-04-25
switchroom update: bun add -g blocks postinstall, leaves claude binary broken


mekenthompson:
## Bug

\`switchroom update\` (after PR #69) runs \`bun add -g @anthropic-ai/claude-code@latest\` to bump Claude CLI. But bun blocks postinstall scripts by default for global installs:

\`\`\`
Blocked 1 postinstall. Run \`bun pm -g untrusted\` for details.
\`\`\`

The \`@anthropic-ai/claude-code\` postinstall (\`node install.cjs\`) is what downloads the platform-native binary (~245MB). Without it, only a stub \`claude.exe\` (500 bytes) lands, and any new claude invocation fails with:

\`\`\`
Error: claude native binary not installed.
\`\`\`

Existing long-running claude processes keep working (they hold the old binary), but ANY restart goes to a broken claude install. So a fleet that runs \`switchroom update\` and then \`switchroom restart\` ends up with all agents dead.

## Repro

1. Run \`switchroom update\` against a fleet
2. Run \`switchroom restart <agent>\`
3. Agent's start.sh tries to spawn claude → "Error: claude native binary not installed."

## Fix

\`bumpClaudeCli\` in \`src/cli/update.ts\` should either:

- Pass \`--trust\` to bun: \`bun add -g --trust @anthropic-ai/claude-code@latest\` (if that's a valid flag)
- Or run \`bun pm -g trust @anthropic-ai/claude-code\` after the install to unblock + run the postinstall
- Or use \`npm install -g\` instead of \`bun add -g\` (npm auto-runs postinstalls)

Verify the fix with: after the bump, run \`claude --version\` and confirm a real version (not the stub error).

## Severity

High — first \`switchroom update\` run on a clean install would silently break the fleet on the next restart.

## Workaround

Operator must run \`bun pm -g trust @anthropic-ai/claude-code\` after \`switchroom update\` until the fix lands.

Caught during a manual \`switchroom update --no-restart\` run. The fleet was OK because no restart happened, but the next restart would have failed.

PR #128, opened 2026-04-27
fix(ci): bypass bktec + soft_fail Eval summary so main goes green

mekenthompson:
## Summary

Two fixes — verified green on [build #86](https://buildkite.com/ken-thompson/switchroom/builds/86):

### 1. Bypass bktec (default to plain vitest)

[Build #77](https://buildkite.com/ken-thompson/switchroom/builds/77) and every \`:vitest: Core tests\` run since #65 fails with **bktec exit code 16**. PR #120 (comma → brace alternation) didn't unblock it — the brace pattern fails just as silently as the comma form did. bktec wraps every non-exec error as exit 16 ([main.go logErrorAndExit](https://github.com/buildkite/test-engine-client/blob/v2.3.2/main.go)), so the real cause could be the file glob, the test-plan API call, or vitest argument construction. Without \`bktec --debug\` we can't tell which. Per the [zzglob docs](https://pkg.go.dev/drjosh.dev/zzglob), our pattern syntax should work.

- Gate bktec on \`SWITCHROOM_USE_BKTEC=1\` (default off). Re-enable later by setting that in cluster env.
- Default path: plain \`bun run test:vitest\`.
- Drop \`parallelism: 2\` — only existed for bktec sharding.

### 2. soft_fail Eval summary + gate scripts with \`|| true\`

The bktec fix exposed a downstream issue. With Core tests now passing, the eval steps run, soft-fail (allowed), and then the **Eval summary** step hard-failed in 1 second. Both \`annotate-evals.sh\` and \`publish-badges.sh\` are documented best-effort with \`exit 0\` at the end, but they're invoked from a YAML \`command: |\` block where buildkite's default shell sets \`-e\`, so any non-zero exit aborts the step.

- \`soft_fail: true\` on the Eval summary step.
- \`|| echo "... (continuing)"\` on each script invocation so an annotate failure doesn't drop the publish-badges call.

## Build #86 final state (this PR)

| Step | Outcome |
|---|---|
| :pipeline: Upload | passed |
| :lock: Type check | passed |
| :vitest: Core tests | **passed** ← was the blocker |
| :telegram: Plugin tests (402) | passed |
| :compass: Skills evals — trigger routing | soft_failed (allowed) |
| :sparkles: Skills evals — quality | soft_failed (allowed) |
| :bar_chart: Eval summary | passed |

GitHub status check: **SUCCESS**

## Follow-ups

- Capture \`bktec --debug\` on a one-off branch with \`SWITCHROOM_USE_BKTEC=1\` to diagnose the original exit-16 cause.
- Investigate why \`Skills evals — quality\` is soft-failing — possibly real eval flakiness, possibly a credential issue on fork PRs.

Refs [#107](https://github.com/switchroom/switchroom/issues/107), [#111](https://github.com/switchroom/switchroom/issues/111)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -97,6 +97,9 @@ steps:
       # to capture debug logs to diagnose. The bktec block below is wired
       # to capture those logs the next time someone flips the env flag —
       # see CHECKLIST in the comment block immediately below.
+      # bktec disabled by default (exit-16 on every Core-tests run since
+      # build #65; tracking issue #107). Re-enable with SWITCHROOM_USE_BKTEC=1
+      # in the cluster env once you're ready to capture --debug logs.
       if [[ "$$SWITCHROOM_USE_BKTEC" == "1" ]]; then
         # ─── bktec re-enable checklist (skills/buildkite-test-engine) ───
         # 1. SWITCHROOM_TEST_ENGINE_API_TOKEN cluster secret must exist with
@@ -118,6 +121,11 @@ steps:
         #    `custom` with a TEST_CMD that substitutes {{testExamples}}.
         #    The custom runner doesn't need RESULT_PATH (per bktec's
         #    config validation in internal/config/validate.go).
+        # 6. **Restore `parallelism: 2` on this step** when re-enabling.
+        #    bktec aborts with "BUILDKITE_PARALLEL_JOB_COUNT was 0" without
+        #    it. The directive was removed in this PR because plain vitest
+        #    runs the suite once, so duplicating jobs would just waste
+        #    compute.
         export BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN="$$(buildkite-agent secret get SWITCHROOM_TEST_ENGINE_API_TOKEN 2>/dev/null || true)"
         if [[ -n "$$BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN" ]]; then
           export BUILDKITE_TEST_ENGINE_DEBUG_ENABLED="true"
@@ -287,9 +295,13 @@ steps:
     # a public badge gist. Neither is load-bearing — every status that
     # matters (build, type-check, tests, evals) already shows on the build
     # page. A red Eval summary should never fail the build, so soft_fail
-    # any exit_status >= 1 (the publish-badges curl PATCH can 404 the gist
-    # on fork PRs that don't have access to the GIST_TOKEN cluster secret).
-    soft_fail: true
+    # the exit codes the scripts can plausibly produce (1 from a script
+    # failure, 2 from buildkite-agent secret/artifact errors). We
+    # deliberately do NOT mask agent loss (-1) or signal kills (143) so
+    # those infrastructure failures still bubble up.
+    soft_fail:
+      - exit_status: 1
+      - exit_status: 2
     command: |
       mkdir -p evals/results
       # Download to "." — buildkite-agent preserves the artifact's original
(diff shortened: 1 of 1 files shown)


mekenthompson (commented):
**Overall:** This is an intentional workaround to make main green — it is not a fix, and should be treated as such.

---

**What's good**

- The bypass is explicit and opt-in, not a silent removal. `SWITCHROOM_USE_BKTEC=1` leaves the bktec path intact and reachable; it is not deleted.
- The bktec block inside the flag is substantially improved over the pre-PR version: `DEBUG_ENABLED=true` is wired, the self-download of v2.3.2 is removed (defers to the hosted agent's pre-installed binary — correct), and both token-missing and binary-missing cases fall back cleanly.
- The soft_fail on Eval summary is tightly scoped: it applies only to that one step, not to the eval trigger/quality steps (which retain specific `exit_status: 1` soft_fail). The `|| echo` guards are correct reasoning — `set -e` was the actual culprit.
- The checklist comment block inside the bktec flag (`# ─── bktec re-enable checklist ───`) is well-written and directly useful for whoever re-enables it.
- The PR description is honest: it names this a workaround, lists the follow-ups, and references the relevant issues (#107, #111).

---

**Issues and concerns**

- **Bypass is permanent by default.** The `SWITCHROOM_USE_BKTEC=1` flag defaults to off everywhere unless someone explicitly sets it in the cluster environment. There is no plan, deadline, or follow-up ticket to actually flip that flag and diagnose bktec. Issue #107 (the tracking issue for the original failure) is still open but has zero comments and no assignee. This pattern ("make green now, re-enable later") has a strong gravitational pull toward "later = never". At minimum, #107 should be updated to reflect this PR's outcome and the specific next action (run a one-off branch with the flag + `--debug`).
- **No inline comment in pipeline.yml explaining the bypass.** The new comment block says *how* to re-enable bktec (checklist), but there is no short "WHY this is disabled by default" comment at the `if [[ "$$SWITCHROOM_USE_BKTEC" == "1" ]]` line itself. A future contributor reading this cold will see the condition and wonder if it is safe to change. The PR description explains the history but PR descriptions are not read when editing YAML six months from now. Suggest: one line above the if-block, e.g. `# bktec disabled by default (exit-16 on every run since build #65; see issue #107)`.
- **`parallelism: 2` is gone.** That is correct given bktec is off. But if someone re-enables bktec without also restoring `parallelism`, bktec will abort with "BUILDKITE_PARALLEL_JOB_COUNT was 0" — the exact failure the original comment described. The re-enable checklist does not mention this. Add it as checklist item 6.
- **bktec validation is what is being bypassed, not just sharding.** bktec does two things: shard test files across parallel jobs (performance) and feed timing data back to the Test Engine suite (analytics). Skipping it means the suite's timing data will stale out, making future re-enablement less useful (cold timing = file-count sharding, not balanced sharding). Low risk but worth noting.
- **`soft_fail: true` (unscoped) vs the other eval steps' `soft_fail: - exit_status: 1` (scoped).** The existing eval steps use the scoped form; this PR uses the unscoped `soft_fail: true` on Eval summary. That means *any* exit status — including signal-killed jobs or agent loss (exit -1, 143) — is silently masked. For a best-effort annotation step this is probably acceptable, but it is inconsistent with the rest of the pipeline and could hide infrastructure problems. Consider `soft_fail: [{exit_status: 1}, {exit_status: 2}]` to match the pattern elsewhere.
- **Blast radius check:** No other steps were reordered, no env vars were added to the global env block, and no new steps were introduced. The diff is confined to the `tests-core` command block and the `Eval summary` step. Blast radius is as-described in the title.

---

**Recommendation**

Ship this as a temporary workaround with the following conditions:

1. Add one inline comment at the `SWITCHROOM_USE_BKTEC` check explaining the bypass and the issue number (`# disabled — exit-16 since build #65, see #107`).
2. Update issue #107 with the current status: bktec is now opt-in, next step is a dedicated debug branch with the flag + `DEBUG_ENABLED=true`.
3. Add `parallelism: 2` restoration to the bktec re-enable checklist (item 6 or a note at the top).
4. Optionally tighten `soft_fail: true` to scoped exit statuses on Eval summary to match pipeline style.

Without #1 and #2, this workaround will sit indefinitely and the CI comment history will be the only record of why it exists.

Issue #129, opened 2026-04-27
vault-broker: simplify + harden (6 items from review)


mekenthompson:
Tracking issue for the items from a review of the recently-shipped vault-broker stack ([PR #112](https://github.com/switchroom/switchroom/pull/112), [#113](https://github.com/switchroom/switchroom/pull/113), [#117](https://github.com/switchroom/switchroom/pull/117)).

Lens: **switchroom should be simple and reliable on a single-user Linux home server.** A few parts of the broker stack earn their complexity (encrypted vault file, materialize-to-temp, e2e harness). Others don't. This issue tracks the work to keep what earns its keep and trim or harden the rest.

## The work

- [ ] **Item 1 — Reframe `schedule.secrets[]` in docs as misconfig protection, not a security boundary.** The ACL prevents a typo letting cron-A read cron-B's keys. It doesn't prevent attack — anyone who can edit cron scripts can also edit the config. Doc-only change.
- [ ] **Item 2 — Auto-start broker via systemd user unit with `Restart=on-failure`.** Today the broker is started by hand with `switchroom vault broker start`. If it dies at 3am, every cron until the next manual unlock either falls back to direct decrypt (often missing in cron context) or fails. Generate `switchroom-vault-broker.service`.
- [ ] **Item 3 — Drop the `/proc/<pid>/exe` interactive fallback in `acl.ts`.** The CLI already has the passphrase from the user; it doesn't need the broker. Removes ~30 LOC and a symlink-fragile equality check.
- [ ] **Item 4 — Surface broker error codes to the CLI.** Today `getViaBroker()` returns `null` for \"broker unreachable\", \"ACL denied\", and \"key not found\" — three very different conditions that get the same UX.
- [ ] **Item 5 — Fail-loud on non-Linux, or document Linux-only.** Current behaviour: silent stderr warning (`peercred ACL is disabled`) on macOS / WSL and the broker keeps running. README markets compliance-by-design; this silent downgrade should be explicit.
- [ ] **Item 6 — Replace `ss -xpn` parsing with `SO_PEERCRED`.** Today's peercred shells out to `ss`, parses the table, walks `/proc/<pid>/cgroup`. The source self-documents the concurrency limitation (\"picks the first connected entry\"). `SO_PEERCRED` via FFI is the kernel API for this. Removes ~150 LOC of regex parsing, fixes the concurrency hazard, drops the runtime dep on iproute2.

## Out of scope here
[...]

PR #130, opened 2026-04-27
fix(vault-broker): simplify + harden (items 1-5 of #129)

mekenthompson:
Closes items 1-5 of [switchroom#129](https://github.com/switchroom/switchroom/issues/129). Five focused changes from a review of the recently-shipped vault-broker stack ([PR #112](https://github.com/switchroom/switchroom/pull/112), [#113](https://github.com/switchroom/switchroom/pull/113), [#117](https://github.com/switchroom/switchroom/pull/117)). Lens: switchroom should be simple and reliable on a single-user Linux box.

Item 6 (replace \`ss -xpn\` parsing with \`SO_PEERCRED\` via FFI) lands separately because it introduces a native dep.

## Item 1 — Reframe \`schedule.secrets[]\` in docs

The per-cron \`secrets:\` allowlist prevents typos (cron-A reading cron-B's keys) and makes secret surface explicit at config-review time. It does **not** prevent attack: anyone who can edit cron scripts can also edit \`switchroom.yaml\` to grant any keys. Documented honestly in [docs/configuration.md](docs/configuration.md) and the schema \`describe()\`. No code change.

## Item 2 — Auto-start the broker + fix a double-prefix bug

The broker systemd unit existed but wasn't actually installed correctly: \`installUnit(\"switchroom-vault-broker\", ...)\` wraps another \`switchroom-\` prefix and writes to \`switchroom-switchroom-vault-broker.service\` — a name no cron timer references. Fixed.

Also added \`startBrokerUnit()\` called from \`installAllUnits()\` when the broker was just installed, so a fresh \`switchroom agent create\` ends up with the broker actually running rather than enabled-but-still-needs-manual-\`systemctl start\`. Best-effort: stderr note on failure.

## Item 3 — Drop the \`/proc/<pid>/exe\` interactive fallback

The interactive ACL fallback compared \`peer.exe\` to \`bunBinDir/switchroom\`. Symlink-fragile, easy to bypass, and unnecessary — the CLI already has the user's passphrase and can read the vault file directly via \`vault get --no-broker\`. Removed the whole fallback path, the \`vault.broker.allow_interactive\` config field, and the \`AclOpts\` test plumbing. The broker now serves cron units only.

Bonus: tightened the deny reason for \"key not in ACL\" — the previous message echoed the allowed-keys list, leaking the cron allowlist via the protocol. Added a regression test.

## Item 4 — Surface broker error codes to the CLI

\`getViaBroker()\` returned \`null\` for three different conditions (unreachable, denied, not-found) and the CLI showed the same UX for all of them. New \`getViaBrokerStructured()\` returns a discriminated union:

\`\`\`ts
type GetResult =
  | { kind: \"ok\"; entry }
  | { kind: \"unreachable\"; msg }
  | { kind: \"denied\"; code; msg }
  | { kind: \"not_found\"; code; msg };
\`\`\`

\`vault get\`'s broker path uses it: a denied request prints the code+reason and falls through to direct decrypt for TTY callers (fails fast for cron); a not-found exits 1 without falling back; unreachable surfaces the underlying errno so cron logs explain why.

Legacy \`getViaBroker()\` kept for the resolver path that doesn't need the discrimination.

## Item 5 — Fail-loud on non-Linux

On macOS / WSL the broker would print a stderr WARN and keep running with mode-0600 as the only access control. The broker's whole point is the cgroup ACL, which is Linux-only — running it without the ACL is a meaningful semantics change that should be opt-in, not the default.

\`VaultBroker.start()\` now refuses to bind on non-Linux unless \`SWITCHROOM_BROKER_ALLOW_NON_LINUX=1\` is set. The error message is actionable. The escape hatch keeps tests working on whatever CI / dev box uses.

## Test plan
- [x] \`npm run lint\` passes
- [x] [src/vault/broker/acl.test.ts](src/vault/broker/acl.test.ts) — 11 cases incl. new \"deny doesn't leak allowlist\" + \"non-cron caller denied\"
- [x] [src/cli/vault-get-broker.test.ts](src/cli/vault-get-broker.test.ts) — new structured-result + non-Linux refusal cases
- [x] [src/agents/cron-broker.test.ts](src/agents/cron-broker.test.ts) — 15/15 pass (broker unit generation untouched)
- [x] [src/config/schema.test.ts](src/config/schema.test.ts) — broker defaults updated for removed \`allow_interactive\` field
- [ ] CI green
- [ ] Manual: install on a Linux box, confirm \`switchroom-vault-broker.service\` ends up running after \`agent create\`

## Notes for review
- This deletes some test cases (the interactive-fallback ones) — that's intentional, not a coverage regression. The behavior they covered no longer exists.
- The double-prefix bug fix (Item 2) means existing installs may have a stale \`switchroom-switchroom-vault-broker.service\` file alongside the new correct one. Worth a quick \`switchroom doctor\` cleanup pass in a follow-up — flagging here so it doesn't get lost.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -105,6 +105,25 @@ Materialized `kind="files"` dirs land under `$XDG_RUNTIME_DIR/switchroom/vault/<
 
 Manage entries with `switchroom vault set <key>`, `switchroom vault get <key>`, and `switchroom vault list`. Multi-line string values are preserved verbatim via piped stdin or `--file <path>`; file-kind entries are set programmatically via `setFilesSecret` (a CLI surface for multi-file set is tracked separately).
 
+### Vault broker (Linux only)
+
+For scheduled tasks that need vault access, switchroom can run a long-lived **vault broker** daemon that holds the decrypted vault in memory after a one-time passphrase entry. Cron scripts then ask the broker for keys instead of prompting for the passphrase on every run. The broker is **Linux-only by design** — its access control relies on cgroup-based systemd unit identification, which doesn't exist on macOS / WSL. On non-Linux platforms `switchroom vault get` always reads the vault file directly with the user's passphrase.
+
+```yaml
+agents:
+  myagent:
+    schedule:
+      - cron: "0 8 * * *"
+        prompt: "morning briefing"
+        secrets: [google_calendar_token, weather_api_key]   # NEW
+```
+
+The `secrets:` array is **misconfiguration protection, not a security boundary**: it prevents a typo in cron-A from accidentally reading cron-B's keys, and it makes the per-cron secret surface area explicit at config-review time. It does not prevent attack — anyone who can edit cron scripts on the host can also edit `switchroom.yaml` to declare any keys, and anyone who has the vault passphrase can read the vault file directly. Frame it as: "the cron-A script that asks for `weather_api_key` was clearly meant to ask for it" — not "the cron-A script can't reach `bank_token` even if compromised."
+
+The broker is started/stopped via `switchroom vault broker {start,stop,status,unlock,lock}`. When `installAllUnits()` runs (called by `switchroom agent create` and similar), a `switchroom-vault-broker.service` user unit is installed with `Restart=on-failure`, so the broker auto-restarts if it crashes and auto-starts at user login.
+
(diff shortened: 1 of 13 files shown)


mekenthompson:
Superseded by #131 — that branch was stacked on this one and contained both sets of commits. Items 1-5 (this PR's content) plus item 6 are all in main now.

PR #133, opened 2026-04-27
test(vault-broker): _testIdentify seam for Linux gated-path coverage

mekenthompson:
## Summary

- Adds `_testIdentify` test hook on `VaultBroker` so unit tests can stub a synthetic `PeerInfo` and exercise the broker's gated `list`/`get` paths on **any** Linux runner — including hosted Buildkite Docker where systemd-user isn't available.
- Two new describe blocks in `server.test.ts`:
  - **allowed cron identity** — happy-path `list`/`get` with the broker treating the test client as `switchroom-myagent-cron-0.service` and matching ACL config.
  - **denied identity (peer=null)** — Linux deny path made explicit.

## Why

The broker's `list`/`get` ops are gated by peercred — on Linux, `identify()` reads `/proc` and `ss`/`SO_PEERCRED` to resolve the caller's systemd cron unit. Under `vitest` the test process isn't a switchroom cron unit, so unit tests can only ever exercise the **deny** path on Linux. Happy-path coverage was therefore non-Linux only (existing tests skip Linux), with real-cron coverage living in `tests/integration/vault-broker-e2e.test.ts` behind `INTEGRATION=1` + systemd-run.

This seam closes the gap: hosted Buildkite Docker now exercises both branches.

## Production codepath unchanged

`_testIdentify` defaults to undefined → broker calls the real `identify()`. The only production caller (`src/cli/vault-broker.ts:120`) constructs `new VaultBroker()` with no opts.

## Test plan

- [x] `npm run lint` passes
- [ ] CI green on Linux (the new `allowed cron identity` block runs there for the first time)
- [x] Existing `list LOCKED after lock` test still works (LOCKED is checked before peercred in `server.ts`)
- [x] Production caller in `src/cli/vault-broker.ts` is unaffected

### DIFF
--- a/src/vault/broker/server.test.ts
+++ b/src/vault/broker/server.test.ts
@@ -38,6 +38,19 @@ const TEST_SECRETS: Record<string, VaultEntry> = {
   },
 };
 
+/**
+ * Deep-clone TEST_SECRETS for each broker. `broker.lock()` mutates entry
+ * values in place (zeros them as a best-effort wipe before GC), so a shallow
+ * copy `{ ...TEST_SECRETS }` would leak that mutation across tests via the
+ * shared entry objects. Concretely: once the "lock wipes secrets" test runs,
+ * subsequent tests that read `foo` get `value: ""` instead of `bar-value`.
+ * The Linux-skipped get tests masked this for a long time; the new
+ * `_testIdentify` happy-path tests below run on Linux and surface the issue.
+ */
+function cloneSecrets(): Record<string, VaultEntry> {
+  return JSON.parse(JSON.stringify(TEST_SECRETS));
+}
+
 // Minimal SwitchroomConfig for broker tests. On Linux the broker uses
 // peercred + ACL to identify cron units; the test process isn't one, so
 // `get` requests are denied. ACL behavior is covered by acl.test.ts; here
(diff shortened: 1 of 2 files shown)


mekenthompson:
**Reviewed and addressed.**

Reviewer flagged two items:
1. **Must-fix:** test failure on shared `VaultEntry` mutation by `lock()` between describe blocks. Already addressed by your `21ba267` follow-up commit (different mechanism — clearing the ACL gate first — but it works; all 15 server tests green locally including the previously failing one).
2. **Should-fix (security defence-in-depth):** `BrokerTestOpts` is publicly exported, and the JSDoc "DO NOT set outside tests" was advisory only. A production caller could in principle pass `_testIdentify` to forge any `PeerInfo` (e.g. an allow-all one), or `_testSecrets` to skip the passphrase/KDF. None of the current production callers do, but a future regression could.

Pushed `1c16a92` adding a runtime guard: constructor throws when any test opt is set without `NODE_ENV=test`. vitest sets that automatically; production builds do not. All 73 broker tests still pass; lint clean; build clean; PII grep at baseline (4 `/home/hindsight/.pg0`).

Optional nits the reviewer flagged (test helper placement, hardcoded PID, skip semantics on non-Linux) are non-blocking — leaving for a follow-up if you want them tightened.

Merging.

Issue #142, opened 2026-04-27
telegram surfaces: quiet, contextual cards (boot + greeting + sub-agent unification)


mekenthompson:
## Problem

Telegram surfaces are noisy and redundant. Two examples Ken called out today:

**Restart:** every restart posts TWO cards — `🎛️ Switchroom boot` (from `telegram-plugin/gateway/boot-card.ts`) and `🎛️ Switchroom · <agent> online` (from `buildSessionGreetingScript` in `src/agents/scaffold.ts`). Together ~15 rows of mostly stable config. Many rows are baked-in at scaffold time (see #60). The transient also produces false-positive reds (systemd `deactivating`, crons mid-re-register) that self-heal in seconds.

**Per turn:** dispatching a sub-agent surfaces it in THREE places — the in-card `[Sub-agents · N running]` block, a `🛠️ Worker dispatched` 1-line reply, and a pinned `🛠️ Background workers (N)` card.

## JTBD framing

The cards exist to answer:

1. *"Did the agent come back up correctly?"* — restart confirmation.
2. *"Can I trust the next reply?"* — auth/plan/memory health.
3. *"What's happening for the request I just sent?"* — contextual progress.

The current implementations also answer non-JTBD: *"What's the full config of this agent?"* (Profile/Tools/Skills/Limits — stable; doesn't change between restarts) and *"What's running across all my requests?"* (we're a single sequential-turn user; there is no "all").

### 1. Restart card — quiet by default

Replace both startup cards with **one** card.

Default state: a single line. e.g. `✅ klanker back up · v0.3.0+43`.

Adds a row **only** when something is genuinely degraded:
- Auth expired or expiring soon
- Rate limited / over plan
- Hindsight unreachable
- Crash recovery (vs planned restart)
- Version drift vs last-seen

Probes fire after a **settle window** (~5–7s, or "first poll succeeded + first timer fired"). Transient reds during the restart window never surface — no `deactivating`, no `crons not registered`. Falls out of design, not a config knob.

Subsumes #60 (no more baked-in rows). Fixes gymbro's false-positive red.

Code touches:
- `telegram-plugin/gateway/boot-card.ts` — reduce to one-line "ready" ack or delete
- `telegram-plugin/gateway/boot-probes.ts` — keep the probe logic, gate on settle
- `src/agents/scaffold.ts` `buildSessionGreetingScript` (lines 317–1066) — delete; replace with the unified card path
[...]

Issue #93, opened 2026-04-25
boot-card: post on every gateway start (planned, crash recovery, fresh) with restart reason


mekenthompson:
## Current behavior
Boot card (#78) only fires when a restart-marker is present — i.e. when the dying gateway wrote a marker for an active chat. Without an active chat or planned-restart marker, the gateway boots silently with `boot: no restart-marker found (clean start or crash recovery path)`. Same gate the legacy banner used.

## Desired behavior (per Ken)
> "Any time an agent restarts, starts, recovers, I want that new start message. Not the old one. The things to be flagged are restart reason though, but in that same card."

Every gateway boot should post the live boot card. The boot type (planned / crash / fresh / OOM-recovery) goes inside the card as a tag/row.

## Why
Operator visibility. Right now a silent gateway crash + auto-restart leaves no trace in Telegram (see related lawgpt-gateway crash bug). Even a fresh first-boot of a brand-new agent gets no surface — the user has to check `journalctl` to know it came up.

## Design sketch
- Decouple "post the boot card" from `readRestartMarker()`. Always post.
- Determine **target chat** from a fallback chain:
  1. Restart marker `chat_id` (if marker present)
  2. Configured `SUBAGENT_OWNER_CHAT_ID` env (if set)
  3. Most recent inbound chat from local SQLite history (if available)
  4. Skip silently if nothing is known (cron-only agent with no DM history)
- Determine **restart reason**:
  - Marker present + recent → "planned restart"
  - No marker, but `clean-shutdown.json` recent → "graceful shutdown + restart"
  - No marker, no clean-shutdown → "crash recovery"
  - No prior gateway state at all → "fresh start"
  - Add as a top-line row in the card: e.g. `🔄 <b>Restart</b>  planned · 0.3s ago`
- Card title can stay `🎛️ Switchroom boot` — the reason row carries the meaning.

## Implementation notes
- The restart-marker writer (`writeRestartMarker`) already runs on graceful shutdown and stamps `chat_id`/`thread_id`/`ack_message_id`. Crash exits don't write it — that's the gap to cover.
- Add a `clean-shutdown.json` write in the SIGTERM handler (separate from the active-chat marker) so we can distinguish planned vs crash.
- For the "no known chat" case, decide: skip (current behavior, less spam) vs post to the agent's primary topic (configurable).

## Acceptance
- Restart lawgpt with no active chat → boot card appears in lawgpt's primary topic with reason "planned restart"
[...]

Issue #43, opened 2026-04-24
bug(progress-card): duplicate pins across turns when sub-agents orphan; local .active-pins.json diverges from Telegram state


mekenthompson:
## Summary

Two separate progress cards were pinned to the same Telegram chat (`8248703757`, clerk's DM) for what should have been a single user turn. Both cards visually present as "pinned" in Telegram simultaneously; local state in `.active-pins.json` only tracks one of them. Telegram and the gateway's local pin registry have diverged.

Screenshot (attached by the user in the incident report) showed two "Working…" cards quote-replying to the same "You do it" inbound message, with elapsed times `01:53` and `03:14`, visible side-by-side in the chat pane and both listed in the pinned-message bar.

Related: #31 (orphaned sub-agent correlation deferring `turn_end`). This is a distinct symptom of the same underlying orphan-correlation fragility.

## Evidence from `clerk/telegram/gateway.log`

Relevant turn sequence around the incident (turns 20–23 on `chatId=8248703757`):

```
progress-card: pinned turnKey=...:20 msgId=5182
progress-card: sub_agent_started agentId=a0ace60f65e2b6c98 correlated=orphan
progress-card: turn_end flush turnKey=...:20
progress-card: turn_end deferred turnKey=...:20 reason=in-flight-sub-agents n=1 agentIds=[a0ace60f65e2b6c98]
progress-card: onTurnComplete firing turnKey=...:20
progress-card: onTurnComplete callback turnKey=...:20
progress-card: unpin turnKey=...:20 msgId=5182              ← logged, but did Telegram honor it?
progress-card: pinned turnKey=...:21 msgId=5188
progress-card: sub_agent_started agentId=a7464bf44fca39ac9 correlated=yes
progress-card: turn_end flush turnKey=...:21
progress-card: unpin turnKey=...:21 msgId=5188
progress-card: pinned turnKey=...:22 msgId=5192
progress-card: sub_agent_started agentId=aba292ed2d11e5cfe correlated=orphan
progress-card: turn_end flush turnKey=...:22
progress-card: turn_end deferred turnKey=...:22 reason=in-flight-sub-agents n=1 agentIds=[aba292ed2d11e5cfe]
progress-card: onTurnComplete firing turnKey=...:22
progress-card: onTurnComplete callback turnKey=...:22
progress-card: unpin turnKey=...:22 msgId=5192              ← same pattern
progress-card: pinned turnKey=...:23 msgId=5197
```

[...]

Issue #31, opened 2026-04-24
bug(progress-card): stale pins not unpinned — background sub-agent correlation orphans the turn_end defer


mekenthompson:
## Summary

Progress card pins on DM chat `8248703757` (clerk agent) are staying pinned indefinitely after the user-visible reply for the turn has been finalized. Three pins are currently in `.active-pins.json` that should have been closed — the oldest is ~10 hours stale.

This is the live production repro of the bug that stale PR `mekenthompson/clerk#34` was trying to fix. That PR is being reapplied to canonical — tracking branch: `fix/progress-card-stale-messageid` (see linked PR once opened).

### `.active-pins.json` snapshot (clerk agent, 2026-04-25 06:16 AEST)

```json
[
  {"chatId":"8248703757","messageId":5113,"turnKey":"8248703757:4","pinnedAt":1777027479288},
  {"chatId":"8248703757","messageId":5126,"turnKey":"8248703757:7","pinnedAt":1777061476078},
  {"chatId":"8248703757","messageId":5130,"turnKey":"8248703757:8","pinnedAt":1777061793840}
]
```

Three pins, three turnKeys. Turn 4 is from 2026-04-24 10:28 — the reply was finalized at that time but the pin was never unpinned.

### Log evidence from `clerk/telegram/gateway.log` (turn 7 — the clearest smoking gun)

```
progress-card: scheduled initial-delay timer turnKey=8248703757:7 delay=30000ms
progress-card: initial-delay timer fired turnKey=8248703757:7
progress-card: pinned turnKey=8248703757:7 msgId=5126
progress-card: tool_use → agent correlation toolUseId=toolu_01Eyreikmv3Dhx2RcgZVsdmw agentId=pending (awaiting sub_agent_started)
progress-card: sub_agent_started agentId=abf63158dd5c4a04a correlated=orphan      ← BUG VECTOR
reply: finalized chatId=8248703757 messageIds=[5128]
progress-card: turn_end flush chatId=8248703757 turnKey=8248703757:7
progress-card: turn_end deferred turnKey=8248703757:7 reason=in-flight-sub-agents n=1 agentIds=[abf63158dd5c4a04a]
```

[...]

PR #144, opened 2026-04-27
refactor(telegram): consolidate sub-agent visibility into progress card (#142 slice 1)

### LINKED ISSUE #60: greeting: replace baked-in version/PR/age text with live probe at SessionStart
## Symptom

The session-greeting card pasted to Telegram on SessionStart shows the **Version** / **Restarted** / **Memory** rows as text that was generated **at the time of the last \`switchroom reconcile\`**, not at the time the greeting actually fires. Spotted live today: an agent's greeting reported \`Version  v0.3.0 · #57 · 21m ago\` even though the running gateway had been started from source files 2+ hours earlier and the on-disk source had since received an additional merged commit (\`ceea0fd\`, post-v0.3.0). The age (\`21m ago\`) and PR badge (\`#57\`) are therefore stale by definition — they encode whatever the source build-info reported during scaffold/reconcile.

## Where it lives

\`src/agents/scaffold.ts\` → \`buildSessionGreetingScript\` writes a bash script with the version row computed at scaffold time. The script is then executed by Claude Code's \`SessionStart\` hook — but the version string is pre-rendered text, not a live read.

\`src/agents/scaffold.ts:1093\` — \`formatVersionRow()\` reads \`VERSION\` / \`COMMIT_DATE\` / \`LATEST_PR\` / \`COMMIT_SHA\` / \`COMMITS_AHEAD_OF_TAG\` from \`src/build-info.ts\` (auto-generated, last refreshed at \`npm run build\`).

The same problem affects any other \"liveness\" claims in the greeting — they're inlined at scaffold time, then become whatever was true *then*, not now.

## Why this matters

The greeting card is the user's primary signal for \"is the agent in a fresh, healthy state?\". When the rows lie about freshness, that signal is corrupted. A user looking at \`v0.3.0 · 21m ago\` reasonably concludes \"this agent is running v0.3.0 and was restarted 21m ago\" — but the truth is \"the greeting was scaffolded 21m ago; the running process started 2h+ ago and is on whatever state the source files were at that earlier process-start time\".

## What \"probed information\" should look like

The greeting bash script should compute its row values at execution time, not at generation time. Concretely:

- **Version**: read \`/proc/<gateway-pid>/cmdline\` to find the exact \`gateway.ts\` path the bun process is running, then \`git -C <repo> describe --tags --always\` for the actual version of the source HEAD that was loaded at process start. Or pin to a manifest written by the gateway on startup that records the SHA it loaded from. (The current static \`VERSION\` import only reflects whatever version was built into the *generating* CLI, not the *running* process.)
- **Restarted**: read the gateway process start time from \`/proc/<pid>\` (the script already does this for the dedupe optimization — surface that value as the \"Restarted\" row instead of the baked-in restart reason).
- **Memory**: a quick \`mcp__hindsight__get_bank_stats\` curl rather than a snapshot from scaffold time.
- **Auth**: probe the credentials file's mtime / expiry rather than pretending to know.

## Acceptance criteria

- Greeting script's row values are computed at exec time from the actual running process and live state.
- Re-running \`switchroom reconcile\` against an unchanged config doesn't change the greeting's *content* (it'll change the next time the script runs).
- Greeting still works when offline (no GitHub probe needed) — version comes from local git or a startup manifest, not a network call.
- Existing \`SWITCHROOM_EVAL_MODE=1\` early-exit and the gateway-start-time dedupe behavior are preserved.

## Priority

P3. Not data-loss, but a trust-erosion bug — the greeting card is supposed to ground the user, and currently it can mislead them about which version is actually running. Pairs with #56 (gateway silently stalls) — both are about \"can the user trust the surface signals\".

## Repro

1. Have an agent scaffolded with the source clone at git tag \`vX.Y.Z\`.
2. Pull additional commits to the source clone after process start (without restarting the gateway / reconciling).
3. \`switchroom restart\` the agent — the SessionStart greeting fires.
4. Observe: \"Version vX.Y.Z\" stamped, even though the just-restarted process loaded the *new* source files which are past \`vX.Y.Z\`.

(Today's example: greeting said \`v0.3.0 · #57\` while \`git describe --tags upstream/main\` returned \`v0.3.0-1-gceea0fd\`.)

### LINKED ISSUE #87: sub-agent activity invisible after parent turn ends (no progress card, no inline notification)
## Symptom

Klanker dispatched a worker sub-agent (file-edit task, expected ~10–15 min). User got a `stream_reply done=true` reply confirming dispatch, then **complete silence** for ~10+ min while the worker ran. No progress card. No "worker still running" indicator. No notification when worker finishes. From the user's POV, the agent went dark.

## Repro

1. Parent agent calls `Agent` tool to dispatch a worker
2. Parent agent immediately follows with `stream_reply` done=true
3. Worker continues running in the background for minutes
4. Telegram has zero visibility into worker progress until it completes (and even completion notification is gated, see below)

## Data — actual log trace

`klanker/telegram/gateway.log`, turn `8248703757:1` (worker dispatch turn at ~22:14:39 UTC):

```
progress-card: scheduled initial-delay timer turnKey=8248703757:1 delay=30000ms
progress-card: initial-delay timer fired turnKey=8248703757:1
progress-card: pinned turnKey=8248703757:1 msgId=2093
progress-card: pin OK turnKey=8248703757:1 msgId=2093
ipc: tool_call tool=get_recent_messages
ipc: tool_call tool=react
progress-card: tool_use → agent correlation toolUseId=toolu_01T7QBgAfdnGByE7hcjMsZGG agentId=pending (awaiting sub_agent_started)
progress-card: sub_agent_started agentId=a051f32494f6e8c7e correlated=orphan pendingSpawns=0 promptSnip="CRITICAL bug fix in..." — NOTE: orphan sub-agents no longer gate parent turn_end defer (#31 fix)
ipc: tool_call tool=stream_reply
progress-card: forceCompleteTurn turnKey=8248703757:1 (external completion signal, e.g. stream_reply done=true)
progress-card: onTurnComplete firing turnKey=8248703757:1
progress-card: unpin turnKey=8248703757:1 msgId=2093
```

So:
- Progress card existed (msgId=2093), pinned at +30s, unpinned when parent called `stream_reply done=true`
- Sub-agent fully visible at IPC layer (`sub_agent_started`, agentId, prompt snippet)
- After parent turn ended, **nothing** else in the log — no further Telegram activity for the orphan worker

## Root causes

**1. Progress-card lifecycle is parent-turn-scoped only.** Per `progress-card driver`, `forceCompleteTurn` fires on `stream_reply done=true` and unpins. There is no handoff to a "sub-agent still running" card. Orphan sub-agents fall off the surface.

**2. Subagent-watcher is silently disabled.** PR #82 hotfixed the historical-replay bug by gating `startSubagentWatcher` on `SWITCHROOM_SUBAGENT_WATCHER=1` (default OFF). Currently the only mechanism that surfaces sub-agent activity to Telegram is gated off, and nothing logs that decision — `grep subagent-watcher gateway.log` returns zero lines, no banner like "watcher disabled because env var not set". Easy to think the watcher ran fine and decided to stay quiet.

**3. The `sub_agent_started → correlated=orphan` path has no Telegram fallback.** The note in the code (`orphan sub-agents no longer gate parent turn_end defer (#31 fix)`) is correct for orchestration semantics but means there's no surface for orphans either.

## Logging fidelity — what's missing for clean RCA

Existing coverage is reasonable for the progress-card lifecycle (`scheduled / fired / pinned / forceCompleteTurn / onTurnComplete / unpin`). Gaps:

- [ ] **Watcher startup decision banner.** Log unconditionally on gateway startup: `subagent-watcher: ENABLED (env=1)` or `subagent-watcher: DISABLED (set SWITCHROOM_SUBAGENT_WATCHER=1 to enable)`. Currently silent when off.
- [ ] **Orphan sub-agent surface decision.** When `sub_agent_started` arrives with `correlated=orphan` AND parent turn is already in `onTurnComplete` state, log: `sub-agent dispatched after parent turn ended — no Telegram surface (watcher=off | watcher=on, will surface)`.
- [ ] **Progress-card render-skip reasons.** Currently `fast-turn suppression turnKey=... (turn ended before initialDelayMs=30000ms)` is logged, but other skip paths (e.g. unpinned-while-still-active, sub-agent-attached-but-no-card-update) aren't. Worth a structured `card-decision` log line per state transition.
- [ ] **Parent-turn-vs-sub-agent timeline correlation.** Right now you have to mentally weave `turnKey=...:N` lines with `agentId=...` lines. A single `parentTurn=8248703757:1 subAgents=[a051f32494f6e8c7e:orphan]` line at turn-end would make this trivial.

## Proposed fix (separate from logging)

When the proper subagent-watcher fix lands (the snapshot-at-boot work that's in flight), default-enable the watcher again — that closes most of this gap. For full coverage, also: keep the progress card pinned (or re-pin a worker-card) while at least one orphan sub-agent is still running; unpin only when the last orphan returns.

## Related

- #64 (sub-agent visibility) — original feature ask
- #70 (PR landed initial watcher) — broke on startup-replay
- #82 (PR gated watcher off) — quick fix, this issue's proximate cause
- #81 (progress-card prose-vs-tools logging) — adjacent diagnostic gap
- #31 (orphan sub-agents no longer defer parent turn_end) — sets the "orphan is fine" semantics this issue rests on

mekenthompson:
## Summary

Slice 1 of #142 — collapse the three Telegram surfaces tracking sub-agent activity into one. The progress card's `[Sub-agents · N running]` block is the canonical surface; the `Worker dispatched` 1-liner and the pinned `Background workers (N)` card are deleted.

## What changed

**Deleted:**
- `renderWorkerCard()` and the pinned-card lifecycle in `telegram-plugin/subagent-watcher.ts`
- `updatePinnedCard` config field and `cardUpdateIntervalMs` knob
- The `Worker dispatched` notification fired on registration
- The gateway-side worker-card pin/edit/unpin handler (`gateway.ts` ~52 lines)
- `maybeSendCardUpdate` and all call sites

**Kept:**
- Historical-vs-active filter (#91) — still gates which entries fire stalls and completions
- `Worker idle` / `Worker done` notifications for state transitions
- The progress card's existing `pendingCompletion` lifecycle keeps the card live past parent turn-end (so post-turn sub-agent visibility is preserved without a separate pinned card)

**Net:** -364 lines / +33 lines across three files.

## Why

Per #142's JTBD framing, "what's running across all my requests?" isn't a real JTBD for a single-user, sequential-turn product — there's no "all". The contextual in-card tracker fully answers "what's happening for the request I just sent?". Two of the three surfaces were duplicate noise.

## Test plan

- [x] `bun test telegram-plugin/tests/subagent-watcher.test.ts` — 14/14 pass
- [x] `bun test telegram-plugin/tests/progress-card-pin-manager.test.ts` — 34/34 pass
- [ ] Manual: send a message that dispatches a sub-agent → verify only the progress card surfaces it; no `Worker dispatched` reply, no pinned worker card
- [ ] Manual: sub-agent outlives parent turn → progress card stays live (existing `pendingCompletion` behaviour)

## Refs

Refs #142. Closes #87.

Followups (separate PRs per #142 slicing):
- Slice 2: quiet-by-default restart card (replaces both startup cards, closes #60, refines #93)
- Slice 3: `/status` slash command for full config audit

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -5192,10 +5192,6 @@ void (async () => {
         if (streamMode === 'checklist') {
           const watcherAgentDir = resolveAgentDirFromEnv()
           if (watcherAgentDir != null) {
-            // Pinned worker card: one message per watcher session,
-            // edited in-place. Managed entirely by the watcher.
-            let workerCardMsgId: number | null = null
-
             subagentWatcher = startSubagentWatcher({
               agentDir: watcherAgentDir,
               sendNotification: (text: string) => {
@@ -5208,55 +5204,6 @@ void (async () => {
                 }).catch((err: Error) => {
                   process.stderr.write(`telegram gateway: subagent-watcher notification failed: ${err.message}\n`)
                 })
-              },
-              updatePinnedCard: (html: string | null) => {
-                const ownerChatId = loadAccess().allowFrom[0]
-                if (!ownerChatId) return
-                if (html === null) {
-                  // No active workers — unpin and delete the card
-                  if (workerCardMsgId != null) {
-                    const msgId = workerCardMsgId
-                    workerCardMsgId = null
-                    // Drop external-pin tracking before issuing the unpin so
-                    // a race with a concurrent service-message capture can't
-                    // re-resurrect the entry. Issue #94.
-                    pinMgr.untrackExternalPin(String(ownerChatId), msgId)
-                    void lockedBot.api.unpinChatMessage(ownerChatId, msgId).catch(() => {})
-                    void lockedBot.api.deleteMessage(ownerChatId, msgId).catch(() => {})
-                  }
-                  return
-                }
-                if (workerCardMsgId == null) {
-                  // Create a new pinned card
-                  void lockedBot.api.sendMessage(ownerChatId, html, {
-                    parse_mode: 'HTML',
-                    link_preview_options: { is_disabled: true },
-                    ...(TOPIC_ID != null ? { message_thread_id: TOPIC_ID } : {}),
-                  }).then((sent) => {
-                    workerCardMsgId = sent.message_id
-                    // Register the worker card with pinMgr BEFORE pinning so
-                    // its `captureServiceMessage` callback recognises the
-                    // service message Telegram emits in response and deletes
-                    // it (matching the existing main-card behaviour).
-                    // Issue #94.
-                    pinMgr.trackExternalPin(String(ownerChatId), sent.message_id)
-                    void lockedBot.api.pinChatMessage(ownerChatId, sent.message_id, {
-                      disable_notification: true,
-                    }).catch(() => {})
-                  }).catch((err: Error) => {
-                    process.stderr.write(`telegram gateway: subagent-watcher card send failed: ${err.message}\n`)
-                  })
-                } else {
-                  // Edit the existing card
-                  void lockedBot.api.editMessageText(ownerChatId, workerCardMsgId, html, {
-                    parse_mode: 'HTML',
-                    link_preview_options: { is_disabled: true },
-                  }).catch((err: Error) => {
-                    const msg = err instanceof GrammyError && err.error_code === 400 &&
-                      /message is not modified/i.test(err.description ?? '') ? '' : err.message
-                    if (msg) process.stderr.write(`telegram gateway: subagent-watcher card edit failed: ${msg}\n`)
-                  })
-                }
               },
               log: (msg) => process.stderr.write(`telegram gateway: ${msg}\n`),
             })
--- a/telegram-plugin/progress-card-driver.ts
+++ b/telegram-plugin/progress-card-driver.ts
@@ -17,7 +17,6 @@
 
 import type { SessionEvent } from './session-tail.js'
 import {
-  hasInFlightSubAgents,
   hasAnyRunningSubAgent,
   initialState,
   reduce,
@@ -489,7 +488,13 @@ export function createProgressDriver(config: ProgressDriverConfig): ProgressDriv
    */
   function maybeCompleteDeferredTurn(cs: PerChatState): void {
     if (!cs.pendingCompletion) return
-    if (hasInFlightSubAgents(cs.state)) return
+    // Gate on ANY running sub-agent (correlated OR orphan). Orphans from
+    // `Agent({run_in_background:true})` only deregister via their own
+    // `sub_agent_turn_end` — the card must stay pinned until then so the
+    // user sees the background work. Closes #87. Historical ghost-pin
+    // risk (#31/#43) is bounded by `closeZombie` on new enqueue +
+    // `maxIdleMs` heartbeat ceiling.
+    if (hasAnyRunningSubAgent(cs.state)) return
     process.stderr.write(`telegram gateway: progress-card: deferred completion firing turnKey=${cs.turnKey} (last sub-agent finished)\n`)
     flush(cs, /*forceDone*/ true)
     completeTurnFully(cs)
@@ -729,16 +734,15 @@ export function createProgressDriver(config: ProgressDriverConfig): ProgressDriv
     // notification. Ken observed ~13 identical "✅ Done" notifications
     // while two parallel review sub-agents were grinding.
     //
-    // Safe to gate on `hasInFlightSubAgents`: the completion paths
+    // Safe to gate on `hasAnyRunningSubAgent`: the completion paths
     // (`completeTurnFully` / `closeZombie` / `maybeCompleteDeferredTurn`)
-    // either (a) ran when !hasInFlightSubAgents or (b) explicitly marked
-    // every running sub-agent as done in the reducer state BEFORE the
-    // final flush. So at the instant the terminal emit fires, the
-    // condition evaluates to "no running sub-agents" and done=true
-    // flows through correctly.
+    // either (a) ran when no sub-agents are running or (b) explicitly
+    // marked every running sub-agent as done in the reducer state BEFORE
+    // the final flush. Including orphans here keeps `done=true` suppressed
+    // while a background dispatch is still active (closes #87).
     const terminal =
       (forceDone || chatState.state.stage === 'done')
-      && !hasInFlightSubAgents(chatState.state)
+      && !hasAnyRunningSubAgent(chatState.state)
     config.emit({
       chatId: chatState.chatId,
       threadId: chatState.threadId,
@@ -984,15 +988,16 @@ export function createProgressDriver(config: ProgressDriverConfig): ProgressDriv
         }
         flush(chatState, /*forceDone*/ event.kind === 'turn_end')
         if (event.kind === 'turn_end') {
-          if (hasInFlightSubAgents(chatState.state)) {
-            // Parent turn ended but correlated sub-agents are still running.
+          if (hasAnyRunningSubAgent(chatState.state)) {
+            // Parent turn ended but at least one sub-agent is still running.
             // Keep the card alive so the sub-agent work stays visible; defer
-            // completion until the last running correlated sub-agent reports
-            // done via its own sub_agent_turn_end (or the parent Agent
-            // tool_result). Orphan sub-agents (parentToolUseId == null) do NOT
-            // gate this defer — they are fire-and-forget for the parent turn's
-            // pin lifecycle (#31 fix). The heartbeat keeps ticking for all
-            // running sub-agents regardless.
+            // completion until the last running sub-agent reports done via
+            // its own sub_agent_turn_end (or the parent Agent tool_result).
+            // Closes #87: orphans from `Agent({run_in_background:true})` now
+            // gate the defer too, so background dispatches stay visible past
+            // parent turn-end. Safety nets: `closeZombie` on new enqueue +
+            // the `maxIdleMs` heartbeat ceiling bound the bad case (orphan
+            // never reports done).
             chatState.pendingCompletion = true
             const correlated: string[] = []
             const orphans: string[] = []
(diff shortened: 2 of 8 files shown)


mekenthompson:
Independent review while triaging open PRs — three things worth surfacing before this merges.

## 1. The orphan-defer change is the load-bearing semantic shift, not just refactor

The diff frames itself as "consolidate sub-agent visibility" but the meaningful change is in [\`progress-card-driver.ts\`](https://github.com/switchroom/switchroom/blob/main/telegram-plugin/progress-card-driver.ts) — flipping the defer gate from \`hasInFlightSubAgents\` (correlated-only) to \`hasAnyRunningSubAgent\` (correlated OR orphan). Orphan sub-agents from \`Agent({run_in_background:true})\` now keep the parent card pinned until they finish.

This **directly reverses PR [#49](https://github.com/switchroom/switchroom/pull/49)'s deliberate carve-out**, which excluded orphans from the defer gate to fix the ghost-pin bug ([#31](https://github.com/switchroom/switchroom/issues/31)/[#43](https://github.com/switchroom/switchroom/issues/43)) — the case where a JSONL-delivery race drops \`sub_agent_turn_end\` for an orphan and the card defers indefinitely.

The PR's mitigation argument is that \`closeZombie\` (on next enqueue) + \`maxIdleMs\` (heartbeat ceiling) bound the bad case. Worth verifying explicitly:

- What's the current \`maxIdleMs\` default? Is it short enough that a ghost-pinned card auto-closes within minutes, not 50+ minutes?
- \`closeZombie\` only fires when a NEW user message arrives. For users who go idle for hours after a \`run_in_background\` dispatch — does the card sit deferred until the heartbeat ceiling triggers?
- Is there a unit test that asserts the ghost-pin scenario — orphan started, parent turn ended, no \`sub_agent_turn_end\` ever — produces a card that auto-closes within the safety window?

If those check out, the change is correct for [#142](https://github.com/switchroom/switchroom/issues/142)'s "in-card surface is authoritative" vision and this is the right approach. Just want it called out as a behavioral-redesign gate-merge rather than buried inside the worker-card cleanup.

## 2. Dead pin-manager API to delete (cherry-pick opportunity from closed [#148](https://github.com/switchroom/switchroom/pull/148))

\`progress-card-pin-manager.ts\` exports \`trackExternalPin\` / \`untrackExternalPin\` — the worker card's pin lifecycle was the only caller. Once this PR lands, both methods are dead code (-38 lines including 2 obsolete tests). Could fold into this PR or open a small follow-up.

The cherry-pickable diff is on the closed [#148](https://github.com/switchroom/switchroom/pull/148) branch — \`progress-card-pin-manager.ts\` and \`tests/progress-card-pin-manager.test.ts\`.

## 3. [#145](https://github.com/switchroom/switchroom/pull/145) interaction

[#145](https://github.com/switchroom/switchroom/pull/145) (currently open) added 6 unit tests pinning the contract of \`hasInFlightSubAgents\` — the function this PR deletes. If [#145](https://github.com/switchroom/switchroom/pull/145) merges first, this PR also needs to drop the new \`describe('hasInFlightSubAgents …')\` block in \`tests/progress-card.test.ts\` to compile cleanly.

Either:
- Land [#145](https://github.com/switchroom/switchroom/pull/145) first → this PR includes the test deletion
- Land this PR first → [#145](https://github.com/switchroom/switchroom/pull/145) needs to retarget those tests at \`hasAnyRunningSubAgent\` (or drop the block entirely since the new gate behavior is well covered in \`progress-card-driver.test.ts:1347+\`)

Flagging so it doesn't surprise the merger.

mekenthompson:
Following up on my [earlier review](https://github.com/switchroom/switchroom/pull/144#issuecomment-4326529038). I checked the code to answer my own three questions:

## ✅ Q1 — \`maxIdleMs\` default and production value

- Driver default: **30 minutes** ([progress-card-driver.ts:386](https://github.com/switchroom/switchroom/blob/main/telegram-plugin/progress-card-driver.ts#L386))
- Production gateway: **5 minutes** ([gateway.ts:5225](https://github.com/switchroom/switchroom/blob/main/telegram-plugin/gateway/gateway.ts#L5225))

5 min is short enough that a ghost-pinned orphan auto-closes well within the user's attention window. Acceptable bound.

## ✅ Q2 — \`closeZombie\` only fires on new enqueue (idle-for-hours case)

Confirmed. \`closeZombie\` is only the "new turn arrives while old is deferred" path. For the user-goes-idle-for-hours case, \`maxIdleMs\` (5 min in prod) is the safety net via the heartbeat zombie ceiling, not \`closeZombie\`. Both paths are covered.

## ⚠️ Q3 — Test for the bad case (orphan never finishing)

This PR adds two excellent tests for the orphan-defer **happy paths** ([test 1](https://github.com/switchroom/switchroom/pull/144/files#diff-progress-card-drivertestts-R1392), [test 2](https://github.com/switchroom/switchroom/pull/144/files#diff-progress-card-drivertestts-R1485)) — both exit the defer via \`sub_agent_turn_end\`.

**Neither tests the bad case**: orphan started, parent turn ended, \`sub_agent_turn_end\` NEVER arrives → \`maxIdleMs\` zombie ceiling auto-closes. The existing zombie tests at line 503+ cover idle-with-no-events generally, but don't specifically pin the orphan-still-running case.

Without that lock, a future change to \`maxIdleMs\` default (raising to 24h, or accidentally disabling it) would silently re-introduce the ghost-pin failure mode that [PR #49](https://github.com/switchroom/switchroom/pull/49)'s correlated-only carve-out was originally designed to prevent ([#31](https://github.com/switchroom/switchroom/issues/31), [#43](https://github.com/switchroom/switchroom/issues/43)).

### Stacked fix branch ready

[\`mekenthompson/switchroom@claude/144-add-orphan-zombie-test\`](https://github.com/mekenthompson/switchroom/tree/claude/144-add-orphan-zombie-test) — single commit on top of this PR's head ([8ba5d72](https://github.com/mekenthompson/switchroom/commit/8ba5d72)). Adds one regression test:

> \`ghost-pin safety net: orphan never reports turn_end → maxIdleMs auto-closes\`

Setup: orphan dispatched → parent \`turn_end\` (defer holds) → no \`sub_agent_turn_end\` ever → advance past 5s (test) maxIdleMs → onTurnComplete fires exactly once with done=true.

Cherry-pick into this PR or merge separately. Test passes clean (1/100 in the file's selective run).

## Status

With Q1+Q2 answered + Q3's test ready to fold in, I'm comfortable with the orphan-defer semantic shift. The other two review points still stand:

- **Dead pin-manager API** (cherry-pick from closed [#148](https://github.com/switchroom/switchroom/pull/148)) — not blocking but a clean follow-up
- **[#145](https://github.com/switchroom/switchroom/pull/145) interaction** — whichever lands second drops the orphaned \`hasInFlightSubAgents\` test block

Issue #92, opened 2026-04-25
lawgpt-gateway: silent crash + auto-restart with no card or banner


mekenthompson:
## Observed
At 2026-04-26 09:45:00 AEST, `switchroom-lawgpt-gateway.service` was terminated and systemd auto-restarted it 3s later (restart counter went 0→1). The agent service `switchroom-lawgpt.service` did NOT restart — only the gateway.

```
Apr 26 09:45:00 systemd[1457]: switchroom-lawgpt-gateway.service: Consumed 4.985s CPU time, 87.3M memory peak, 0B memory swap peak.
Apr 26 09:45:03 systemd[1457]: switchroom-lawgpt-gateway.service: Scheduled restart job, restart counter is at 1.
Apr 26 09:45:03 systemd[1457]: Started switchroom-lawgpt-gateway.service - switchroom telegram gateway (lawgpt).
```

## Two problems

1. **Cause unknown.** No error/exception/fatal log preceded the termination. The gateway just exited and systemd respawned it. We need a final log line stamping the exit reason (uncaught exception, IPC partner death, OOM, signal received) so this kind of event is debuggable from the journal alone.

2. **Silent recovery.** The post-restart gateway logged `boot: no restart-marker found (clean start or crash recovery path)`, took the no-banner branch, and never told the user the gateway bounced. From Ken's perspective the gateway just disappeared and reappeared — no signal that anything happened.

## Why it's a problem
- Gateway crashes are bugs we want to know about. A crash that's silent at runtime AND silent on the operator surface (Telegram) is the worst kind — it gets normalized.
- The `restart-marker present` branch is the only thing that posts a card right now. A clean process death without a marker → silent.

## Requested fix
- **Stamp exit reason on shutdown.** Wrap the gateway's main loop with try/catch + signal handlers so the last log line before exit always identifies the cause.
- **Always post a card on boot, not just on planned-restart-with-marker.** See related issue (boot card should reflect every start: planned restart, crash recovery, fresh start), with the card showing the restart *reason* (planned / crash / OOM / unknown).

## Evidence
- Local host: pixsoul-ubuntu
- Gateway journal: `journalctl --user -u switchroom-lawgpt-gateway.service --since 09:36 --until 09:46`
- Agent journal: agent service did NOT restart (`NRestarts=0`, `Active: active (running) since 09:36:16`)

## Related
[...]

Issue #30, opened 2026-04-24
Phase 4c: wire operator-events end-to-end (client sender + watchdog + /status enrichment)


mekenthompson:
Phase 4b (#29) shipped the callback handler, IPC protocol extension, and history store. The operator-events taxonomy is now built out, but not yet wired end-to-end. This issue tracks the remaining work to close Phase 4 entirely.

## Scope

1. **Client-side IPC sender.** `ipc-client.ts` gained a `sendOperatorEvent` implementation in #29, but nothing in the Claude-side bridge actually calls it. Wire `session-tail.ts`'s existing `onOperatorEvent` callback to call `bridge.sendOperatorEvent(...)`. Tests: classifier-emits → IPC-forwards → gateway-records.

2. **Gateway IPC handler.** `createIpcServer({...})` is missing the `onOperatorEvent` option. Add it, pass a callback that calls `recordOperatorEvent(...)` from operator-events-history. Add `validateClientMessage` stricter check: regex-validate `msg.agent` against `^[a-z0-9][a-z0-9_-]{0,50}$` and allowlist `msg.kind` to the `OperatorEventKind` union before dispatching.

3. **`/status` enrichment.** Foreman's `buildFleetSummary` in `foreman-handlers.ts` currently doesn't surface the last-event-per-agent. Import `formatLastEventLine` from operator-events-history and append it under each agent line. Test that events within TTL show up, expired ones don't.

4. **Systemd uptime watchdog.** The `agent-restarted-unexpectedly` event kind is in the taxonomy but never emits. When the gateway sees a "recovered from unexpected restart" banner (gateway.ts around line 2649), cross-check `systemctl --user show switchroom-<agent> -p NRestarts,ActiveEnterTimestamp` — if the systemd restart count increased since the last-seen value, emit the event via the operator-events pipeline. This closes the 2026-04-21 lessons-learned loop (IPC flaps falsely triggering restart banners).

5. **Full `swap-slot` / `add-slot` callback wiring.** #29 shipped these as stubs that reply with the CLI command. Implement the real dispatch: `op:swap-slot:<agent>` calls `switchroom auth use <agent> <next-healthy-slot>`; `op:add-slot:<agent>` starts the auth-add flow (ForceReply + URL). Both refresh the auth dashboard on success.

[...]

PR #147, opened 2026-04-27
feat(operator-events): crash-event on boot + NRestarts watchdog (#92, #30)

mekenthompson:
## Summary

PR 2 of 5 in the cluster A+B (silent failures + operator-events) workstream. **Stacks on [#146](https://github.com/switchroom/switchroom/pull/146)** — the operator-events pipeline that PR carries is the channel this PR's two new producers send through.

Adds two producers:

1. **Boot-card crash detection ([#92](https://github.com/switchroom/switchroom/issues/92))** — when the gateway reboots after a crash but `resolveBootChatId` returns null (no recent activity), it used to silently skip ([gateway.ts:5517](telegram-plugin/gateway/gateway.ts:5517)). Now it always emits an `agent-crashed` operator event so the operator sees \`💥 Agent crashed: <name>\` in chat.

2. **NRestarts watchdog ([#30](https://github.com/switchroom/switchroom/issues/30) task 4)** — new module polls systemd's `NRestarts` for the agent unit (default 30s). When the count ticks up without a corresponding restart-pending marker, emits an `agent-restarted-unexpectedly` operator event. Closes the 2026-04-21 lessons-learned loop where IPC flaps falsely triggered the gateway's recovery banner.

## Refactor

Pulls the body of PR #146's `onOperatorEvent` IPC handler into a top-level `emitGatewayOperatorEvent(event)` helper. All three producers — IPC, boot crash, watchdog — funnel through this single point so cooldown / history / Telegram-broadcast logic stays in one place.

## What's in this PR vs. #146

This PR's branch is based on `feat/issue-30-operator-events-wiring`, so the GitHub diff against `main` includes #146's changes. **The new content unique to this PR:**

- New: [`telegram-plugin/gateway/restart-watchdog.ts`](telegram-plugin/gateway/restart-watchdog.ts) — pure module
- New: [`telegram-plugin/tests/restart-watchdog.test.ts`](telegram-plugin/tests/restart-watchdog.test.ts) — 18 tests
- Modified: [`telegram-plugin/gateway/gateway.ts`](telegram-plugin/gateway/gateway.ts) — extract helper, boot crash event, watchdog wiring

You can review just this PR's commit by viewing the latest commit on the branch directly.

## Knobs

- `SWITCHROOM_RESTART_WATCHDOG_POLL_MS` (default 30000) — set to 0 to disable
- Watchdog only activates when `SWITCHROOM_AGENT_NAME` is set
- `execFileSync` argv mode for the systemctl call (no shell — same security pattern as the rest of the gateway)

## Test plan

- [x] `npm run lint` (tsc --noEmit) — clean
- [x] All 185 operator-events / IPC validator / restart-watchdog tests pass
- [x] 18 new restart-watchdog tests cover: parsing happy/sad paths, decision branches (first tick, no change, increase, planned-restart suppression, multi-restart, daemon-reload counter reset), and integrated polling with execShow/parse failure resilience
- [ ] Manual: `kill -9` the gateway → next boot posts \`💥 Agent crashed\` card to allowlisted chat (cooldown protects against crash loops)
- [ ] Manual: `systemctl --user restart switchroom-<agent>` from terminal (no /restart marker) → \`🔄 Agent restarted unexpectedly\` card within 30s
- [ ] Manual: user issues /restart → no `agent-restarted-unexpectedly` event (planned-restart marker suppresses it)

Refs [#92](https://github.com/switchroom/switchroom/issues/92), [#30](https://github.com/switchroom/switchroom/issues/30) task 4.

### DIFF
--- a/telegram-plugin/bridge/bridge.ts
+++ b/telegram-plugin/bridge/bridge.ts
@@ -353,6 +353,27 @@ if (sessionTailEnabled) {
       cwd: sessionCwd,
       log: (msg) => process.stderr.write(`telegram bridge: ${msg}\n`),
       onEvent: forwardSessionEvent,
+      onOperatorEvent: (ev) => {
+        // Phase 4c: forward Anthropic API errors to the gateway so it can
+        // post the operator card + record into the /status history. The
+        // gateway resolves the destination chat from its access allowlist
+        // (operator events are agent-level, not tied to a specific user
+        // message), so chatId is left empty here.
+        if (!ipc || !ipc.isConnected()) return
+        try {
+          ipc.sendOperatorEvent({
+            type: 'operator_event',
+            kind: ev.kind,
+            agent: AGENT_NAME,
+            detail: ev.detail.slice(0, 1000),
+            chatId: '',
+          })
+        } catch (err) {
+          process.stderr.write(
+            `telegram bridge: sendOperatorEvent failed kind=${ev.kind}: ${(err as Error).message}\n`,
+          )
+        }
+      },
     })
     process.stderr.write(
       `telegram bridge: session tail watching ${sessionTailHandle.getActiveFile() ?? '(no active file yet)'}\n`,
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -66,6 +66,13 @@ import {
   deleteFromHistory, query as queryHistory, getLatestInboundMessageId,
 } from '../history.js'
 import { parseQueuePrefix, parseSteerPrefix, formatPriorAssistantPreview, formatReplyToText } from '../steering.js'
+import {
+  renderOperatorEvent,
+  shouldEmitOperatorEvent,
+  type OperatorEvent,
+  type OperatorEventKind,
+} from '../operator-events.js'
+import { recordOperatorEvent } from '../operator-events-history.js'
 
 /**
  * Truncation cap for the `reply_to_text` channel-meta attribute (issue #119).
@@ -139,6 +146,8 @@ import type {
   SessionEventForward,
   PermissionRequestForward,
   HeartbeatMessage,
+  ScheduleRestartMessage,
+  OperatorEventForward,
   InboundMessage,
 } from './ipc-protocol.js'
 import { writePidFile, clearPidFile } from './pid-file.js'
@@ -1093,6 +1102,75 @@ const ipcServer: IpcServer = createIpcServer({
         success: true,
         waitingForTurn: true,
       });
+    }
+  },
+
+  onOperatorEvent(_client: IpcClient, msg: OperatorEventForward) {
+    // Phase 4c: bridge has detected an Anthropic API error (or synthetic
+    // gateway-side event). Render the operator card + post to every chat
+    // on the access allowlist, deduped by per-agent per-kind cooldown.
+    //
+    // chatId on the wire is currently always empty — the bridge doesn't
+    // know which user-chat the error should attribute to (errors aren't
+    // tied to a specific inbound message). We broadcast to the same set
+    // of chats that receive permission requests so the right operator
+    // sees it.
+    const kind = msg.kind as OperatorEventKind
+    const agent = msg.agent
+
+    if (!shouldEmitOperatorEvent(agent, kind)) {
+      process.stderr.write(
+        `telegram gateway: operator-event suppressed (cooldown) agent=${agent} kind=${kind}\n`,
+      )
+      return
+    }
+
+    const event: OperatorEvent = {
+      kind,
+      agent,
+      detail: msg.detail,
+      suggestedActions: [],
+      firstSeenAt: new Date(),
+    }
+
+    try {
+      recordOperatorEvent(event)
+    } catch (err) {
+      process.stderr.write(
+        `telegram gateway: recordOperatorEvent failed agent=${agent} kind=${kind}: ${(err as Error).message}\n`,
+      )
+    }
+
+    let rendered: ReturnType<typeof renderOperatorEvent>
+    try {
+      rendered = renderOperatorEvent(event)
+    } catch (err) {
+      process.stderr.write(
+        `telegram gateway: renderOperatorEvent failed agent=${agent} kind=${kind}: ${(err as Error).message}\n`,
+      )
+      return
+    }
+
+    const access = loadAccess()
+    if (access.allowFrom.length === 0) {
+      process.stderr.write(
+        `telegram gateway: operator-event no-allowlist agent=${agent} kind=${kind} (recorded only)\n`,
+      )
+      return
+    }
+
+    process.stderr.write(
+      `telegram gateway: operator-event posting agent=${agent} kind=${kind} to ${access.allowFrom.length} chat(s)\n`,
+    )
+    for (const chat_id of access.allowFrom) {
+      void bot.api.sendMessage(chat_id, rendered.text, {
+        parse_mode: 'HTML',
+        reply_markup: rendered.keyboard,
+      }).catch(e => {
+        process.stderr.write(
+          `telegram gateway: operator-event send to ${chat_id} failed agent=${agent} kind=${kind}: ${e}\n`,
+        )
+      })
     }
   },
 
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -2989,6 +2989,26 @@ function isSelfTargetingCommand(name: string): boolean {
 // ─── Restart marker ───────────────────────────────────────────────────────
 type RestartMarker = { chat_id: string; thread_id: number | null; ack_message_id: number | null; ts: number }
 
+/**
+ * In-memory timestamp of the most recent planned restart written by this
+ * gateway process. The restart-watchdog reads this (with a freshness
+ * window) instead of `readRestartMarker()` to decide whether a NRestarts
+ * uptick is user-initiated.
+ *
+ * The disk marker is unsafe for that purpose because the bridge-reconnect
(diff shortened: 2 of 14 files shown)


mekenthompson:
Independent code review while triaging the open PR queue. Skim of the merged-in [#146](https://github.com/switchroom/switchroom/pull/146) base; deep review of the new content unique to this PR. Tests pass clean (25/25) and \`tsc --noEmit\` is green.

## Summary

The watchdog and crash-event modules are well-designed: pure decision function with DI, argv-mode \`execFileSync\` (no shell injection), \`unref()\`-ed timer, fault-tolerant ticks, fault-tolerant emit. Test coverage on \`decideWatchdogTick\` and \`parseSystemdShowOutput\` is thorough.

**One real correctness bug** (false-positive race) and a few smaller comments below. The race deserves a fix before merge; everything else is suggestion-level.

## 🔴 Bug: false-positive race between bridge-reconnect and watchdog tick

The watchdog's \`isPlannedRestartFresh: () => readRestartMarker() != null\` ([gateway.ts:5597](https://github.com/switchroom/switchroom/blob/feat/issue-92-crash-aware-boot-card/telegram-plugin/gateway/gateway.ts#L5597)) relies on the disk marker to suppress emits during user-initiated \`/restart\`. But the marker is **cleared by the bridge-reconnect handler** at [gateway.ts:1013](https://github.com/switchroom/switchroom/blob/feat/issue-92-crash-aware-boot-card/telegram-plugin/gateway/gateway.ts#L1013), which fires within ~1-2s of agent boot. The watchdog tick interval is 30s. So the typical \`/restart agent\` timeline is:

\`\`\`
t=0s    user types /restart        → writeRestartMarker
t=0s    detached `systemctl restart switchroom-<agent>` spawned
t=2s    agent boots, bridge reconnects → clearRestartMarker (bridge-reconnect handler)
t=5s    watchdog tick               → marker is GONE → NRestarts delta=1
                                    → emits 'agent-restarted-unexpectedly'
\`\`\`

**Almost every \`/restart agent\`** lands a false-positive watchdog event. The 5-min cooldown bounds the spam to one card per 5 min, but the card is misleading — the user just typed \`/restart\` and gets a \"restarted unexpectedly\" warning.

The watchdog's own JSDoc acknowledges this:
> \"The marker is NOT a perfect signal — but combined with the cooldown inside the operator-events pipeline (5min per agent+kind), spurious emissions are bounded.\"

I'd push back on accepting that — the failure mode is \"user got a misleading card every time they used a normal feature\". Cooldown bounds frequency but doesn't eliminate the wrongness.

### Suggested fix

Track planned restarts in-memory instead of re-reading the disk marker each tick:

\`\`\`ts
// Module-level in gateway.ts
let lastPlannedRestartAt: number | null = null

// When /restart writes the marker (gateway.ts:3657, 3754):
writeRestartMarker(...)
lastPlannedRestartAt = Date.now()

// In the watchdog wiring:
isPlannedRestartFresh: () => {
  if (lastPlannedRestartAt == null) return false
  return Date.now() - lastPlannedRestartAt < 60_000
},
\`\`\`

The 60s window comfortably covers systemd kill→restart→bridge-reconnect (typically ~2-5s) plus the next watchdog tick. After 60s, any further NRestarts increment is genuinely unexpected. Doesn't depend on disk-marker lifetime.

### Test gap

The unit suite tests \`isPlannedRestartFresh: () => true\` and \`() => false\` in isolation, but doesn't exercise the disk-marker path. A regression test for the race would call the gateway's actual \`isPlannedRestartFresh\` callback after a simulated \`/restart\` → bridge-reconnect → watchdog-tick sequence. Worth adding once the in-memory fix lands.

## 🟡 Smaller suggestions

### 1. Consider relaxing \`parseSystemdShowOutput\`

Currently both \`NRestarts\` AND \`ActiveEnterTimestampMonotonic\` are required for the parser to return a result; missing either → \`null\` → tick skipped. The decision function only uses \`nRestarts\`, so the timestamp-required check is over-strict. A future systemd version that drops or renames \`ActiveEnterTimestampMonotonic\` would silently disable the watchdog. Suggest: make activeEnter optional (default 0) and decide on \`nRestarts\` alone.

[restart-watchdog.ts:55](https://github.com/switchroom/switchroom/blob/feat/issue-92-crash-aware-boot-card/telegram-plugin/gateway/restart-watchdog.ts#L55)

### 2. Synchronous first \`tick()\` blocks gateway boot

[restart-watchdog.ts:187](https://github.com/switchroom/switchroom/blob/feat/issue-92-crash-aware-boot-card/telegram-plugin/gateway/restart-watchdog.ts#L187) calls \`tick()\` synchronously, which calls \`execFileSync('systemctl', …)\` with a 5s timeout. If systemctl is slow (boot under load, dbus contention) the gateway's startup blocks for up to 5s before the bot starts polling. Suggest: \`setImmediate(tick)\` or schedule the first tick at \`pollIntervalMs\` instead of inline.

### 3. \`agent: process.env.SWITCHROOM_AGENT_NAME ?? '-'\` for the boot-crash event

[gateway.ts:5514](https://github.com/switchroom/switchroom/blob/feat/issue-92-crash-aware-boot-card/telegram-plugin/gateway/gateway.ts#L5514) defaults to literal \`'-'\` when the env var is unset. The cooldown is keyed on \`(agent, kind)\` — if multiple agentless gateways share an operator-events history (unlikely today but possible), they'd alias under \`'-'\` and cooldown each other. Minor; could log a stderr warning instead.

### 4. NIT: detail string says \"since last poll\"

\`'systemd auto-restarted the agent service \${delta} times since last poll'\` ([restart-watchdog.ts:105](https://github.com/switchroom/switchroom/blob/feat/issue-92-crash-aware-boot-card/telegram-plugin/gateway/restart-watchdog.ts#L105)) — when delta>1 across a cooldown window, \"last poll\" can be 5+ minutes ago, not 30s. Cosmetic.

## ✅ Things that are correct

- **No shell injection.** \`execFileSync('systemctl', ['--user', 'show', unit, …], {…})\` is argv-mode; no shell metachar risk even if \`agentName\` were untrusted.
- **5s timeout on systemctl exec.** Bounded.
- **\`setInterval(tick, …).unref()\`** correctly conditional on \`typeof === 'function'\`.
- **\`stdio: ['ignore', 'pipe', 'pipe']\`** correctly avoids stdin attachment.
- **\`Number.isFinite(n) && n >= 0\`** guard rejects NaN, Infinity, negative.
- **Daemon-reload counter reset is handled** (delta < 0 → no emit, baseline rebased).
- **Cooldown bounds crash-loop spam** — verified via the operator-events pipeline.
- **\`emitGatewayOperatorEvent\` extraction** is clean — single point for cooldown + record + broadcast across all three producers (IPC, boot crash, watchdog).
- **\`reason === 'crash'\` gating** correctly suppresses planned-restart noise from the boot-crash event (the disk-marker race only affects the watchdog, not the boot path — gateway boot consumes the marker as part of its own boot flow).

## Test recommendation

After the in-memory \`lastPlannedRestartAt\` fix, add an integration test covering the full \`/restart agent\` → bridge-reconnect → watchdog-tick sequence so future regressions on this race surface red instead of by user report.

mekenthompson:
**Fix branch ready** for the marker-race issue in my [earlier review](https://github.com/switchroom/switchroom/pull/147#issuecomment-4326584322): [\`mekenthompson/switchroom@claude/147-fix-marker-race\`](https://github.com/mekenthompson/switchroom/tree/claude/147-fix-marker-race) — single commit on top of this PR's head ([57dd8e6](https://github.com/mekenthompson/switchroom/commit/57dd8e6)).

\`tsc --noEmit\` clean. Existing 18/18 \`restart-watchdog.test.ts\` tests still pass (the watchdog module itself didn't change — the fix is in gateway.ts).

## What it does

Module-level \`lastPlannedRestartAt: number | null\` set inside \`writeRestartMarker()\` after the disk write succeeds. Watchdog's \`isPlannedRestartFresh\` now checks that timestamp against a 60s freshness window. Survives bridge-reconnect clearing the disk marker.

\`\`\`diff
+ let lastPlannedRestartAt: number | null = null
+ const PLANNED_RESTART_FRESHNESS_MS = 60_000

  function writeRestartMarker(marker: RestartMarker): void {
    const p = restartMarkerPath(); if (!p) return
    try {
      writeFileSync(p, JSON.stringify(marker))
+     lastPlannedRestartAt = Date.now()
      process.stderr.write(...)
    } catch (err) { ... }
  }

+ function isPlannedRestartFresh(now: number = Date.now()): boolean {
+   if (lastPlannedRestartAt == null) return false
+   return now - lastPlannedRestartAt < PLANNED_RESTART_FRESHNESS_MS
+ }

  // In the watchdog wiring (gateway.ts:5597):
- isPlannedRestartFresh: () => readRestartMarker() != null,
+ isPlannedRestartFresh: () => isPlannedRestartFresh(),
\`\`\`

## How to use

- **Cherry-pick** into this PR: \`git cherry-pick 57dd8e6\` from \`feat/issue-92-crash-aware-boot-card\`, or merge the branch directly.
- **Or merge separately** after this PR lands — the patch is just one commit on top.

## Test gap I'm leaving

The new \`isPlannedRestartFresh\` is a pure timestamp check but uses a module-private \`let\` so it can't be unit-tested without exposing a setter or refactoring. The existing \`decideWatchdogTick\` tests continue to cover the decision-function contract (with \`recentPlannedRestart: true/false\` injected). For full integration coverage, a future test would simulate \`writeRestartMarker\` → bridge-reconnect → watchdog-tick using a real (or fake) \`isPlannedRestartFresh\` from gateway.ts. Worth doing once the fix lands; intentionally out of scope for this minimal patch.

The other suggestions from my review (relax the parser, async first tick, agentless gateway label, "since last poll" cosmetic) are not in this branch — they're suggestion-level, not blocking.

Issue #101, opened 2026-04-26
progress-card: formatDuration <1s HTML parse error crashes gateway (sendMessage 400, sibling of #86)


mekenthompson:
## Observed

`switchroom-lawgpt-gateway.service` has been crash-looping every ~14 sec since ~08:04 today (4+ hours, 14+ restarts on the gateway counter, many more on the agent process). Smoking gun in journal:

```
Apr 26 08:04:33 telegram gateway: unhandled rejection: GrammyError: Call to 'sendMessage' failed!
  (400: Bad Request: can't parse entities: Unsupported start tag "1s" at byte offset 132)
Apr 26 08:04:33 telegram gateway: shutdown.clean_marker_skipped signal=unhandledRejection (crash path — banner will fire on next boot)
```

## Cause

`formatDuration()` in `telegram-plugin/card-format.ts` returns the literal string `<1s` for sub-second durations. Per its own docstring (line 8): *"Duration format: `<1s` for sub-second"*.

PR #89 (closes #86) caught this in **one** call site — `subagent-watcher.ts` rendering the worker card. But `progress-card.ts` calls `formatDuration()` **11+ more times** without HTML escaping, all of which can hit the same bug:

```
telegram-plugin/progress-card.ts:831:  const elapsed = formatDuration(now - state.turnStartedAt)
telegram-plugin/progress-card.ts:858:  const gap = formatDuration(opts.stuckMs)
telegram-plugin/progress-card.ts:929:  const dur = formatDuration(age)
telegram-plugin/progress-card.ts:972:  const dur = formatDuration(now - item.startedAt)
telegram-plugin/progress-card.ts:978:  const dur = formatDuration(now - item.startedAt)
telegram-plugin/progress-card.ts:986:  const dur = formatDuration(item.finishedAt - item.startedAt)
telegram-plugin/progress-card.ts:1091: const dur = formatDuration(end - sa.startedAt)
telegram-plugin/progress-card.ts:1098: const elapsed = formatDuration(now - sa.startedAt)
telegram-plugin/progress-card.ts:1106: const curDur = formatDuration(now - cur.startedAt)
```

When a turn elapsed < 1s gets dropped into the progress-card HTML and sent to Telegram, the `<1s` is parsed as a malformed `<1s>` HTML tag → 400 → unhandled rejection → gateway exits → systemd restart-on-failure → loop.

## Why #89 didn't catch it

[...]

PR #148, opened 2026-04-27
feat(telegram): unify sub-agent surface to in-card tracker only (closes #87, partial #142)

### LINKED ISSUE #60: greeting: replace baked-in version/PR/age text with live probe at SessionStart

### LINKED ISSUE #87: sub-agent activity invisible after parent turn ends (no progress card, no inline notification)

mekenthompson:
## Summary

PR 2 of [#142](https://github.com/switchroom/switchroom/issues/142). Deletes two redundant sub-agent surfaces so each sub-agent appears in exactly one place — the main progress card's in-card \`[Sub-agents · N running]\` block, which already stays alive past parent turn-end via the driver's \`pendingCompletion\` gate.

**Pre-#142** every sub-agent dispatch produced THREE surfaces:
1. The in-card \`[Sub-agents · N running]\` row
2. A \`🛠️ Worker dispatched\` inline reply
3. A pinned \`🔧 Background workers (N)\` card

**Post-PR 2**: just (1).

## What goes

- \`renderWorkerCard()\` in [telegram-plugin/subagent-watcher.ts](telegram-plugin/subagent-watcher.ts) (entire function — its surface is gone)
- The \`🛠️ Worker dispatched\` inline-reply block at the end of \`registerAgent()\`
- \`updatePinnedCard\` config field, \`maybeSendCardUpdate\` function, and the \`lastCardUpdate\` / \`pendingCardUpdate\` / \`cardUpdateIntervalMs\` state
- The full \`updatePinnedCard\` callback wiring in [telegram-plugin/gateway/gateway.ts](telegram-plugin/gateway/gateway.ts) (~80 lines: \`workerCardMsgId\` lifecycle, pin/unpin/edit for the worker card, \`pinMgr.trackExternalPin\` / \`untrackExternalPin\` tracking)
- The \`trackExternalPin\` / \`untrackExternalPin\` API on [telegram-plugin/progress-card-pin-manager.ts](telegram-plugin/progress-card-pin-manager.ts) — the worker card was its only consumer
- Twelve \`renderWorkerCard\` unit tests + two external-pin tests

## What stays

- \`WorkerEntry\` registry + JSONL tail + stall detection — the watcher still surfaces stall warnings (\`⚠ Worker idle: …\`) and completion notifications (\`✓ Worker done: …\`) as inline messages. Those are transition signals, not per-dispatch chrome.
- Historical-files snapshot (the [#82](https://github.com/switchroom/switchroom/pull/82) fix) — still gates stall detection so restarts don't fire 60s-stall warnings for every long-dead session in JSONL history.
- The \`pendingCompletion\` driver path that keeps the main card alive past parent turn-end (locked in [progress-card-driver.test.ts:1347+](telegram-plugin/tests/progress-card-driver.test.ts:1347)).

## What's added

One focused unit test in [progress-card-driver.test.ts](telegram-plugin/tests/progress-card-driver.test.ts) locking the contract this PR depends on: while the parent turn is in \`pendingCompletion\` and a sub-agent is still running, the heartbeat fires renders with \`done=false\` (so they're edits, not new sendMessage calls per [#101](https://github.com/switchroom/switchroom/issues/101)), and only the terminal emit on the last \`sub_agent_turn_end\` carries \`done=true\` + fires \`onTurnComplete\` exactly once.

## Issue refs

- Closes [#87](https://github.com/switchroom/switchroom/issues/87) — sub-agent activity invisible after parent turn ends. The in-card tracker now stays alive past parent turn-end via the existing \`pendingCompletion\` gate; this PR deletes the alternative surface (\`🔧 Background workers\` card) that was a workaround for the gap.
- Partial [#142](https://github.com/switchroom/switchroom/issues/142) (PR 2 of 3). PR 1 will rewrite the boot card and delete \`buildSessionGreetingScript\` (closes [#60](https://github.com/switchroom/switchroom/issues/60)). PR 3 will move the deleted greeting content into a \`/status\` command extension.

## Acceptance check

- [x] \`tsc --noEmit\` clean
- [x] All progress-card-driver tests pass (98/98), including the new heartbeat-during-pendingCompletion lock-in test
- [x] All progress-card-pin-manager tests pass (32/32) after dropping the two external-pin tests
- [ ] CI green on Linux (Windows path-join breaks the watcher test harness's mock fs — pre-existing, also fails on \`origin/main\`)
- [ ] Manual: send a message that dispatches a sub-agent → exactly ONE card in chat. In-card tracker shows the sub-agent. No \"Worker dispatched\" reply. No pinned \"Background workers\" card. Sub-agent outlives parent turn → SAME card stays alive heartbeating until done.

## Diff stat

\`\`\`
6 files changed, 125 insertions(+), 484 deletions(-)
\`\`\`

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -5465,16 +5465,17 @@ void (async () => {
 
         // Background sub-agent visibility watcher. Watches the subagents/
         // directory under each session dir for new agent-<id>.jsonl files
-        // and surfaces live activity to Telegram via a pinned card +
-        // inline notifications. Only started when a valid agentDir is known
-        // (gate on streamMode=checklist for progress-card parity).
+        // and surfaces stall/completion transitions as inline messages.
+        // Per #142, the per-dispatch "Worker dispatched" reply and the
+        // pinned "🔧 Background workers" card were deleted — those
+        // duplicated information already shown in the main progress card's
+        // in-card `[Sub-agents · N running]` block, which now stays alive
+        // past parent turn-end via the driver's pendingCompletion gate.
+        // Only stall + completion notifications remain (they signal
+        // transitions, not per-dispatch chrome).
         if (streamMode === 'checklist') {
           const watcherAgentDir = resolveAgentDirFromEnv()
           if (watcherAgentDir != null) {
-            // Pinned worker card: one message per watcher session,
-            // edited in-place. Managed entirely by the watcher.
-            let workerCardMsgId: number | null = null
-
             subagentWatcher = startSubagentWatcher({
               agentDir: watcherAgentDir,
               sendNotification: (text: string) => {
@@ -5487,55 +5488,6 @@ void (async () => {
                 }).catch((err: Error) => {
                   process.stderr.write(`telegram gateway: subagent-watcher notification failed: ${err.message}\n`)
                 })
-              },
-              updatePinnedCard: (html: string | null) => {
-                const ownerChatId = loadAccess().allowFrom[0]
-                if (!ownerChatId) return
-                if (html === null) {
-                  // No active workers — unpin and delete the card
-                  if (workerCardMsgId != null) {
-                    const msgId = workerCardMsgId
-                    workerCardMsgId = null
-                    // Drop external-pin tracking before issuing the unpin so
-                    // a race with a concurrent service-message capture can't
-                    // re-resurrect the entry. Issue #94.
-                    pinMgr.untrackExternalPin(String(ownerChatId), msgId)
-                    void lockedBot.api.unpinChatMessage(ownerChatId, msgId).catch(() => {})
-                    void lockedBot.api.deleteMessage(ownerChatId, msgId).catch(() => {})
-                  }
-                  return
-                }
-                if (workerCardMsgId == null) {
-                  // Create a new pinned card
-                  void lockedBot.api.sendMessage(ownerChatId, html, {
-                    parse_mode: 'HTML',
-                    link_preview_options: { is_disabled: true },
-                    ...(TOPIC_ID != null ? { message_thread_id: TOPIC_ID } : {}),
-                  }).then((sent) => {
-                    workerCardMsgId = sent.message_id
-                    // Register the worker card with pinMgr BEFORE pinning so
-                    // its `captureServiceMessage` callback recognises the
-                    // service message Telegram emits in response and deletes
-                    // it (matching the existing main-card behaviour).
-                    // Issue #94.
-                    pinMgr.trackExternalPin(String(ownerChatId), sent.message_id)
-                    void lockedBot.api.pinChatMessage(ownerChatId, sent.message_id, {
-                      disable_notification: true,
-                    }).catch(() => {})
-                  }).catch((err: Error) => {
-                    process.stderr.write(`telegram gateway: subagent-watcher card send failed: ${err.message}\n`)
-                  })
-                } else {
-                  // Edit the existing card
-                  void lockedBot.api.editMessageText(ownerChatId, workerCardMsgId, html, {
-                    parse_mode: 'HTML',
-                    link_preview_options: { is_disabled: true },
-                  }).catch((err: Error) => {
-                    const msg = err instanceof GrammyError && err.error_code === 400 &&
-                      /message is not modified/i.test(err.description ?? '') ? '' : err.message
-                    if (msg) process.stderr.write(`telegram gateway: subagent-watcher card edit failed: ${msg}\n`)
-                  })
-                }
               },
               log: (msg) => process.stderr.write(`telegram gateway: ${msg}\n`),
             })
--- a/telegram-plugin/progress-card-pin-manager.ts
+++ b/telegram-plugin/progress-card-pin-manager.ts
@@ -111,21 +111,6 @@ export interface PinManager {
     pinnedMessageId: number
     serviceMessageId: number
   }): void
-  /**
-   * Register a pin made outside the per-turn `considerPin()` path so
-   * `captureServiceMessage()` will recognise its service message and
-   * delete it. Used by the worker / sub-agent card (issue #94), which
-   * pins through the gateway directly rather than the progress-card
-   * pin candidate flow. Idempotent — calling twice with the same
-   * (chatId, messageId) is a no-op.
-   */
-  trackExternalPin(chatId: string, messageId: number): void
-  /**
-   * Drop an external pin from tracking. Call when the corresponding
-   * pinned message is unpinned/deleted by its owner so the manager
-   * doesn't keep a stale reference. Idempotent.
-   */
-  untrackExternalPin(chatId: string, messageId: number): void
   /** Test-only: snapshot the currently-pinned turnKeys. */
   pinnedTurnKeys(): ReadonlyArray<string>
   /** Test-only: look up the pinned message id for a turnKey. */
@@ -172,13 +157,6 @@ export function createPinManager(deps: PinManagerDeps): PinManager {
   // unpin path can also scrub the service message if the capture-delete
   // somehow failed.
   const serviceMessages = new Map<string, number>()
-  // Pins that the manager DIDN'T make through `considerPin()` but should
-  // still recognise so their "Clerk pinned …" service message gets
-  // deleted. Issue #94: the worker / sub-agent card pins via the gateway
-  // directly; without this set, captureServiceMessage would skip its
-  // service message and the user sees the system-message noise that the
-  // main card already suppresses. Keyed by `${chatId}:${messageId}`.
-  const externalPins = new Set<string>()
   // Fire-and-forget promises we want tests to be able to drain.
   const inFlight = new Set<Promise<unknown>>()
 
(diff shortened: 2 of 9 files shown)


mekenthompson:
Closing in favor of [#144](https://github.com/switchroom/switchroom/pull/144), which I missed when I started this. Both PRs target the same worker-card deletion + close [#87](https://github.com/switchroom/switchroom/issues/87), but #144 came first (1.5h earlier) and goes deeper:

**Why [#144](https://github.com/switchroom/switchroom/pull/144) is the better implementation.** It also flips the defer gate from \`hasInFlightSubAgents\` (correlated-only) to \`hasAnyRunningSubAgent\` (correlated OR orphan). That makes the in-card surface authoritative for orphan sub-agents from \`Agent({run_in_background:true})\` too — the proper closure of [#87](https://github.com/switchroom/switchroom/issues/87). My PR was the conservative version that just deleted the redundant surface and kept PR #49's orphan carve-out.

**One unique cleanup worth cherry-picking from this PR.** The \`trackExternalPin\` / \`untrackExternalPin\` API on \`progress-card-pin-manager.ts\` had the worker card as its only consumer — once #144 lands, it's dead code (-38 lines including 2 obsolete tests). Either fold into #144 or open a small follow-up. Diff is at [`progress-card-pin-manager.ts`](https://github.com/switchroom/switchroom/pull/148/files#diff-progress-card-pin-managerts) on this branch.

The lock-in test I added in [`progress-card-driver.test.ts`](https://github.com/switchroom/switchroom/pull/148/files#diff-progress-card-drivertestts) (heartbeat keeps emitting renders during the deferred-completion window with \`done=false\`) is also worth keeping — but #144 already adds two similar tests at lines 1391/1453, so it's likely redundant.

Apologies for the duplicate.

Issue #205, opened 2026-04-28
EPIC: Vault security — Telegram-2FA, cron permission grants, audit log


mekenthompson:
## Status (rewritten 2026-05-01 after audit)

**Phases 1A / 1B / 1C — SHIPPED.** Per-entry vault scope, audit log, and the broker `list` ACL gate are all live in production. The remaining vault-security work is now narrowly scoped:

### Done

- ✅ #8 — Per-entry `scope.allow` / `scope.deny` (`src/vault/vault.ts:75-95`)
- ✅ #206 — Audit log (`src/vault/broker/audit-log.ts`)
- ✅ #207 — Broker `list` op honours peercred + ACL + scope (`src/vault/broker/server.ts:454-587`)
- ✅ #344 — Same as #207, dup
- ✅ #346 — `shouldInstallBrokerUnit` gate (`src/agents/systemd.ts:374-381`)
- ✅ #265 items 1+2+3 (vault-grant wizard hardening) — shipped on `chore/overnight-review-fixes`

### Remaining

- 🔬 **#161 — Telegram-2FA for vault unlock** (Phase 2). Needs design proposal first. Audit recommendation: capability-token model — repurpose the existing `src/vault/grants.ts` to mint a session-unlock token gated by a Telegram approval callback. Cheapest path because `grants.ts` already implements signed/revocable tokens.
- 📋 **`vault doctor` / audit-review CLI** (Phase 4). `src/vault/doctor.ts` exists but only checks key referencing — would benefit from an audit-log-tail surface and an ACL-coverage report.

### Decision

Recommend closing this umbrella once the two remaining items are filed as their own narrow tickets (or merged into #161). The original epic has done its job.

mekenthompson:
Forensic audit (2026-05-01) recommends partial closeout + epic body update.

**Already shipped (close the sub-issues):**

- #8 — Per-entry vault scope (`src/vault/vault.ts:75-95`)
- #206 — Audit log (`src/vault/broker/audit-log.ts`)
- #207 — `list` op honours peercred + ACL + scope (`src/vault/broker/server.ts:454-587`)
- #344 — Same as #207 (already commented above)
- #346 — `shouldInstallBrokerUnit` gate (`src/agents/systemd.ts:374-381`)

**Still open (carry the work):**

- #161 — Telegram-2FA for vault unlock (Phase 2 — needs design proposal first)
[...]

PR #223, opened 2026-04-28
docs(#161): Telegram-as-2FA for vault unlock — design proposal

### LINKED ISSUE #161: Investigate Telegram-as-2FA for vault unlock (passwordless via bot prompt)
## Idea

The current vault model requires a passphrase to unlock the broker. This means:

- Operator has to remember a passphrase
- Passphrase is typed into a terminal (currently un-masked, see sister issue)
- After daemon restart, agents are blocked until someone SSH's in to unlock

The Telegram plugin already has a confirmation flow for sensitive actions (slash commands, MCP tool prompts). What if we used Telegram itself as the second factor for unlocking the vault?

## Sketch

When an agent (or another caller) needs a vault key while the broker is locked:

1. Broker emits a "pending unlock" event, keyed by request id.
2. Plugin sees the pending request, sends the user a Telegram message: "Agent <name> is requesting vault key <X>. Approve unlock? [Yes / No / Always for this session]"
3. User taps Yes → plugin signals broker to unlock with a stored device-bound secret (TouchID-style: secret never leaves the host, Telegram confirms identity). Broker stays unlocked for a TTL (e.g. 1h) or until lock is signaled.
4. If the user is offline / no response within N minutes, fall back to existing passphrase prompt.

## Why this is interesting

- **Passwordless from the operator's POV** — they tap a button instead of typing a secret.
- **Audit trail** — every unlock approval is a Telegram message, easy to review.
- **Phishing-resistant** — unlock approval requires the bound Telegram account, not just knowledge of a passphrase.
- **Headless-friendly** — works the same whether the operator is on their laptop, phone, or away from a terminal.

## Open questions

- What's the on-disk secret bound to? (Per-host random key, encrypted by Telegram-bound key derivation? Hardware-backed via TPM where available?)
- Bootstrap: how does the FIRST unlock happen, before Telegram pairing? Probably keep the current passphrase flow as the fallback / setup path.
- Granularity: per-key approval vs. per-session unlock? Latter is the usable default; former is overkill for most agents.
- Recovery if the bound Telegram account is lost: passphrase should remain the break-glass.

This is a research/design issue, not a build ticket — start with a written proposal before any code.

mekenthompson:
Closes #161 (the design-doc deliverable; implementation PRs land as follow-ups). Phase 2 of vault security epic #205.

Per the issue: "This is a research/design issue, not a build ticket — start with a written proposal before any code."

## Summary

The proposal recommends shipping Telegram-as-2FA for vault unlock as an **opt-in config flag** in Phase 1, with passphrase remaining the default. The operator binds their Telegram account once (PIN exchange), then approves unlock prompts via Telegram instead of typing the passphrase each time. The passphrase stays as break-glass for recovery.

**Storage:** OS Keyring (libsecret/Keychain) primary, encrypted file (\`~/.config/credstore.encrypted/vault-dbs\`) fallback for headless cron scenarios where the keyring isn't available.

**Threat-model net:** Stronger against scrollback/shoulder-surf/keystroke-logger attacks; introduces dependency on Telegram security; root compromise stays equally bad in both models.

**Phasing:** Opt-in via \`vault.unlock_method: "telegram"\` flag (Phase 1) → keyring/TPM hardening + rate limiting (Phase 2) → default Telegram (Phase 3, after deployment confidence).

**Implementation scope (Section 8):** 6 PRs, ~6-8 weeks for Phase 1 — broker RPC for approval, config schema + bind command, Telegram plugin button UI, audit log enhancements, docs/examples, autoUnlock compatibility.

**5 open questions** flagged with proposed answers — keyring availability in systemd sessions, cron blocking behavior, multi-host binding, 24/7 ops scenarios, recovery codes.

## Review

Read the full doc at \`docs/vault-telegram-2fa-proposal.md\`. After approval, the proposal becomes the spec; PR 1 ("Core Broker Changes — Telegram Unlock RPC") is the natural next worker dispatch.

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/docs/vault-telegram-2fa-proposal.md
+++ b/docs/vault-telegram-2fa-proposal.md
@@ -1,0 +1,528 @@
+# Telegram-as-2FA for Vault Unlock: Design Proposal
+
+## 1. Threat Model
+
+### Current Passphrase Model
+The existing vault security rests on a single factor: the operator knows a strong passphrase. Attack surface:
+
+- **Shoulder-surfing / scrollback capture**: Passphrase was visible on terminal before #160; now masked, but still a single factor.
+- **Breakage/friction**: Operators who forget the passphrase must perform key rotation (re-encrypt entire vault with new passphrase). This friction may incentivize weak passphrases or note-taking (both worse).
+- **Host compromise (non-root)**: A process running as the same user (but outside switchroom cgroups) can read `~/.switchroom/vault.enc` and attempt offline brute-force. The scrypt KDF (N=32768, ~100ms) adds cost, but does not prevent determined offline attacks.
+- **Broker memory read**: If an attacker gains root or debugger access post-unlock, the plaintext vault in broker memory is exposed for the unlocked duration (default: 1 hour TTL).
+
+### Proposed Telegram 2FA Model
+**What it protects against:**
+- **Keyboard sniffing / scrollback**: No passphrase is entered on the terminal. Reduces shoulder-surf and scrollback risk to keystroke-logger level.
+- **Operator availability friction**: Operators don't memorize a second factor — they tap "Approve" on a device they already use. Reduces incentive for weak passphrases.
+- **Phishing**: The approval is bound to the operator's Telegram account (which runs on a separate device). Hijacking the terminal does not grant unlock ability.
+
+**What it does NOT protect against:**
+- **Host compromise (root)**: Root can still read vault.enc, intercept unlock RPCs, or extract any device-bound secret stored on the host. Root is root.
+- **Hijacked Telegram account**: If the operator's Telegram session is compromised (stolen phone, token-dumped), the attacker can approve unlocks. The 2FA only helps if Telegram is harder to compromise than the terminal user account.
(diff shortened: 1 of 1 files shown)


mekenthompson:
Closing this proposal as superseded.

The Telegram-2FA design (passwordless broker unlock with device-bound secret + Telegram approval) was originally drafted to address two concerns: (1) sparse/unattended cron access patterns, and (2) eliminating passphrase entry on the operator's terminal.

Concern 1 was solved via a different mechanism: the **capability-tokens** track shipped today as #225 / #242 / #244 / #245. Crons get long-lived signed grants (mintable via Telegram inline-keyboard wizard, revocable from chat) and don't require operator-tap or passphrase per request.

Concern 2 is acceptable as-is — operator types passphrase via `/vault unlock`, message is deleted from chat immediately, never persisted to local history.db, never scanned by secret-detect, never logged. Audited and confirmed (28 Apr 2026).

If a future operator scenario reopens the case for fully-passwordless unlock, this proposal can be revived from the git history at `docs/vault-telegram-2fa-proposal.md` (in this branch). For now, the capability-tokens flow is the canonical "Telegram-native, button-tap-approval" path.

PR #351, opened 2026-04-29
chore(cli): remove user-facing 'agent reconcile' — surface is now update+restart only

mekenthompson:
## Summary

Removes the public `switchroom agent reconcile <name>` subcommand from the CLI. The user-facing surface is now two verbs only: **`update`** and **`restart`**. Per Ken's instruction (2026-04-29).

Internal `reconcileAgent` / `reconcileAndRestartAgent` functions are **preserved** — they're still called by `restart` and `update`, so functional behavior is unchanged. Only the user-typed CLI surface shrinks.

## Drive-by bug fix

Removing standalone `agent reconcile` also closes a half-state bug discovered live during yesterday's school-cron port (2026-04-29):

`switchroom agent reconcile <name>` rendered the cron `.sh` scripts but did **not** install the corresponding systemd `.timer`/`.service` units when new crons were added to switchroom.yaml. Users got broken half-state config with no warning. With the subcommand gone, the only paths to apply config changes are `restart` (renders units) and `update` (renders + pulls + reinstalls), so the bug is no longer reachable from the CLI surface.

Note: the underlying `installScheduleTimers` function is still only called by `agent create`, not by `restart`. That's a **separate** bug and should be filed/fixed in a follow-up — adding a new cron to yaml + running `restart` should install missing unit files. Out of scope here.

## What changed

- `src/cli/agent.ts`: removes the `agent.command("reconcile <name>")` subcommand registration (~160 lines). Internal `reconcileAgent` import preserved.
- `src/cli/doctor.ts`: updates fix-it suggestions to point at `switchroom agent restart` instead of `reconcile --restart`.
- `src/cli/memory.ts`, `src/cli/vault-broker.ts`: minor cleanups of references.

## Out of scope (follow-up PR needed on telegram-plugin)

The Telegram bot still has a deprecated `/reconcile` slash command and references in:
- `telegram-plugin/server.ts:4881-4894` (deprecated handler routing to `/update`)
- `telegram-plugin/server.ts:5354` (`/help` listing)
- `telegram-plugin/server.ts:5360` (`/grant` description says "and reconcile")
- `telegram-plugin/README.md` lines 37, 290, 292
- `telegram-plugin/welcome-text.ts:219`

Will be a separate PR scoped to the Telegram surface.

## Test plan

- [ ] `bun run build` succeeds (verified locally)
- [ ] `switchroom agent reconcile clerk` returns "unknown command 'reconcile'" (verified locally)
- [ ] `switchroom agent restart <name>` still works end-to-end
- [ ] `switchroom update` still works end-to-end
- [ ] No regression in `switchroom doctor` output


mekenthompson:
Closing as already-merged. Commit 0a5917c (this PR's only commit) landed on upstream/main directly via push/cherry-pick at some earlier point — content is in v0.4.0 already. Branch can be deleted.

PR #508, opened 2026-05-01
docs(streaming): design doc for stream-json daemon mode (Path C)

mekenthompson:
## Summary

DRAFT design doc for **Path C — migrate switchroom agents to stream-json daemon mode**. The path that would deliver Claude.ai-style per-token streaming.

This PR is **doc only, no code**. The intent is to capture all the validation evidence from this session into a single forensic document so the team can make an informed call on whether to commit a 2-4 week architectural spike.

## Context

This came out of trying to figure out why the streaming UX work in PR #486 (PTY-tail through bridge → gateway) shipped but didn't deliver per-token streaming. Investigation found:

1. PR #486's PTY-tail extractor doesn't match current Claude Code TUI output — confirmed broken via fixture test in PR #507
2. The codebase comment in `session-tail.ts:5` and `pty-tail.ts:5` claiming Claude Code doesn't support `--output-format=stream-json` in `--channels` mode is **wrong as stated** — `--print --output-format=stream-json --include-partial-messages` works AND emits real Anthropic-API per-token deltas
3. But `--dangerously-load-development-channels` doesn't expose the channel's MCP tools to the model in `--print` mode, so Path C requires re-architecting inbound + outbound (it's not a flag change)

## Doc structure

10 sections:
- §1 Live validation against Claude Code 2.1.126 (5 verified findings + 1 negative)
- §2 Proposed architecture diagram (bridge-daemon owns claude lifecycle)
- §3 Per-feature impact analysis — 19 switchroom features mapped classic→daemon, 5 flagged MEDIUM/HIGH risk
- §4 Migration strategy (recommends per-agent opt-in)
- §5 Spike plan with 8 binary acceptance items
- §6-7 Open questions + risk register
- §8 Decision points for review
- §9 Alternatives considered (Path A heartbeat, Path B sidecar, V1Extractor fix, prompt-only steering, OpenClaw model)
- §10 Recommendation: ship Path A heartbeat now, scope Path C as a tracked design + spike

## What this is NOT

- Not a commitment to ship Path C
- Not advocacy — risks are listed honestly (HIGH: permission prompts, OAuth)
- Not a replacement for `docs/streaming-deterministic.md` (April 2026 notes); companion to it

## Recommendation in the doc

> Ship Path A (heartbeat placeholder) immediately. File this design doc as an architectural issue. Schedule a 1-2 week spike against the 8 acceptance items. Spike outcome is binary: green-light migration, or shelve and stick with Path A long-term.

[...]

PR #514, opened 2026-05-01
feat(streaming): heartbeat placeholder — design + §3 minimum implementation

mekenthompson:
## Summary

DRAFT design doc for **Path A — heartbeat placeholder updates**. The simpler ship-now answer to the *\"I get no experience until the status card happens\"* complaint.

This PR is **doc only, no code**. Will follow with implementation PR after design review.

## Why this exists

After PR #508's spike on Path C (stream-json daemon mode) revealed a hard blocker on permission prompts, the recommendation was to ship Path A immediately and defer Path C until the permission policy is decided. This is the design for that immediate ship.

## Key contents

- **§3 Minimum viable shape (zero extractors)** — pure elapsed-time counter on the placeholder. No new dependencies. Even if every enrichment source breaks, user always sees moving text.
- **§4 Optional enrichment via existing sources** — recall.py's update_placeholder + session-tail's tool_use events. Both already shipping in production. No new extractor introduced.
- **§9 Risk register** — no HIGH risks; failure modes degrade to today's behaviour
- **§14 Implementation estimate: 1-2 days**

## Self-review notes

I'll add the critical self-review as a PR comment after this opens. Same forensic standard I applied to the Path C doc — strengths AND weaknesses called out honestly.

### DIFF
--- a/telegram-plugin/docs/heartbeat-placeholder-design.md
+++ b/telegram-plugin/docs/heartbeat-placeholder-design.md
@@ -1,0 +1,530 @@
+# heartbeat placeholder — design
+
+Status: **DRAFT — design only, no code**.
+Author: writeup by `assistant` agent, 2026-05-01.
+Companion: `docs/stream-json-daemon-mode.md` (the deferred Path C alternative).
+
+---
+
+## 0. TL;DR
+
+The user's UX complaint: between the `🔵 thinking` placeholder appearing
+(~500ms after inbound) and the model's final reply (~5–25s later), the
+chat goes silent. No visible activity. The placeholder text doesn't
+change. Hindsight's `📚 recalling memories` → `💭 thinking` transitions
+fire only when Hindsight is configured AND only between inbound and the
+model's first content token — not during the long model TTFT or during
+tool-call waits.
+
+This document specifies a heartbeat that updates the placeholder every
+3-5 seconds with the elapsed wait time and (optionally) the latest
+semantic activity. The user always sees a moving signal — the chat
+never feels dead.
+
+The design deliberately rejects PTY-tail-style TUI extraction (the
+broken pattern from PR #486) and any new dependency on parsing
+external-system output. Every input source the heartbeat consumes
+already exists in production today. The minimum viable shape uses zero
+extractors and produces a pure elapsed-time counter; the optional
+enrichment uses session-tail (already shipping) without introducing
+new fragility.
+
+Scope: low-risk shipped-in-a-day work. ~50-100 LOC + tests. No
+architectural change.
+
+---
+
+## 1. Goal
+
+A user sending a message to a switchroom agent should see **visible
+progress** on the placeholder draft message at all times during the
+turn:
+
+- **T+~500ms**: `🔵 thinking` (existing pre-alloc)
+- **T+~3s**: `🔵 thinking · 3s`
+- **T+~6s**: `🔵 thinking · 6s` (heartbeat tick — NEW)
+- **T+~9s**: `🔧 reading CLAUDE.md · 9s` (optional — semantic enrichment)
+- **T+~12s**: `🔍 grep · 12s`
+- **T+~25s**: [final reply consumes the placeholder]
+
+The hard requirement: **no period > 5s where the placeholder text
+hasn't changed.** This is the OpenClaw "no silent gap > 2s" rule from
+issue #303 applied to the chat surface.
+
+The soft requirement: the moving text should be **informative when
+possible** (tool labels, recall progress) but degrades gracefully to
+just elapsed time when the enrichment source is unavailable.
+
+## 2. Non-goals
+
+- **Per-token streaming** — that's Path C, deferred (see
+  `stream-json-daemon-mode.md`). Heartbeat-driven updates are
+  semantic chunks, not character streams.
+- **Replacing recall.py's `update_placeholder` IPC** — that flow stays
+  as-is; heartbeat coexists with it.
+- **Parsing Claude Code's TUI rendering** — explicitly retired by PR
+  #507. Heartbeat consumes only typed events from existing sources.
+- **Server-side rendering improvements** (markdown, formatting, etc.)
+  — out of scope.
+
+## 3. The minimum viable shape (zero extractors)
+
+The simplest version: a per-chat timer that edits the placeholder
+every N seconds with the current elapsed wait.
+
+### 3.1 Lifecycle
+
+1. **Pre-alloc fires** (existing): `gateway.ts:3854` creates the
+   placeholder draft with text `🔵 thinking`. Stores chatId →
+   `{draftId, allocatedAt}` in `preAllocatedDrafts` map.
+
+2. **Heartbeat starts**: a per-chat `setTimeout` chain starts. First
+   tick fires at `HEARTBEAT_INTERVAL_MS` (default 5s) after pre-alloc.
+
+3. **Each tick**:
+   - If chat no longer has a pre-alloc entry (consumed or
+     turn-end-cleanup) → cancel timer, return.
+   - Compute `elapsedMs = Date.now() - preAllocatedDrafts.get(chatId).allocatedAt`.
+   - Format as `🔵 thinking · 5s` / `🔵 thinking · 12s` / `🔵 thinking · 1m 5s`.
+   - Edit the draft via `sendMessageDraftFn(chatId, draftId, text)`.
+   - Schedule next tick at `+HEARTBEAT_INTERVAL_MS`.
+
+4. **Heartbeat stops** (any of):
+   - Pre-alloc consumed by `reply` / `stream_reply` (existing
+     `preAllocatedDrafts.delete(chatId)` in `gateway.ts:1875`).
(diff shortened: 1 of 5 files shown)


mekenthompson:
## Critical self-review

Putting on the same critical-reviewer hat I used for #508. Strengths AND weaknesses, no advocacy.

### What's solid

1. **Clear scope ladder (§3 minimum → §4 enrichment)** — the zero-extractor minimum delivers a real UX win even if every enrichment source fails. This addresses the user's stated concern about extractor fragility head-on.
2. **Failure-mode tables in §3.4 and §4.5** — every "what if X breaks?" answer is "degrades to today's behaviour" or better. No regression vector.
3. **Performance analysis in §8 is concrete** — actual numbers (~0.25 edits/sec, well under Telegram caps), not hand-wave.
4. **§7 test plan covers all five categories** — pure formatter, lifecycle, architectural pin, e2e behavioural, integration. The architectural pin pattern was proven valuable in PR #500/507.
5. **§9 risk register has no HIGH risks** — and that's actually true, not aspirational. The worst case is "placeholder freezes like today."
6. **§13 decision points are real questions, not rhetorical** — defaults can be tuned, but the structure is clear.

### What's weak

#### Weakness 1 — §4.2 hand-waves the session-tail integration point

I describe "subscribe via the existing event ingest path (the progress-card driver currently consumes these)" but don't say HOW the subscription works. Is it a callback? A shared event bus? A second consumer of the same JSONL file? The progress-card driver is already a complex consumer — adding a second consumer might be subtle.

**Fix**: read the progress-card driver's session-tail integration before implementation, document the subscription pattern in this section, OR mark §4.2 as "deferred to spike, ship §3 minimum first."

#### Weakness 2 — §3.5 claims a "real UX win" with no measurement

I assert that even pure elapsed time delivers value. This is plausible but unmeasured. Maybe users find a counter without semantic context *more* annoying than today's static text (uncertainty signal vs no signal). I'm guessing.

**Fix**: ship §3 minimum first, gather one week of real user reaction (Ken in his own DMs), THEN decide whether to add §4 enrichment. The doc currently bundles both — should be split into two PRs.

#### Weakness 3 — coordination with `update_placeholder` is glossed

§3.4's "last-write-wins" entry says heartbeat and `update_placeholder` from recall.py race, but "next heartbeat tick re-asserts the right text shortly." That's actually a problem: between the recall.py write and the next heartbeat tick (up to 5s), the placeholder text differs from what heartbeat thinks it is. Then heartbeat overwrites recall.py's update. User might see flash: `📚 recalling memories` → `🔵 thinking · 5s`.

**Fix**: §4.1 already proposes `currentPlaceholderLabel` map — strengthen it. Heartbeat should ALWAYS read the latest label before composing text. recall.py writes to the label map (not directly to the draft); next heartbeat tick reads it. This eliminates the race. Section §6.3 implies this but §3 doesn't make it explicit.

#### Weakness 4 — no reasoning for the 5s interval default

§3 picks 5s. §13 acknowledges it could be tuned. But there's no analysis of *why* 5s vs 3s vs 7s. Is it Telegram rate limit? User perception? Battery on mobile clients? Just made up?

**Fix**: add a sub-section "why 5s" with the actual reasoning. My honest reason: it's a guess that feels right. Ship with 3s for the first week, see if anyone complains, tune from there.

#### Weakness 5 — §8 ignores compounding from `update_placeholder`

I calculated heartbeat traffic in isolation. recall.py adds 2 more edits per turn. If a future hook adds another caller (e.g., a sub-agent dispatch hook fires update_placeholder), they all stack. At some point we're at 10+ edits per chat per turn — still under cap, but worth noting the per-chat ceiling exists.

**Fix**: add a per-chat budget cap at the IPC layer ("max N edits per minute per chat"; drop edits beyond"). Or document the assumption: "operators adding new update_placeholder callers should consider the cumulative edit rate."

#### Weakness 6 — §6 doesn't address the gateway restart case

If gateway restarts mid-turn:
- `preAllocatedDrafts` map is empty (in-memory only)
- Heartbeat timer is gone
- The Telegram draft itself still exists (Telegram-side state)
- Next inbound creates a NEW pre-alloc → orphaning the old draft

§3.4 mentions this as "acceptable — boot-clear from PR #500 handles the worst case" but boot-clear only handles `turn-active.json`, not pre-alloc drafts. Telegram has zombie drafts left over.

**Fix**: investigate whether gateway boot needs a sweep of pre-alloc drafts (similar to the boot-clear of `turn-active.json`). Probably needs a `preAllocatedDrafts.json` sidecar OR explicit Telegram cleanup on boot.

#### Weakness 7 — heartbeat content vs draft transport mismatch

`sendMessageDraftFn` is the API call. But for groups/topics that don't have draft transport, the placeholder doesn't exist. Heartbeat correctly doesn't start in those cases (per §11). But the doc should call this out earlier (in §1) — heartbeat works wherever pre-alloc works, which is DMs + groups in current code, NOT forum topics. Some users may expect heartbeat in forum topics and be confused.

**Fix**: add a one-liner in §1 about scope: "wherever pre-alloc fires today (DMs + groups via PR #491; forum topics excluded pending separate work)."

#### Weakness 8 — §7.4 e2e test is vague

"shows N edits over M simulated seconds" — what N and M? Without specific numbers the test is just a placeholder.

**Fix**: spec it concretely. Example: `it('emits exactly 3 sendMessageDraft edits between t=0 and t=15s with intervalMs=5000')`.

#### Weakness 9 — no rollback plan

§10 talks about migration / rollout (purely additive, opt-out via knob). But what if the heartbeat causes an unforeseen issue in production (e.g., Telegram rate-limits the bot temporarily, or the heartbeat keeps a draft alive past its 48h Telegram editability window)?

**Fix**: add a §10.1 "rollback plan" — set the env var `SWITCHROOM_PLACEHOLDER_HEARTBEAT_MS=0` and restart agents. No code revert needed. Doc should mention this explicitly so an on-call has the path.

#### Weakness 10 — §14 implementation estimate is optimistic

"1-2 days" is a clean-room estimate. Real development includes design review, code review iterations, CI debugging, and the inevitable "oh wait the existing X integration is messier than I thought." A more honest estimate is **2-4 days** with adequate buffer. 1-2 days is what it'd take in a perfect world.

**Fix**: bump §14 to 2-4 days with the buffer rationale.

### Summary

**Quality: 8/10 as a starting point. 9/10 if Weaknesses 3 and 6 are addressed before implementation.**

The doc is honestly better than the Path C doc was on first pass — fewer load-bearing assumptions, more concrete failure analysis, no hidden recommendation bias. The main remaining issues are:

- **Weakness 3 (race coordination)** — load-bearing, must be specified before code
- **Weakness 6 (gateway restart cleanup)** — could ship without it but accumulates Telegram drafts
- **Weakness 2 (split into two PRs)** — would actually de-risk things; ship §3 first

### Recommendation for next steps

1. Fix Weakness 3 (specify the label-map coordination concretely)
2. Fix Weakness 6 (decide on pre-alloc draft boot-cleanup)
3. Optionally: split into two PRs — §3 minimum first (3 day estimate), §4 enrichment second (2 day estimate)
4. Ship the §3 minimum implementation

If the team wants to ship faster: address Weaknesses 3 and 6 as inline comments in this doc, ship one combined implementation PR for §3+§4. Trade-off: bigger blast radius if something goes wrong.

mekenthompson:
## Implementation shipped on this PR

Following the review/design/review cycle, §3 minimum implementation now ships on this branch alongside the (revised) design doc.

### Implementation summary

- **`telegram-plugin/placeholder-heartbeat.ts`** — pure helper module with formatter + lifecycle. ~220 LOC, all dependencies injected for testability.
- **`telegram-plugin/gateway/gateway.ts`** — wired `startPlaceholderHeartbeat` at pre-alloc success, `cancelPlaceholderHeartbeat` at all 3 `preAllocatedDrafts.delete` sites. `SWITCHROOM_TG_PLACEHOLDER_HEARTBEAT_MS` env var honored for rollback.
- **49 new tests**: 40 lifecycle/formatter cases, 6 architectural pin cases, all passing.

### Doc revisions (commit `ef991c5`)

Addressed 9 of 10 self-review weaknesses from earlier comment:
- Weakness 3 (race coordination) → §4.6 explicit single-source-of-truth pattern
- Weakness 4 (5s rationale) → §5.1 three-constraint analysis
- Weakness 5 (per-chat budget) → §8.0 numerical analysis
- Weakness 6 (gateway restart cleanup) → §3.4 documented as unchanged-from-today
- Weakness 7 (forum-topic scope) → §1 explicit
- Weakness 8 (concrete e2e specs) → §7.4 N=3 calls between t=0 and t=15s
- Weakness 9 (rollback playbook) → §10.1 operator instructions
- Weakness 10 (realistic estimate) → §14 split into PR 1 + PR 2 with 2-3 day each
- Weakness 2 (split into two PRs) → §14 explicitly proposes; this PR is PR 1 (§3 minimum)

### Verification

```
$ npm run lint        # clean
$ cd telegram-plugin && bun test
…
 2914 pass / 0 fail across 149 files (up from 2865 / 147)
```

### Next step

After merge: build + restart agents, verify the heartbeat ticks visibly in DMs by sending a tool-using prompt.

Issue #546, opened 2026-05-01
bug: Telegram replies posted twice — one formatted, one raw


mekenthompson:
## Symptom

Some agent turns post **two messages** to Telegram for a single reply: one rendered with HTML formatting (markdown converted, code blocks, bold, etc.) and a duplicate with raw/unformatted text. User sees both.

Observed in DM chats. Specific occurrence pattern not yet pinned down — could be tied to `stream_reply` finalisation, parallel `reply` + answer-stream materialisation, or a race between MCP tool and gateway answer-stream paths.

## Prior work

- PR #357 (merged, debug branch) traced every Telegram API write to identify DM dup source. Findings unclear from memory — needs re-read.
- PR #253 retracts answer-stream preview when reply path wins; possibly incomplete coverage.
- PR #523 awaits prior draft stream finalize on fresh-turn supersession (related supersession path).
- Epic #424 covers silent-failure RCA — this bug fits there.

## Required work

1. **Deep RCA** — instrument enough to reconstruct which code path posted each of the two messages in a real reproduction. Read PR #357's debug findings, PR #253, PR #523, and the current `gateway.ts` / `bridge.ts` / `progress-card-pin-manager.ts` / answer-stream materialisation code. Identify the exact race / dual-path that lets two writes through for one logical reply.

2. **Test coverage** — add failing tests that reproduce the dup at the unit and integration level before the fix. The bug should be expressible as a test that fails on main and passes after the fix. Cover both DM and forum-topic chats if the symptom differs.

3. **Fix** — minimal code change that resolves the race / dedupes the path. No backwards-compat shims; if there's a wrong path, delete it.

4. **Verify** — `bun run build` clean, all tests pass, manual reproduction script confirms single message posted.

## Out of scope

- Don't refactor the gateway broadly; minimal targeted fix.
- Don't add feature flags.

## Context for reviewer

Affects every user. High signal/noise and trust impact — duplicate messages make the bot look broken. Should be near top of the bug queue.

🤖 Filed by clerk

mekenthompson:
## Fresh forensic incident — 2026-05-02 08:27 AEST

Reproduced live in chat \`8248703757\` while debugging #550 (watchdog false positive).

**Timing:**
- 08:27:58 — assistant turn started in response to user msg #643 ("I didn't trigger it, debug why")
- Assistant emitted preamble text: \`Found it. Watchdog did fire.\` (one short line, as guidance recommends before tool calls)
[...]

PR #622, opened 2026-05-03
hotfix(gateway): declare outboundDedup (every reply was ReferenceError-ing)

mekenthompson:
## Production smoking gun

From clerk's gateway.log:

\`\`\`
telegram channel: reply: invoked chatId=... preview=\"stream_reply just threw \`outboundDedup is not defined\` — looks like a regression\"
...
preview=\"Tried twice — both \`stream_reply\` and \`reply\` threw \`outboundDedup is not define...\"
\`\`\`

User-facing symptom: agent thumbs-ups and doesn't respond. Every \`reply\` / \`stream_reply\` tool call threw at runtime. The agent's prose then quoted the error inside ANOTHER reply call — which also threw.

## Root cause

PR #599 (#546 fix, commit \`5bed5b7\`) added 4 read sites of \`outboundDedup.check\` / \`.record\`:
- \`executeReply\` line 2111
- turn-flush line 2433
- \`executeStreamReply\` lines 2452, 2579
- turn-flush line 3770

…but never added the module-level declaration \`const outboundDedup = new OutboundDedupCache()\`. The import at line 51 brings in the class; nobody instantiates it. Every read throws \`ReferenceError\`.

Latent in main since 2026-05-03 (when #599 merged). Just hit clerk because clerk had a fresh gateway restart and the agent tried to reply.

## Why testing missed it

- **tsc --noEmit didn't catch it.** Need to investigate why — undeclared identifiers should fail strict mode. Hypothesis: a global type-augmentation or implicit any pattern in this file masks it. Filing a follow-up to enable a stricter check.
- **No test imports gateway.ts directly.** The 7000-line module has no integration test that exercises the IPC tool-call dispatch path. The dedup cache has 23 unit tests on the cache itself; the dedup-suppress controller branch has isolated tests; the actual call site path through \`executeReply\` has zero runtime coverage.
- **No CI smoke that exercises the IPC flow.** The waiting-UX harness drives session events but not tool_calls.

## Fix

Declare the module-level singleton at the same state block as \`activeStatusReactions\`, etc. Three lines.

## Follow-ups (will file separately)

1. Regression test: load gateway.ts, simulate an inbound + reply tool call, assert no ReferenceError.
2. tsconfig audit: figure out why tsc didn't catch this and tighten the rule.


mekenthompson:
Superseded by #625 — the original branch was based on feat/global-auth-accounts which swept in 4 unrelated auth commits. PR #625 is a clean cherry-pick onto main with only the hotfix + lint check. Closing this one.

PR #669, opened 2026-05-04
telegram(plugin): /auth v3a — accounts-first layout + per-account drill-down

mekenthompson:
## v3a — accounts-first /auth dashboard layout + per-account drill-down

Reorders the `/auth` Telegram dashboard around the account-pool mental model: shared accounts are the primary unit, slots are demoted. Per-account actions (Reauth/Remove) move into a drill-down sub-view to keep the main board scannable.

### What ships

- `buildDashboardText` renders accounts above slots; `Anthropic accounts (N)` header, per-account rows with health badge, then `Slots (M)` section
- `buildDashboardKeyboard` reordered: account drill-down buttons → slot rows (unchanged) → pending-flow recovery → quota row → refresh
- 4 new `CallbackAction` kinds: `account-view`, `account-rm`, `account-rm-confirm`, `account-reauth`
- 3 new exported builders: `buildAccountSubViewText`, `buildAccountSubViewKeyboard`, `buildAccountRemoveConfirmKeyboard`
- Old `account-enable` / `account-disable` callbacks preserved for back-compat with on-disk messages
- `account-reauth` shows a toast "Reauth not yet wired — coming in v3b" — the underlying CLI verb doesn't exist yet
- All buttons apply the existing 64-byte `callback_data` budget guard
- Sub-views use `editMessageText` / `editMessageReplyMarkup` (no new messages); single `answerCallbackQuery` per path; CLI failure on remove early-returns so the user sees the error OR a refresh, not both

### Deferred to v3b

- Per-account reauth flow (needs `auth account reauth <label>` CLI verb)
- Per-account quota numbers in the dashboard body (needs cache layer; quota fetch is expensive)
- Account rename (CLI verb + Telegram input flow)
- Fallback history surface (needs durable backing store; current is in-memory)
- Soonest-resetting fallback strategy (current is first-healthy-in-listing-order)

### Tests

81/81 passing in `auth-dashboard.test.ts`. New tests cover the drill-down round-trip, cancel path, remove confirm, long-label budget fallback, and account-not-found path.

### DO NOT MERGE without operator approval — UX redesign

### DIFF
--- a/src/auth/account-store.ts
+++ b/src/auth/account-store.ts
@@ -285,49 +285,6 @@ export function removeAccount(label: string, home: string = homedir()): void {
   rmSync(accountDir(label, home), { recursive: true, force: true });
 }
 
-/* ── Rename ──────────────────────────────────────────────────────────── */
-
-/**
- * Rename an account on disk by moving its directory.
- *
- * Both labels are validated. Refuses when the source doesn't exist or
- * the destination already does (to avoid silent merges that would
- * lose the destination's credentials).
- *
- * Atomic at the filesystem level: rename(2) is atomic within a single
- * filesystem, and accounts always live under one parent dir.
- *
- * The caller is responsible for downstream side effects:
- *   1. Rewriting `agents.<name>.auth.accounts` lists in switchroom.yaml
- *      to swap the old label for the new one.
- *   2. Re-fanning out credentials to enabled agents (the per-agent
- *      mirror at `<agentDir>/.claude/credentials.json` is unchanged
- *      content-wise — just the source-of-truth label moved — so a
- *      restart isn't strictly required, but a fanout pass keeps the
- *      broker's view consistent).
- */
-export function renameAccount(
-  oldLabel: string,
-  newLabel: string,
-  home: string = homedir(),
-): void {
-  validateAccountLabel(oldLabel);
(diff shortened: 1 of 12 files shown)


mekenthompson:
Follow-up commits address the reviewer punch list:

- **#1 (blocker, double `answerCallbackQuery`)** — `account-view` not-found path now calls `answerCallbackQuery` exactly once, with the error text. Found path: empty toast + edit. (folded into commit f4de099a)
- **#2 (blocker, rm error path)** — `account-rm-confirm` runs the CLI directly and early-returns on any error before calling `sendAuthDashboard`. User sees error OR refresh, not both. (folded into commit f4de099a)
- **#3 (scope drift)** — entire `account-rename` feature reverted (commit 07bd32da). 9 out-of-scope files restored to main.
- **#4 (Cancel button guard)** — `buildAccountRemoveConfirmKeyboard` Cancel button now applies the same `Buffer.byteLength <= CALLBACK_BUDGET_BYTES` guard as the Yes button (commit f4de099a).
- **#5 (not-found test)** — added test in `auth-dashboard.test.ts` for the `account-view` deleted-account branch (commit 23bfc746).

Final diff: 3 files, all in telegram-plugin scope. 81/81 tests pass.

Still draft.

PR #677, opened 2026-05-04
refactor(progress-card): unified close path + convergence test

mekenthompson:
## Summary

Collapses the three per-chat close paths in `progress-card-driver.ts` (`completeTurnFully`, `closeZombie`, Gap-8 stalled-close) into a single parametrised `closePerChat(cs, reason: 'turn-end' | 'zombie' | 'stalled')`.

This is **PR-C1** of a 2-PR series — refactor + convergence test only. PR-C2 will follow with the broader test backfill (bg-carry lifecycle, pin failure paths, edit-timestamps budget, dispose preservePending, sidecar partial-write, snapshot extras, 2-threads-in-same-chat, memory bounds).

Stacks on top of #673 (firePin/phaseFor) and #674 (eviction off heartbeat), both already merged.

## What changed

- New `export type CloseReason = 'turn-end' | 'zombie' | 'stalled'`
- New `closePerChat(cs, reason)` unifies prelude (timer cleanup, force-close subagents for zombie/stalled, `prepareSilentEndSuppression`, `flush(forceDone, stalledClose=reason==='stalled')`, and `completeTurnFully` tail).
- `closeZombie` retained as a thin alias for call-site readability.
- All 5 close call-sites routed through `closePerChat`: heartbeat zombie ceiling, Gap-8 stalled, `forceCompleteTurn`, ingest `turn_end` no-subagents path, `maybeCompleteDeferredTurn`.

## Invariants preserved

- Zombie passes `stalledClose=false`; stalled passes `stalledClose=true`. The header rendering distinction stays.
- Zombie does NOT clear `pendingSyncEchoes` — late echoes after zombie close still resolve.

## Tests

New `tests/progress-card-close-paths-converge.test.ts` — 5 cases asserting all three close reasons converge on the same final state (timers null, chat deleted, heartbeat stopped if last, baseKey cleaned).

- `tsc --noEmit`: clean
- `check-plugin-references.mjs`: clean
- Targeted `progress-card-*` + `two-zone-*`: 145/145 pass
- Full `vitest run`: 4569 pass + 39 pre-existing unrelated flakes (boot-self-test, bridge-watchdog, cli.issues — matches the allowlist confirmed on PR-B)

## Follow-up

PR-C2 will add the 10 remaining tests from the original test-backfill plan.

### DIFF
--- a/telegram-plugin/progress-card-driver.ts
+++ b/telegram-plugin/progress-card-driver.ts
@@ -51,6 +51,18 @@ import {
 export type ApiFailureKind = 'permanent_4xx' | 'transient' | 'benign'
 
 /**
+ * Reason a per-chat card is being closed. Used by the unified
+ * `closePerChat` helper to drive the small set of behavioural deltas
+ * between paths (sub-agent force-close, stalled-render flag).
+ *
+ *   - 'turn-end' : normal completion — no in-flight sub-agents.
+ *   - 'zombie'   : abandonment via heartbeat maxIdle ceiling or
+ *                  new-enqueue force-close.
+ *   - 'stalled'  : Gap-8 deferred-completion timeout expired.
+ */
+export type CloseReason = 'turn-end' | 'zombie' | 'stalled'
+
+/**
  * Failure descriptor reported back to the driver after an async emit fails.
  * The outer layer (server.ts) inspects the raw Telegram error and classifies
  * it before calling `reportApiFailure`.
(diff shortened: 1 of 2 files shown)


mekenthompson:
Reviewer flagged a minor behavior-expansion to document: the two no-subagents `turn-end` call-sites (`forceCompleteTurn` no-subagents path and ingest `turn_end` no-subagents path) now also invoke `prepareSilentEndSuppression` + an explicit `flush(forceDone=true)` via `closePerChat`. Previously they jumped straight to `completeTurnFully` without those. `prepareSilentEndSuppression` is idempotent (no harm); the extra `flush` is a no-op when the reducer's prior emit already rendered the terminal frame. Intentional unification — flagging for archaeologists.

PR #760, opened 2026-05-06
fix(telegram): materialize bot token from vault at runtime (#758)

### LINKED ISSUE #758: Materialize TELEGRAM_BOT_TOKEN from vault at runtime; eliminate plaintext .env at rest
## Problem

The Telegram gateway reads \`TELEGRAM_BOT_TOKEN\` from \`<agent>/telegram/.env\` at startup and hard-fails (\`TELEGRAM_BOT_TOKEN required\`) if the file is missing. The vault-reference resolver (\`src/vault/resolver.ts\`) is shipped and \`switchroom.yaml\` configs already use \`vault:<key>\` references for \`bot_token\`, but the resolver only runs on the **install/scaffold** path (\`writeAgentEnv\` in \`src/cli/onboarding.ts:163\`). On agent restart, no resolution happens — the gateway just reads the on-disk \`.env\`.

This means:
- The plaintext token still lives at rest in \`<agent>/telegram/.env\` even when configured as a vault reference.
- Backups, snapshots, and the trash dir capture plaintext tokens. (We just found one in \`~/.trash/klanker-migration-20260424-110807/...\` with a literal token.)
- You can't actually delete the \`.env\` and recover via restart — the gateway hard-fails. Confirmed empirically: gymbro's gateway hit \`start-limit-hit\` after 10 systemd retries when its \`.env\` was removed.

The recently-merged RFC A (PR #756) was framed as \"complete the operational migration\" on the assumption the runtime path already pulled from vault. It doesn't. RFC A's premise is wrong; that PR is essentially a no-op until this issue ships.

## Expected behaviour

On agent start:
1. If \`TELEGRAM_BOT_TOKEN\` is set in environment, use it.
2. Otherwise, look up \`bot_token\` from the agent's resolved switchroom config.
3. If it's a \`vault:\` reference, resolve it via the broker (or \`--no-broker\` direct read if broker unreachable).
4. Inject the resolved token into the gateway process environment. Do **not** write it to \`.env\`.
5. If vault is locked / unreachable, fail with a clear error pointing the user at \`switchroom vault unlock\` — not at \"set it in .env\".

After this ships, \`<agent>/telegram/.env\` should be deletable for any agent whose config uses a \`vault:\` ref, and restart should succeed.

## Out of scope
- Other secrets (Drive OAuth, etc.) — separate issue.
- Removing existing \`.env\` files automatically — that's an operator decision after this lands.
- Approval kernel (PR #756 RFC B).

## Acceptance
- [ ] Restarting any agent whose \`bot_token\` is a \`vault:\` ref succeeds with no \`.env\` file present.
- [ ] Gateway never writes the resolved token to disk.
- [ ] Clear error message when vault is locked at agent-start time.
- [ ] Existing agents with plaintext \`.env\` still work (env-set takes precedence).
- [ ] Test coverage on the \"vault unreachable\" failure path.

## Context
- Discovered while executing RFC A operational migration on Ken's deployment, 2026-05-06.
- Resolver exists at \`src/vault/resolver.ts\`; gateway entry point reads \`.env\` directly (grep for \`TELEGRAM_BOT_TOKEN required\` for the failure call-site).

🤖 Filed by Claude (klanker)

mekenthompson:
## Summary

Closes #758. Telegram gateway and foreman now resolve `TELEGRAM_BOT_TOKEN` from vault at startup instead of requiring a plaintext `.env` file. Token stays in-memory; `.env` becomes optional for any agent whose config uses a `vault:` reference.

## Changes
- New `src/telegram/materialize-bot-token.ts` — env → config literal → broker → direct-decrypt fallback. In-memory only. Custom `BotTokenMaterializeError` with reason codes (locked / unreachable / denied / not_found / config / unknown).
- `gateway.ts` and `foreman.ts` wired to call the new resolver before the legacy `.env` check; clear error pointing to `switchroom vault unlock` when locked.
- Existing `writeAgentEnv` left untouched (separate follow-up to gate it on `isVaultReference`).

## Reviewed
Fresh Opus reviewer: APPROVE. Two non-blocking nits filed as follow-ups.

## Test plan
- [x] 9 new tests pass (`materialize-bot-token.test.ts`)
- [x] `npm run lint:tsc` clean
- [x] Full vitest run: 292 pass / 4 fail — same 4 baseline failures as `origin/main` (unrelated `execFileSync(BUN, …)` tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## [Unreleased]
 
+### Added
+
+- **Webhook ingest hardening (#714)** — two defenses added to
+  `src/web/webhook-handler.ts` before auto-dispatch ships:
+  - **Dedup by `X-GitHub-Delivery`**: per-agent LRU (1000 entries, 24h
+    retention) backed by `~/.switchroom/agents/<agent>/telegram/webhook-dedup.json`.
+    Replay returns 200 `{ok:true,deduped:true}` and skips JSONL append.
+    Generic source has no delivery header — dedup is skipped silently.
+  - **Per-source token-bucket rate limit**: default 60 rpm (burst 60),
+    configurable via `channels.telegram.webhook_rate_limit.rpm` in
+    switchroom.yaml. Exceeding the limit returns 429 with `Retry-After`.
+    First throttle event per `(agent, source)` per 60s window is written
+    to `<agent>/telegram/issues.jsonl` for Telegram visibility.
+  - `webhook_rate_limit` added to `TelegramChannelSchema` in
+    `src/config/schema.ts`; cascades via the existing channels deep-merge.
+
 ## v0.6.7 — 2026-05-05
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,10 @@
     retention) backed by `~/.switchroom/agents/<agent>/telegram/webhook-dedup.json`.
     Replay returns 200 `{ok:true,deduped:true}` and skips JSONL append.
     Generic source has no delivery header — dedup is skipped silently.
-  - **Per-source token-bucket rate limit**: default 60 rpm (burst 60),
-    configurable via `channels.telegram.webhook_rate_limit.rpm` in
-    switchroom.yaml. Exceeding the limit returns 429 with `Retry-After`.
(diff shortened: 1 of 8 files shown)


mekenthompson:
Superseded by clean rebase — see new PR.

PR #762, opened 2026-05-06
feat(approval-kernel): IPC broker + SQLite kernel + Telegram card primitive (RFC B Phase 1)

mekenthompson:
## Summary

Implements **RFC B Phase 1** — the approval kernel: a single source of truth for human-in-the-loop approvals across the Switchroom surface area.

**What's in:**
- **Kernel core** (`src/vault/approvals/`) — SQLite-backed approval store with canonical request schema (zod-validated), TTL/expiry, idempotent decisions, audit trail.
- **Broker IPC** (`src/vault/broker/`) — `approval.create`, `approval.list`, `approval.decide`, `approval.get` request types over the existing peercred-authenticated unix socket. Rate caps applied per-client.
- **Telegram card primitive** (`telegram-plugin/gateway/approval-card.ts`) — renders a pending approval as an inline-keyboard card; `approval-callback.ts` routes Approve/Deny taps back to the kernel.
- **`/approvals` commands** (`approvals-commands.ts`) — list/decide from chat without leaving Telegram.
- **Migration doc** (`src/vault/approvals/MIGRATION.md`) — playbook for moving legacy approval callsites.

**What's deferred (Phase 2, per RFC §12):**
- Migration of legacy approval callsites (grants resolver, vault unlock prompt, etc.) onto the kernel — tracked separately.
- Short-poll wait helper for CLI consumers (blocking \`approval.wait\` API) — RFC C reconciler depends on this; will land as a follow-up issue.

## Reviewer verdict

APPROVE (after fix-loop addressing schema, rate caps, and status collision blockers — captured in the second commit).

## Tests

- `bun test src/vault/` → **190 pass / 20 fail / 2 errors** (matches pre-RFC-B baseline; the 20 failures are unrelated legacy flakes).
- `npm run lint:tsc` → clean for all touched files.

## Branching note

This is a clean cherry-pick of `feat/approval-kernel-rfc-b` (`35e5bc27`) onto `upstream/main` to avoid divergence in the mekenthompson fork.

PR #770, opened 2026-05-06
fix(gateway): tighten materializeBotToken catch + align approval-callback signature

mekenthompson:
Two reviewer/worker-found fixes in already-shipped code:

1. **#761 follow-up**: outer try/catch in `gateway.ts` and `foreman.ts` only swallows `ERR_MODULE_NOT_FOUND` now. Programming errors in materializeBotToken propagate with original messages instead of being masked by '.env' fallback.

2. **#762 latent bug**: `approval-callback.ts` was calling `approvalRecord` with the pre-RFC-B-v2 signature (`granted: bool`, `approver_user_id`). Updated to v2 signature (`decision: ApprovalDecisionMode`, `granted_by_user_id`). Bug didn't fire in production yet because no callsite wires `apv:*` taps to real flows; would have on next migration wave.

## Test plan
- [x] tsc clean for changed files
- [x] approval kernel tests still pass

### DIFF
--- a/telegram-plugin/foreman/foreman.ts
+++ b/telegram-plugin/foreman/foreman.ts
@@ -116,9 +116,28 @@ try {
 // ─── Bot token ────────────────────────────────────────────────────────────
 // Issue #758: when bot_token is a `vault:` ref and no .env was written,
 // materialize it from the vault at startup (in-memory only).
-let TOKEN: string
+//
+// The outer try/catch is narrowed (post-#761 review) to ONLY catch
+// ERR_MODULE_NOT_FOUND from the dynamic import. Other errors (e.g. throws
+// from inside materializeBotToken that aren't BotTokenMaterializeError)
+// must propagate so we don't mask real bugs behind the legacy "set in .env"
+// hint.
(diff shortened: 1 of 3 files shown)


mekenthompson:
**APPROVE** (Reggie / automated review)

- **foreman.ts + gateway.ts**: Catch narrowing is correct — `ERR_MODULE_NOT_FOUND` and `MODULE_NOT_FOUND` (legacy Node alias) are the only codes swallowed; everything else re-throws with original message. Two-phase pattern cleanly separates dynamic-import failure from `materializeBotToken` errors. No issue.
- **approval-callback.ts**: `approvalRecord` call now matches the v2 signature exactly — `decision: ApprovalDecisionMode` (replaces `granted: boolean`) and `granted_by_user_id: number` (replaces `approver_user_id: string`, `telegram-plugin/gateway/approval-callback.ts:92`). All four switch cases assign `decision`; TypeScript exhaustiveness holds via the discriminated union. `approver_set` correctly stays `string[]` via `String(granted_by_user_id)`.
- **Minor**: error hint text simplified to `"(token-materialization helper not found)"` — intentional and accurate for the `null` module path.

Safe to merge.

Issue #714, opened 2026-05-05
webhook ingest: x-github-delivery dedup + per-source rate limit


mekenthompson:
## Problem

The webhook ingest path in `src/web/webhook-handler.ts` (added in #577) ships without two defenses that are needed before any auto-dispatch flips on:

1. **No replay/duplicate protection.** GitHub redelivers (network blip, manual replay); handler currently appends a fresh JSONL line for each. Any downstream auto-trigger would fire twice.
2. **No rate limit.** A misconfigured webhook source, or a leaked secret, could fire thousands of POSTs/min — each invokes HMAC verify, JSONL append, and (post-dispatch-feature) a `claude -p` spawn. Subscription quota disappears fast.

## Scope

- **Dedup by `X-GitHub-Delivery`**: track last N (default 1000) IDs in-memory + on-disk LRU at `~/.switchroom/agents/<agent>/telegram/webhook-dedup.json` (24h retention). On hit: return 200 with `{ok:true, deduped:true}`, skip JSONL append.
- **Per-source rate limit**: token-bucket per `(agent, source)`. Default 60/min, configurable via `channels.telegram.webhook_rate_limit` in switchroom.yaml. Beyond cap: 429 + first throttle event written to `issues.jsonl`.
- Unit tests for both.

## Out of scope

- Dispatch / auto-trigger (separate issue).
- Per-event filtering or templating (separate issue).

## Acceptance criteria

- [ ] Replayed delivery → 200 deduped, no second JSONL line.
- [ ] Burst >60/min → 429 with `Retry-After`; first throttle event lands in `issues.jsonl`.
- [ ] Existing `src/web/webhook-*.test.ts` still passes.
- [ ] CHANGELOG entry under `[Unreleased]`.

mekenthompson:
Tracked under epic #717.

mekenthompson:
Implementation in PR #718

PR #774, opened 2026-05-06
fix: cron DM routing for dm_only agents

mekenthompson:
## Summary
- Both `scaffoldAgent` and `reconcileAgent` were passing `telegramConfig.forum_chat_id` to `buildCronScript`, so scheduled tasks for `dm_only` agents (e.g. clerk) routed output to the forum supergroup instead of the user's DM.
- Now both call sites use the `userId ?? forum_chat_id` fallback (`cronChatId` / `reconCronChatId`).
- Adds `src/agents/cron-dm-routing.test.ts` to pin the contract.

## Notes on Bug 2 (vault broker for python skills)
The brief also mentioned that `mail.py` and `ms_graph_token.py` called `switchroom vault get --no-broker`. Inspection of the current files at `/home/kenthompson/.switchroom/skills/{mail,calendar}/scripts/` shows no `--no-broker` flag — they already use plain `switchroom vault get`, which auto-uses the broker socket from `$SWITCHROOM_VAULT_BROKER_SOCK`. No skill edits required.

## Test plan
- [x] `npx vitest run src/agents/` — 226/226 pass
- [x] New test `cron-dm-routing.test.ts` covers both `userId-present` and `userId-undefined` cases


mekenthompson:
**REQUEST_CHANGES** — Reggie (code reviewer)

**Cron fix (scaffold.ts:1925, 3059–3061):** Correct. Both call sites now use `userId ?? forum_chat_id` instead of passing `forum_chat_id` directly. The pre-existing `cronChatId` variable at line 1921 confirms the fix is a straightforward one-liner that was already computed but unused.

**Bug — dedup/rate-limit ordering (`webhook-handler.ts:352–370`):**
`dedupStore.check()` has a side effect: it registers the delivery ID on first call. But this happens *before* the rate-limit check. If a first-seen delivery is rate-limited (returns 429), its ID is already stored in the dedup store — GitHub's retry with the same delivery ID will return 200/deduped and the event is silently dropped forever. Fix: move the rate-limit check before the dedup check, so a rejected delivery is never registered.

**Scope mismatch — 741 of 798 changed lines are undescribed:**
PR title/body cover only the cron DM fix, but `webhook-handler.ts` (+199 lines), `webhook-handler.test.ts` (+485 lines), `schema.ts` (+16 lines), and the `CHANGELOG.md` entry are all webhook hardening (#714 work) with no mention in the PR description. The CHANGELOG `[Unreleased]` section describes webhook changes only — the cron fix has no entry. Either split the PR or describe the webhook work and add a cron entry to CHANGELOG.

**Weak test for the core fix (`cron-dm-routing.test.ts:54–57`):**
`resolveCronChatId` is a local helper defined inside the test file, not imported from `scaffold.ts`. A regression at the actual call sites (`scaffoldAgent`, `reconcileAgent`) would not be caught by this test. Consider testing those functions directly, or at minimum importing the routing logic from the module under test.

PR #798, opened 2026-05-07
rfc(docker): r2 — address security BLOCK + ops + implementation findings

mekenthompson:
Substantial revision of the Docker multi-container RFC addressing all findings from three independent Opus reviewers (security BLOCK + operational + implementation REQUEST_CHANGES).

## Headline changes

**Container identity model** (security BLOCK fix). Original RFC's Phase 0 asked the wrong question — the broker ACL gates on `peer.systemdUnit` (cgroup-derived), structurally absent in containers. Redesigned around per-agent unix-socket directories at `/run/switchroom/broker/<agent>/sock` with mode 0700 and per-agent UIDs. Identity resolved via `getsockname()`, no HMAC tokens, file permissions enforce the boundary. Same model for the approval kernel.

**Scheduler architecture** (implementation gap). Compose has no native scheduler. New singleton scheduler container reads the existing config cascade and dispatches via `docker exec`, owns its own audit DB, doesn't hold vault material.

**Process tree clarification.** Was described as 4-way supervision (tini + start.sh + tmux + claude + gateway). Reality is single-spine: `tini → tmux → start.sh → claude → MCP children (incl. gateway)`. RFC updated.

## Operational fixes
- `stop_grace_period: 45s` in compose snippet (Docker default 10s would SIGKILL mid-drain)
- Resource defaults updated from measured RSS: klanker 6g, conversational 1.5g
- Watchdog fixture target raised from 15+ to 30+ with explicit trap-zone, rate-cap, fingerprint, mmin coverage
- Mac/Win wake-audit + log-grep gets a non-journald path (stdout-tailing shim → bind-mounted JSONL)
- Phase 4 update flow commits to rolling per-agent restart, not parallel recreate
- Migration tool gets bidirectional `chown` for round-trip support

## Effort re-baseline
~2770 agent-min Opus 4.7 (was 2000): P0 180, P1 600, P2 280, P3 750, P4 360, P5 600.

## Review
Independent fresh Opus reviewer: APPROVE. All 14 actionable fixes verified landed correctly with file:line citations. No new defects introduced. Engineering content not in the fix list unchanged.

### DIFF
--- a/reference/rfc-docker-multi-container.md
+++ b/reference/rfc-docker-multi-container.md
@@ -84,26 +84,87 @@ The OpenClaw wedge — stock `claude` CLI under your Pro/Max subscription, not a
                  │   └────────────────┘  │ container   │  │
                  │                       │ (singleton) │  │
                  │   ┌────────────────┐  └─────────────┘  │
-                 │   │ agent-…        │                   │
-                 │   └────────────────┘                   │
+                 │   │ agent-…        │  ┌─────────────┐  │
+                 │   └────────────────┘  │ switchroom- │  │
+                 │                       │ cron        │  │
+                 │                       │ (singleton, │  │
+                 │                       │  docker.sock)│ │
+                 │                       └─────────────┘  │
                  └────────────────────────────────────────┘
 ```
 
-### Per-agent container
+### Per-agent container internals
 
-PID 1 is `tini`. tini reaps zombies, forwards signals, and exits clean. Under tini, `start.sh` (rendered from `profiles/_base/start.sh.hbs`, today 383 lines) does what it does now: env setup, tmux session spawn, claude attach. Inside tmux, `claude --continue` runs against `CLAUDE_CONFIG_DIR=/state/.claude/<agent>`. The telegram gateway runs as a sibling process inside the same container — same as today's per-agent `telegram-plugin/server.ts` (171 lines). Same FIFO queue, same stream-reply path, same tmux-send-keys interrupt handler.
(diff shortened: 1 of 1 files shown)


mekenthompson:
**Reggie (automated review) — REQUEST_CHANGES**

Security BLOCK resolved. Per-agent socket directory model is sound: `getsockname()`-derived identity, OS permissions as gatekeeper, compose generator enforcement + unit test + doctor check as layers. Phase 0 matrix redesign is better than the old peercred matrix. Operational gaps (grace period, resource defaults, scheduler, log abstraction, round-trip migration) all addressed with specifics.

Two new findings not currently in the Risks section:

**1. Scheduler + broker root privilege** (`reference/rfc-docker-multi-container.md`, compose snippet ~L212)

`switchroom-cron` runs `user: "0:0"` with `/var/run/docker.sock` mounted. Docker socket access is host-root-equivalent regardless of mount flags — the scheduler can issue arbitrary Docker API commands (not scoped to exec-only). `vault-broker` also runs `0:0` for its entire lifetime to `chown` per-agent directories; an init-container or entrypoint privilege-drop would scope that to setup only. Neither is mentioned in the Risks section. Operators need to understand this trust boundary before implementation.

**2. `docker.sock:ro` is cosmetic** (`reference/rfc-docker-multi-container.md`, compose snippet ~L213)

The `:ro` flag on the scheduler's socket mount prevents unlinking the socket file on the filesystem but does not restrict Docker API operations (exec, start, stop, pull). The Risks section should clarify the actual privilege level so the HMAC-fallback threat model is correctly framed if that path is ever needed.

**Fix:** One paragraph in the Risks section covering both. No redesign needed — scheduler-needs-docker-exec is a known constraint of the singleton-scheduler-via-docker-exec architecture choice. Document it honestly so operators can make an informed decision (e.g. whether to run a docker-socket-proxy in front of the scheduler on sensitive hosts).

PR #841, opened 2026-05-08
feat(progress-card): plain-text parent rows, drop tool-name prefix and <code> styling

mekenthompson:
## Summary

Parent zone of the two-zone progress card now renders rows as bare human text. The raw tool name and the monospace `<code>` wrapping around the label are gone — the label IS the row.

When a tool emits no label (TodoWrite, Task housekeeping, etc.) we fall back to a humanised tool name from a small map in `tool-labels.ts`; `mcp__*` tools fall back to the existing "Server: action" form via a newly-exported `mcpDisplayName`.

The `◉` glyph alone signals the in-flight row — dropped the `<b>` on the prefix.

## Before / after

```
Before: ● Bash <code>Patch UML to guard hass.resources</code>
After:  ● Patch UML to guard hass.resources

Before: ◉ <b>Bash</b> <code>Reloading HA core</code>
After:  ◉ Reloading HA core

(no label, common tool)
After:  ● updating tasks

(no label, mcp tool)
After:  ● Telegram: reply
```

Label is HTML-escaped (it's now bare text in an HTML message). The truncate-to-80 cap stays.

## Test plan

- [x] `telegram-plugin/tests/two-zone-snapshot-extras.test.ts` updated and extended (label-present row, in-flight label row, no-label common-tool fallback, no-label mcp tool, HTML-escape).
- [x] `npx vitest run telegram-plugin/tests/two-zone-*` — all 40 tests green.
- [x] `npm run lint:tsc` clean.
- [x] Pre-existing failures on `main` in `tests/cli.issues.test.ts`, `tests/boot-self-test.test.ts`, `tests/run-hook.test.ts` are unrelated to this change.

### DIFF
--- a/telegram-plugin/tests/two-zone-snapshot-extras.test.ts
+++ b/telegram-plugin/tests/two-zone-snapshot-extras.test.ts
@@ -100,13 +100,16 @@ describe('PR-C2: two-zone card snapshot extras', () => {
     // 12 items, cap 8 → 4 hidden.
     expect(out).toContain('(+4 earlier)')
     // The visible bullets are the LAST 8 (slice(-8) → f4..f11).
-    expect(out).toContain('<code>f11.ts</code>')
-    expect(out).toContain('<code>f4.ts</code>')
-    // f3 (the latest hidden) must not appear as a bullet code block.
-    expect(out).not.toContain('<code>f3.ts</code>')
+    // f11 is the in-flight bullet (stage=run, last index) → ◉.
+    expect(out).toContain('◉ f11.ts')
+    expect(out).toContain('● f4.ts')
+    // f3 (the latest hidden) must not appear as a bullet.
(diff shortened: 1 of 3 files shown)


mekenthompson:
Rebased `ux/card-plain-rows` onto `upstream/main` — the apparent webhook-hardening drift was fork divergence (upstream already has #714 merged in a different shape). PR is now UX-only: 3 files, single commit `706a6207`.

- `telegram-plugin/tests/two-zone-snapshot-extras.test.ts`
- `telegram-plugin/tool-labels.ts`
- `telegram-plugin/two-zone-card.ts`

Targeted test green: `vitest run telegram-plugin/tests/two-zone-snapshot-extras.test.ts` → 8/8 passed.

No second PR needed for the webhook commits.

PR #905, opened 2026-05-10
feat: host-shell access to v0.7 vault broker (operator socket)

mekenthompson:
## Why

Eight host-shell CLI verbs were broken under v0.7 docker mode because the broker only bound sockets at \`/run/switchroom/broker/<agent>/sock\` (inside the container, mode 0600 chowned to the per-agent UID) and the host CLI defaulted to \`~/.switchroom/vault-broker.sock\` (v0.6 host-side path) which doesn't exist anymore. Every host-shell broker call returned \"broker unreachable\":

| Verb | Pre-fix |
|---|---|
| \`switchroom vault broker {status,unlock,lock}\` | \`{\"running\":false}\` while broker was up |
| \`switchroom vault doctor\` | false-negative |
| \`switchroom vault auto-unlock {status,poll}\` | false-negative |
| \`switchroom agent restart [--name|all]\` | preflight false-negative blocked restart |
| \`switchroom vault {get,list}\` | broker path dead; resolver fell back to direct decrypt |

## Design

Add a host-shell-reachable \"operator\" socket via the existing path-as-identity model. Third identity kind alongside agent and null:

\`\`\`
host:      ~/.switchroom/broker-operator/sock           (mode 0600, chowned to operator UID)
          ↑ docker bind mount
container: /run/switchroom/broker/operator/sock         (broker binds + chowns)
\`\`\`

The bind path tells \`peercred.socketPathToIdentity\` this is the operator. peercred's own UID-match gate would deny here (host UID never matches the broker container's root UID), so peercred is bypassed for this listener — trust comes from the bind path + chown + 0600 file mode.

## What changed (8 slices)

1. **peercred** — new \`socketPathToIdentity()\` returns \`{kind:\"agent\",name} | {kind:\"operator\"}\`. Backward-compat \`socketPathToAgent()\` returns null for the operator path so the per-agent enumerator skips it. \`isReservedAgentName(\"operator\")\` is true; the agent allocator throws if anyone tries to use it. New \`unlockSocketFor()\` is the single source of truth for unlock-pair derivation — handles both v0.6 flat shape (\`foo.sock → foo.unlock.sock\`) and v0.7 subdir shape (\`operator/sock → operator/unlock\`).
2. **broker server** — new \`bindOperatorListener()\` binds data + unlock pair, chowns to operator UID. \`_handleRequest\` threads an \`isOperator\` flag that routes to operator-mode dispatch: audit \`caller=\"operator\"\`, peer-null fail-closed skipped, grant-mgmt cron-deny skipped, list/get apply entry scope with \`agentSlug=\"operator\"\` (default-deny on agent-scoped keys).
3. **compose generator** — new \`operatorUid\` option emits \`SWITCHROOM_BROKER_OPERATOR_UID\` env + the host-bind volume on the broker service block. Omitting it preserves pre-fix behavior exactly.
4. **apply** — captures \`SUDO_UID\` (or \`process.getuid()\`) and threads as \`operatorUid\`. Pre-creates \`~/.switchroom/broker-operator\` so docker doesn't auto-create as root.
5. **CLI broker client** — \`resolveBrokerSocketPath()\` and \`getSocketPath()\` now defer to a runtime-aware default: under \`isDockerRuntime()\` prefer the operator socket, fall back to the v0.6 path otherwise.
6. **preflight + bot-token messages** — pre-fix instructed operators to run \`switchroom vault broker unlock\`, which under docker with a downed broker also failed. Now: reachable-but-locked still says that (works), unreachable + docker-mode says \`docker compose up -d\` + Telegram \`/vault unlock\` or \`docker exec\`.

## Test plan

- [x] \`npm run lint\` — clean
- [x] \`bun test telegram-plugin/tests/\` — 3199 pass / 0 fail
- [x] \`npx vitest run\` — 4472 pass / 40 skip / 0 fail across 290 files
- [x] 78 new unit assertions: peercred-socket-path (operator identity, reservation, unlockSocketFor server+client agreement), compose-generator (operator bind/env emission with/without operatorUid, absolute-path baking under homeDir override)

### Verification on the box

After merge:
\`\`\`bash
bun run build
sudo HOME=/home/kenthompson PATH=\"\$PATH\" \"\$(which switchroom)\" apply --non-interactive
docker compose -p switchroom -f ~/.switchroom/compose/docker-compose.yml up -d --remove-orphans
switchroom vault broker status              # was: {\"running\":false}
                                             # now: {\"running\":true,\"unlocked\":true,...}
switchroom agent restart all                 # was: \"vault broker is unreachable\"
                                             # now: succeeds
\`\`\`

## Out of scope (deferred)

- \`--as-agent <name>\` flag for operator \`get\` of agent-scoped keys. Default-deny is the safe starting point; add when a real use case surfaces.
- Removing the v0.6 host-side default fallback. Kept indefinitely for legacy installs.
- Docker-mode guard on \`switchroom vault broker start/stop\` — those start a v0.6-style host daemon (useless under v0.7 but not worse than pre-fix).

## Hard constraints honored

- No changes to agent containers, agent compose blocks, or the wire protocol
- No \`docker exec\` from CLI as a fallback (option (b) was rejected upstream)
- All bind paths are absolute under sudo (\`homeDir\` baked at apply time)

### DIFF
--- a/src/agents/compose.ts
+++ b/src/agents/compose.ts
@@ -30,6 +30,7 @@
 import { createHash } from "node:crypto";
 import type { SwitchroomConfig, AgentConfig } from "../config/schema.js";
 import { resolveAgentConfig } from "../config/merge.js";
+import { isReservedAgentName } from "../vault/broker/peercred.js";
 
 /** UID range reserved for agent containers. 999 slots — practical fleet limit. */
 export const AGENT_UID_MIN = 10001;
@@ -85,6 +86,17 @@ export function resolveResourceDefaults(
  * never shift (which would require a chown sweep over their state).
  */
 export function allocateAgentUid(name: string): number {
+  // Names reserved by other identity kinds (today: "operator", used for
+  // the host-shell broker socket) cannot be used as agent names.
+  // Refusing here at allocation rather than letting a same-named agent
+  // silently collide with the operator socket — which would forge an
+  // identity from the broker's POV.
+  if (isReservedAgentName(name)) {
+    throw new Error(
+      `agent name '${name}' is reserved by switchroom for another identity kind ` +
+      `(see vault/broker/peercred.ts:RESERVED_AGENT_NAMES). Pick a different name.`,
+    );
+  }
   const hash = createHash("sha256").update(name).digest();
   const u32 = hash.readUInt32BE(0);
   const range = AGENT_UID_MAX - AGENT_UID_MIN + 1;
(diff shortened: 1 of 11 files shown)


mekenthompson:
Superseded by #991 (rebased onto current upstream/main + missing src/runtime-mode.ts added). This branch had gone stale and a naive merge would have reverted ~12K lines (CHANGELOG, the entire #969 vault UX epic). The new PR preserves the original design intent + commit message.

PR #908, opened 2026-05-10
fix(docker): mount skills+credentials, repair scheduler cascade

### LINKED ISSUE #907: Container migration: agent sandboxes missing bind-mounts for ~/.switchroom/{skills,credentials}; cron schedule also stopped firing post-cutover
## Summary

After the systemd → Docker container migration for agent sandboxes, gymbro's cron tasks are largely broken. Two distinct regressions:

1. The agent container only bind-mounts `~/.switchroom/agents/<name>/`, `~/.switchroom/logs/<name>/`, and `~/.switchroom/switchroom.yaml` from the host. Shared paths the agent needs — `~/.switchroom/skills/` and `~/.switchroom/credentials/` — are NOT mounted, so:
   - All cron prompts that call `python3 ~/.switchroom/skills/<x>/scripts/...` fail (calendar, mail).
   - Reconcile-time symlinks at `agents/<name>/.claude/skills/{calendar,garmin,home-assistant,...}` dangle inside the container (target path doesn't exist in-sandbox).
   - Filesystem-resident credentials (e.g. `GARMIN_TOKEN_DIR=/home/<user>/.switchroom/credentials/garmin-tokens`, `HA_SSH_KEY=~/.switchroom/credentials/ha-ssh-key-*`) point at non-existent paths inside the container.
   - Vault-resident secrets still work because the broker socket *is* bound (`/run/switchroom/broker`) — that asymmetry is the smoking gun.

2. The cron schedule itself stopped firing on May 8 ~05:51 UTC, coinciding with the migration. No scheduler process is visible inside the agent container; no entries in `~/.switchroom/agents/gymbro/telegram/issues.jsonl` between May 8 ~05:51 UTC and a manual test fired today (May 10). Possible the host-side scheduler did not re-register agent timers post-cutover, or it isn't running at all.

## Repro / forensic evidence (gymbro, switchroom 0.7.7)

\`/proc/self/mountinfo\` inside the gymbro container shows only:
\`\`\`
~/.switchroom/agents/gymbro       → /state/agent (rw)
~/.switchroom/logs/gymbro         → /var/log/switchroom (rw)
~/.switchroom/switchroom.yaml     → /state/config/switchroom.yaml (ro)
docker volume switchroom_kernel-gymbro-sock → /run/switchroom/kernel
docker volume switchroom_broker-gymbro-sock → /run/switchroom/broker
\`\`\`

\`ls ~/.switchroom/\` inside container only shows \`agents/\` and \`logs/\` — no \`skills/\`, no \`credentials/\`.

\`ls .claude/skills/garmin/\` → \`No such file or directory\` (dangling symlink).

Manual test of all 6 cron-N.sh scripts: each exits 0 (heartbeat) but records issues like:
- \`cron:cron-0::morning-briefing-data-unavailable\` — *"Garmin + calendar scripts not reachable in cron sandbox; only agents/ + logs/ visible"*
- \`cron:cron-4::garmin-skill-missing\` — *"~/.switchroom/skills/garmin symlink broken"*

\`clean-shutdown.json\` shows \`{"signal":"SIGTERM","reason":"systemctl: external restart"}\` at May 8 16:53 UTC — consistent with cutover timing.

## Suggested fix

1. Add bind-mounts to the agent container template (read-only is fine):
   - \`~/.switchroom/skills/\` (ro)
   - \`~/.switchroom/credentials/\` (ro)
2. Make the reconcile step that creates \`agents/<name>/.claude/skills/<x>\` symlinks container-aware (either fail loudly when the target won't be visible, or copy the skill in instead of symlinking).
3. Verify the host-side scheduler service that fires \`cron-N.sh\` is running and re-registered all agent timers post-migration. Add a startup self-check that lists registered agent timers.

## Workaround

For users who only need Telegram delivery, manually patch the cron-N.sh chat_id values back to the user's DM if reconcile rewrote them to a stale forum chat_id. (Survives until next reconcile.)

mekenthompson:
## Summary

Closes #907 — three regressions from the systemd → docker cutover, fixed in one cohesive PR.

**1. Missing bind-mounts in agent container template** (`src/agents/compose.ts`)
Add ro bind-mounts for `~/.switchroom/skills/` and `~/.switchroom/credentials/` at their canonical host paths. Skill scripts (`python3 ~/.switchroom/skills/<x>/scripts/...`) and filesystem-resident credentials (`GARMIN_TOKEN_DIR`, `HA_SSH_KEY`) now resolve identically inside the container as on the host. Mirrors the existing dual-mount pattern used for `agents/`, `logs/`, `.claude/projects/`. Test added in `tests/docker/compose-generator.test.ts`.

**2. Reconcile-time symlink reachability assert** (`src/agents/scaffold.ts`)
With (1) fixed, `<agentDir>/.claude/skills/<name>` symlinks targeting `<skills_dir>/<name>` resolve in-container. Add a post-create `existsSync(dest)` check so any future regression surfaces as a loud reconcile-time warning rather than a silent skill-missing failure when a cron prompt fires inside the sandbox.

**3. Scheduler stopped firing post-cutover — root cause + visibility** (`src/scheduler/dispatch.ts`, `src/agent-scheduler/index.ts`)
- **Root cause**: `collectScheduleEntries` walked the raw `config.agents[name].schedule` and skipped entries declared in `defaults.schedule` or in a profile's `extends`-chain. The in-agent scheduler then exited 0 with "no schedule entries", and the start.sh supervisor burned its 10-restart-in-60s budget then gave up. Fix: route through `resolveAgentConfig` so the cascade is applied.
- **Visibility**: startup self-check log now always names every resolved entry's cron expression, so the regression would have been visible by `grep agent-scheduler.log` rather than silent.
- **Restart-burn guard**: when the config legitimately yields zero entries or no chat target, idle-sleep instead of exit-0 so the supervisor keeps the process alive (observable in `ps`) and recovers on the next container restart.

Regression test added in `src/scheduler/collect-entries.test.ts`.

## Test plan

- [x] `bun test src/scheduler src/agent-scheduler tests/docker/compose-generator.test.ts` — 107 pass, 1 pre-existing fail and 2 pre-existing harness errors unrelated to this PR (verified by stashing changes and re-running)
- [x] New regression tests pass for cascade-resolution and skills/credentials mounts
- [ ] Operator: verify `agent-scheduler.log` shows the new startup self-check line on next gymbro restart
- [ ] Operator: verify `ls /home/<user>/.switchroom/skills` works inside an agent container after `switchroom apply`

## Deferred / follow-ups

None — all three regressions are fixed end-to-end. Future hardening (out of scope): a doctor-check that asserts both new bind mounts are present in the generated compose, mirroring `checkAgentSocketMounts`.

### DIFF
--- a/src/agent-scheduler/index.ts
+++ b/src/agent-scheduler/index.ts
@@ -178,6 +178,22 @@ export function resolveChannelTarget(
   };
 }
 
+/**
+ * Park the process indefinitely without exiting. Used by main() when the
+ * config is unusable (missing forum chat or zero resolved entries) — see
+ * #907 root cause: exiting in those cases burns the start.sh supervisor's
+ * restart budget and the scheduler silently disappears. Wakes on SIGTERM
+ * so container shutdown still proceeds cleanly.
+ */
+async function idle(): Promise<void> {
+  await new Promise<void>((resolve) => {
+    const onSignal = () => resolve();
+    process.on("SIGTERM", onSignal);
+    process.on("SIGINT", onSignal);
+  });
+  process.exit(0);
+}
+
 export async function main(): Promise<void> {
   const agentName = process.env.SWITCHROOM_AGENT_NAME;
   if (!agentName) {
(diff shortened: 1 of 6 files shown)


mekenthompson:
Superseded:

- **Fix 1 (skills/credentials bind mounts)** — addressed by #912 (merged as d7ba15d7), with conditional emit via existsSync to handle vault-only operators with no `~/.switchroom/credentials/` dir.
- **Fix 2 (reconcile-time symlink reachability check)** — different approach taken: #912 fixes the underlying tilde-path resolution via a `$HOME/.switchroom` symlink in start.sh's docker preamble (#910), so the symlinks actually resolve in-container without needing a separate reachability assert.
- **Fix 3 (scheduler cascade)** — addressed by #917, which carries forward your cascade analysis verbatim.

Closing as fully superseded; everything you flagged here is now in main.

Issue #1002, opened 2026-05-11
agent add: scaffold leaves a placeholder `groups: {"-100..."}` entry that 404s on boot for DM-only agents


mekenthompson:
## Symptom

A freshly-created agent with `--topology dm` has a `groups` section in `access.json` pointing at a chat the bot is not in:

```json
{
  "dmPolicy": "allowlist",
  "allowFrom": [...],
  "groups": {
    "-1003852747971": {
      "requireMention": false,
      "allowFrom": []
    }
  }
}
```

The gateway then probes that chat at boot and gets a 404:

```
telegram gateway: boot-probe-failed: chatId=-1003852747971 reason="400 Bad Request: chat not found"
telegram gateway: unhandled rejection (log_only): GrammyError: Call to 'getChat' failed! (400: Bad Request: chat not found)
```

The bot still works for DMs (this isn't fatal), but the unhandled-rejection log line is alarming and the boot-probe failure muddies any subsequent debugging.

## Repro

```bash
SWITCHROOM_BOT_TOKEN=<token> switchroom agent add testagent \
  --profile default \
  --topology dm \
  --bot-username <bot-username> \
  --allow-from <user-id>

cat ~/.switchroom/agents/testagent/telegram/access.json
# Note the groups: { "-100..." : {...} } entry that was never asked for

docker logs switchroom-testagent 2>&1 | grep -i 'boot-probe'
# boot-probe-failed: chatId=-100... reason="400 Bad Request: chat not found"
```

## Root cause (likely)

The scaffold either:
- Copies a literal chat_id from a default-profile template (a real chat id that's no longer valid), or
- Has a placeholder generator that writes a deterministic but non-existent supergroup id.

For DM-topology agents, the `groups` section should be empty.

## Fix

In the access.json template / wizard, gate the `groups` block on `topology === "forum"`. For DM-topology, write `groups: {}`.

Belt-and-suspenders: the gateway's boot-probe should treat 404 as soft-fail and just remove the dead entry from its in-memory copy (with a log line at info level, not unhandled-rejection).

## Impact

Cosmetic for healthy DM agents, but noisy in the log and confusing for operators trying to diagnose why a bot isn't responding (sends them looking for a supergroup that doesn't exist).

## Encountered while

UAT setup (#866 Phase 2a). Investigating "why doesn't the bot reply to DMs", the boot-probe 404 was the loudest log entry and looked like the real cause (it wasn't — see #BUG-allowFrom-number for the actual cause).

[...]

Issue #1001, opened 2026-05-11
agent add: `--allow-from <user_id>` writes a NUMBER, gateway requires a STRING (silent — only the gateway log surfaces it)


mekenthompson:
## Symptom

`switchroom agent add <name> --allow-from <user_id>` writes the user_id into `access.json:allowFrom` as a **number**:

```json
{
  "dmPolicy": "allowlist",
  "allowFrom": [8248703757],
  "groups": {}
}
```

The gateway then refuses to honour it:

```
telegram gateway: access.json: allowFrom contains non-string entries — all values must be quoted strings (e.g. ["8248703757"]). Fix and reload. Treating as empty.
```

Result: the bot is running but silently ignores all DMs from the intended user. Only an operator who knows to read the gateway log will figure out why.

## Repro

```bash
SWITCHROOM_BOT_TOKEN=<token> switchroom agent add test-harness \
  --profile default \
  --topology dm \
  --bot-username meken_switchroom_test_bot \
  --allow-from 8248703757

cat ~/.switchroom/agents/test-harness/telegram/access.json
# {
#   "allowFrom": [8248703757]   ← number, should be "8248703757"
#   ...
# }

docker logs switchroom-test-harness 2>&1 | grep allowFrom
# telegram gateway: access.json: allowFrom contains non-string entries ...
```

## Root cause

The wizard parses `--allow-from <user_id>` as an integer (via `Number(...)` or similar) and writes it into the JSON as a JS number. The gateway's `access.json` reader enforces a string type for two reasons:

1. JS numbers lose precision past 2^53; Telegram user_ids can exceed this in future.
2. Bot API responses already return user_ids as strings in some contexts; treating allowFrom as a `Set<string>` makes equality checks uniform.

## Fix

Two changes:

1. In `src/agents/setup-dm.ts` (or wherever `--allow-from` is written to `access.json`), serialize as a string:
   ```ts
   - allowFrom: [Number(opts.allowFrom)],
   + allowFrom: [String(opts.allowFrom)],
   ```
2. Add a schema-level validation in the writer so writing a numeric `allowFrom` entry throws at the source rather than producing a quietly-broken access.json.

## Impact

Silent failure mode: agent boots, looks healthy, ignores every DM. Discoverable only via reading the gateway log. Affects every operator who passes `--allow-from` at agent add time (the docs encourage this — see SETUP.md §5 for #866 UAT).

## Encountered while

[...]

Issue #1000, opened 2026-05-11
vault: `vault set --no-broker` breaks the directory-layout symlink, re-introducing the v0.7.12 divergence


mekenthompson:
## Symptom

After a clean `switchroom apply` that has the canonical v0.7.12+ vault layout in place:

```
~/.switchroom/vault.enc -> vault/vault.enc      (symlink)
~/.switchroom/vault/vault.enc                   (regular file, 38580 bytes)
```

Running `switchroom vault set --no-broker <key>` on the host (interactive passphrase entry) writes the new vault state to the LEGACY path as a regular file, replacing the symlink:

```
~/.switchroom/vault.enc                          (regular file, 40654 bytes) ← divergent
~/.switchroom/vault/vault.enc                    (regular file, 38580 bytes) ← stale, broker reads this
```

The broker keeps reading from `vault/vault.enc` (the canonical directory entry), so it never sees the freshly-added keys. On the next `switchroom apply` the divergence guard fires and refuses to proceed.

## Repro

Reliable: had it happen twice in the same UAT setup session today.

```bash
# Starting state: vault.enc is a symlink to vault/vault.enc
ls -la ~/.switchroom/vault.enc ~/.switchroom/vault/
# vault.enc -> vault/vault.enc
# vault/vault.enc (real file)

switchroom vault set telegram-test-bot-token --file /tmp/tok.txt --format string --allow test-harness
# (passphrase entered, set succeeds)

ls -la ~/.switchroom/vault.enc ~/.switchroom/vault/
# vault.enc  (regular file, 40654 bytes)   ← symlink GONE
# vault/vault.enc  (still 38580 bytes, untouched)
```

## Root cause (likely)

`vault set --no-broker` uses an atomic-rename pattern: write to a temp file, `rename(tmp, vault.path)`. When `vault.path` is a symlink, `rename` REPLACES the symlink with the new regular file rather than following it. This is the same shape of bug that the PR #955 migration helper was supposed to address; the runtime save path appears to skip the helper.

## Fix

In `saveVault` (or wherever the atomic-rename happens), resolve `vault.path` through `realpath`/`lstat` first; if it's a symlink, rename onto the target of the symlink, not the link itself. Or: switch to writing into the directory-layout path (`vault/vault.enc`) when the v0.7.12+ layout is detected, leaving the legacy symlink alone.

## Impact

[...]

PR #1007, opened 2026-05-11
fix: four UAT-setup papercuts (closes #999/#1000/#1001/#1002)

### LINKED ISSUE #999: vault: passphrase prompt writes to stdout instead of stderr, gets eaten by $(...) substitution
## Symptom

Any pattern that wraps `switchroom vault get --no-broker <key>` in `$(...)` command substitution silently fails:

```bash
$ export TELEGRAM_API_HASH=$(switchroom vault get --no-broker telegram-uat-api-hash)
Error: Passphrase cannot be empty
$ echo $?
0
$ echo "$TELEGRAM_API_HASH"
(empty)
```

The operator sees a confusing "Passphrase cannot be empty" because the **passphrase prompt is written to stdout** (captured into the variable, never displayed), so they type nothing.

Downstream, code that does `getMe(TELEGRAM_API_ID, TELEGRAM_API_HASH)` with an empty hash gets back **"API ID invalid"** from Telegram — a fully misleading error since the API ID is correct.

## Repro

```bash
$ unset SWITCHROOM_VAULT_PASSPHRASE
$ X=$(switchroom vault get --no-broker any-known-key)
Error: Passphrase cannot be empty
$ echo "X=$X"
X=
```

## Root cause

The vault CLI's passphrase prompt at `src/cli/vault.ts:99` writes to `process.stdout` rather than `process.stderr`. In a `$(...)` subshell, stdout is captured by the parent, so the prompt is never visible to the operator.

## Fix

Change the prompt path to write to `process.stderr`. Stderr is the right destination for interactive prompts by Unix convention — it lets stdout stay clean for the actual secret payload (which IS what the caller wants captured).

```ts
- process.stdout.write(prompt);
+ process.stderr.write(prompt);
```

Same change in the hidden-input path. Verify the existing happy-path (non-piped invocation) still shows the prompt to the user — `stderr` defaults to the TTY, so it does.

## Impact

This papercut was hit **3 distinct times** in a single UAT-setup session (each time producing a non-obvious downstream error). High user-facing confusion factor. Low fix-effort.

## Encountered while

Setting up the UAT harness end-to-end (#866 Phase 2a). The first failure was setting the test bot token; the second was reading `telegram-uat-api-hash` for `bun run uat:login`; the third was during `switchroom agent add` after env vars had been unset. Each time the actual cause (subshell-eats-prompt) was opaque from the visible error.

## Related
- #988 — broker unlock false-timeout (separate UX bug in the same vault CLI)

mekenthompson:
Four small, independent fixes bundled. Each closes a distinct issue; bundled because no single fix exceeds ~15 LOC and they all live in the operator-onboarding path.

### #999 — vault passphrase prompt to stderr (was stdout)

\`switchroom vault get --no-broker <key>\` is canonically called via \`\$(...)\` substitution to capture the secret. Pre-fix the passphrase prompt wrote to stdout, was captured by the subshell, never reached the operator → \"Passphrase cannot be empty\" with no visible prompt. Three downstream errors in a single UAT session disappear.

### #1000 — vault saveVault preserves symlinked layout

The v0.7.12+ layout has \`~/.switchroom/vault.enc -> vault/vault.enc\`. Pre-fix \`renameSync(tmp, path)\` replaced the symlink with a regular file; broker kept reading the (untouched) target, next \`switchroom apply\` tripped the divergence guard. \`lstat\` → \`realpathSync\` resolves before rename; tmp stays in the resolved file's directory so atomic-rename semantics are preserved.

### #1001 — allowFrom entries always strings

Both \`buildAccessJson\` writers (\`onboarding.ts\` and \`scaffold.ts\`) now \`String(userId)\` defensively. Gateway rejects numeric \`allowFrom\` entries as 'non-string'; pre-fix a numeric userId from a legacy \`~/.switchroom/user.json\` could silently disable DM routing.

### #1002 — DM-topology agents omit the groups block

Pre-fix DM-topology agents got a groups entry pointing at the global \`telegram.forum_chat_id\` (or v0.7 sentinel \"0\") which the gateway boot probe 404'd on every restart. Fixed in three places:

- \`add-orchestrator.ts\` step 5b passes \`dmOnly: opts.topology === \"dm\"\`
- \`onboarding.ts:buildAccessJson\` skips groups when \`dmOnly\` OR forumChatId is \`\"\"\` / \`\"0\"\`
- \`scaffold.ts:buildAccessJson\` adds the same empty/sentinel guard alongside the existing \`agentConfig.dm_only\` gate

## Test plan

- [x] 9 new unit tests across 3 files — all pass
- [x] \`tsc --noEmit\` clean
- [x] \`npm run build\` clean
- [x] \`add-orchestrator.test.ts\` updated for new \`opts\` arg — 29/29 pass
- [x] Fresh-process reviewer pass (separate session) — APPROVE (after I reverted one unrelated stash-pop artifact — \`assertions.ts\` — that landed in the first push)

## Refs

Closes #999, #1000, #1001, #1002.

### DIFF
--- a/src/agents/add-orchestrator.test.ts
+++ b/src/agents/add-orchestrator.test.ts
@@ -856,6 +856,7 @@ describe("addAgent — forum topology (#543 acceptance)", () => {
       "777",
       "-1001234567890",
       42,
+      { dmOnly: false },
     );
     // The mock captured the right shape — verify it on disk too.
     const fs = require("node:fs") as typeof import("node:fs");
@@ -901,6 +902,7 @@ describe("addAgent — forum topology (#543 acceptance)", () => {
       "888",
       "-1009999999999",
       undefined,
+      { dmOnly: false },
     );
   });
 
@@ -942,6 +944,7 @@ describe("addAgent — forum topology (#543 acceptance)", () => {
       "555",
       "",
       undefined,
+      { dmOnly: true },
     );
   });
 });
--- a/src/agents/add-orchestrator.ts
+++ b/src/agents/add-orchestrator.ts
@@ -356,15 +356,19 @@ export async function addAgent(opts: AddAgentOpts): Promise<AddAgentResult> {
   }
 
   // ── Step 5b: write access.json ───────────────────────────────────────────
-  // DM mode: only the per-user `allowFrom` matters; the `groups` block is
-  // written with an empty key and ignored by the gateway.
+  // DM mode: only the per-user `allowFrom` matters; passing `dmOnly: true`
+  // tells writeAccessJson to omit the `groups` block entirely so the
+  // gateway's boot probe doesn't 404 on a synthetic empty-string chat id
+  // (#1002).
   // Forum mode: the operator supplied --forum-chat-id, which becomes the
   // group key, and (optionally) --topic-id which scopes the policy to a
   // single forum topic.
   const agentDir = created.agentDir;
   const forumChatId = opts.topology === "forum" ? (opts.forumChatId ?? "") : "";
   const forumTopicId = opts.topology === "forum" ? opts.forumTopicId : undefined;
-  writeAccessJson(agentDir, userId, forumChatId, forumTopicId);
+  writeAccessJson(agentDir, userId, forumChatId, forumTopicId, {
+    dmOnly: opts.topology === "dm",
+  });
   log(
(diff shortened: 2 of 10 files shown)


mekenthompson:
Superseded by replacement PR — rebased onto current upstream/main (which now includes #1005's assertions.ts dedupe) and the stray stash-pop artifact removed.

Issue #926, opened 2026-05-10
Telegram `/update apply` is broken on docker installs — no docker binary in agent container


mekenthompson:
## Bug

PR #924 shipped \`/update apply\` (Telegram-side host update). It works for the dry-run path (\`/update\` → \`switchroom update --check\`), but **the apply path is broken end-to-end on docker installs**.

## Reproducer

\`\`\`
$ docker exec switchroom-<agent> which docker
(not found, exit 1)
$ docker exec switchroom-<agent> ls /var/run/docker.sock
(no such file or directory)
$ docker exec switchroom-<agent> which systemd-run
(not found)
\`\`\`

The agent container has the switchroom CLI baked in (per CLAUDE.md), but no docker binary, no docker socket, no systemd-run. When \`/update apply\` calls \`spawnSwitchroomDetached(['update', ...])\`:

- ✅ \`--check\` dry-run works (no docker shell-outs)
- ✅ \`apply-config\` step works (re-execs the in-container switchroom CLI)
- ❌ \`pull-images\` step fails: \`docker: command not found\`
- ❌ \`recreate-containers\` step fails: same
- The \`notifyDetachedFailure\` handler catches the early exit, posts \`❌ update failed (exit 127): docker: not found\` back to chat

## Root cause

The systemd-run cgroup-escape design in \`spawnSwitchroomDetached\` (\`gateway.ts:5631-5638\`) was for the systemd-host era. In docker mode, even with cgroup escape, the spawned process is still inside the agent container — it doesn't reach the host's docker daemon.

For \`/update apply\` to work end-to-end, one of:

1. **Bind-mount /var/run/docker.sock into agent containers.** Worst option — sec regression, breaks the per-agent isolation principle.
2. **Host-side update daemon** that listens on a UDS the agent gateway can call. Cleanest but new infra (~few hours).
3. **Document /update apply as 'requires host CLI access'** — the current state, but with a clean error message instead of a confusing 'docker: not found'.

## Immediate fix (this issue)

Option 3: gate \`/update apply\` in the Telegram handler with a runtime docker-availability check. If \`docker\` is not on PATH (or \`/var/run/docker.sock\` is absent), refuse with:

\`\`\`
❌ /update apply needs docker access from inside the agent container, but
   it's not available (docker binary not on PATH).

   On docker installs, run \`switchroom update\` from the host shell.

   Tracked as #XXX (proper fix is a host-side update daemon).
\`\`\`

[...]

PR #1145, opened 2026-05-12
ci(uat): promote fuzz harness to a Buildkite PR gate (#1132 / #1134 follow-up)

mekenthompson:
## Summary

- Adds the `:robot: UAT fuzz (real Telegram)` step to `.buildkite/pipeline.yml`, gating PRs that touch `telegram-plugin/`, `src/agents/`, `telegram-plugin/uat/`, or `vitest.uat.config.ts` on the second-pass fuzz harness (PRs #1132 / #1134 + `fuzz-human-style-dm.test.ts`).
- Targets the self-hosted `uat-host` queue, sequential per `concurrency_group: switchroom/uat-fuzz`. Hosted agents cannot host the `test-harness` Docker agent or the bearer-equivalent driver MTProto session.
- Pulls four cluster secrets (`TELEGRAM_API_ID`, `TELEGRAM_API_HASH`, `TELEGRAM_UAT_DRIVER_SESSION`, `TELEGRAM_TEST_BOT_USERNAME`); soft-skips with a warning annotation if any are missing so secret rollout can land separately.
- `describe.skip`s the two JTBD scenarios that failed in the #1132 overnight run (`jtbd-interrupt-marker-dm`, `jtbd-rapid-followup-dm`) — both suspected real bugs, neither should gate every PR until classified.
- `.buildkite/README.md` documents the self-hosted agent setup, the four secrets, and provisioning/rotation via the Buildkite REST API. `telegram-plugin/uat/SETUP.md` adds an "CI gate" section with the per-scenario in-CI / local-only classification.

## JTBD served

JTBD-floor: *"user sends a message, user gets a reply within sensible latency"*. The fuzz harness directly asserts that floor across 56 invariant cases (15 + 18 + 23). Moving from manual to PR-gating means future regressions in the gateway / silent-end / pacing layer are caught **before** merge — the chain PR #1132 surfaced (three independently-shippable ghosting bugs) was a manual catch overnight.

## Verification

- `bun lint` clean in the worktree.
- `python3 yaml.safe_load` clean on the new `pipeline.yml`.
- An independent reviewer agent ran against the branch and verdict was APPROVE; both nits it flagged (stale header comment in `pipeline.yml`; bogus `buildkite-agent secret create` example in the README) are folded in as a follow-up commit on this branch.

The fuzz step itself cannot run end-to-end until the self-hosted `uat-host` agent is installed on the UAT host and the four cluster secrets are provisioned (instructions in `.buildkite/README.md`). The step is designed to soft-skip safely with a warning annotation in the interim so this PR doesn't block other PRs from landing.

Verification target (post-merge): **10 consecutive green CI runs on `main`**. Tracked manually until then.

## Test plan

- [ ] Install the `buildkite-agent` on the UAT host with `tags="queue=uat-host"` and confirm it appears in the BK Agents list.
- [ ] Provision the four cluster secrets via the REST API per `.buildkite/README.md`.
- [ ] Land this PR.
- [ ] Open a no-op PR that touches `telegram-plugin/` and confirm the `:robot: UAT fuzz (real Telegram)` step runs against the self-hosted agent and goes green.
- [ ] Watch the next ten merges to `main` and track the step's green rate. If any flake, classify and follow up.

## Out of scope

- Wiring the cheap non-fuzz scenarios (`silent-end-recovery-dm`, `jtbd-status-query-dm`) into the gate — defer to a follow-up once flake telemetry exists on the fuzz step.
- Root-causing the two skipped JTBD scenarios — separate investigation.
- A host-side update daemon so `/update apply` works from in-Telegram (the existing `#926` thread).

### DIFF
--- a/.buildkite/README.md
+++ b/.buildkite/README.md
@@ -43,6 +43,7 @@ The agent box needs:
 | Type check | `bun lint` (= `tsc --noEmit`) | Any `*.ts`, `tsconfig.json`, package/lockfile |
 | Core tests | `bun run test:vitest` | `src/`, `tests/`, `telegram-plugin/**/*.ts`, vitest config, lockfiles |
 | Plugin tests | `cd telegram-plugin && bun test` | `telegram-plugin/**` |
+| UAT fuzz | `bun run test:uat fuzz-` (self-hosted `uat-host` queue) | `telegram-plugin/**`, `src/agents/**`, `telegram-plugin/uat/**`, `vitest.uat.config.ts` |
 | Trigger evals | `python3 evals/run_trigger.py --parallel 5` | `skills/**/SKILL.md`, eval framework, profile CLAUDE.md |
 | Quality evals | `python3 evals/run_quality.py --parallel 5` | `skills/**/SKILL.md`, eval framework, dataset |
 | Eval summary | `annotate-evals.sh` | Always (after eval steps); no-ops if no result files |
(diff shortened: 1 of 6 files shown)


mekenthompson:
## CI status: pipeline gate works; main is pre-existing red

After the fragment-split commit (`26cbc799`), Buildkite **accepts the pipeline upload cleanly** — `Queue 'uat-host' does not exist` no longer surfaces because the queue reference is hidden inside `.buildkite/pipeline.uat.yml` and only loaded when `SWITCHROOM_UAT_GATE_ENABLED=true`. The `:robot: UAT fuzz — gate` step shows up as `broken` in BK build [#1713](https://buildkite.com/ken-thompson/switchroom/builds/1713), which is BK's term for "if-condition false → skipped" — i.e. exactly what we want pre-rollout.

The build still goes red, but on **two pre-existing failures on `main`** that this PR doesn't touch:

1. `tests/uat-driver.test.ts` — 36 cases failing with `Invalid session string (version = undefined)` from `@mtcute/core`. Looks like an mtcute version bump tightened session-string validation; the test's mock session no longer parses.
2. `src/vault/flock-concurrent.test.ts` — `Cannot use import statement outside a module` from a `worker_threads.eval()` callsite. ESM-vs-CJS resolution inside the worker eval body.

Both fail identically on `main` HEAD (`cd934de2`) — see BK [#1712](https://buildkite.com/ken-thompson/switchroom/builds/1712) (and #1710, #1708, #1707, #1702 — main has been red since at least last week). Cross-check:

```
$ curl /v2/.../builds/1712 | grep -E "Invalid session|Cannot use import"
→ Invalid session string (version = undefined)   ← same as this PR
→ Cannot use import statement outside a module   ← same as this PR
```

This PR's diff is CI-config + UAT scenario `describe.skip`s only — no `tests/uat-driver.test.ts` change, no `src/vault/flock.ts` change. The PR is mergeable on its own merits; the red is environmental.

Suggestion: either (a) merge this PR and address the pre-existing failures in a follow-up, or (b) land the two fixes first and then re-trigger this build. The verification target ("10 consecutive green CI runs on main post-merge") can't progress until main itself goes green, regardless of this PR.

PR #1149, opened 2026-05-12
test(uat): fuzz breadth across CC-1 / CC-2 / CC-3 / CC-7 negatives

mekenthompson:
## Summary

**Goal #2** — turn the status-ask cause-class catalog into fuzzy UAT testing validatable against the real test-harness Telegram bot.

The Goal #1 series (PRs #1144 / #1146 / #1147 / #1148, all merged) shipped one dedicated regression-locked scenario per cause class. Those pin the canonical case with deep assertions. This PR adds a **breadth probe per class**: many prompt shapes exercising the same load-bearing invariant. If a regression slips past the dedicated test (a variant it doesn't cover), the fuzz cases catch it.

## New file

`telegram-plugin/uat/scenarios/fuzz-status-ask-dm.test.ts`

| Cause class | Cases | Invariant |
|---|---|---|
| **CC-1** reaction lifecycle terminal | 5 | Last `+` reaction op by tail-after-reply is in `{👍, 💯, 🎉}` |
| **CC-2** mid-turn silent | 4 | Every non-last bot message has `silent=true`; last has `silent=false` |
| **CC-3** silence-poke ladder | 2 | First reply lands in [70s, 200s] window (soft at 80s case; firm-window-extended at 200s case) |
| **CC-7** near-miss negatives | 8 | Status-ask-shaped strings that don't match the classifier still reach the agent, get a sensible reply, no leak, no crash |

Each `describe` block has the load-bearing invariant at top and varies inputs in a `_CASES` table — easy to extend.

## Out of scope (parked in catalog)

- **CC-4** framework-fallback wording — 5min wedge per case, not fuzz-friendly. Cheap unit snapshot test would suffice.
- **CC-5** subagent flag leak — needs a controlled gateway-abort path that's intrusive to mock at UAT level.
- **CC-8** boot card on real crash vs. clean-shutdown marker — needs a restart-harness extension.

## Wall-clock

Full file run ≈ 30 min sequential (UAT vitest config pins `maxForks: 1` because Telegram rate-limits are per-bot global). CC-3 cases dominate at ~3.5 min each. Suggest running fuzz cause-class blocks selectively (e.g. \`bun run test:uat -- fuzz-status-ask -t "CC-1"\`) when iterating.

## Test plan

- [ ] \`bunx tsc --noEmit\` clean ✅ (verified locally).
- [ ] \`bun run test:uat -- fuzz-status-ask\` against a live test-harness agent — ~30 min full run.
- [ ] Confirm scenarios fail loudly with useful trails when the underlying invariant breaks (smoke-check by inverting an assertion locally).

### DIFF
--- a/docs/status-ask-cause-classes.md
+++ b/docs/status-ask-cause-classes.md
@@ -142,7 +142,8 @@ Fuzz extensions land as a single follow-up PR (PR 4) once 1–3 are green.
 - **2026-05-13** — PR #1144 merged: CC-6 (delete stale `progress-card-dm.test.ts`) + CC-1 (real L1 reaction lifecycle UAT in `reactions-dm.test.ts`).
 - **2026-05-13** — PR #1146 merged: CC-3 (silence-poke wire-path UAT in `silence-poke-soft-dm.test.ts` — forces 90s silent tool churn, asserts first reply lands in [70s, 200s] window).
 - **2026-05-13** — PR #1147 merged: CC-2 (mid-turn `disable_notification` UAT in `midturn-silent-dm.test.ts` + `ObservedMessage.silent` exposure on the harness type).
-- **2026-05-13** — this PR: catalog doc lands on `main` for future agents to reference; classifier-variants fuzz block adds to `fuzz-extended-dm.test.ts` covering CC-7 at the fuzz level.
+- **2026-05-13** — PR #1148 merged: catalog doc lands on `main` for future agents to reference; classifier-variants fuzz block adds to `fuzz-extended-dm.test.ts` covering CC-7 at the fuzz level.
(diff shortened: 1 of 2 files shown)


mekenthompson:
Addressed review feedback in `7709ca51`:

1. **CC-1 "two-message reply" case dropped** — confirmed the structural flaw (`expectMessage(/\S/)` returns on the soft-commit, terminal-done can't have landed in the 8s tail). Left a NOTE in the case table describing what a proper breadth probe of this shape would need (final-message predicate / quiescence helper) for a follow-up PR.
2. **CC-3 ceiling bumped** from `MAX + 40_000` (240s) to `MAX + 80_000` (280s) for the >100s sleep case. Walk: single 200s sleep returns at t≈200s with both soft + firm pokes piggybacked on the same tool result; reply drafts at 205-230s normally; new ceiling absorbs Telegram/mtcute/model jitter. `timeoutMs` budget (290s) still gives `expectMessage` headroom so a broken wire surfaces as a ceiling-assertion failure (clear) not a timeout.
3. **Nit fixed** — "error-path: refuse politely" → "potentially-refusal prompt" with a comment clarifying the case tests reaction lifecycle, not refusal content.

`bunx tsc --noEmit` clean.

Issue #1163, opened 2026-05-13
Agent self-service config: schedule.d/ overlays + agent-config broker MCP


mekenthompson:
## Problem

Agents cannot manage their own crons or skills today. `/state/config/switchroom.yaml` is bind-mounted read-only into each agent container (EROFS on write); `~/.switchroom/skills/` is operator-owned. Every cron addition or skill install requires the operator to hand-edit the yaml on the host and run `switchroom apply`.

Recent friction:
- An agent (`clerk`) tried to add two new cron entries (kids-email monitor + daily digest) and hit EROFS — handed the user a paste-block instead.
- A week earlier an agent hit the same wall editing `~/.switchroom/skills/<name>/*`.

User wants agents to self-service this with safety rails.

### Phase 1 — overlay merge (`schedule.d/`, `skills.d/`)

- New writable per-agent overlay tree at `~/.switchroom/overlays/<agent>/{schedule.d,skills.d}/`.
- Loader globs `*.yaml` from those dirs and **appends/unions** into the agent's resolved config before Zod validation. Append-only — overlays cannot override main `switchroom.yaml`.
- Bind-mount `overlays/<agent>` rw into each agent container at `/state/overlays/self` (self-only).
- New schema field `switchroom.agents_overlay_dir` (default `~/.switchroom/overlays`).
- New narrow `OverlayDocSchema` rejects unknown top-level keys early.

### Phase 2 — `agent-config` broker MCP

- Host-side daemon mirroring `vault-broker` (Unix socket, peercred caller identity).
- MCP tools: `switchroom_schedule_add` / `_remove` / `_list`, `switchroom_skill_install` / `_remove` / `_list`, `switchroom_config_get`.
- Validates via imported `ScheduleEntrySchema`; shells out to `switchroom agent reconcile <name>`; writes via `.staging/` + dry-run reconcile + atomic rename.
- Audit log at `~/.switchroom/logs/config-broker.jsonl`.
- **Approval-card flow** (reuse `vault_request_access` plumbing as a distinct `config_request_apply` op) triggered by: new secret reference, cross-agent write, non-bundled skill source, cron expression < 5min, total cron count > 20.
- Skill install allow-list: `bundled:<name>` + `git+https://...@<pinned-sha>` only. No `npm install` / `pip install` at install time.
- Concurrency: per-agent serialization queue + global reconcile flock.

## Grounded against source (spot-checked in `switchroom.js` bundle)

[...]

PR #1185, opened 2026-05-13
feat(agents): hot-apply cron-only reconciles without container restart

mekenthompson:
## Summary
- Hot-applies cron-only config changes (schedule.d edits) without bouncing the agent container — avoids killing in-flight Claude sessions on every schedule edit.
- Falls back to full container recreate when non-cron config changes (env, mounts, image) are detected.
- Part of the agent-config series (issue #1163). Builds on Phase A (PR #1182, merged) and lands ahead of the broker write tools (Phase C) and overlay loader (Phase B, local).

## Test plan
- [x] \`npm run build\` clean
- [x] \`npm test\` — branch-tip baseline: 5620 passed / 96 failed (across 20 files). Failures are pre-existing environmental issues (vault broker socket path, docker-fleet / compose v2 absence, scheduler IPC, cron auth, static-binary install), confirmed against base \`b660380\` in the Phase B baselining work — not introduced by this branch.
- [ ] Manual: edit a cron entry on a live agent; assert systemd unit rewrites without container recreate.
- [ ] Manual: edit a non-cron field; assert fallback to full recreate fires.

### DIFF
--- a/src/agents/cron-hot-reload.test.ts
+++ b/src/agents/cron-hot-reload.test.ts
@@ -1,0 +1,67 @@
+/**
+ * Phase F (switchroom#1163) — `applyCronChangesHot` & `classifyChangeKind`.
+ *
+ * The helper itself is deliberately minimal: cron scripts on the host
+ * bind-mount are already rewritten by `reconcileAgent`, so the in-
+ * container scheduler sees them on the next fire without a docker
+ * touch. These tests pin the contract — particularly that no docker /
+ * systemctl side effects sneak in — so a future refactor can't quietly
+ * resurrect the container bounce.
+ */
+import { describe, expect, it } from "vitest";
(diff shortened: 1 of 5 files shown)


mekenthompson:
Rebased onto current main — the 9 stale commits from the pre-#1182-merge base are gone. Diff is now scoped to the hot-cron change only (5 files, +355/-4).

Reviewer (independent opus pass) verdict on the hot-cron logic itself: approve-with-followups. Findings:
- Minor: cron-script writes aren't atomic-rename — a cron fire racing a partial write could exec a truncated script. Pre-existing race, but the hot path now makes it the primary delivery mode. Followup.
- Minor: `ipcSignalled` is a dead-code seam today and the test pins it false. Cosmetic; refactor when the IPC signalling lands.

Phase A per-agent quarantine intact. Token preflight correctly bypassed on the hot path (no restart → no token re-read).

PR #1187, opened 2026-05-13
feat(#1163): overlay loader — schedule.d/skills.d storage backend

mekenthompson:
## Summary
- Adds the writable per-agent overlay loader: globs `*.yaml` from `~/.switchroom/overlays/<agent>/{schedule.d,skills.d}/` and append-merges into the resolved config before Zod validation.
- Narrow `OverlayDocSchema` rejects unknown top-level keys early — overlays cannot override main `switchroom.yaml`.
- Compose emit pre-creates the per-agent overlay dir so the bind-mount target exists on first boot.
- This is the **storage backend** for the agent-config broker (Phase A merged as #1182). Per the reviewer-flipped sequencing in #1163, broker write tools (`schedule_add` / `_remove`) land on top of this — not exposed to agents as an independent surface in this PR.

## Test plan
- [x] `npm run build` clean
- [x] 21/21 new tests pass (5 schema + 16 loader): precedence, missing dirs, malformed YAML, schema rejection, merge semantics
- [x] Pre-existing failures verified against `origin/main` — same 19 environmental failures (vault-broker socket path, docker-compose v2 absence, etc.) — not introduced here
- [ ] Manual: drop a schedule.d/*.yaml into an agent's overlay dir; assert next reconcile picks it up via Zod validation

## Series context
Part of #1163. Sequence: A (#1182, merged) → **B (this PR)** → F (#1185, merged) → D / C / E (planned).

### DIFF
--- a/src/agents/compose.ts
+++ b/src/agents/compose.ts
@@ -1236,6 +1236,15 @@ function emitAgentService(
   try {
     mkdirSync(`${hostHomeForChecks}/.switchroom/audit/${a.name}`, { recursive: true });
   } catch { /* best-effort */ }
+  // Phase B (switchroom #1163): pre-create the per-agent overlay
+  // directory so the agent-config write tools (Phase C) have a writable
+  // landing zone. The whole ~/.switchroom/agents/<name>/ tree is already
+  // bind-mounted rw at lines above (the dual-mount), so no separate
+  // volume entry is needed — just the host-side dir, owned by the
+  // operator umask before docker auto-creates it as root.
+  try {
+    mkdirSync(`${hostHomeForChecks}/.switchroom/agents/${a.name}/schedule.d`, { recursive: true });
+  } catch { /* best-effort */ }
   // Agent-config audit log (rw) — the read-only agent-config MCP broker
   // (src/mcp/agent-config/server.ts) appends one JSONL row per tool call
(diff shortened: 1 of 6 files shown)


mekenthompson:
Reviewer (independent opus pass) verdict: **approve-with-followups**, no blockers.

Findings confirmed:
- Append-only enforced (`overlay-loader.ts:128` push-only, `OverlayDocSchema.strict()` rejects unknown top-level keys).
- Silent-self-grant hole closed at the loader: entries with non-empty `secrets:` are dropped with a warning, neighbours preserved (`overlay-loader.ts:138-150`, test `overlay-loader.test.ts:175-194`).
- Per-agent quarantine: dual try/catch — per-file and per-agent isolation, validated.
- Glob order deterministic.

Non-blocking follow-ups for the next PR:
1. `compose.ts:1244` mkdir inherits operator umask — recommend explicit `mode: 0o700` for overlay dirs, and chown to per-agent UID once that exists. Only `schedule.d` is pre-created; `skills.d` not yet (fine for B; wire when Phase E lands).
2. Verify `ScheduleEntrySchema` is not `.passthrough()` upstream — would allow overlay entries to smuggle unknown fields past the strict overlay-doc layer.
3. `OVERLAY_SOURCE` symbol is stamped but unread; wire consumers in Phase C/D.
4. `applyAgentOverlays` warnings are dropped by `loadConfig` — `console.warn` only today. Future structured-log thread-through.

Issue #26, opened 2026-04-24
installScheduleTimers: partial-failure leaves orphaned timer files


mekenthompson:
LOW-priority edge case flagged by the security reviewer on PR #25.

## Context

`createAgent()` in `src/agents/create-orchestrator.ts` pushes a timer-rollback entry onto the stack AFTER `installScheduleTimers()` returns successfully. That's correct for the normal case. But `installScheduleTimers` loops and calls `writeFileSync` per timer — if it throws mid-loop (e.g. disk full after timer 0 written but before timer 1), it never returns, so the rollback entry is never pushed. Orphaned `.timer` + `.service` files for units written before the failure remain on disk.

## Severity

LOW. The failure mode is exotic (disk full mid-install) and the orphaned files are harmless unit files that won't auto-start. But the widened rollback in #19/#20 implies cleanup will happen, and for this edge case it doesn't.

## Fix

Either:
1. Make `installScheduleTimers` transactional — rollback its own writes on throw before re-raising.
2. Replace the single `withRollback(() => installScheduleTimers(...))` in `create-orchestrator.ts` with a per-timer loop that pushes a rollback entry after each successful write.

Option 2 is less invasive; option 1 is cleaner separation of concerns.

## Context

- Security reviewer on PR #25 flagged this
- Not a regression — same behavior existed before #19's rollback widening; just becomes visible now that we advertise cleanup semantics

Issue #24, opened 2026-04-24
Phase 4a follow-up: callback_data agent-name safety before Phase 4b handler


mekenthompson:
Follow-up from the retroactive review of Phase 4a (which shipped as part of PR #22's squash merge). Must land before Phase 4b wires the `callback_query` handler.

## Concern (latent defect)

`telegram-plugin/operator-events.ts` builds `callback_data` values with raw agent names:

```ts
callback_data: `op:reauth:${ev.agent}`
callback_data: `op:swap:${ev.agent}`
// etc.
```

The callback handler will parse these by splitting on `:` to get `['op', action, agent]`. If an agent name ever contains a colon (or other delimiter we introduce later), the parse mis-fires. Today agent names are operator-controlled and typically match `[a-z0-9_-]+`, so no active exploit — but it's a latent defect that breaks cleanly in Phase 4b when the handler gets written.

## Fix options

- **A.** URL-encode the agent name in `callback_data` (and decode in the handler).
- **B.** Validate agent names against a strict slug regex (`^[a-z0-9][a-z0-9_-]*$`, max 62 chars) at `OperatorEvent` construction time, then trust them through the protocol.

Option B aligns with the slug-validation work proposed in #21. Recommend folding this into that issue.

## Other retro-review notes (non-blocking, not worth separate issues)

- `tests/operator-events.test.ts` `"never throws"` suite covers 14 inputs but misses boxed `new Number(42)` and objects with throwing getters. Outer try/catch in `classifyClaudeError` handles both; tests would be belt-and-braces.
- `tests/operator-events-session-tail.test.ts` uses `setTimeout(150ms, 300ms)` polling. Flaky under load. Replace with event-driven polling.
- `operator-events.fixtures.json` — confirmed clean of `sk-ant-`, `sk-proj-`, Telegram bot token shapes. No live secrets.

## Never-throws verification (for the record)

Retro reviewer confirmed `classifyClaudeError` does not throw on: `null`, `undefined`, numbers, booleans, strings, arrays, deep-nested objects, `Error` instances, Symbols, functions, circular objects, and objects with null/primitive `error` fields. All fall through to `unknown-4xx` via the outer try/catch. Core contract holds.

## Context

- PR #22 merged 2026-04-24 (includes Phase 4a unintentionally)
- Blocks Phase 4b (operator-events callback wiring + IPC disconnect + systemd uptime diff)

Issue #21, opened 2026-04-24
Harden createAgent: name slug validation + dead code + schedule timer rollback


mekenthompson:
Three convergent findings from PR #20 reviewers. None blocking Phase 2, but should land before Phase 3 foreman bot exposes the `createAgent` surface to end-user input.

## 1. Name slug validation at orchestrator entry (medium)

`createAgent(opts)` in `src/agents/create-orchestrator.ts` uses `name` directly in `resolve(agentsDir, name)` (line ~210) without re-validating the slug. Today the yaml schema enforces `[a-z0-9][a-z0-9_-]*` at reload time, but this is defense-in-depth that fires late:
1. `writeAgentEntryToConfig(configPath, name, profile)` — writes whatever name is passed.
2. `loadConfig(configPath)` — schema validation fires, throws if bad.
3. But the yaml is already corrupted with an entry that fails schema, and the rollback push happens AFTER the reload, so it won't unwind.

Fix: validate `name` with `^[a-z0-9][a-z0-9_-]{0,62}$` as the very first check in `createAgent()`, before any disk writes. Same for `completeCreation()`.

More urgent than pre-#20 because the Phase 3 foreman bot will expose this path to raw user input.

## 2. Dead variable `wroteYamlEntry` (low)

`src/agents/create-orchestrator.ts:180` — `wroteYamlEntry` is set but never read. Both reviewers flagged it independently. Delete it.

## 3. Schedule timer rollback gap (low, pre-existing)

The widened rollback in #20 covers yaml/scaffold/systemd/env/auth, but `installScheduleTimers` / `enableScheduleTimers` have no rollback push. If they throw, previously-stacked rollbacks fire, but partially-installed timer units survive on disk. Fix: add timer-unit uninstall push inside the existing `withRollback` wrapper.

## Context

- PR #20 merged 2026-04-24
- Phase 3 foreman (#15) makes this more urgent

PR #1241, opened 2026-05-14
fix(install): round-4 — UID alignment for log dir + auth code

mekenthompson:
## Summary

Follow-up to merged PRs #1231 / #1234 / #1237. Phase 3 re-validation on a fresh VM surfaced two more UID-alignment holes that prevented an end-to-end install from producing a working agent.

## Findings fixed

**#21 — `alignAgentUid` skipped the log dir because it didn't exist yet.** The first-pass `alignAgentUid` call inside `apply`'s scaffold loop runs BEFORE `ensureHostMountSources` creates `~/.switchroom/logs/<agent>`, so the `existsSync` gate at `scaffold.ts:194` silently skipped the log dir. `ensureHostMountSources` then `mkdir`'d it as the host operator UID. `start.sh` inside the container runs as the per-agent UID, bind-mounts that dir to `/var/log/switchroom`, and hits "Permission denied" trying to write supervisor logs — the gateway/autoaccept/scheduler sidecars never start. Fixed by a second-pass `alignAgentUid` loop right after `ensureHostMountSources`. **Verified working on the VM** — `~/.switchroom/logs/<agent>` is now `10074:10074` after apply.

**#22 — `switchroom auth code` left OAuth tokens host-owned.** `auth code` runs as the host operator UID and writes tokens into `<agentDir>/.claude/`, undoing the alignAgentUid work apply did earlier. The containerised agent (per-agent UID) then can't read `.oauth-token` and `start.sh` hits EACCES. Fixed via a new `realignAfterHostWrite(name, agentDir)` helper in `auth.ts` that calls `alignAgentUid` after `submitAuthCode` completes. **Verified working on the VM** — `.oauth-token` is now `10074:10074` after auth code.

## What's still not working — beyond install-validation scope

Even with the UID alignment correct and `CLAUDE_CODE_OAUTH_TOKEN` set in the agent process's env, Claude Code 2.1.141 (the bundled version) shows the "Select login method" prompt at boot instead of consuming the env token. Verified via `cat /proc/<claude-pid>/environ` — the var is set with a valid `sk-ant-oat01-...` value. This is a Claude/switchroom integration question (likely involving `--dangerously-load-development-channels server:switchroom-telegram` mode's auth handling) that goes beyond the install-validation scope; tracking separately.

Related: autoaccept-poll's `PROMPTS` table doesn't include a rule for "Select login method" — even if Claude legitimately needs to be driven through it on first run, autoaccept would need to dispatch the keystroke. Easy follow-up if the integration path requires it.

Other follow-ups from this round:
- #17 — `setup --foreman` writes config but compose generator ignores it (foreman service never reaches docker). Separate larger PR.
- #24 — `install-deps.sh` should also install `expect` (doctor flags it as missing).
- #26 — bundled `examples/switchroom.yaml` puts 4 agents on one bot token, triggering Telegram 409 polling conflicts.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `vitest run tests/setup.test.ts src/agents/scaffold.test.ts` — 43 tests pass
- [x] On a fresh VM: apply → log dir is `10074:10074` (was host-uid-owned before this fix)
- [x] On a fresh VM: auth login → auth code → `.oauth-token` is `10074:10074` (was host-uid-owned before this fix)
- [ ] End-to-end "agent replies in Telegram" — blocked on the separate OAuth-handoff investigation noted above

### DIFF
--- a/examples/switchroom.yaml
+++ b/examples/switchroom.yaml
@@ -45,12 +45,17 @@ defaults:
       format: html
   tools:
     allow: [all]
-  env:
-    SWITCHROOM_AUDIT_URL: "https://audit.example"
-  hooks:
-    PreToolUse:
-      - command: "/opt/switchroom-audit.sh"
-        timeout: 5
+  # Example: custom hooks. Uncomment + drop a script of your own at the
+  # referenced path to log every tool call before it runs. Left commented
+  # because the placeholder script doesn't ship — without one, Claude
+  # surfaces a non-blocking hook error on every tool call (noisy but
+  # harmless). Install-validation finding #29.
+  # env:
+  #   SWITCHROOM_AUDIT_URL: "https://audit.example"
+  # hooks:
+  #   PreToolUse:
+  #     - command: "/opt/switchroom-audit.sh"
+  #       timeout: 5
   # Bundled skills that ship with switchroom. `humanizer` removes AI-writing
   # patterns from agent replies before they reach Telegram (29 patterns from
   # Wikipedia's "Signs of AI writing" guide). `humanizer-calibrate` is its
--- a/src/cli/apply.ts
+++ b/src/cli/apply.ts
@@ -497,6 +497,41 @@ export async function runApply(
   // back to interactive unlock).
   await ensureHostMountSources(config);
 
+  // ── 2b. Re-align per-agent log dir ownership ─────────────────────
+  //
+  // `ensureHostMountSources` just created `~/.switchroom/logs/<agent>`
+  // as the host operator UID (because mkdir runs as the operator).
+  // The per-agent scaffold loop above already called `alignAgentUid`,
+  // but at that point the log dir didn't exist yet (existsSync gate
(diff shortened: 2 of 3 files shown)


mekenthompson:
Closing as superseded by #1254 (RFC H — single-writer auth-broker).

This PR's host-UID-vs-agent-UID round-trip helpers (`chownToHostForWrite` + `realignAfterHostWrite`) are correct for the current filesystem-token-handoff model but become dead code after #1254 merges. RFC H replaces the per-agent token files with a broker-mediated credential plane (Decisions 5 + 6): agents request tokens via socket with peercred-validated identity, switchroom no longer copies tokens across UIDs.

The install-validation finding #21 (apply-time log-dir alignment skipped because dir didn't exist yet) is still real and worth landing — opening a smaller focused PR for just that change. Reviewer approval on the round-4 PR stands as evidence the change is correct; just no longer needed.

Install-validation report at #1253 already cross-references this PR — will update the report once #1254 lands.

Issue #1395, opened 2026-05-16
sec(WS3): broker IPC & path-as-identity — audit findings


mekenthompson:
WS3 of epic #1389. **Static source analysis only** (production-adjacent shared host; no active testing). Code-grounded against upstream/main. Audit complete; **fresh-reviewer gate pending** before any remediation.

## Findings

| Sev | Title | file:line |
|---|---|---|
| **HIGH** | approval-kernel `approval_consume`/`approval_revoke`/`approval_record` have **no listener-identity ACL** — keyed only on wire-supplied `request_id`/`decision_id` | `src/vault/approvals/kernel-server.ts:410-449`, `kernel.ts:332,396` |
| **MEDIUM** | vault-broker `mint_grant` uses unbounded wire `req.agent` as an unsanitized path component in a root `mkdirSync`+`writeFileSync`+`renameSync`; the `server.ts:2035` "writes to ~/.switchroom/agents/<agent>" comment is factually wrong (container-internal dead path / doc-drift) | `src/vault/broker/server.ts:2035,2044-2049`; schema `src/vault/broker/protocol.ts:117` |
| **LOW** | `scaffold.ts` `chown -R` over the agent-owned, bind-mounted tree relies on GNU `-P` default (no explicit `--no-dereference`); A1-sibling, currently safe but invariant is tool-inherited not code-asserted | `src/agents/scaffold.ts:251,264` |
| INFO | 32-bit `request_id` (`randomBytes(4)`) — compounds the HIGH's guessability; **primary owner WS4** | `src/vault/approvals/kernel.ts:88` |
| INFO | hostd `agent_exec` allowlists only `argv[0]`, passes full `req.args.argv` to `docker exec`; **primary owner WS5** | `src/host-control/server.ts:727-740` |

### HIGH — cross-agent approval tampering

[...]

PR #1487, opened 2026-05-18
fix(vault-broker): fall through to standing ACL when a presented token is unusable (get/list)

mekenthompson:
## Summary

**PR 1 of 2 — Fix A.** Deterministic fix for the failure class that stranded the `gymbro` agent (and is a latent footgun for every agent): a stale/rotted `~/.switchroom/agents/<agent>/.vault-token` made the broker **more restrictive than presenting no token at all**. The `get`/`list` handlers treated token presence as terminal (`validateGrant` → `grant-invalid` → audit+return) and never fell through to the no-token per-agent-socket standing-ACL path (`checkAclByAgent` → `schedule.secrets[]`). That cascaded into agents firing `vault_request_access` approval cards whose minted grant wrote a *new* token that re-broke the standing ACL.

This brings `get`/`list` into line with the **already-in-production `put` precedent** (`validateGrantForWrite`: `grant-expired`/`grant-revoked` hard-deny; `grant-invalid`/`grant-write-not-allowed` fall through to path-as-identity).

**Read-path divergence (owner-approved):** only `grant-revoked` hard-denies on `get`/`list` — revocation is a deliberate operator kill-switch. `grant-invalid`, `grant-key-not-allowed` **and** `grant-expired` fall through (expiry is time passing, not an operator decision; an unusable token must never be more restrictive than no token when the agent has a standing grant).

## Invariant & security

`allow iff (valid token grants key) OR (standing ACL grants key) else deny`. Identity stays path-as-identity (closure-captured socket path), never wire-supplied. A garbage token grants nothing it couldn't get presenting no token. Independently reviewed at the **plan stage** (`PLAN-VERDICT: SOUND-WITH-CHANGES`); all required changes (B1/B3/B4) folded in.

**B1 (highest-risk detail, folded in):** on fall-through the token-branch deny audit is **deleted**, not just the `return` — the single terminal audit row is written by the no-token ACL path. Exactly one audit row per request; the #1433 hash-chain stays intact (verified by test).

## Test plan

- [x] `tsc --noEmit` clean
- [x] New `server-token-fallthrough.test.ts`: revoked → hard-deny + `method:grant`; invalid/expired → fall-through with a **single non-grant** audit row (B1); valid-token + `list` parallels
- [x] 2 existing `server-grants.test.ts` cases updated to the new (intentional) behaviour; revoked/valid regression unchanged
- [x] **105 broker tests pass, 0 fail** (`bun test` — token-fallthrough + grants + write-grants + server + audit-log + acl)
- [ ] CI green

## Scope / sequencing

Fix A only. **Fix B** (`vault_request_access` becomes standing-ACL-aware, using a **broker-probe** instead of gateway-side `checkAclByAgent` to avoid the broker↔gateway config-drift hole the plan review flagged) is a **separate PR that MUST land after this** (B's short-circuit is wrong pre-A). WS3 / #1395-adjacent — flagging for security review.


mekenthompson:
Closing — this branch is on the pre-history-rewrite timeline (DIRTY, unmergeable). Recreated byte-identically against the new upstream/main and re-reviewed (APPROVE) as **#1496**. Auto-merge disabled here; no action on this PR.

PR #1490, opened 2026-05-18
feat(skill-author): agent-scope skill authoring (PR A)

mekenthompson:
## Scope (PR A)

Adds four tools that let an agent author its own skills under
`~/.switchroom/agents/<agent>/.claude/skills/<slug>/` — the same dir
the bundled-skill reconciler symlinks into. Agent-authored real dirs
coexist with bundled-skill symlinks (the reconciler explicitly leaves
non-symlink dirs alone).

**New CLI:** `switchroom skill create|edit|read|delete`
**New MCP tools:** `skill_create`, `skill_edit`, `skill_read`, `skill_delete`
**Extended:** `skill_list` now returns `editable_scopes: ["agent"]`

## Safety rails

- Path allowlist: `SKILL.md`, `README.md`, `scripts/*.{sh,py}`, `assets/*`, `reference/*.md` (max depth 3).
- Per-file 256 KiB, per-skill 2 MiB, max 50 files.
- Atomic temp-dir → rename for create; O_EXCL temp + rename for edit.
- Symlink TOCTOU defence: `lstat` + `realpath` check that resolved path stays under the skill's scope root. Refuses to follow / write through any symlink.
- Optimistic-concurrency token (dir mtime in ms) on edit — `E_SKILL_VERSION_STALE` on mismatch.
- YAML frontmatter validation on SKILL.md: `name == slug`, `description` 1..1024 chars, no duplicate keys.
- Cron-source denial: `SWITCHROOM_TURN_SOURCE=cron` → `E_SKILL_AUTHOR_REQUIRES_INTERACTIVE`.
- `skill_delete` refuses if path is a symlink (that's a bundled-skill install — caller should use `skill_remove`).

## Error codes

`E_SKILL_SCOPE_DENIED`, `E_SKILL_INVALID_NAME`, `E_SKILL_INVALID_PATH`, `E_SKILL_INVALID_FRONTMATTER`, `E_SKILL_FILE_TOO_LARGE`, `E_SKILL_BUNDLE_TOO_LARGE`, `E_SKILL_ALREADY_EXISTS`, `E_SKILL_VERSION_STALE`, `E_SKILL_AUTHOR_REQUIRES_INTERACTIVE`, `E_SKILL_NOT_FOUND`.

## Architectural decisions (not in the brief)

- **Per-file exit codes** (11..15) for the new author errors so callers can disambiguate `version-stale` vs `interactive-required` vs `file-too-large` without parsing stderr JSON.
- **`spawnSyncWithStdin` helper** added to the MCP shim — `skill_create`'s `files` map and `skill_edit`'s `content` would otherwise blow argv length limits at 256 KiB.
- **Shared `appendAudit` re-export** from `skill-common.ts` rather than direct import from `agent-config.ts` in the new author CLI — leaves future swap-points open.
- **No file-extension allowlist for `assets/`** — depth-3 `assets/<sub>/<file>` allowed for fonts / images / blobs. Names still character-class-restricted.
- Pre-existing test-suite failures (`tests/boot-self-test`, `tests/deps`, `tests/host-control`, etc.) are infrastructure-level and unrelated to this PR. All touched tests pass.

## Deferred

- **PR B:** Global scope (operator-owned bundled pool), operator-overwrite guard, `skill_list.editable_scopes` adding `"global"`.
- **PR C (indefinite):** Publishing flow.

## Test plan

- [x] `src/cli/agent-config-skill-author.test.ts` — 10 tests covering happy path, path traversal, allowlist, frontmatter mismatch, duplicate keys, symlink TOCTOU, stale version, cron denial, symlink-delete denial, already-exists.
- [x] `src/mcp/agent-config/server.author.test.ts` — 3 tests covering registration, descriptions, required-fields schema.
- [x] `src/agents/reconcile-default-skills.test.ts` still green.
- [x] `src/cli/agent-config.test.ts`, `src/mcp/agent-config/server.test.ts`, `src/cli/agent-config-skill-write.test.ts` all still green.
- [x] `npm run lint:tsc` clean.


mekenthompson:
Review feedback addressed in 22478e3.

**B1 (blocker — identity-pin bypass)** — `resolveAgent()` now enforces: any explicit `--agent` REQUIRES `SWITCHROOM_AGENT_NAME` to be set AND to match. Omitted `--agent` requires the env-pin set too (no anonymous calls). New error code `E_AGENT_PIN_REQUIRED` (exit 16). Applied uniformly to `skillCreate`/`Edit`/`Read`/`Delete` via a shared `agentErrCode()` mapper. Tests: unset-env, mismatched-env, matching-env happy path, and a regression that all four entry points enforce the rule.

**B2 (blocker — parent-dir symlink TOCTOU)** — two-layer fix:
1. `skillCreate` walks every component of `agentScopeSkillsRoot(...)` before staging, refusing on any `isSymbolicLink()` hit. Catches the "swap `.claude/skills` for a symlink to `/tmp`" attack before we materialize a stage dir inside attacker-controlled territory.
2. `safeWriteFile` adds a per-component `lstatSync` walk from `scopeRoot` down to `dirname(absPath)` (walking the UNRESOLVED path — `realpathSync` would have already followed the symlinks we're trying to detect).

Comments document the best-effort caveat (a parent can still be swapped between walk and open) and the tighter invariant that bounds it: O_EXCL temp file opened in `parent`, immediate `rename` onto `absPath` in the SAME parent. Test confirms an outside dir stays empty when a parent component is symlinked.

**S1 — symlink-precheck inverted** — `if (st)` → `if (st && st.isSymbolicLink())`. Now only triggers for an actual stale symlink at the slug path.

**S2 — `spawnSyncWithStdin` timeout** — bumped 15s → 30s, plus `maxBuffer: 8 MiB`. Comment notes: 2 MiB bundle + JSON.parse + per-file validation + atomic temp-dir + per-component lstat + fsync on rename can exceed 15s on a loaded box; better to wait than spuriously kill a legitimate create.

**S5 — version-token semantics doc** — `skill_edit` tool description now states: "Token is per-skill, not per-file — any change anywhere in the skill bumps it. Re-read (via `skill_read`) before retrying on E_SKILL_VERSION_STALE."

**E2E test added** — `server.author.test.ts` spawns the real CLI (`bun bin/switchroom.ts skill create --from-stdin`) with a sandboxed `HOME`, verifies the round-trip parses, and asserts the SKILL.md lands on disk. Skips cleanly if `bun` isn't on PATH.

Out of scope for this fix-up commit (per the original task constraints): global scope, quota enforcement, anything outside the 5 PR files.

Targeted tests: 31/31 pass (`agent-config-skill-author.test.ts`, `server.test.ts`, `server.author.test.ts`).

PR #1493, opened 2026-05-18
fix(gateway): vault_request_access short-circuits when standing ACL already covers the key (#1487 Fix B)

mekenthompson:
## Summary

**PR 2 of 2 — Fix B (cascade-stopper). Depends on #1487 (Fix A) — must merge AFTER it.**

When an agent calls `vault_request_access` (or an operator taps Approve) for a **read** key the agent's standing `schedule.secrets[]` ACL already covers, the gateway now short-circuits: **no card, no mint, no `.vault-token` written**. Minting that token is exactly what shadowed the standing ACL pre-#1487 (the gymbro trap) and is plain redundant post-#1487 — this makes the harmful path *structurally unreachable*, so no skill/prompt wording can re-trigger it. The model no longer has to "remember" anything; the system enforces it.

## The B2 design point (from the plan's security review)

Coverage is decided **by probing the broker AS THE AGENT** — a no-token `listViaBroker()` over the per-agent socket (path-as-identity; the gateway runs in the agent's container, so the broker attributes the probe to this agent and applies *its own* loaded config + post-#1487 ACL logic). **Deliberately NOT** a gateway-side `checkAclByAgent`/config read: the gateway can hold newer config than the broker, so a config-derived "covered" could be wrong where the broker still denies ("covered here, denied there"). `list` returns only ACL-visible key **names** — never secret values. **Read scope only** (the standing ACL confers read, not write). Guard is on **both** the request path (before the card is staged) and the operator-approve path (before `mintGrantViaBroker`). **Fail-open** on probe error (a redundant card is harmless; a suppressed one is not).

## Ordering (hard requirement)

Must land **after #1487**. B's "you already have it — just read it directly" guidance is only true once A makes a present-but-unusable token fall through to the standing ACL. **Do not enable auto-merge on this until #1487 is merged.**

## Test plan

- [x] `tsc --noEmit` clean
- [x] Source-pattern assertions in `vault-request-access-tool.test.ts` (this flow's established, non-behaviourally-unit-testable style): probe-before-card, probe-before-mint, **broker-not-gateway-config (B2)**, read-scope-only, fail-open
- [x] **48 tests pass, 0 fail** incl. `vault-grant-union` / `vault-request-access-unlock-resume` / `vault-grant-auto-resume` / `vault-approval-posture` regression
- [ ] CI green
- [ ] merged only after #1487

WS3 / #1395-adjacent — flagging for security review.


mekenthompson:
Closing — pre-history-rewrite timeline. Recreated against the new upstream/main and re-reviewed as **#1497** (must land after #1496). No action on this PR.

PR #1696, opened 2026-05-23
chore: remove orphan commands/ markdown stubs

mekenthompson:
## Summary

Removes `commands/{setup,start,status,stop}.md` — four short markdown stubs with zero inbound references.

## Why

These look like leftover scaffolding from an earlier plugin-metadata design. The CLI verbs are wired through `src/cli/index.ts` (Commander) and documented in `docs/cli-reference.md`. `.claude-plugin/plugin.json` does not reference these files. No code, docs, or CI depends on them.

## Test plan

- [x] `grep -rn 'commands/\(setup\|start\|status\|stop\)' --include='*.json' --include='*.md' --include='*.ts'` → no hits outside CHANGELOG
- [ ] CI: lint / vitest / bun-test (no source-code changes — should be green)

## Risk

None. Pure markdown deletion.


mekenthompson:
Closing — my 'zero inbound references' claim was wrong.

The reviewer agent caught real load-bearing coupling I missed:

- `tests/plugin-onramp-manifest.test.ts` explicitly pins all four files (their existence, frontmatter, body content + minimum length, and specific CLI references) — deletion would fail the unit-test suite.
- `.claude-plugin/plugin.json:4` and `.claude-plugin/marketplace.json:14` advertise these as the plugin's slash commands (`/switchroom:setup`, `/switchroom:start`, `/switchroom:stop`, `/switchroom:status`) to users browsing the marketplace.
- `docs/publishing.md:24-31` documents `commands/*.md` as a shipped artifact.
- `README.md:77` shows users typing `/switchroom:setup` in the quick-start.
- The files themselves (87+39+31+29 lines) are operational playbooks, not stubs.

My grep missed these because I searched for `commands/setup` etc. instead of the slash-command form `/switchroom:setup` — that's the convention Claude Code plugins auto-discover by, which is exactly why no `.ts/.json` file ever needs to name the markdown files directly.

The CI green was a false positive: the vitest path filter skipped the `unittest` shard because the diff was `commands/*.md` only, so the pinning test never ran. Same class as the `project_gha_required_checks_sentinel` memory.

Retiring the plugin's slash-command surface would be a separate product decision (remove advertised features from a published plugin), not a cleanup. Not pursuing.

PR #1733, opened 2026-05-24
fix(gateway): vault_request_save routes through attest_via_posture under telegram-id

mekenthompson:
## Summary

Closes the UX gap where tapping Save on a `vault_request_save` card surfaced a misleading "🔒 Vault is locked" message even when the broker had been auto-unlocked for hours.

#1115 follow-up. The save handler now uses `attest_via_posture: true` under `telegram-id` mode (just like `mint_grant` does since #1115). Closes the tracked follow-up named in the existing `gateway.ts:11784` comment.

## Surfaced 2026-05-24

Clerk asked for `forward-email/api-key` via `vault_request_save`. Operator tapped Save. Gateway responded with "🔒 Vault is locked, run `/vault unlock`". The broker was unlocked at the time (uptime 1h, 78 keys, all per-agent socket pairs ok). The misleading text was the gateway's **per-chat passphrase cache** being empty — not the broker. Audit log confirmed zero `op:put` lines for the requested key — every save attempt short-circuited on the empty cache.

## Root cause

The #1115 first-cut added a telegram-id silent-save shortcut that pulled an in-memory passphrase. The reviewer flagged it as a bypass surface (claude in the same container could exfiltrate). The follow-up withdrew the shortcut but left telegram-id mode falling through to the standard passphrase-cache path. **Telegram-id installs never run `/vault unlock`** (auto-unlock at boot), so the cache stays empty forever and every `vault_request_save` card wedges.

The original PR comment (`telegram-plugin/gateway/gateway.ts:11784`) named the proper resolution as a tracked follow-up: *"routing it through broker-IPC attest_via_posture is a tracked follow-up."*

## Architecture

Already in place:
- Broker server supports `put` with `attest_via_posture: true` (`src/vault/broker/server.ts:1448-1500`) — added by #1115 follow-up rev 3.
- Broker client `putViaBroker` accepts the flag (`src/vault/broker/client.ts:557-600`).

Gap: the CLI shell-out (`defaultVaultWrite`) didn't expose the flag.

## Fix

- New `defaultVaultWritePosture(slug, value)` helper calls `putViaBroker` directly with `attest_via_posture: true`. Returns the same `VaultWriteResult` shape as `defaultVaultWrite` for drop-in substitution. Wraps broker error shapes in `VAULT-BROKER-*` prefixes so the existing `parseVaultCliError` / `renderVaultCliError` stack still surfaces actionable host hints on failure.
- `handleVaultRequestSaveCallback` branches on `VAULT_APPROVAL_AUTH_MODE`:

| Mode | Path | Passphrase needed |
|---|---|---|
| `telegram-id` | `defaultVaultWritePosture` (new) | No |
| `passphrase` | cached-passphrase + `defaultVaultWrite` (legacy) | Yes |

- Card text corrected from "🔒 Vault is locked" to "🔒 Passphrase not cached for this chat" — the broker is independent of the chat's passphrase cache; the old text conflated them.

## Scope

Only `handleVaultRequestSaveCallback`. The other two save-ish callsites (`handleVaultDeferCallback`, stash-command auto-save) keep their existing cached-passphrase paths — they're operator-initiated chat commands where `/vault unlock` is a natural prerequisite, not agent-initiated approval cards. Tracked follow-up if the same UX wedge surfaces there.

## Test plan

- [x] `vitest run telegram-plugin/tests/` — 2745/2745 pass
- [x] `bun test telegram-plugin/tests/` — 3331/3331 pass
- [x] `tsc --noEmit` clean
- [x] `check-vault-test-hermeticity.mjs` clean
- [x] New `vault-write-posture.test.ts` (7 cases) pins the helper's contract
- [x] Existing `vault-approval-posture.test.ts` updated with new pins; invariant #3 preserved
- [ ] **Live canary after release**: on clerk's chat, ask the agent to retry `vault_request_save` for `forward-email/api-key`; tap Save; verify the secret lands (audit log shows `op:put, key:forward-email/api-key, result:allowed`) without any `/vault unlock` first

## Risk

Low. The change is gated on `VAULT_APPROVAL_AUTH_MODE === 'telegram-id'` so passphrase-mode installs see zero diff. The new `attest_via_posture` flag has been broker-side since #1115 follow-up rev 3 (in production via the `mint_grant` path). The handler keeps the same outcome shape (success / failure / wake-agent) so downstream code is unchanged.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -332,7 +332,7 @@ import {
 import { runPipeline } from '../secret-detect/pipeline.js'
 import { StagingMap } from '../secret-detect/staging.js'
 import { maskToken } from '../secret-detect/mask.js'
-import { defaultVaultWrite, defaultVaultList } from '../secret-detect/vault-write.js'
+import { defaultVaultWrite, defaultVaultList, defaultVaultWritePosture } from '../secret-detect/vault-write.js'
 import { parseVaultCliError, renderVaultCliError } from '../secret-detect/vault-error.js'
 import { recentDenialsFromAuditLog, type RecentDenial } from './recent-denials.js'
 import { detectSecrets } from '../secret-detect/index.js'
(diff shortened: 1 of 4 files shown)


mekenthompson:
Reviewer's blocker addressed in 9254ffb4: replaced the `vi.mock` cross-runner-pollution pattern with a dependency-inject `deps?` seam on `defaultVaultWritePosture`. Production callers omit `deps` and get the live broker client. The test passes mock functions directly.

Verified cross-runner:
- vitest run: 7/7 ✓
- bun test: 7/7 ✓  
- Full plugin sweep `bun test telegram-plugin/tests/`: 3331 pass / 0 fail / 0 error

Also noting the non-blocking nit re `postureMintAgents` allowlist: my operator config (~/.switchroom/switchroom.yaml) already has `[clerk, klanker, gymbro]` allowlisted, so the live-canary scenario will exercise the success path. Operators new to telegram-id mode who haven't allowlisted their agents will hit `VAULT-BROKER-DENIED (DENIED): agent 'X' is not on vault.broker.postureMintAgents` — surfaces via existing renderer as broker_denied with the existing remedy. Worth a docs nudge in a follow-up if telegram-id adoption picks up; not in scope here.

PR #1766, opened 2026-05-25
fix(#1762): config_propose_edit silent E_DENIED on oversize diff

mekenthompson:
## Summary

Fixes #1762. Two bugs in the `config_propose_edit` approval path that conspired to swallow Telegram's 4096-char `sendMessage` limit as an opaque `E_DENIED: operator denied …` — even though the operator never saw a card.

- **Card body overflow** — `buildConfigApprovalCardBody` inlined the full unified diff into a single `<pre>` block. Diffs > ~4 KB 400'd at the Bot API; the handler caught the null return and auto-replied `verdict: "deny"`. Now: truncate inline to ~3000 chars with a sentinel, and ship the full diff as a `.patch` document attachment on the same card. Approve/Deny buttons preserved.
- **Discarded reason (higher leverage)** — `approval-gateway` dropped the gateway-side `reason` field and the server rendered every deny as a misleading "operator denied". Plumbed `reason` + a new `denySource: "operator" | "dispatch_failure"` discriminant through `ApprovalResult`. Server maps `dispatch_failure` denies to a distinct `E_APPROVAL_DISPATCH_FAILED` error code.

Related: #1758 — this is the same opaque-failure class the structured error envelope work is addressing.

## Files

- `telegram-plugin/gateway/config-approval-handler.ts` — pre-flight oversize check, `truncateDiffForCard`, optional `postAttachment` dep, `denySource` on dispatch-failure denies.
- `telegram-plugin/gateway/ipc-protocol.ts` — extend `ConfigApprovalResolvedEvent` with `denySource`.
- `telegram-plugin/gateway/gateway.ts` — wire `postAttachment` against `bot.api.sendDocument` for `.patch` upload.
- `src/host-control/approval-gateway.ts` — propagate `reason` + `denySource` from the wire to `ApprovalResult`; tag every internal fail-closed branch with `denySource: "dispatch_failure"`.
- `src/host-control/server.ts` — new exported `formatConfigApprovalDenyError` helper; `handleConfigProposeEdit` emits `E_APPROVAL_DISPATCH_FAILED` when `denySource === "dispatch_failure"`, otherwise `E_DENIED: … : <reason>`.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm run lint` clean (one pre-existing `@mtcute/node` plugin-references error baseline-on-main, unrelated to this change)
- [x] New + relevant existing tests pass (27/27 in scope):
[...]

Issue #1762, opened 2026-05-25
config_propose_edit: silent E_DENIED when diff body > 4096 chars (Telegram Bot API limit)


mekenthompson:
## Repro

Admin agent calls `mcp__hostd__config_propose_edit` with a unified diff whose rendered card body exceeds Telegram's 4096-char `sendMessage` limit. Caller receives `E_DENIED (approval_id=…)` in ~300ms — **operator never sees a card**.

Two real occurrences tonight:
- approval_id `6792eaae` @ 00:52:05Z
- approval_id `e95fc9e1` @ 00:55:12Z

Both with the same ~4.7KB clerk-cron-prompts diff.

## Root cause

Traced end-to-end. Smoking gun in `/var/log/switchroom/gateway-supervisor.log`:

```
telegram gateway: config-approval postCard failed:
Call to 'sendMessage' failed! (400: Bad Request: message is too long)
```

Flow:
1. `buildConfigApprovalCardBody` (`telegram-plugin/gateway/config-approval-handler.ts:62-75`) inlines the entire `unifiedDiff` into a single `<pre>` block.
2. `postCard()` (same file, ~L100) calls Bot API `sendMessage`. Body > 4096 → 400 error → returns null.
3. Handler at L121-130 catches null and auto-replies `verdict: "deny"`, `reason: "Telegram sendMessage failed"` back over IPC.
4. `src/host-control/approval-gateway.ts:117-130` resolves with `verdict: "deny"`.
5. `src/host-control/server.ts:1263-1268` builds `E_DENIED: operator denied config_propose_edit (approval_id=…)` — **and discards `ApprovalResult.reason`**, so the caller has no idea this wasn't an actual operator tap.

## Two bugs in one path

**Bug 1 — card body overflow:** No chunking, no attachment-fallback, no truncation. Comment at `server.ts:1206-1212` explicitly punts on "attachment fallback for very large diffs". That shortcut now bites for any non-trivial yaml edit. RFC §3.4 contemplated diffs up to ~63 KiB.

**Bug 2 — misleading error:** `"operator denied"` is a lie when the operator never saw anything. Discarding the gateway-side `reason` field means every future silent-deny is a mystery. **Higher leverage to fix — without it, even fixing bug 1 leaves us blind to the next class of dispatch failure.**

## Fix scope

Small, two files:

[...]

PR #1768, opened 2026-05-25
fix(#1767): drive-write approval card oversize-body overflow

### LINKED ISSUE #1767: drive-write-approval: latent HTML-escape overflow twin of #1762
## Context

PR #1766 fixed silent E_DENIED on the config-approval card path (issue #1762) by adding a rendered-body (post-HTML-escape) length cap with binary-search line-snap truncation in `telegram-plugin/gateway/config-approval-handler.ts`.

Reviewer flagged a likely sibling bug during the re-review:

> `telegram-plugin/gateway/drive-write-approval.ts` has its own `postCard` path with its own body builder. If it renders user-supplied content with HTML escaping, it has the same latent 5x-inflation overflow bug (`&` → `&amp;` etc.) and will silently 400 from Telegram's 4096-char `sendMessage` limit — surfacing to the agent as an opaque deny.

## Ask

1. Audit `drive-write-approval.ts` for the same overflow pattern: any HTML-escaped user content (file titles, diff bodies, reason fields) wrapped in framing that could exceed 4096 chars after escape.
2. If confirmed, apply the same fix shape as #1766: cap on the rendered body, line-snap binary-search truncation, ship the full content as a `.patch` attachment, surface a structured deny reason if `postCard` fails.
3. Consider factoring the cap+truncate helper out of `config-approval-handler.ts` so both paths share one implementation.

## Out of scope nits also flagged in #1766 re-review (low priority)

- Oversize detection in `config-approval-handler.ts:259` uses substring match on the sentinel literal instead of the `DIFF_SENTINEL` constant.
- Unreachable empty-`bestDiff` edge case in the binary search could render bare `<pre></pre>` with no sentinel.
- Missing test for single unbroken line exceeding the cap (no `\n` to snap to).

mekenthompson:
## Summary

Follow-up to #1766. Drive-write approval cards had the same latent overflow pattern that bit `config_propose_edit` in #1762: `buildDiffPreviewCard` HTML-escapes `docTitle` + every body line with no upstream length cap, so a long Drive doc title, anchor `displayName`, or agent-supplied summary can render past Telegram's 4096-char `sendMessage` limit once `&` / `<` / `>` inflate up to 5x. Telegram then returns a generic 400 that surfaces as a silent `E_DENIED` from the agent's perspective.

- Extract the rendered-body cap + line-snap binary-search helper into a new shared module `telegram-plugin/gateway/oversize-card-body.ts` exposing `truncateRawToFit({raw, render, cap, sentinel, hardLimit?})`.
- `config-approval-handler.ts` now uses the helper; `buildConfigApprovalCardBody` returns `{body, truncated}` and the dispatcher keys oversize detection off the boolean instead of substring-matching the sentinel.
- `drive-write-approval.ts` runs `card.text` through the helper between `buildCard` and `postCard`; on truncation, logs the event. If `postCard` still fails after shrinking, surfaces a structured deny reason (`"Telegram sendMessage failed even after oversize-body truncation"`).
- Rolls in the three #1766 review nits while in the file: exported `DIFF_SENTINEL` constant; removed the unreachable `bestDiff=""` edge case (helper guards it); added a test for a single unbroken line exceeding the cap (no `\n` to snap to) to lock in the char-truncation fallback.

Closes #1767.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `vitest run telegram-plugin/gateway/{config-approval-handler,drive-write-approval,oversize-card-body,diff-preview-card}.test.ts` — 56/56 pass
- [x] New shared-helper tests cover no-op-when-fits, binary-search-shrink (5x inflate), line-snap (no mid-line cut), char-truncation fallback (single long line), defensive hard-cut (framing-alone overflow)
- [x] New drive-write oversize tests cover truncation-before-post, no-op-when-fits, structured-deny-reason on post-failure-after-truncation
- [x] All pre-existing config-approval / drive-write / diff-preview-card tests still pass against the refactor

### DIFF
--- a/telegram-plugin/gateway/config-approval-handler.test.ts
+++ b/telegram-plugin/gateway/config-approval-handler.test.ts
@@ -67,7 +67,7 @@ afterEach(() => {
 
 describe("buildConfigApprovalCardBody", () => {
   it("HTML-escapes the diff body so `<` / `&` can't break out of the <pre> block", () => {
-    const body = buildConfigApprovalCardBody({
+    const { body } = buildConfigApprovalCardBody({
       agentName: "klanker",
       reason: "<script>",
       unifiedDiff: "a & b <c>",
@@ -80,7 +80,7 @@ describe("buildConfigApprovalCardBody", () => {
     // 3000 `&` chars escape to 15000 `&amp;` chars — far past 4096.
     // The post-escape cap MUST kick in and truncate the rendered body.
     const evilDiff = "&".repeat(3000);
-    const body = buildConfigApprovalCardBody({
+    const { body } = buildConfigApprovalCardBody({
       agentName: "klanker",
       reason: "test",
       unifiedDiff: evilDiff,
@@ -91,7 +91,7 @@ describe("buildConfigApprovalCardBody", () => {
 
   it("rendered body stays under 4096 when raw diff is all `<` (5x escape)", () => {
     const evilDiff = "<".repeat(3000);
-    const body = buildConfigApprovalCardBody({
+    const { body } = buildConfigApprovalCardBody({
       agentName: "klanker",
       reason: "test",
       unifiedDiff: evilDiff,
@@ -102,7 +102,7 @@ describe("buildConfigApprovalCardBody", () => {
 
   it("clips an unbounded operator-supplied `reason` to ~500 chars with ellipsis", () => {
     const longReason = "x".repeat(2000);
-    const body = buildConfigApprovalCardBody({
+    const { body } = buildConfigApprovalCardBody({
       agentName: "klanker",
       reason: longReason,
       unifiedDiff: "small",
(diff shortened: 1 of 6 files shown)


mekenthompson:
Addressed all three reviewer concerns in e51e063f: (1) HIGH — moved primary defense to per-field intake clipping in `buildDiffPreview` (`docTitle` 200, `displayName` 150) so the helper's entity-bisect fallback is unreachable on user-controlled fields; (2) MED — added `.txt` attachment fallback when truncation fires, wired `postAttachment` in gateway.ts mirroring config-approval's `.patch` flow; (3) LOW — `truncateDiffForCard` now returns `{out, truncated}` and dispatcher keys oversize off the boolean. tsc clean, 83 targeted tests pass.

PR #1992, opened 2026-05-29
feat(web): containerize dashboard + webhook receiver (switchroom-web)

mekenthompson:
## Summary

Phase 3 of the webhook Docker-native migration. Packages the `switchroom web` server (dashboard + GitHub-webhook receiver) as an image-pinned container in its own compose project (`switchroom-web`), replacing the legacy `switchroom-web.service` systemd unit.

- **`docker/Dockerfile.web`** — thin `switchroom/base` extension running `switchroom web` under bun as uid 1000; copies `dist/cli/switchroom.js` + `dist/cli/ui`; tini entrypoint; no docker socket / caps.
- **`src/cli/webd.ts`** — `install`/`status`/`uninstall` lifecycle. Compose project `switchroom-web` at `~/.switchroom/web/docker-compose.yml`. `install` refuses to write a root-uid compose (silent-503 guard).
- **`switchroom update`** — opt-in `refresh-web` step gated on a new `web_service.managed` config flag (default **false**) so existing systemd installs aren't surprised.
- **CI** — `web` leg added to the `build-dependents` matrix.
- **Docs** — `webhook-ingest.md` Deployment section + `cli-reference.md` `webd` entry.

## Why

The systemd unit ran the server straight off the shared source checkout. Any other agent's worktree activity could delete the hoisted root `node_modules` out from under the long-lived process, so the next restart crash-looped on ENOENT. Pinning it to an image makes it immune to repo churn and brings the last systemd holdout in line with the rest of the fleet.

Three load-bearing properties of the compose shape, each forced by what the web service is:
- `network_mode: host` — the server must own host loopback `127.0.0.1:8080`, where the cloudflared tunnel (webhooks) and `tailscale serve` (dashboard) both reach it; the dashboard CSRF gate trusts the `Tailscale-User-Login` header only on loopback (PR #1380).
- runs as the **operator uid** — the receiver's forward to each agent's `webhook.sock` is peercred-gated to `{agent uid, SWITCHROOM_WEBHOOK_RECEIVER_UID = operator uid}`; any other uid → 503.
- **no docker socket, no added caps** — minimal surface for the one internet-facing component.

## Rollback safety

The systemd unit is intentionally NOT removed by this PR or by `uninstall`. Cutover is a deliberate manual operator step (container up → verify → `systemctl --user stop && disable switchroom-web.service`). If the container misbehaves, re-enable the unit and you're back on the source-checkout server in seconds.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `src/cli/webd.test.ts` — pins compose invariants (host net, operator uid, no docker socket, no caps, no-new-privileges, symlink-safe `:ro` yaml mount, tag pinning, determinism)
- [x] `src/cli/update.test.ts` — `refresh-web` gating (managed flag, `--skip-images`, re-exec to `webd install`, non-zero failure); step-list expectations updated
- [x] config schema tests green (new `web_service.managed` flag)
- [ ] After merge + release: build `switchroom-web` image, canary on reggie (stage → cut over → verify CF webhook 202 + edge-403 matrix + dashboard via Tailscale → disable unit)

## Follow-up (not in this PR)

- Ruleset `16470166` PUT to add `build-dependents (web, docker/Dockerfile.web)` as a **gating** required check (the leg runs now but won't gate until then).

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -169,6 +169,7 @@ jobs:
           - { name: kernel,       file: docker/Dockerfile.kernel }
           - { name: hostd,        file: docker/Dockerfile.hostd }
           - { name: auth-broker,  file: docker/Dockerfile.auth-broker }
+          - { name: web,          file: docker/Dockerfile.web }
           # hindsight is NOT here — it has its own `build-hindsight`
           # job (PR #1310) that runs in parallel with build-base
           # because Dockerfile.hindsight extends upstream
--- a/docker/Dockerfile.web
+++ b/docker/Dockerfile.web
@@ -1,0 +1,77 @@
+# switchroom/web — dashboard + GitHub-webhook receiver container (Phase 3).
+#
+# Replaces the `switchroom-web.service` systemd unit that ran the web
+# server straight off the shared source checkout
+# (`bun bin/switchroom.ts web`). That arrangement was fragile: any other
+# agent's worktree activity could delete the hoisted root node_modules
+# out from under the long-lived process, so the NEXT restart crash-looped
+# on ENOENT even though the code was unchanged. Pinning the receiver to an
+# image makes it immune to repo churn and brings the last systemd holdout
+# in line with the rest of the fleet (vault-broker, kernel, hostd, agents
+# are all containers).
+#
+# What this container runs:
+#   `switchroom web --port 8080 --bind 127.0.0.1`
+#   - The web dashboard (served from /opt/switchroom/dist/cli/ui).
+#   - The GitHub-webhook receiver (POST /webhook/:agent/:source), which —
+#     when an agent has channels.telegram.webhook_via_gateway: true —
+#     forwards verified+rendered events to that agent's in-container
+#     gateway over ~/.switchroom/agents/<agent>/telegram/webhook.sock.
+#
+# Why host networking (set in the compose file, not here):
+#   Two host-side consumers reach the server on 127.0.0.1:8080 and that
+#   address must NOT change:
+#     - cloudflared tunnel (hooks.switchroom.ai → 127.0.0.1:8080) for
+#       GitHub webhooks.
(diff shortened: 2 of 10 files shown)


mekenthompson:
APPROVE

Phase 3 web-service containerization. Traced the security boundary, the compose invariants, and the update gating end-to-end — all sound. The three load-bearing properties (host net for loopback:8080, operator-uid for the peercred forward ACL, no-socket/cap_drop-ALL minimal surface) are each pinned by a test in `src/cli/webd.test.ts` and enforced in `renderWebComposeFile` (`src/cli/webd.ts:570-653`). `doInstall` refuses to write a root-uid compose (`src/cli/webd.ts:692-703`), killing the exact silent-503 class this migration targets. Default-off `web_service.managed` (`src/config/schema.ts:863-880`) means existing systemd installs are untouched, and the systemd unit is deliberately left in place for instant rollback.

Non-blocking notes:

- `src/cli/update.ts:343` — `refresh-web` re-execs `switchroom webd install` with no `--tag`, so it pins the web container to `:latest` (DEFAULT_IMAGE_TAG, `src/cli/webd.ts:524`) while the rest of the fleet runs the release-pinned tag `switchroom update` is rolling to. On a pinned-release update the web image will drift to `:latest` and skew from the fleet. Consider threading the resolved release tag into the re-exec, or document that managed web tracks `:latest` by design.

- `src/cli/webd.ts:803` — `doUninstall` runs `docker compose down`, which removes the container and the project network, not just a stop; the user-facing copy and the `uninstall` command description (`src/cli/webd.ts:845`, "Stop the web container") say "stop". Cosmetic — the compose file is preserved so re-install is clean, but the wording undersells what `down` does.

- CI: the `web` leg is added to `build-dependents` (`.github/workflows/docker-images.yml:9`) but, per the PR body, will not gate until the ruleset PUT lands. Acknowledged as follow-up — flagging so it is not forgotten.

Rationale: additive, default-off, image-pinned, minimal-surface, and rollback-safe; the two code notes are version-hygiene and copy, neither blocks merge.

PR #1997, opened 2026-05-29
fix(scaffold): export SWITCHROOM_HANDOFF_SHOW_LINE before the gateway fork

mekenthompson:
## Summary
- `SWITCHROOM_HANDOFF_SHOW_LINE` is read **only** by the telegram-plugin gateway (it gates the visible `↩️ Picked up where we left off …` line on the first reply after a restart).
- In the docker preamble the gateway is forked in the **outer** pass, *before* the `exec tmux` re-entry. The export lived in the inner-pass section **after** that fork, so the gateway never inherited it.
- Net effect: `session_continuity.show_handoff_line: false` has been a **silent no-op on every docker agent** since v0.7. Verified live — the running gateway's `/proc/<pid>/environ` had no such var even with the flag set.

## Fix
- Move the export ahead of the `if docker …` runtime branch (still gated on `handoffEnabled`) so it covers the docker outer fork pass, the inner tmux pass, and the v0.6 non-docker path in one place.
- Mirrors the existing `TELEGRAM_STATE_DIR` hoist (line ~33), which solved the identical fork-ordering problem for the gateway.
- Removes the now-redundant late export.

## Why (JTBD)
Serves **you-hold-the-leash** / config-honesty: a documented config field that silently does nothing erodes trust in the cascade. The flag now actually suppresses the banner where set (e.g. an autonomous webhook-driven review bot, where the handoff line is pure noise).

## Test plan
- [x] New regression test: rendered start.sh exports `SWITCHROOM_HANDOFF_SHOW_LINE` **before** the gateway fork, and exactly once.
- [x] Existing assertions preserved (`=true` default present; absent when handoff disabled).
- [x] `vitest run tests/scaffold.test.ts` → 179 passed.
- [x] `tsc --noEmit` clean.
- [x] `scaffold.regenerate-start-sh` + `lifecycle.docker` green.

## Risk
Low. Single env-export relocation in a template; gating unchanged; no runtime-behavior change other than the gateway now seeing the var it was always meant to.

### DIFF
--- a/profiles/_base/start.sh.hbs
+++ b/profiles/_base/start.sh.hbs
@@ -26,6 +26,21 @@
 # `src/agents/lifecycle.ts:attachAgent` expect — the contract is the
 # same one v0.6 systemd has always honored, just enforced inside the
 # container instead of by the host's user systemd manager.
+
+{{#if handoffEnabled}}
+# The telegram-plugin gateway reads SWITCHROOM_HANDOFF_SHOW_LINE to
+# decide whether to prepend the visible "↩️ Picked up where we left
+# off …" line on the first reply after a restart. It MUST be exported
+# *before* the gateway is forked in the docker preamble below (and
+# before the tmux re-exec), otherwise the gateway — the sole consumer —
+# never inherits it and session_continuity.show_handoff_line:false
+# silently no-ops on every docker agent. Living here, ahead of the
+# runtime branch, covers docker (both the outer fork pass and the inner
+# tmux pass) and the v0.6 non-docker path in one place. Gated on
+# handoffEnabled so a handoff-disabled agent emits no handoff env at all.
+export SWITCHROOM_HANDOFF_SHOW_LINE={{#if handoffShowLine}}true{{else}}false{{/if}}
+{{/if}}
+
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

Correct, minimal fix for a real silent no-op.

- `profiles/_base/start.sh.hbs:41` — the `export SWITCHROOM_HANDOFF_SHOW_LINE` now sits ahead of the `if [ "$SWITCHROOM_RUNTIME" = "docker" ]` branch (line ~42), so it is in the environment before the gateway is forked in the docker outer pass, survives the `exec tmux` re-entry, and still covers the v0.6 non-docker path. This is the same hoist pattern already proven for `TELEGRAM_STATE_DIR`. Root cause was accurate: the old export at the former line ~528 was after the fork, so the gateway (sole consumer) never inherited it.
- `profiles/_base/start.sh.hbs:30` — wrapping the export in `{{#if handoffEnabled}}` is a behavior change from the old unconditional emit, but the correct one: a handoff-disabled agent has no handoff line to suppress, so emitting no handoff env is right, not a regression.
- `tests/scaffold.test.ts:1221-1250` — regression test asserts the export precedes `_switchroom_supervise gateway` and that exactly one export survives (no stale late-pass duplicate). That exactly-one assertion is the part that would have caught the original bug and guards the relocation.

Rationale: single env-export relocation in a template, gating semantics improved, ordering proven by test; no other runtime behavior changes.

PR #1999, opened 2026-05-30
fix(telegram): hold 👍 reaction until background workers finish

mekenthompson:
## Summary
- The terminal 👍 reaction was painted at the parent turn's `turn_end`, **independent of background sub-agent worker lifecycle**. A turn that dispatched a worker (Agent/Task) and ended showed 👍 — reading as "done / nothing happening" — while the worker kept running and the model's own reply said work was ongoing. (Operator-observed on a live agent.)
- Add `StatusReactionController.hold()`: freeze on a **working** glyph (✍️/⚡), suppress stall promotion (🥱/😨) for the held window, stay non-terminal so `finalize()` still works.
- `finalizeStatusReaction` defers the `'done'` terminal while `countRunningWorkers() > 0` (registers the held controller); the sub-agent watcher's `onFinish` calls `promoteDeferredDoneReactions()` to flip to 👍 once no workers remain running.

## Why
JTBD `know-what-my-agent-is-doing`: the ambient reaction must reflect *current activity*, not the parent turn boundary. Showing 👍 while a worker runs is a false "finished" signal.

## How it's safe
- **Instance-guarded**: promote no-ops if a newer turn replaced the held controller for that key (the new turn owns the reaction).
- **Race guard**: after holding, re-check `countRunningWorkers()` and promote immediately if the last worker finished between the count and the registration.
- **Errors finalize regardless** — a failed/aborted turn never waits on a worker.
- **No-hang backstop**: the watcher's silent-stall synthesis sets `state='done'` → fires `onFinish` (≤5 min after a worker goes silent), so a held reaction always resolves to 👍.
- Historical (boot-leftover) and already-terminal workers are excluded from the running count.

## Test plan
- [x] `status-reactions.test.ts` — new `hold()` suite: suppresses stall, promotes read/thinking → working glyph, `finalize()` still reaches 👍, no double-paint, no-op after finalize. Green under **both** vitest and bun.
- [x] Existing reaction suites (active-reactions, sweep, reply-terminal, allowed-filter) green.
- [x] `tsc --noEmit` clean; `check-bot-api-wrapping.sh` clean.
- [ ] UAT + fleet rollout (post-merge).

## Risk
Low. Pure additive controller method + a guarded branch in one finalize path. No change when there are no running workers (the dominant case) — `countRunningWorkers()` returns 0 and behavior is identical to today.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1096,6 +1096,18 @@ if (!STATIC) setInterval(checkApprovals, 5000).unref()
 const chatThreadMap = new Map<string, number>()
 const activeStatusReactions = new Map<string, StatusReactionController>()
 const activeReactionMsgIds = new Map<string, { chatId: string; messageId: number }>()
+// Reactions whose terminal 👍 is deferred because a background sub-agent
+// worker was still running when the parent's `turn_end` fired. Painting 👍
+// then would read as "done / nothing happening" while the worker keeps
+// going. The controller is `hold()`-frozen on a working glyph; the entry
+// is promoted to 👍 by `promoteDeferredDoneReactions` when the watcher's
+// `onFinish` reports the last worker complete. Keyed by statusKey; the
+// stored controller instance guards against a newer turn superseding the
+// held one (promote no-ops if `activeStatusReactions[key] !== ctrl`).
+const deferredDoneReactions = new Map<
+  string,
+  { chatId: string; threadId: number | undefined; ctrl: StatusReactionController }
+>()
 
 // #546 — outbound content-dedup window. PR #599 introduced the four read
 // sites (`outboundDedup.check` / `.record` in executeReply, executeStreamReply,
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- `gateway.ts:1918-1929` (finalizeStatusReaction) — deferring the terminal 👍 when `countRunningWorkers() > 0` is correct: it does NOT call `purgeReactionTracking`/remove from `activeStatusReactions`, so the held controller stays live for `promoteDeferredDoneReactions` to match on `live === ctrl`. The post-register re-check (`if (countRunningWorkers() === 0) promoteDeferredDoneReactions()`) closes the window where the last worker fires onFinish between the count and the map insert. Sound.
- `gateway.ts:1948-1963` (promoteDeferredDoneReactions) — instance guard (`live === ctrl`) drops a held entry superseded by a newer turn, and `deferredDoneReactions.clear()` at the end flushes stale superseded entries too, so no map leak. Correct.
- `subagent-watcher.ts:685 / :1117` set `entry.state = 'done'` BEFORE `onFinish` is invoked at `:956`. This is the load-bearing ordering: when `onFinish`→`promoteDeferredDoneReactions` runs, the just-finished worker already reads `done` and is excluded from `countRunningWorkers()` (which counts `state === 'running' && !historical`), so the promote can't be blocked by the very worker whose completion triggered it. No hang. Verified.
- `status-reactions.ts:341` — `resetStallTimers` early-returns on `this.held`, and `hold()` clears existing stall timers, so 🥱/😨 is suppressed for the held window while `finalize()`/`cancel()`/`finishWithState` all clear `held`. Non-terminal hold is preserved correctly.
- `status-reactions.ts:243-253` — `hold()` no-ops after `finished` (cannot resurrect) and only enqueues a working glyph when not already on one (no double-paint). Test suite at `tests/status-reactions.test.ts:343-422` pins exactly these: stall suppression, read→working promotion, deferred finalize→👍, no double-paint, post-finalize no-op.

Rationale: additive controller method plus one guarded branch in the single 'done' finalize chokepoint; zero behavior change when no workers run (the dominant path). The one hang risk — onFinish seeing its own worker as still-running — is foreclosed by the watcher's state-before-callback ordering. Errors finalize regardless, and the watcher's ≤5min silent-stall synthesis is the no-hang backstop. Merge.

PR #2000, opened 2026-05-30
feat(telegram): live worker-activity feed for background sub-agents

mekenthompson:
## Summary

Background sub-agents (`run_in_background: true`) decouple from the parent turn — when the turn ends nothing surfaces the worker's ongoing jsonl activity, so a long background worker reads as **silence** (the exact gap an operator hit watching a dispatched worker go quiet).

This surfaces that activity the same way the main agent's live answer does: **one regular Telegram message per worker that edits in place** as work happens (current tool + short summary + elapsed), finalized with a tool-count + duration recap on completion.

Layout (running → terminal):
```
🔧 Worker · research competitors
⚡ Bash grep -r "pricing" (3 tools · 00:42)
  ↳ scanning vendor pages…
        ↓ (edits in place)
✅ Worker done · research competitors
5 tools · 01:18
```

## Why

- **JTBD "you hold the leash":** a dispatched background worker going quiet is the awareness gap. This restores live awareness without babysitting.
- **Not card chrome.** The pinned progress card was deleted in #1126 and "Chat IS the artifact" bars re-adding widgets. This is framework-authored content in a normal growing chat message — indistinguishable from "the agent is typing live", the same shape as visible-answer-stream.

## What

- **New `worker-activity-feed.ts`** — pure `renderWorkerActivity` + an injected `BotApiForWorkerFeed`, mirroring `issues-card.ts`. One edit-in-place message per worker keyed by jsonl agent id, with: first-paint delay (trivial sub-second workers stay silent — handback covers them), min-edit-interval throttle + body-dedup, per-worker serialization, 429 cooldown, message_id-drift re-post, forced terminal edit on finish.
- **Watcher `onProgress` enriched** with `lastTool` + `toolCount` — no new callback / signature change (the legacy bucket relay ignores them).
- **Gateway wiring** is watcher-driven (survives turn end), BACKGROUND-gated, behind `SWITCHROOM_WORKER_ACTIVITY_FEED`. When ON it supersedes the coarse 5-min bucket relay to avoid double-surfacing the same beat.

Flag-gated **default-off** (visible-answer-stream convention) — validated via UAT before flipping the default.

## Test plan

- [x] `worker-activity-feed.test.ts` — 16 tests: render (running/starting/done/failed/escape/singular), first-paint delay, dedup, throttle, forced terminal edit, finish no-op when unposted, drop, 429 cooldown, stale-edit re-post, threadId forwarding. Green under **both** vitest and bun.
- [x] subagent-watcher suite (44 tests) still green after the `onProgress` enrichment.
- [x] `tsc --noEmit`, `check-bot-api-wrapping.sh`, `check-plugin-references.mjs` clean.
- [ ] UAT against test-harness with the flag ON before flipping default.

## Risk

Low. New surface is behind `SWITCHROOM_WORKER_ACTIVITY_FEED` (default off) — zero behaviour change until explicitly enabled. The only always-on change is the additive `onProgress` payload fields, which the existing bucket relay ignores.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -53,6 +53,7 @@ import { OutboundDedupCache } from '../recent-outbound-dedup.js'
 import { createInboundCoalescer, inboundCoalesceKey } from './inbound-coalesce.js'
 import { StatusReactionController } from '../status-reactions.js'
 import { DeferredDoneReactions } from '../reaction-defer.js'
+import { createWorkerActivityFeed } from '../worker-activity-feed.js'
 import { isTelegramReplyTool, isTelegramSurfaceTool } from '../tool-names.js'
 import { appendActivityLabel } from '../tool-activity-summary.js'
 import { toolLabel } from '../tool-labels.js'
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- `worker-activity-feed.ts:629-666` (doUpdate) — first-paint gate (`elapsedMs < firstPaintMin`), body-dedup, and `minEditInterval` throttle are correctly ordered, and a failed edit resets `messageId=null`/`lastBody=null` to re-post on the next tick rather than hammering inside a cooldown. Sound rate-limit hygiene.
- `worker-activity-feed.ts:697-716` (update) + `:717-729` (finish) — per-worker promise `chain` serializes ticks so two rapid cues can't double-send, and `finish` deletes the handle in `.finally`. The one theoretical hazard — a late `update` re-posting a running line after `finish` painted the terminal recap — cannot occur: the watcher gates `onProgress` on `entry.state === 'running'` (`subagent-watcher.ts:678`), so no progress cue is emitted post-terminal. Safe by construction.
- `worker-activity-feed.ts:507-537` (renderWorkerActivity) — every user-controlled field (description, tool name, sanitised arg, summary) is `truncate`-then-`escapeHtml`'d in the right order, so a mid-entity cut can't emit broken HTML. Tested at `tests/worker-activity-feed.test.ts:266-278`.
- `gateway.ts:17530-17552` (onProgress) — when `workerFeedEnabled`, the feed update runs and the function returns BEFORE `decideSubagentProgress`, so the same beat isn't double-surfaced (in-message edit + injected 'still working' inbound). Correct supersession; flag-off path is byte-identical to today.
- `gateway.ts:17396-17410` (onFinish) — forced terminal edit is gated on `workerFeedEnabled` and maps `outcome === 'failed' ? 'failed' : 'done'` (so a stale 'orphan' boot row still finalizes an already-posted message). `finish` no-ops when `messageId == null`, so trivial sub-second workers never post (handback covers them).
- `subagent-watcher.ts:686-690` — `onProgress` enriched with `entry.lastTool`/`entry.toolCount` (populated at `:649`/`:654`); additive payload fields, legacy bucket relay ignores them. No signature break.

Non-blocking nits:
- `gateway.ts:17539` — chat fallback `fleetChatId || (loadAccess().allowFrom[0] ?? '')` can resolve to `''`, producing a wasted sendMessage that throws-and-logs. Harmless (caught), but a guard would skip the dead call.
- `worker-activity-feed.ts:730-732` — `drop()` is exported but has no gateway call site in this diff; if the reaction-defer supersession path should discard a superseded worker's live message, it's currently unwired. Dead API, not a bug.
- `worker-activity-feed.ts:668-677` — a worker that finishes while inside a 429 cooldown leaves its message frozen on the last running render (documented 'stale but harmless'). With the feed superseding the bucket relay, that's a permanent 'running' line for that one worker until next session. Acceptable for a default-off feature.

Rationale: net new surface is entirely behind `SWITCHROOM_WORKER_ACTIVITY_FEED` (default off) — zero behavior change until enabled — and the only always-on change is the additive `onProgress` payload the legacy relay ignores. Rate-limiting, dedup, serialization, 429 cooldown, stale-edit re-post, and first-paint suppression are all handled and pinned by 16 tests under both vitest and bun. The terminal-revert race is foreclosed by the watcher's running-state gate. Clean. Merge; flip the default only after the UAT noted in the description.

PR #2002, opened 2026-05-30
fix(telegram): worker-feed shows real task description + live UAT

mekenthompson:
## Summary

Follow-up to #2000 (live worker-activity feed). Two changes:

1. **Real task description in the feed header.** The feed rendered
   `🔧 Worker · sub-agent` because the subagent-watcher never reassigns
   its `'sub-agent'` default from the worker jsonl. The dispatch-time
   description already lives in the registry `subagents.description`
   row, so onProgress + onFinish now prefer it. Header reads
   `🔧 Worker · <real task>`.

2. **Adds the mtcute UAT scenario** that exercises the feed end-to-end:
   first paint → in-place edit → terminal recap.

## Why

The original #2000 UAT used a worker doing three silent `sleep 20`s.
That tripped the **test-harness** stall override
(`SWITCHROOM_SUBAGENT_STALL_MS=5000` / `_STALL_TERMINAL_MS=10000`, set
in switchroom.yaml for PR #1110): a worker silent >15s gets a
*synthesized* terminal `turn_end` mid-flight, which flips the watcher
entry to `done` and **suppresses every later onProgress** — so the feed
never painted (root-caused live: handback fired at +15s of a 20s sleep,
before the worker's real `end_turn`). Production thresholds (60s/300s)
don't hit this, but the UAT worker must stay active. The new scenario
drives ~10 short narrated steps (~2s gaps) so the jsonl keeps ticking
under the 5s stall floor, and documents the coupling in a comment.

## Test plan

- [x] `worker-activity-feed.test.ts` 17/17 (vitest + bun)
- [x] `tsc --noEmit` clean
- [x] mtcute UAT `jtbd-worker-activity-feed-dm` **green** on
      test-harness v0.14.15 — paint at 00:09 (`⚡ Bash echo step-1
      (2 tools · 00:09)`), in-place edit, `✅ Worker done` terminal recap
- [ ] description polish re-validated live during post-release canary

## Risk

Low. Behaviour-gated behind `SWITCHROOM_WORKER_ACTIVITY_FEED` (default
off, test-harness only). The description change is a string-source swap
(registry value preferred over the watcher default); one extra column
on an already-issued indexed query. No change when the flag is off.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -17396,21 +17396,6 @@ void (async () => {
                 // handback gating below, so it must run before any early
                 // return. Cheap no-op when nothing is deferred.
                 deferredDoneReactions.promote()
-                // #PR2 live worker-feed: force the terminal recap edit on
-                // the worker's live message. No-op when no message was ever
-                // posted (trivial workers stay silent; handback covers them).
-                // 'orphan' is a stale boot row, not a fresh completion — map
-                // it to 'done' so an already-posted message still finalizes.
-                if (workerFeedEnabled) {
-                  void workerActivityFeed.finish(agentId, {
-                    description,
-                    lastTool: null,
-                    toolCount,
-                    latestSummary: resultText,
-                    elapsedMs: durationMs,
-                    state: outcome === 'failed' ? 'failed' : 'done',
-                  })
-                }
                 // IO: resolve the fleet chat id and the background flag.
                 // The DECISION (gating + inbound build) is delegated to
                 // the pure `decideSubagentHandback` so it is unit-tested
@@ -17418,6 +17403,10 @@ void (async () => {
                 // `subagent-handback-decision.test.ts`.
                 let fleetChatId = ''
                 let isBackground = false
+                // Dispatch-time task description from the registry row —
+                // see the onProgress note; used for the feed's terminal recap
+                // header so it matches the running header ("· <real task>").
+                let dispatchDesc = ''
                 try {
                   const fleets = progressDriver?.peekAllFleets() ?? []
                   for (const f of fleets) {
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `gateway.ts:17422-17431` (onProgress) & `:17439-17448` (onFinish) — the SELECT now pulls `description` alongside `background`. Verified safe: `description TEXT` is a base column of the `subagents` table (`registry/subagents-schema.ts:190`), present at CREATE TABLE — not a late ALTER like `jsonl_agent_id` — so the widened query can't throw 'no such column' on any DB generation. Correct root-cause fix: the watcher never reassigns its `'sub-agent'` default, and the dispatch-time description lives in the registry row.
- `gateway.ts:17556` & `:17460` — `dispatchDesc || description` fallback is right: a null/empty registry description (column is nullable, `?? ''`) degrades to the watcher default rather than rendering blank. Both the running header and the terminal recap now use the same source, so they stay consistent.
- `gateway.ts:17449-17467` — the `workerActivityFeed.finish()` block is relocated below the turnsDb read so `dispatchDesc` is available, but it still runs AFTER `deferredDoneReactions.promote()` and BEFORE `decideSubagentHandback`. The #1999 reaction-promotion ordering and the finish-before-handback invariant are both preserved. No regression.
- `uat/scenarios/jtbd-worker-activity-feed-dm.test.ts:146-167` — the scenario correctly documents and works around the test-harness stall coupling (`SWITCHROOM_SUBAGENT_STALL_MS=5000`/`_TERMINAL_MS=10000` from #1110): ~10 paced ~2s steps keep the jsonl ticking under the 5s floor so no synthesized mid-flight terminal suppresses onProgress. Accurate diagnosis of why #2000's three silent sleep-20s never painted.

Non-blocking nits:
- `uat/...feed-dm.test.ts:205-211` — the mid-flight 'live edit' check only asserts the message still exists (`not.toBeNull()`), not that the body changed between first paint and the 12s mark; the load-bearing movement proof is `doneText !== before` at `:231`. Acknowledged in the comment as soft. Fine, but a body-delta assertion at the mid-point would make 'edits in place' a hard guarantee rather than inferred.
- The widened SELECT adds one column to an already-issued indexed lookup — negligible, noted for completeness.

Rationale: tightly-scoped follow-up to #2000. The only behavioral change is a string-source swap (registry description preferred over the watcher's 'sub-agent' default), gated behind `SWITCHROOM_WORKER_ACTIVITY_FEED` (default off), with the schema column verified to exist so the path can't silently break. Plus a well-reasoned live UAT. 17/17 unit + green mtcute UAT on the harness. Merge.

PR #2004, opened 2026-05-30
fix(telegram): harden worker-feed real-description wiring + docs/comment sweep

mekenthompson:
## Summary
- Pin the #2002 worker-feed real-description behavior with a fast unit test so it can't silently regress (today it's only covered by the slow, non-gating UAT).
- Extract the duplicated inlined raw-SQL registry read in `gateway.ts` onProgress/onFinish into the existing typed `getSubagentByJsonlId` + a new DB-free pure helper `resolveWorkerFeedDispatch` (DRYs the two call sites, gives a regression pin).
- Sweep the stale comment on `subagent-watcher.ts:77` that misled readers into thinking `WorkerEntry.description` carries the real task name (it's a generic `'sub-agent'` placeholder, never reassigned).
- Update JTBD (`reference/know-what-my-agent-is-doing.md`), user docs (`docs/telegram-plugin.md` + env table), and agent guidance (`CLAUDE.md`).

## Why
The worker-activity feed's headline value is the **real dispatch task** in the header (`🔧 Worker · <task>`). That wiring (registry `description` → feed) had no fast test pinning it — a refactor could swap it back to the watcher's `'sub-agent'` label and only the slow UAT would catch it. This adds the pin and removes the comment that would actively mislead the next agent into sourcing the header from the watcher.

## Test plan
- [x] `tsc --noEmit` EXIT 0
- [x] `check-bot-api-wrapping.sh` clean (gateway.ts line shifts absorbed)
- [x] vitest: worker-feed-dispatch (6) + worker-activity-feed (17) + subagent-progress-inbound-builder (18)
- [x] bun: worker-feed-dispatch (6) + subagent-tracker-hooks (10)
- [ ] UAT `jtbd-worker-activity-feed-dm` (pre-rollout gate)

## Risk
Low — behavior-preserving gateway refactor onto an already-tested helper plus a new pure helper; no flag-state or default change (feed stays default-off fleet-wide).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -826,6 +826,17 @@ JTBDs — see "Design contract" above.)
 - **Config resolution** → `src/config/merge.ts` + `docs/configuration.md`.
 - **Progress card** → `telegram-plugin/stream-reply-handler.ts` +
   `telegram-plugin/card-format.ts` + `docs/telegram-plugin.md`.
+- **Background worker-activity feed** (flag
+  `SWITCHROOM_WORKER_ACTIVITY_FEED`, default off) → live edit-in-place
+  message per `run_in_background` sub-agent. Render in
+  `telegram-plugin/worker-activity-feed.ts`; wired in
+  `telegram-plugin/gateway/gateway.ts` `onProgress`/`onFinish`. The feed
+  header's real task description comes from the registry `subagents`
+  row via `telegram-plugin/gateway/worker-feed-dispatch.ts`
+  (`resolveWorkerFeedDispatch`, pinned by `worker-feed-dispatch.test.ts`)
+  — the watcher only carries a generic `'sub-agent'` placeholder, so do
+  NOT source the header from `subagent-watcher.ts`. E2E gate:
+  `telegram-plugin/uat/scenarios/jtbd-worker-activity-feed-dm.test.ts`.
 - **Auth** → `src/auth/accounts.ts` (slots) + `src/auth/manager.ts`
   (OAuth). Telegram `/auth` routing: `telegram-plugin/auth-slot-parser.ts`.
 - **Runtime inspection** → `switchroom debug turn <agent>` (prompt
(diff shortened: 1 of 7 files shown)


mekenthompson:
APPROVE

- Behavior-preserving refactor verified. The two inlined `SELECT background, description FROM subagents WHERE jsonl_agent_id = ?` reads in `gateway/gateway.ts` (old onFinish/onProgress) are replaced by `getSubagentByJsonlId` + `resolveWorkerFeedDispatch`. The typed reader does `SELECT *` + `mapSubagentRow`, which maps `background !== 0` → boolean (`registry/subagents-schema.ts:308`), semantically identical to the old `row.background === 1`. No behavior change.
- Description fallback preserved exactly. `resolveWorkerFeedDispatch` returns `feedDescription: (sub?.description ?? "") || watcherDescription` (`gateway/worker-feed-dispatch.ts:233`), which equals the old `dispatchDesc || description`. `isBackground: sub?.background ?? false` matches the old default-false on a missing row. Verified against the 6 pins in `tests/worker-feed-dispatch.test.ts:294-334` (real desc wins, null/empty/missing all fall back, missing row never flips a foreground turn background).
- Best-effort safety preserved. Both call sites keep the `try/catch { /* best-effort */ }` around the DB read, and the pre-DB default `resolveWorkerFeedDispatch(null, description)` (`gateway.ts:120`, `:168`) means a DB hiccup keeps the watcher label and `isBackground=false` rather than throwing out of the terminal/progress handler. Feed stays gated on `isBackground` (`:182`), unchanged.
- Stale comment correctly removed. `WorkerEntry.description` doc (`subagent-watcher.ts:73-78`) + the init note (`:870-872`) now accurately state it is a permanent `sub-agent` placeholder never reassigned from the jsonl, and that the gateway sources the real header from the registry via `resolveWorkerFeedDispatch`. Matches the actual init at `:873`.
- Docs accurate. `docs/telegram-plugin.md` env-table row, the feed section, the JTBD `reference/know-what-my-agent-is-doing.md` entry, and the `CLAUDE.md` subsystem map all correctly state the flag is default-off and the header sources from the registry row, not the watcher.

Rationale: pure DRY-onto-an-existing-typed-helper plus a new pure DB-free helper and a fast regression pin for the #2002 real-description wiring; no flag-state or default change, feed stays default-off. Merge.

PR #2006, opened 2026-05-30
test(telegram): fuzz worker-feed dispatch resolution (20k random rows)

mekenthompson:
## Summary
Adds a seeded, randomized property sweep for `resolveWorkerFeedDispatch` (the #2002 worker-feed header wiring), which previously had only 6 example-based cases.

## Why
The operator asked, fairly, whether the worker-feed hardening was *fuzz*-tested — it wasn't. The helper is a pure function that decides whether the feed header shows the real dispatch task or falls back to the watcher's generic `sub-agent` placeholder. That precedence is exactly the bug #2002 fixed, so it deserves randomized coverage, not just hand-picked examples.

## What
- 20k iterations, deterministic `mulberry32(seed)` PRNG so any failure is reproducible from the printed seed/context.
- Adversarial description pool: `null`, `''`, the literal `'sub-agent'`, whitespace-only, control chars, multi-byte unicode/emoji, `'0'`/`'false'` truthiness traps, a 4 KB task, embedded newlines/quotes/tags — crossed with random background/null-row/watcher combos.
- Invariants asserted: real registry description always wins the header; otherwise feedDescription === the watcher label untouched; `isBackground` mirrors the row and is `false` on a registry miss; types stay concrete; pure/deterministic.
- **Mutation-verified**: inverting the helper precedence (watcher wins) turns the sweep red, so it genuinely guards the regression.

## Test plan
- [x] vitest: 7/7 pass
- [x] bun: 7/7 pass (100,010 expect() calls)
- [x] tsc --noEmit clean
- [x] mutation check: precedence inversion → red

## Risk
None — test-only addition; runtime helper untouched. Does not ship to agents (tests aren't in the runtime bundle), so no release/rollout needed.

### DIFF
--- a/telegram-plugin/tests/worker-feed-dispatch.test.ts
+++ b/telegram-plugin/tests/worker-feed-dispatch.test.ts
@@ -61,3 +61,80 @@ describe('resolveWorkerFeedDispatch (#2002 regression pin)', () => {
     expect(resolveWorkerFeedDispatch(null, '').isBackground).toBe(false)
   })
 })
+
+// Deterministic PRNG (mulberry32) so a failing case is reproducible from the
+// printed seed rather than flaking once and vanishing.
+function mulberry32(seed: number): () => number {
+  let a = seed >>> 0
+  return () => {
+    a |= 0
+    a = (a + 0x6d2b79f5) | 0
+    let t = Math.imul(a ^ (a >>> 15), 1 | a)
+    t = (t + Math.imul(t ^ (t >>> 7), 61 | t)) ^ t
+    return ((t ^ (t >>> 14)) >>> 0) / 4294967296
+  }
+}
+
+// Adversarial description fragments the gateway might hand us — empty, the
+// literal watcher placeholder, whitespace-only, control chars, multi-byte
+// unicode/emoji, and a pathologically long task. A registry description is
+// only "real" (must win the feed header) when it's a non-empty string.
+const DESC_POOL: (string | null)[] = [
+  null,
+  '',
+  'sub-agent',
+  '   ',
+  '\n\t',
+  'Background ten-step worker',
+  'Crawl the repo for dead code',
+  '🔧 deploy · staging',
+  'café — naïve façade',
+  '0',
+  'false',
+  'a'.repeat(4096),
+  'line1\nline2\nline3',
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

Test-only PR: appends a seeded randomized property sweep to `telegram-plugin/tests/worker-feed-dispatch.test.ts` (worker-feed-dispatch.test.ts:64-140). 1 file, +77, runtime helper `resolveWorkerFeedDispatch` untouched.

- Reference invariant matches production exactly. The helper from #2004 returns `feedDescription: (sub?.description ?? '') || watcherDescription` and `isBackground: sub ? sub.background : false`. The test's `realDescription = !rowMissing && typeof description === 'string' && description.length > 0` (worker-feed-dispatch.test.ts:118-119) is logically identical to the `||` truthiness rule: every non-empty string is truthy, so whitespace-only `'   '`, the literal `'sub-agent'`, and the `'0'`/`'false'` traps in DESC_POOL (worker-feed-dispatch.test.ts:80-94) all assert description-wins, while `''`/`null` assert watcher-fallback. No divergence on any pool input.
- isBackground assertion correct: `rowMissing ? false : background` (worker-feed-dispatch.test.ts:114) matches the helper returning `false` on a registry miss — a miss can't promote a foreground turn to background.
- Determinism is real. `mulberry32(seed)` (worker-feed-dispatch.test.ts:66-75) is pure, seeded 1..20000, no `Date`/`Math.random`; CI is reproducible and the printed `ctx` makes any failure debuggable. `Math.floor(rng()*arr.length)` stays in-bounds since `rng()` is normalized to [0,1).
- PR body claims mutation verification (precedence inversion -> red), vitest 7/7, bun 7/7, tsc clean — consistent with the assertions; invariant #3 (worker-feed-dispatch.test.ts:124) is what catches the inversion.

Non-blocking nit: a fixed seed range over a 14x6 fixture cross-product saturates the input space well before 20k iterations, so most runs repeat. It's a thorough table test branded as a fuzzer; coverage is genuine but '20k random rows' oversells novelty per iteration. Fine to keep.

Rationale: test-only, deterministic, reference impl provably equivalent to the production helper across all branches; strengthens regression coverage of the #2002 fix. Merge.

PR #2011, opened 2026-05-30
feat(telegram): resuming beat + awaiting-reaction across permission gates

mekenthompson:
## Summary

Closes the "did my Allow tap do anything?" UX gap on permission cards. A permission gate suspends the turn **inside** the bridge's permission call, so no hook events reach the status-reaction controller while the operator decides. Two problems followed:

1. After ~30s/90s the **stall watchdog painted 🥱/😨** on a turn that was healthily waiting on a human — it looked dead.
2. After tapping Allow/Deny, **nothing visibly signalled** that the agent received the verdict and resumed; the operator had to poke the agent to know it was still working.

## What changed (framework-side only — no model narration)

This is consistent with the #1988 decision to drop framework chatter: the framework owns the *beat*, not the words.

- **`status-reactions.ts`** — new `awaiting` state (🙏) via `setAwaiting()`. It parks the reaction immediately **and suspends the stall watchdog** (a turn blocked on the operator is not stalled). Unlike `hold()`, it does **not** set `held`, so the next working transition re-arms the watchdog — a genuine *post-approval* hang still promotes to 🥱/😨.
- **`gateway.ts`** — `onPermissionRequest` parks the active turn's reaction on 🙏. Every verdict path (button allow/deny, always-allow `asn`/`asb`, `/allow`·`/deny` slash, TTL auto-deny) calls `resumeReactionAfterVerdict()` to flip 🙏 → working the instant the turn un-parks. `currentTurn` still points at the suspended turn (no `turn_end` fires mid-gate), so the helper addresses the right controller.
- **`gateway.ts`** — the allow/deny card edit now appends a deterministic `· ▶️ <agent> resuming…` beat, so the operator sees the agent acknowledge the verdict and continue. **👍 still only lands on `turn_end`** — done is never premature.

## Why this is the right shape

`turn_end` (the Stop hook) is the only thing that paints 👍, and it can't fire while claude is parked inside the permission call — so "done" already correctly waits for post-approval work to finish. The real defect was purely *feedback*: the wait looked like a stall and the resume was invisible. Both are fixed at the reaction/card layer without re-introducing the within-turn nudge ladder.

## Test plan

- [x] `status-reactions.test.ts` — 4 new tests: 🙏 emits immediately; stall suppressed for 120s while awaiting; watchdog **re-arms** (🥱) once a working transition resumes; no-op after `finalize`. (30/30 pass)
- [x] `tsc --noEmit` clean
- [x] `check-bot-api-wrapping.sh` clean (no allowlist drift)
- [x] always-allow + permission + reaction suites green (119 tests across 9 files)
- [ ] UAT canary on test-harness before fleet rollout (interactive permission card flow)

## Risk

Low. Additive reaction state + a card-text append; no change to verdict plumbing, persistence, or `turn_end` finalize. Worst case if `currentTurn` was already nulled (swept turn): `resumeReactionAfterVerdict()` no-ops.

## JTBD

`know-what-my-agent-is-doing` — the status reaction must reflect real turn activity; "waiting on you" and "back to work" are now legible instead of masquerading as a stall.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1954,6 +1954,24 @@ function paintStatusReactionError(chatId: string, threadId: number | undefined):
   ctrl.setError()
 }
 
+/**
+ * Flip the current turn's status reaction off 🙏 (awaiting-approval) back
+ * to a working glyph once a permission verdict has been dispatched. The
+ * turn was suspended *inside* the bridge's permission call, so `currentTurn`
+ * still points at it; the verdict un-parks claude and it resumes the SAME
+ * turn. `setThinking()` re-arms the stall watchdog that `setAwaiting()`
+ * suspended, so a genuine post-approval hang still promotes to 🥱/😨, and
+ * it is replaced by the real tool glyph (✍/⚡) as soon as the resumed turn
+ * fires its next PreToolUse. Non-terminal — 👍 still waits for `turn_end`.
+ */
+function resumeReactionAfterVerdict(): void {
+  const turn = currentTurn
+  if (turn == null) return
+  activeStatusReactions
+    .get(statusKey(turn.sessionChatId, turn.sessionThreadId))
+    ?.setThinking()
+}
+
 function resolveThreadId(chat_id: string, explicit?: string | number | null): number | undefined {
   if (explicit != null) return Number(explicit)
   return chatThreadMap.get(chat_id)
@@ -2876,6 +2894,9 @@ const pendingStateReaper = setInterval(() => {
       // dispatchPermissionVerdict so it's buffered+redelivered too if
       // the bridge is also offline at sweep time.
       dispatchPermissionVerdict({ type: 'permission', requestId: k, behavior: 'deny' })
+      // The auto-deny un-parks the suspended turn — flip 🙏 → working so
+      // it doesn't sit on the awaiting glyph (or stall) after the timeout.
+      resumeReactionAfterVerdict()
       process.stderr.write(
         `telegram gateway: permission TTL expired — auto-deny request=${k} ` +
         `tool=${v.tool_name} (no operator response in ` +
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

Additive reaction-state change (+136/-1, 3 files). Verdict plumbing, persistence, and turn_end finalize are untouched; this is purely feedback-layer. The core invariant — 👍 only ever lands on turn_end — is preserved, so "done" can't fire prematurely during a gate.

- `status-reactions.ts:138` — `setAwaiting()` correctly pairs `skipStallReset: true` + `clearStallTimers()` so the watchdog is genuinely suspended (not just reset), and crucially does NOT set `finished`/`held`, so the next `setThinking()`/`setTool()` re-arms it. The re-arm path is the load-bearing claim and `status-reactions.test.ts:186` proves it (🙏 → setThinking → 🥱 after 30s). Good.
- `status-reactions.ts:138` — early `if (this.finished) return` makes it a no-op post-finalize, verified at `tests/status-reactions.test.ts:210`. No resurrection of a closed controller.
- `gateway.ts:19` — `resumeReactionAfterVerdict()` addresses `currentTurn`, while `onPermissionRequest` (`gateway.ts:49`) parks via the passed-in `activeTurn`. In the gate scenario these are the same controller (claude is parked inside the permission call, no turn_end fires, so currentTurn can't advance) — correct as written. Nit: the asymmetry (activeTurn to set, global currentTurn to clear) is a latent footgun if the gateway ever gains concurrent-turn sessions; a one-line assertion or shared accessor would harden it. Non-blocking.
- `gateway.ts:72` + `gateway.ts:100` — the always-allow (asn/asb) path can call `resumeReactionAfterVerdict()` once eagerly (pre-persistence) and again in the dispatch callback. Double `setThinking()` is idempotent so harmless, but confirm the eager call at :72 is intentional (it's there to un-park before the hostd persistence round-trip, which the comment states). Fine.
- `gateway.ts:90` — `resumeBeat` HTML-escapes `SWITCHROOM_AGENT_NAME` via `escapeHtmlForTg` before appending to the card label, matching the escaping discipline of the surrounding scope-commit paths. No injection regression.
- `gateway.ts:36` — TTL auto-deny path resumes the reaction so an expired card doesn't strand the turn on 🙏. Correct; pairs with the existing auto-deny dispatch.

Rationale: the defect was purely feedback (a human-wait looked like a stall; resume was invisible), and the fix lands entirely at the reaction/card layer with 4 targeted tests, clean tsc, and 119 green across the permission/reaction suites. Merge after the UAT canary checkbox closes.

PR #2014, opened 2026-05-31
test(telegram): UAT scenario for forwarded-burst coalescing

mekenthompson:
## Summary

Adds `jtbd-forwarded-burst-dm` — the end-to-end UAT gate for the A1 inbound coalescing shipped in **v0.14.18 (#2007)**. Until now the merge logic was covered only by unit + fuzz tests; nothing exercised the real inbound → gateway coalescer → claude → outbound path over a live Telegram chat.

## Why

A forwarded burst / split paste is one thought. The scenario fires a 3-part burst near-simultaneously (concurrent sends keep inter-arrival under the default 500ms window), with the instruction in the final part asking the agent to echo every token it received **in this single incoming message**.

- Coalesced → one turn sees `ALPHA`/`BRAVO`/`CHARLIE` in one message → names all three.
- Fanned out → the instruction's message carries only its own token → names just one.

Anchoring on "this single message" sidesteps conversation-history bleed, which makes a naive combine-the-facts assertion useless as a coalescing probe (later turns carry prior turns in history). Assertion is set-membership (order-independent) since concurrent sends may arrive in any order.

## Test plan

- [x] `tsc --noEmit` clean
- [x] Live against `test-harness` on v0.14.18: **3/3 green** (~18s each), one shared-context reply naming all tokens every run
- [x] Forensic warning path for partial fan-out (>1 token-bearing reply)

## Risk

None — new test file only, no source changes. It is a live-Telegram UAT (not part of the gating unit suite), so it runs in the `ci-uat` workflow / on-demand, consistent with the other `jtbd-*` scenarios.

Serves: `reference/steer-or-queue-mid-flight.md` ("Forwarded burst / split paste" UAT prompt).

### DIFF
--- a/telegram-plugin/uat/scenarios/jtbd-forwarded-burst-dm.test.ts
+++ b/telegram-plugin/uat/scenarios/jtbd-forwarded-burst-dm.test.ts
@@ -1,0 +1,158 @@
+/**
+ * JTBD scenario — forwarded burst / split paste coalesces into ONE turn.
+ *
+ * Serves: `reference/steer-or-queue-mid-flight.md` — the "Forwarded
+ * burst / split paste" UAT prompt. When several messages land in quick
+ * succession from the same sender (a forward of 3-4 messages, or a long
+ * paste Telegram split into chunks), inbound coalescing must merge them
+ * into a SINGLE Claude turn with shared context — not reply to each
+ * fragment in isolation.
+ *
+ * This is the end-to-end gate for the A1 coalescing work shipped in
+ * v0.14.18 (#2007). The merge logic itself is covered by unit + fuzz
+ * tests (`inbound-coalesce.test.ts`, `pending-inbound-buffer.test.ts`),
+ * but only this scenario exercises the real inbound → gateway coalescer
+ * → claude → outbound path over a live Telegram chat.
+ *
+ * ## How the signal is constructed
+ *
+ * Naively asking the agent to "combine facts from several messages"
+ * does NOT distinguish coalesced from fanned-out: even when each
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

- telegram-plugin/uat/scenarios/jtbd-forwarded-burst-dm.test.ts:1-158 — new live-Telegram UAT for the #2007 inbound coalescing. The probe design is the right one: anchoring the instruction on "this SINGLE incoming message" (lines 51-56) sidesteps conversation-history bleed, which would make a naive combine-the-facts assertion useless since fanned-out later turns still carry prior turns in history. Set-membership assertion (line 119) is correct given concurrent sends arrive in any order.
- :89 — observer started before the burst (correct, since observeMessages only sees live updates), drained into a latest-by-id map so streamed edits collapse to final text. Sound.
- :96 — Promise.all concurrent sends to stay inside the 500ms coalesce window; the rationale (serial round-trips could blow the window on a slow link) is right.
- :126-145 — partial-fan-out forensic: warns (not fails) when >1 token-bearing reply appears but one happened to be complete. Reasonable — avoids flaking on a benign partial while still surfacing the signal.
- Non-blocking nit: timeouts are wall-clock (ANSWER_TIMEOUT_MS 40s, outer +20s). As a live-model UAT outside the gating suite that's fine, but it's inherently latency-sensitive — if test-harness TTFO regresses this could flake. Acceptable for the ci-uat lane.

Rationale: new test file only, zero source changes, no gating-suite impact; the coalescing assertion is methodologically sound and was reported 3/3 green live on v0.14.18. Merge.

PR #2015, opened 2026-05-31
feat(telegram): A2 multi-attachment / album inbound coalescing

mekenthompson:
## Summary

Fold a forwarded **album** (`media_group_id`) or a **text + multi-image burst** into a single Claude turn, gated behind a new per-agent cap `channels.telegram.coalesce.max_attachments` (**default 1 = today's single-attachment behaviour, byte-for-byte preserved**). Follow-on to A1 inbound coalescing (#2007 / v0.14.18).

## Why

A1 coalesces a burst into one turn but carries **at most one attachment** — a second photo/document or an album part bypasses to its own turn (so nothing is dropped, but a forwarded 3-photo album fans out into 3 turns). A2 lets an operator opt into folding the whole album/burst into one shared-context turn. JTBD: *“I forward a batch of photos and the agent reasons over all of them together.”*

## What changes (only active when the cap is raised)

- **Cap gate.** `bufferedAttachmentKeys` becomes a per-key **count** (was a presence `Set`). The `media_group_id` album bypass and the second-attachment bypass in `handleInboundCoalesced` lift up to the cap; attachments past the cap still bypass to their own turn → no media loss.
- **Merge.** The coalescer merge collects the first attachment as **primary** plus **capped extras** (arrival order preserved).
- **Resolve + meta.** `handleInbound` downloads the extra photos and emits **numbered meta** (`image_path_2`, `attachment_file_id_2`, …) plus `attachment_count`.
- **Bridge instructions.** Updated so the agent processes every numbered sibling, not just the first.
- **Latent fix** in `pending-inbound-buffer` `mergeRun`: when the media-bearing entry is **not** the run anchor, its attachment meta is now spliced into the merged meta. Previously the agent saw no `image_path` on that buffered+merged path — A2's numbered fields made the gap worse.
- Cap/ordering + numbered-meta logic extracted to **pure helpers** (`coalesce-attachments.ts`) so they're unit-testable without `loadAccess()`/IPC.
- New config knob documented (`docs/telegram-features.md`, `docs/configuration.md`, plugin README).

## Test plan

- [x] `coalesce-attachments.test.ts` — cap enforcement, arrival order, floor-at-1, numbered meta for photo/document/mixed extras.
- [x] `pending-inbound-buffer.test.ts` — meta splice when media ≠ anchor; no splice needed when media **is** anchor.
- [x] `tsc --noEmit` clean; `check-bot-api-wrapping.sh` + `check-plugin-references.mjs` clean.
- [x] Full telegram-plugin bun suite green (3697 pass / 0 fail); scaffold + merge config vitest green (265 pass).
- [ ] **UAT on test-harness with `max_attachments` raised** — forwarded-album scenario, before any fleet rollout.

## Risk

Low at default. The default cap of 1 reproduces the exact pre-A2 single-attachment path (album + second-attachment bypasses unchanged). All new behaviour is behind `max_attachments > 1`. The numbered-meta rendering relies on claude-code's `<channel>` meta→tag passthrough (same mechanism A1 uses for the primary attachment).

Unrelated pre-existing vitest infra flakes on this shared host (run-hook/issues/bundler subprocess + Hindsight HTTP 502) are not touched by this change.


mekenthompson:
APPROVE

A2 multi-attachment coalescing — clean, gated behind `coalesce.max_attachments` (default 1 = byte-for-byte A1 behaviour). Logic, cap accounting, and the latent merge-meta fix all check out. Merge once CI is green.

- `gateway.ts:297,299-301` — `bufferedAttachmentKeys` correctly migrated from presence-`Set` to count-`Map`; `coalesceMaxAttachments()` floors at 1. Cap accounting is consistent: the overflow probe at `gateway.ts:390` (`(count ?? 0) >= maxAttachments`) and the increment at `gateway.ts:404` mean a window buffers exactly `max` attachments before the next one bypasses to its own turn — matches `splitCoalescedAttachments`' `slice(0, max(1,max))` cap (primary + max-1 extras = max total). No media loss path.
- `gateway.ts:370` — album (`media_group_id`) bypass now gated on `maxAttachments <= 1`. At default cap each part keeps its own turn (unchanged); raised cap lets parts share the coalesce key and the overflow bypass at :390 catches parts past the cap. Correct.
- `coalesce-attachments.ts:193` — `Math.max(1, maxAttachments)` guards the cap-0/negative strip-the-primary footgun; covered by test at `coalesce-attachments.test.ts:588`.
- `pending-inbound-buffer.ts:489-499` — the real latent fix: when the media-bearing entry isn't the run anchor, attachment meta (`image_path`, `attachment_*`, numbered siblings, `attachment_count`) is now spliced into the merged meta. Regex `/^(image_path|attachment_)/` at :467 covers `attachment_count` and all numbered keys. Without this the agent saw no `image_path` on the buffered+merged path; A2's numbered fields would have widened the gap. Good catch, well-tested (:669).
- Bridge instructions (`bridge.ts:139`) updated to tell the agent to process every numbered sibling, not just the first — necessary for the feature to actually reach the model.

Nit (non-blocking):
- `gateway.ts:427-440` — `attachmentCount` adds `extraResolved.length` unconditionally, but a failed extra-photo download pushes `{imagePath: undefined, attachment: undefined}` (line 431), which `buildExtraAttachmentMeta` emits nothing for. So on a download failure `attachment_count` can overcount the numbered fields actually present, and the agent hunts for an `image_path_N` that isn't there. The primary path doesn't have this (count gated on `imagePath != null || attachment != null` at :439). Consider filtering `extraResolved` to resolved-only before counting. Edge case, slow-link only.

Rationale: default cap reproduces the pre-A2 single-attachment path exactly (album + second-attachment bypasses unchanged), all new behaviour is behind `max_attachments > 1`, helpers are pure + unit-tested, and the diff fixes a pre-existing meta-loss bug on the buffered-merge path. UAT-on-raised-cap is still unchecked in the PR body — fine to merge at default, but run that forwarded-album scenario before any fleet rollout raises the cap.

PR #2016, opened 2026-05-31
feat(telegram): deferred safe-boundary `!` interrupt (default-off)

mekenthompson:
## Summary

Adds an opt-in `channels.telegram.interrupt.safe_boundary` flag (default **false**). When on, a `!`-prefix interrupt that lands **mid-tool-call** is deferred to a clean boundary instead of `C-c`-ing the agent partway through a `Write`/`Bash`.

- Flag off (default) → historical synchronous SIGINT, byte-for-byte unchanged.
- Flag on + no tool in flight → fires immediately (already a clean boundary).
- Flag on + tool in flight → park the interrupt (ack ⚡), wait for `tool_result` (last open tool drains) or `turn_end`, then SIGINT + deliver the replacement body as a fresh turn.
- Bounded by `interrupt.max_wait_ms` (default 8000) — a long tool never strands the user.
- Rapid repeated `!` while one is parked **coalesce** into a single interrupt carrying the latest body.
- A bare `!` (halt-now, no body) always fires immediately, even with the flag on.

Serves `reference/steer-or-queue-mid-flight.md` — directly closes its "Unsafe interrupt point" anti-pattern ("Mid-tool-call is not 'amend time.'").

## Why

We can't pause claude's internal loop (unmodified-CLI constraint) — the only levers are SIGINT via tmux and observing the session JSONL. The "safe boundary" is derived from the session-event stream the gateway already receives: a `tool_use` opens an unsafe window, its matching `tool_result` closes it; `turn_end`/`enqueue` clear the slate. That in-flight tracker + the fire-now/defer decision live in a pure, fuzzable module (`telegram-plugin/gateway/interrupt-defer.ts`).

## Design notes / constraints

- Single pending slot (one claude per agent, singleton `currentTurn`) — matches the existing mid-turn architecture. Cross-thread coalescing into one slot is an accepted v1 edge under the singleton model.
- Empty `!` stays synchronous by design: it's an explicit halt, not an amend.
- Deferred timeout fires the SIGINT even if a tool is still open — the user explicitly asked to interrupt; the cap trades a tiny mid-tool-C-c risk for a guaranteed response.

## Plumbing

Mirrors the coalesce flag exactly: zod field in `src/config/schema.ts` → projection in `src/agents/scaffold.ts` (→ `access.json`) → `Access` type + `readAccessFile` passthrough + `loadAccess()` read in `gateway.ts`.

## Test plan

- [x] `interrupt-defer.test.ts` — 17 tests: tracker open/close/parallel/turn_end/enqueue clear, sub-agent/non-tool ignore, missing toolUseId, unknown-id no-op; `decideInterruptTiming` truth table; `resolveInterruptMaxWaitMs` floor.
- [x] `scaffold.test.ts` — +3: projects `interrupt.{safe_boundary,max_wait_ms}`, `false` not dropped as falsy, omitted when unset.
- [x] `tsc --noEmit` clean; `check-bot-api-wrapping.sh` clean; `check-plugin-references.mjs` clean.
- [x] `telegram-plugin` `tests/` suite 3714 pass / 0 fail; `vitest` scaffold (186) + merge (82) green. (UAT scenarios skip — no `TELEGRAM_API_ID` in CI worktree.)
- [ ] UAT: "Unsafe interrupt point" scenario with flag on — owed before fleet rollout (feature is default-off so safe to ship dormant).

## Risk

Low. Default-off; the off-path is the existing code unchanged. New behavior is gated entirely behind `safe_boundary === true`.


mekenthompson:
APPROVE

Deferred safe-boundary `!` interrupt — clean, gated entirely behind `interrupt.safe_boundary === true` (default false = byte-for-byte historical synchronous SIGINT). Pure decision core is well-isolated and thoroughly tested; gateway wiring mirrors the existing coalesce-flag plumbing. Merge once CI is green; run the UAT before enabling on any agent.

- `interrupt-defer.ts:30-76` — `ToolFlightTracker` is sound: per-`toolUseId` Set, `turn_end`/`enqueue` clear the slate (correctly handles a killed turn that never emits the trailing `tool_result`), sub-agent events ignored because the parent `Task` tool_use already represents the wait. Truth table in `decideInterruptTiming` (:90) is minimal and correct. 17 unit tests cover it.
- `gateway.ts:4255-4262` — boundary check runs after every session event and fires the parked interrupt the instant the last in-flight tool drains. `fireDeferredInterrupt` (:1426) nulls the slot + clears the timer before any await, so timer-vs-boundary can't double-fire. Idempotent, correct.
- `gateway.ts:9928-9956` — park/coalesce path: rapid repeated `!` replace the stashed body but preserve `registeredAt`/`deadlineTimer`, so the wait stays bounded. Good. Empty `!` stays synchronous (`!interrupt.emptyBody` gate at :8996) — correct, halt-now is not amend-time.

Nits / watch (non-blocking, both bounded by `max_wait_ms`):
- `gateway.ts:8989-9000` vs `:9928` — the defer DECISION reads `isMidToolCall()` early in `handleInbound`, but the slot is PARKED later at the delivery site after intervening awaits. If the in-flight tool finishes inside that await window, the `tool_result` boundary event fires while `pendingDeferredInterrupt` is still null (missed), so the interrupt waits out the full `max_wait_ms` timer instead of firing at the boundary that already passed. Bounded latency only, not a correctness break. Could re-check `isMidToolCall()` at park time and fire immediately if already clean.
- Same window enables a rare double-fire: a 2nd `!` arriving after the tool completed but before the missed boundary drained the 1st could take the synchronous path (`deferInterrupt=false`) and SIGINT+deliver while the 1st sits parked with a live timer → two SIGINTs + two deliveries. Requires precise timing (tool_result not yet processed by the ipc handler) and is gated off-by-default; worth a guard (e.g. clear/replace any pending slot on a synchronous fire).

Rationale: default-off means the shipped off-path is the existing code unchanged; all new behaviour is behind the flag; the pure core is fuzzable and tested; the cross-thread single-slot limitation and timeout-fires-anyway tradeoffs are documented and accepted under the singleton `currentTurn` model. The two race windows are latency/edge issues bounded by `max_wait_ms`, not media/state loss, and only reachable with the flag on.

PR #2017, opened 2026-05-31
feat(telegram): grow worker-activity feed narrative instead of one collapsing line (#2009)

### LINKED ISSUE #2009: Worker feed: stream the worker's narrative text so it reads more like the main agent's live answer
## Observed

Watching clerk dispatch a background `@researcher` worker (drafting the #11 sync spec), the worker-activity feed card appeared to freeze: it sat on `⚡ Bash … (13 tools · 01:00)` for ~98s before jumping to `⚡ Write personal-os-sync-spec.md (16 tools · 02:58) ↳ Spec written.`

This was **not a bug** — the gateway log confirms a real stall + recovery:

```
23:25:09 subagent-watcher: stall detected for abb92cc38c5906e5f (idle 60s): sub-agent
23:25:46 subagent-watcher: stall cleared (activity resumed after 98s — re-arming detection)
```

The sub-agent went quiet for ~98s mid-`Write` of the long doc. The elapsed timer kept ticking; only the current-tool line froze because no new JSONL row crossed.

## Why the worker feed looks different from the main agent's message

Two different data sources:

| | Main agent message | Worker feed |
|---|---|---|
| Source | bridge **live token stream** (the agent's words as generated) | sub-agent **on-disk JSONL**, polled ~1/s by the watcher |
| Granularity | token-level prose growing live | event-level: `⚡ <tool> (N tools · MM:SS) ↳ latest line` |
| During a long single tool call | prose keeps flowing | freezes on the tool; only the timer moves |

`worker-activity-feed.ts:6-16` documents that the feed deliberately mirrors the *form* of the main answer (an edit-in-place message, "indistinguishable from the agent typing live") but it **cannot tap a Task-tool sub-agent's live token stream** — the JSONL is the only observability surface for a background worker, and it records only completed events.

## Proposed improvement

Make the worker feed read closer to the main agent's live answer by surfacing the worker's **narrative text more fully**:

- Today `WorkerActivityView.latestSummary` shows a single capped line. Let the worker's narrative grow (like the main answer's prose) instead of collapsing to one line, so a long step reads as "live typing" rather than a frozen tool label.
- Consider a "still working" affordance during a known-long single tool (the elapsed timer alone isn't obviously *progress* vs *stall* to the user) — e.g. a subtle spinner/ellipsis on the tool line, distinct from the existing ⚠️ stall indicator.

## Hard limit (out of scope / non-goal)

The worker feed can **never** be token-smooth or gap-free during a single long tool call — there is no token-level stream for Task-tool sub-agents; the JSONL is event-granular (~1/s at best, silent during a long single tool). This issue is about narrowing the *perceived* gap, not eliminating the source-level difference.

## Source pointers

- `telegram-plugin/worker-activity-feed.ts` — `renderWorkerActivity` + `WorkerActivityView` (the render snapshot; `latestSummary` is the one-line cap to expand)
- `telegram-plugin/subagent-watcher.ts` — JSONL poll + stall detection that feeds the cues
- Main-answer counterparts for parity reference: `stream-reply-handler.ts`, `answer-stream.ts`, `draft-stream.ts`

## JTBD

Serves *you hold the leash* (awareness of in-flight work) and the "Chat is the artifact" principle — the worker should read like the agent working, not a status widget.

mekenthompson:
## Summary

Makes the live **worker-activity feed** read like the main agent's
growing answer instead of a flickering one-liner. The feed (the
edit-in-place message that surfaces a *background* sub-agent's progress)
previously replaced its single `↳` summary line on every watcher tick.
This accumulates distinct narrative lines into a growing `↳` block.

Closes #2009.

## Why

#2009: a long background worker's feed collapsed each new narrative onto
the prior one, so the operator saw one line mutate rather than a trail of
what the worker had done — the opposite of the "chat is the artifact" feel
the main agent's live answer already has.

## What changed

- `WorkerActivityView` gains an optional `narrativeLines?: string[]`.
- `renderWorkerActivity` renders a growing block of `↳` lines when
  `narrativeLines` is present/non-empty; otherwise it falls back to the
  single `latestSummary` line (**full back-compat** for direct callers).
- The feed manager accumulates `latestSummary` into a per-worker
  `narrative[]`, deduped against the preceding line and **capped to the
  last 6**. Accumulation runs *before* the throttle / cooldown /
  first-paint gates, so a line is never lost during a held tick — it
  surfaces on the next edit that fires.
- **No gateway change**: the manager builds the block from the
  per-emission `latestSummary` the watcher already passes on each
  `sub_agent_text`.

## Scope / non-goals

- The hard limit from the issue stands: during a *single long tool* the
  watcher emits no narrative, so the feed legitimately holds. This is
  event-granular surfacing, not token smoothing — we never fabricate
  progress.
- Skipped the optional animated "still working" spinner the issue floated
  under "Consider": an animating frame defeats the body-dedup that
  protects Telegram's edit budget. Deliberate scope call.

## Test plan

- [x] `vitest run telegram-plugin/tests/worker-activity-feed.test.ts` — 25 pass (was 14; +11 for growing block, dedup, cap, throttle-still-accumulates, HTML-escape, back-compat fallback).
- [x] `tsc --noEmit` clean.
- [x] `node scripts/check-plugin-references.mjs` clean.
- [ ] UAT `jtbd-worker-activity-feed-dm` after batched release (flag-gated `SWITCHROOM_WORKER_ACTIVITY_FEED=1`).

## Risk

Low. Render-only + manager-internal accumulation behind the existing
worker-feed flag; no gateway/IPC surface touched; single-line behavior
unchanged when `narrativeLines` is absent.

### DIFF
--- a/telegram-plugin/tests/worker-activity-feed.test.ts
+++ b/telegram-plugin/tests/worker-activity-feed.test.ts
@@ -92,6 +92,42 @@ describe('renderWorkerActivity', () => {
     expect(out).toContain('⚠️ <b>Worker failed</b>')
   })
 
+  it('grows a narrative block when narrativeLines is present', () => {
+    const out = renderWorkerActivity(
+      view({
+        latestSummary: 'newest only — should be ignored',
+        narrativeLines: ['read the brief', 'scanned vendor A', 'scanned vendor B'],
+      }),
+    )
+    expect(out).toContain('↳ <i>read the brief</i>')
+    expect(out).toContain('↳ <i>scanned vendor A</i>')
+    expect(out).toContain('↳ <i>scanned vendor B</i>')
+    // The single-line latestSummary fallback is NOT used when a block is present.
+    expect(out).not.toContain('newest only')
+    // Three narrative lines → three ↳ lines.
+    expect(out.match(/↳/g) ?? []).toHaveLength(3)
+  })
+
+  it('falls back to latestSummary when narrativeLines is empty', () => {
+    const out = renderWorkerActivity(view({ narrativeLines: [], latestSummary: 'one line' }))
+    expect(out).toContain('↳ <i>one line</i>')
+    expect(out.match(/↳/g) ?? []).toHaveLength(1)
+  })
+
+  it('drops blank narrative lines from the block', () => {
+    const out = renderWorkerActivity(
+      view({ narrativeLines: ['kept', '   ', 'also kept'] }),
+    )
+    expect(out).toContain('↳ <i>kept</i>')
+    expect(out).toContain('↳ <i>also kept</i>')
+    expect(out.match(/↳/g) ?? []).toHaveLength(2)
+  })
+
+  it('escapes HTML inside narrative lines', () => {
+    const out = renderWorkerActivity(view({ narrativeLines: ['a <b>x</b> & y'] }))
+    expect(out).toContain('a &lt;b&gt;x&lt;/b&gt; &amp; y')
+  })
+
   it('escapes HTML in description, tool, arg, and summary', () => {
     const out = renderWorkerActivity(
       view({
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

Render-only + manager-internal accumulation, fully back-compat, well-tested. Closes #2009. Merge once CI is green.

- `worker-activity-feed.ts:131-140` — `renderWorkerActivity` grows a `↳` block from `narrativeLines` when present/non-empty, else falls back to the single `latestSummary` line. Direct callers that don't pass `narrativeLines` are unaffected (back-compat preserved). HTML-escaped + truncated per line, same as the old path. Correct.
- `worker-activity-feed.ts:240-249` — `accumulateNarrative` dedups only against the immediately-preceding line (`h.narrative[length-1] === line`), not the whole block. That's the right call for the watcher's tick re-emission, but note an A→B→A sequence legitimately re-adds A (a real revisit), which is intended per the docstring. Cap-to-6 via `splice` is correct.
- `worker-activity-feed.ts:253-256` — accumulation runs BEFORE the cooldown/throttle/first-paint gates, so a line held during a throttled tick still lands in `narrative[]` and surfaces on the next edit that fires. This is the key correctness property and it's directly tested (`worker-activity-feed.test.ts` throttle case). Good.
- `worker-activity-feed.ts:257` — `renderWorkerActivity({ ...view, narrativeLines: h.narrative })` passes the live accumulated array; `narrative: []` initialized on handle creation (:336). No unbounded growth (capped at 6), no gateway/IPC surface touched.

Rationale: +11 tests cover the growing block, dedup, 6-line cap, throttle-still-accumulates, HTML-escape, and the empty→`latestSummary` fallback; behaviour is unchanged when `narrativeLines` is absent; the change is behind the existing `SWITCHROOM_WORKER_ACTIVITY_FEED` flag. The documented non-goal (no narrative emitted during a single long tool, so the feed legitimately holds) is honest — event-granular surfacing, not fabricated progress. UAT (`jtbd-worker-activity-feed-dm`) still owed but safe to ship behind the flag.

PR #2019, opened 2026-05-31
test(telegram): guard permission-verdict resume-reaction wiring

mekenthompson:
## Summary
Adds a structural regression guard pinning the permission-card resume beat: every `dispatchPermissionVerdict(...)` callsite in `gateway.ts` must be paired with a `resumeReactionAfterVerdict()` call (within 15 lines).

## Why
The v0.14.18 regression (fixed in v0.14.19) was exactly this: the free-text `y <id>`/`no <id>` reply path dispatched the verdict to un-park the suspended `claude` turn but never flipped the 🙏 awaiting glyph back to a working glyph — so answering a permission card by text reply left the card visually frozen even though the turn had resumed.

This bug class is **invisible to UAT**: the mtcute driver cannot observe Telegram reactions (driver disconnect drops them). The `StatusReactionController` state machine is well unit-tested in `status-reactions.test.ts`, but the per-path *wiring* in `gateway.ts` had no coverage. There are 5 independent verdict paths (button tap, always-allow, `/allow`·`/deny`, TTL auto-deny, free-text reply); a 6th could be added tomorrow and silently drop the resume again.

## Test plan
- [x] `vitest run telegram-plugin/tests/permission-verdict-resume-guard.test.ts` — 3/3 pass on current source.
- [x] Negative check: deleting any one `resumeReactionAfterVerdict()` makes the guard flag the orphaned dispatch callsite.
- [x] `tsc --noEmit` clean.
- [x] Confirmed file runs under the CI vitest sentinel (not excluded; not in the hardcoded bun lists, so no vitest/bun import conflict).

## Risk
Low — test-only, no runtime change. Static-text guard, so it pins the file-level invariant but won't catch a semantic break *inside* `resumeReactionAfterVerdict()` (that's the controller unit tests' job). The 15-line window has headroom over the widest real gap (~9 lines).

### DIFF
--- a/telegram-plugin/tests/permission-verdict-resume-guard.test.ts
+++ b/telegram-plugin/tests/permission-verdict-resume-guard.test.ts
@@ -1,0 +1,86 @@
+/**
+ * Structural pin for the permission-card resume beat.
+ *
+ * What broke (and the bug this guards against): when the operator
+ * answers a permission card, the suspended `claude` turn un-parks and
+ * resumes the SAME turn — the gateway must flip the awaiting glyph
+ * (🙏) back to a working glyph so the operator sees progress instead
+ * of a stuck card. That flip is `resumeReactionAfterVerdict()`.
+ *
+ * The verdict can arrive down several independent paths (button tap,
+ * always-allow, `/allow`·`/deny`, TTL auto-deny, free-text `y <id>`/
+ * `no <id>` reply, …). Every one of them calls
+ * `dispatchPermissionVerdict(...)` to un-park the turn — but the resume
+ * glyph flip is a *separate* call right next to it. The free-text-reply
+ * path shipped the dispatch WITHOUT the resume (fixed in v0.14.19), so
+ * answering via a text reply left the card frozen on 🙏 even though the
+ * turn was running. The controller-level behaviour is covered by
+ * `status-reactions.test.ts` ("setAwaiting" + watchdog re-arm); mtcute
+ * UAT cannot observe reactions at all, so this static pin is the only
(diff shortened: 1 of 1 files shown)


mekenthompson:

- `telegram-plugin/tests/permission-verdict-resume-guard.test.ts:34-40` — the callsite scan (`dispatchPermissionVerdict\s*\(` minus the `function …` definition) correctly resolves the 5 real verdict paths in `gateway/gateway.ts` (dispatch at 2990, 9106, 12003, 15649, 15844) and excludes the definition at 3989. Verified against live source.
- `…test.ts:46` — `RESUME_WINDOW = 15` has real headroom: the actual dispatch→resume gaps are 3 / 5 / 1 / 9 / 7 lines (widest is the slash-command path at 15649→15658 = 9). 15 leaves slack without letting an unrelated resume 'cover' a different block's dispatch. Good calibration.
- `…test.ts:78-81` — the third assertion pins `resumeReactionAfterVerdict` existence, closing the hole where deleting the helper would make the pairing vacuously pass. Sound.
- Nit (non-blocking) `…test.ts:31` — the regex matches the literal token in prose, so a future comment writing `dispatchPermissionVerdict(...)` (with a paren) inside a `//` block would register as a phantom callsite and could false-fail. Today's only comment mention (gateway.ts:2988) lacks the paren so is safely ignored; worth a `// not in a comment` note if it ever bites.

Rationale: test-only, zero runtime change, and it closes a real UAT blind spot — mtcute can't observe reactions, so this static pin is the only guard against a new verdict path dropping the resume glyph flip (the exact v0.14.18→.19 regression). Claims in the PR body check out against live source.

PR #2020, opened 2026-05-31
test(uat): ignore worker-feed messages in memory-recall matcher

mekenthompson:
## Summary
Makes the `jtbd-memory-survives-restart-dm` UAT robust to the worker-activity feed, which became default-on fleet-wide in v0.14.19.

## Why
The recall scenario matched the first non-empty bot message with `expectMessage(/\S/)`, then asserted the recalled `SWITCHROOM_UAT_MEM_*` token appears verbatim. Now that a background sub-agent can post a `🔧 Worker · …` feed message into the DM at any time, that matcher can latch onto the feed's first paint instead of the agent's reply — failing the token check against feed text and surfacing as a **false "memory miss"** rather than a real regression. The capture phase has the same exposure (wrong TTFO measurement).

## What changed
- `telegram-plugin/uat/assertions.ts` — new canonical `WORKER_FEED_RE` + `isWorkerFeedMessage()` (single source of truth for the feed's rendered shape).
- `jtbd-memory-survives-restart-dm.test.ts` — capture + recall matchers now use a feed-excluding predicate (`/\S/` AND not a feed message). The hard token-presence assertion is unchanged, so a genuine memory miss still fails.
- `telegram-plugin/uat/feed-matcher.test.ts` — new bun unit test pinning the helper (the live UAT needs sudo + a real Telegram session, so it can't run in CI; this is the CI-verifiable floor). Registered in `test:bun`.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `lint:bun-test-imports` clean (uat/ is excluded from vitest; bun:test import is correct there)
- [x] `bun test telegram-plugin/uat/feed-matcher.test.ts` → 6 pass
- [ ] CI green
- [ ] (optional) live `bun run --cwd telegram-plugin test:uat jtbd-memory-survives-restart-dm` on the harness

## Risk
Low — UAT-only. No runtime/product code touched. The predicate is strictly narrower than `/\S/` only for messages that match the feed header shape.


mekenthompson:

- `telegram-plugin/uat/assertions.ts:24` — `WORKER_FEED_RE` is a sound single source of truth for the feed shape; the emoji-anchored branches (`🔧 Worker`, `✅ Worker done`, `⚠️ Worker failed`) are tight. Good that the helper and the worker-feed scenario both consume it rather than re-deriving.
- `telegram-plugin/uat/assertions.ts:24` (non-blocking nit) — the emoji-less `Worker (?:done|failed)` alternative is the one broad branch: a real agent reply that literally contains 'worker done' / 'worker failed' in prose would be excluded as feed. Harmless in *this* scenario (the recall reply is a `SWITCHROOM_UAT_MEM_*` token, and the capture reply is free-form but won't hit that phrase), but if this matcher is reused elsewhere, anchor that branch to line-start or keep it emoji-gated. The PR's negative test at feed-matcher.test.ts:31 already pins the 'mentions worker' case, so the risk is bounded.
- `jtbd-memory-survives-restart-dm.test.ts:67-69` — `isReply = /\S/ && !isWorkerFeedMessage` is exactly the right narrowing; the hard token-presence assertion downstream is untouched, so a genuine memory miss still fails loudly (no masking of real regressions). Applied to both capture and recall matchers — correct, since the capture phase had the same TTFO-mismeasurement exposure.
- `package.json` (test:bun) — `feed-matcher.test.ts` is registered in the bun list (uat/ is vitest-excluded, bun:test import is correct), giving the helper a CI-verifiable floor even though the live UAT needs sudo + a real Telegram session. Right call.

Rationale: UAT-only, zero runtime/product code, and it fixes a real false-'memory miss' failure mode introduced when the worker feed went default-on in v0.14.19 — the predicate is strictly narrower than `/\S/` only for feed-shaped messages, and the token assertion that catches real regressions is preserved.

PR #2021, opened 2026-05-31
feat(telegram): default-on inbound coalescing, safe-boundary interrupt, worker feed

mekenthompson:
## Summary

Flips the three behavioral features shipped **default-off** in v0.14.20 to **default-on**, so they become the standard fleet behaviour with no flag/opt-in required.

| Feature | Old default | New default | Opt-out |
|---|---|---|---|
| A2 album coalescing (#2015) | `max_attachments=1` | `10` (full Telegram album) | set `max_attachments: 1` |
| `!` safe-boundary interrupt (#2016) | `safe_boundary=false` | `true` (defer to clean boundary, ≤8000ms) | set `safe_boundary: false` |
| Worker-activity feed (#2009) | flag off (`=1` to enable) | ON | `SWITCHROOM_WORKER_ACTIVITY_FEED=0` |

## Why

Operator directive: these three should be default behaviour, not flagged. They were sequenced default-off across PRs #2009/#2015/#2016 for a staged rollout; this PR completes that ramp.

## Implementation

- Each default is extracted into a small **pure, exported resolver** (`resolveCoalesceMaxAttachments`, `resolveSafeBoundaryEnabled`, `isWorkerActivityFeedEnabled`) mirroring the existing `resolveInterruptMaxWaitMs` seam, then called from the gateway in place of the inline expressions.
- Schema `describe()` text + `AccessFile` doc comments updated to document the new defaults and opt-outs.
- Scaffold projection unchanged (it still only projects when config sets a value; the default lives in the gateway), so the "field absent when unset" scaffold tests stay correct.

## Test plan

- [x] New unit tests pin each new default (undefined→10 / undefined→true / undefined→on) plus the opt-out and floor cases
- [x] `vitest` coalesce-attachments + interrupt-defer + worker-activity-feed: 63 pass
- [x] `vitest` scaffold.test.ts: 186 pass (projection unaffected)
- [x] `npm run lint` clean (tsc, plugin-references, bot-api-wrapping, bun-test-imports, no-pii, vault-hermeticity, no-broadcast)
- [ ] **Fleet UAT post-release** — #2015 album + #2016 interrupt paths go LIVE for the first time; UAT must exercise both (not just the worker feed) before rollout

## Risk

Medium — two paths (album coalescing, deferred interrupt) run in production for the first time at these defaults. Both are config-revertible without a code change. Worker feed already ran fleet-wide via the flag.

### DIFF
--- a/src/config/schema.ts
+++ b/src/config/schema.ts
@@ -666,14 +666,13 @@ export const TelegramChannelSchema = z
           .optional()
           .describe(
             "Maximum number of media attachments carried into ONE coalesced " +
-            "Claude turn. Default 1 — a second photo/document/voice within the " +
-            "coalesce window (or an album / media_group_id) starts its own turn, " +
-            "preserving the historical single-attachment behaviour. Raise to let " +
-            "a forwarded album or a text+multi-image burst arrive as one turn; " +
-            "the agent then sees numbered attachment fields (image_path, " +
-            "image_path_2, …). Excess attachments beyond the cap spill into the " +
-            "next turn. Each attachment is downloaded, so a high cap on a slow " +
-            "link delays turn start."
+            "Claude turn. Default 10 — a full Telegram album (media_group caps " +
+            "at 10) or a text+multi-image forwarded burst arrives as a single " +
+            "turn; the agent sees numbered attachment fields (image_path, " +
+            "image_path_2, …). Set 1 to restore the historical " +
+            "single-attachment-per-turn behaviour. Excess attachments beyond " +
+            "the cap spill into the next turn. Each attachment is downloaded, " +
+            "so a high cap on a slow link delays turn start."
           ),
       })
       .optional()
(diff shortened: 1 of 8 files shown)


mekenthompson:
APPROVE — merge once CI is green; do NOT skip the live fleet UAT (see last bullet).

- `telegram-plugin/gateway/coalesce-attachments.ts:39-47` — the new `DEFAULT_MAX_ATTACHMENTS = 10` const and `resolveCoalesceMaxAttachments` export were inserted *between* the `splitCoalescedAttachments` JSDoc block (which ends '…silently dropping the only attachment.') and the `splitCoalescedAttachments` function it documents — that docblock is now orphaned from its function. Cosmetic only: move the two new exports above the docblock (or give them their own placement) so the 'floored at 1' JSDoc sits on `splitCoalescedAttachments` again. Non-blocking.
- `telegram-plugin/gateway/coalesce-attachments.ts:44` — `Math.max(1, configured ?? 10)` is correct: default 10 (Telegram media_group cap), 0/negative floored to 1 so the only attachment is never stripped. Pinned by the new floor tests. Good.
- `telegram-plugin/gateway/interrupt-defer.ts:104-106` — `resolveSafeBoundaryEnabled = configured !== false` correctly makes undefined→true and only an explicit `false` opt out; matches the schema/doc-comment edits at src/config/schema.ts:690 and gateway.ts AccessFile doc. Consistent.
- `telegram-plugin/worker-activity-feed.ts:38-40` — `isWorkerActivityFeedEnabled = envVal !== '0'` flips the env semantics from the old `=== '1'`. Note the asymmetry: a pre-existing operator who set the var to anything non-empty-non-'0' (e.g. 'false', 'off') to *think* they disabled it under the old `=== '1'` rule would now have it ON. Documented opt-out is exactly `=0`, so this is acceptable, but worth a release-note line since the only safe disable token changed. Non-blocking.
- `telegram-plugin/gateway/gateway.ts:9007, 3152, 17578` — all three inline expressions correctly swapped for the resolvers; no stray old-default callsite left. Scaffold projection untouched (default lives in the gateway), so the 'field absent when unset' scaffold tests stay valid as the PR claims.
- Product risk (not a code defect): album coalescing (#2015) and deferred `!` interrupt (#2016) execute in production for the FIRST time at these defaults — they only ever ran flag-off before. The worker feed already ran fleet-wide via the flag, but the other two have no live mileage. Both are config-revertible without a code change. The post-release fleet UAT MUST exercise the album fold-in and the mid-tool-call interrupt path, not just the worker feed, before rollout.

Rationale: pure, well-tested default flips behind stable resolver seams, lint+tsc clean, fully revertible via config; the only real exposure is operational (two paths going live), which the PR correctly flags and gates behind UAT.

PR #2022, opened 2026-05-31
test(uat): skip tool-activity feed in memory-recall matcher

mekenthompson:
## Summary
Extends the memory-survives-restart recall matcher (#2020) to also skip the **tool-activity feed**, not just the worker-activity feed.

## Why
#2020 made the recall matcher ignore `🔧 Worker · …` feed messages. But the live UAT still failed: on any turn that uses tools — memory recall does — `renderActivityFeed` (`telegram-plugin/tool-activity-summary.ts:214`) paints `→ Finding the right tool` as its own bot message before the real answer lands. The `/\S/`-based matcher latched onto it, and the verbatim-token assertion failed against feed text rather than catching a genuine memory miss.

Observed live failure before this fix:
```
[memory-survives] capture phase: TTFO=8691ms, reply length=24
× ... CONTRACT FAILED: token SWITCHROOM_UAT_MEM_5177007E298E6CF8 not present.
  Reply was: "→ Finding the right tool"
```

## What changed
- `assertions.ts`: new `isActivityFeedMessage` — a message qualifies only when **every** non-empty line starts with `→ ` or `✓ ` (so a real reply that merely contains an arrow is never misclassified).
- `jtbd-memory-survives-restart-dm.test.ts`: `isReply` now excludes both feeds.
- `feed-matcher.test.ts`: +6 unit cases pinning `isActivityFeedMessage`.

A genuine recall miss now surfaces as a timeout instead of a false latch — the correct behavior for a regression guard.

## Test plan
- [x] `bun test telegram-plugin/uat/feed-matcher.test.ts` — 12 pass
- [x] `tsc --noEmit` clean
- [x] **Live UAT** `jtbd-memory-survives-restart-dm` now PASSES (recall TTFO ~27s, token round-trips) — same scenario that failed pre-fix

## Risk
Test-only change (UAT harness + its unit test). No runtime/agent code touched.

### DIFF
--- a/telegram-plugin/uat/assertions.ts
+++ b/telegram-plugin/uat/assertions.ts
@@ -35,6 +35,35 @@ export function isWorkerFeedMessage(m: ObservedMessage): boolean {
   return WORKER_FEED_RE.test(m.text);
 }
 
+/**
+ * A single tool-activity-feed line as rendered by
+ * `renderActivityFeed` (telegram-plugin/tool-activity-summary.ts): the
+ * in-progress step is `→ <label>`, finished steps are `✓ <label>`, and a
+ * long turn gets a `✓ +N earlier…` header. Telegram strips the bold/italic
+ * wrapping, so the observed text is just the marker glyph + label.
+ */
+const ACTIVITY_FEED_LINE_RE = /^[→✓]\s/u;
+
+/**
+ * True when `m` is the live tool-activity feed (the one-message list of
+ * "what the agent is doing this turn") rather than the agent's reply. A
+ * message qualifies only when EVERY non-empty line is an activity line —
+ * so a real reply that merely contains an arrow is never misclassified.
+ *
+ * Recall/reply scenarios must skip this in addition to
+ * {@link isWorkerFeedMessage}: on a turn that uses tools, the feed paints
+ * `→ Finding the right tool` as its own bot message before the real answer
+ * lands, and an `expectMessage(/\S/)` would otherwise latch onto it.
+ */
+export function isActivityFeedMessage(m: ObservedMessage): boolean {
+  const lines = m.text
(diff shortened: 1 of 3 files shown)


mekenthompson:

- `telegram-plugin/uat/assertions.ts:43` — `ACTIVITY_FEED_LINE_RE = /^[→✓]\s/u` anchored at line start is the right shape; combined with the every-line test it can't be tripped by a reply that contains an arrow mid-sentence (pinned by feed-matcher.test.ts:73). Good.
- `telegram-plugin/uat/assertions.ts:54-61` — `isActivityFeedMessage` requiring EVERY non-empty line to match (and returning false on zero lines) is the correct conservative choice: it only suppresses a message that is wholly feed chrome, so a genuine reply is never misclassified. The empty-message guard is pinned at feed-matcher.test.ts:78.
- `jtbd-memory-survives-restart-dm.test.ts:71-72` — `isReply` now ANDs both feed exclusions; the downstream verbatim-token assertion is untouched, so a real recall miss still fails (now as a timeout rather than a false latch — correct for a regression guard, as the PR states).
- Nit (non-blocking) `telegram-plugin/uat/assertions.ts:54` — a one-line agent reply that happens to be exactly `→ …` or `✓ …` (e.g. the agent literally answers with a single arrow-prefixed line) would be classified as feed. Not reachable in this scenario (recall reply is a `SWITCHROOM_UAT_MEM_*` token, capture reply is free-form prose), so harmless — worth a comment if this matcher gets reused for a scenario whose expected reply could start with → or ✓.

Rationale: test-only UAT harness change, no runtime code; closes a real false-'memory miss' (the live failure where `→ Finding the right tool` latched the matcher), the detector is strictly conservative, and the live UAT now passes per the PR's test plan.

Issue #865, opened 2026-05-09
Phase 1: mtcute driver bootstrap + session vaulting


mekenthompson:
Bootstrap the mtcute MTProto user-account driver and persist its session string to vault.

Part of epic #863.

## Tasks
- [ ] Add `mtcute` (or `@mtcute/node`) to `telegram-plugin` deps.
- [ ] Build `telegram-plugin/uat/driver.ts` — wraps mtcute client with: send text, send voice, observe messages/edits/reactions/pins in a target chat+topic, attach files.
- [ ] Build a one-time interactive login script (`bun run uat:login`) that prompts for phone + 2FA, captures the session string, and writes it to vault as `telegram-uat-driver-session`.
- [ ] Document the 2FA re-login playbook in `telegram-plugin/uat/SETUP.md` so it's not rediscovered at 2am.
- [ ] **Security:** session string is bearer-equivalent to the user account. Never log, never echo, never commit. Treat with same care as a bot token.

## Acceptance
- `bun run uat:login` produces a valid session string, vaulted.
- A smoke driver script can connect, send a message to the test supergroup, and observe the bot's reply.

Depends on: phase-1 supergroup + driver user-account membership.

mekenthompson:
Closed: shipped. mtcute driver lives in telegram-plugin/uat/, with operator login flow (`bun run uat:login`) documented in CLAUDE.md.

PR #2023, opened 2026-05-31
test(telegram): album-coalescing e2e UAT regression gate

mekenthompson:
## Summary
Adds the end-to-end UAT regression gate for inbound **album coalescing**, the one e2e coverage hole left by #2021 (which flipped `coalesce.max_attachments` default `1 → 10`, making albums fold into a single Claude turn by default).

- `driver.sendAlbum()` — the mtcute driver previously had no way to send a Telegram `media_group`; this posts N photos as one group, caption on the first item (matches real client behaviour).
- `fixtures/album/{red,green,blue}.jpg` — 3 tiny solid-colour 320×240 JPEGs (~674 B each), committed so the gate runs without a generation step.
- `jtbd-album-coalescing-dm.test.ts` — sends a 3-photo album asking the agent how many image files it received **in the one message**, asserts the reply is `3`.

## Why
Before #2021, an album bypassed coalescing (each part its own turn) and the agent only ever saw one image per turn. Post-flip it must see all three together. Nothing exercised a real `media_group` end-to-end, and the driver couldn't even send one — so a silent revert of album coalescing would pass every existing gate. A reply of **3** proves the album coalesced; a non-coalescing gateway answers **1**.

## Test plan
- [x] `tsc --noEmit` (strict) on the new scenario — clean
- [x] `check-bun-test-imports` — clean (imports from `vitest`, not `bun:test`)
- [ ] Live UAT run against `test-harness` during the v0.14.21 canary (this scenario doubles as the canary)

## Risk
Test-only — no runtime/src changes. Worst case is a flaky UAT scenario, which is non-gating (UAT is off the default CI critical path).

Part of: https://github.com/switchroom/switchroom/issues/865


mekenthompson:
REQUEST_CHANGES — the gate's reply matcher can latch onto feed chrome, which both flakes the test and can mask the very regression it guards.

- `telegram-plugin/uat/scenarios/jtbd-album-coalescing-dm.test.ts:54-57` — the reply predicate `(m) => !m.text.trimStart().startsWith('→') && /\d/.test(m.text)` is strictly weaker than the canonical feed matchers and is unsafe now that BOTH feeds are default-on (#2021):
  1. It does not exclude the **worker-activity feed**, which is default-on and carries digits: `✅ Worker done · 10 tools · 1:03`, `🔧 Worker · … · 0:12` all satisfy `!startsWith('→') && /\d/`. If one paints before the agent's answer the matcher latches onto it and the `/\b3\b|three/` assertion fails against feed text → **false fail / flake**.
  2. It only skips the `→` in-progress line, not the `✓`-prefixed activity lines — e.g. the `✓ +3 earlier…` header (tool-activity-summary.ts) starts with `✓`, contains the digit 3, and would latch AND coincidentally satisfy `\b3\b` → **false PASS that masks a silent album-coalescing revert**. That defeats the purpose of the gate.
  Fix: reuse the canonical helpers instead of an inline matcher — `isWorkerFeedMessage` is already on this branch (assertions.ts:34, from #2020) and `isActivityFeedMessage` lands in #2022. Rebase on #2022 and make the predicate `(m) => !isWorkerFeedMessage(m) && !isActivityFeedMessage(m) && /\d/.test(m.text)` (mirroring the `isReply` seam in jtbd-memory-survives-restart-dm.test.ts). Until #2022 merges, at minimum AND in `!isWorkerFeedMessage(m)`.

- `telegram-plugin/uat/driver.ts:660-674` — `sendAlbum` looks correct: `InputMedia.photo` with caption on item 0 only, `sendMediaGroup`, returns all ids. `replyTo = opts?.replyTo ?? opts?.messageThreadId` mirrors the existing sendText convention. No issue.

- `telegram-plugin/uat/scenarios/jtbd-album-coalescing-dm.test.ts:30-36` — committed fixtures + existence pre-check with a regenerate hint is a clean, hermetic approach. The 3→'agent reports 3' assertion correctly distinguishes coalesced (3) from non-coalescing (1). Good gate design once the matcher is fixed.

Rationale: test-only and non-gating, so not a BLOCK — but as written the matcher can both flake (worker-feed latch) and, worse, return a false PASS on a `✓ +N earlier…` header that happens to contain 3, silently masking an album-coalescing regression. The canonical matchers built in #2020/#2022 exist precisely to prevent this; reuse them.

PR #2029, opened 2026-05-31
fix(telegram): close two background-worker handback gaps

mekenthompson:
## Summary

A background worker (`Task` + `run_in_background:true`) finishes by firing the subagent-watcher's `onFinish`, which wakes the agent to reply to the user in its own voice (Option B handback). Two paths silently dropped that handback — the user dispatches a background job and is **never told the result**.

- **Gap 1 — restart survival.** The boot scan tags every JSONL present at startup `historical` to suppress replay of prior-session workers. But it also tagged workers still **running** at boot — ones dispatched in a prior gateway life and still in-flight across a restart / fleet rollout. Those completed with outcome `orphan` (which the handback gate drops) **and** were skipped by stall synthesis (`checkStalls` ignores historical entries), so a silent death never recovered either. Net: dispatched worker + any gateway bounce (incl. a routine fleet rollout) + worker finishes = user never told. **Fix:** a file still `running` at boot is promoted to a live entry (`historical=false`) so it gets progress nudges, the stall-synthesis safety net, and a real `completed`/`failed` handback. A file already `done` at boot stays suppressed (regression-guarded).
- **Gap 2 — failure honesty.** The `failed` outcome was dead code (no caller set it), so every dead worker was reported `completed`. **Fix:** a terminal error line in the worker's own transcript (model API failure / quota exhaustion / crash) flips the terminal outcome to `failed` and carries the error detail into the handback result slot.

## Why

Serves vision outcome 2 — *you hold the leash: awareness + control*. A delegated background job that finishes (or dies) without telling the operator breaks the always-available promise. The restart-survival gap is the worse of the two: a plain gateway bounce or fleet rollout — routine operations — strands any in-flight worker.

## Implementation notes

- Gap 1's promotion re-runs the **idempotent `jsonl_agent_id` backfill**, because the handback's `isBackground` lookup is keyed on it; an unlinked row (prior life crashed before the link persisted) would mis-resolve the promoted worker as foreground and drop the handback for a *different* reason.
- Gap 2 reuses `detectErrorInTranscriptLine` — the same classifier the operator-event path uses. `terminal:true` excludes in-flight retries (a 529 mid-backoff is `terminal:false`); routine tool-level `is_error` results never reach the check (they parse as `sub_agent_tool_result`). So neither a transient retry nor a grep-found-nothing trips a false failure.
- Errored workers stay in `running` state (not a new `failed` state) so stall synthesis still applies if they then go silent; the terminal outcome is computed at the `done` transition. The audit row (`recordSubagentEnd`) is kept consistent on both the turn_end and stall-synthesis paths.
- **Subscription-honest:** all visibility stays pure jsonl-tail → render. No model call added.

## Test plan

- [x] 6 new tests in `subagent-watcher-handback-gaps.test.ts` (3 per gap, incl. a done-at-boot regression guard and a mid-run tool-error negation)
- [x] Green under **both** vitest and `bun test`
- [x] `tsc --noEmit` clean
- [x] Adjacent suites still pass: `subagent-watcher-stall-terminal`, `subagent-handback-decision`, `subagent-handback-inbound-builder` (29 total)

## Risk

Low. The behavior change is scoped to (a) in-flight-at-boot workers, previously a guaranteed silent-drop, and (b) terminal-error classification of a previously-dead outcome. Done-at-boot suppression (the legitimate use of `historical`) is unchanged and regression-guarded.

### DIFF
--- a/telegram-plugin/subagent-watcher.ts
+++ b/telegram-plugin/subagent-watcher.ts
@@ -40,7 +40,7 @@ import {
 } from 'fs'
 import { basename, join } from 'path'
 import { homedir } from 'os'
-import { projectSubagentLine, sanitizeCwdToProjectName } from './session-tail.js'
+import { projectSubagentLine, sanitizeCwdToProjectName, detectErrorInTranscriptLine } from './session-tail.js'
 import { sanitiseToolArg } from './fleet-state.js'
 import { escapeHtml, truncate } from './card-format.js'
 import { bumpSubagentActivity, recordSubagentStall, recordSubagentResume, recordSubagentEnd, reapStuckRunningRows } from './registry/subagents-schema.js'
@@ -142,6 +142,21 @@ export interface WorkerEntry {
    * dead, the file is just left over from a prior session.
    */
   historical: boolean
+  /**
+   * True once a TERMINAL error line — a model API failure / quota
+   * exhaustion / crash, NOT an in-flight retry or a routine tool-level
+   * `is_error` result — has been observed in this worker's own
+   * transcript. Drives the `failed` terminal outcome so the handback
+   * tells the user the delegated work did NOT complete, instead of
+   * dressing a dead worker up as `completed`. Classified by
+   * `detectErrorInTranscriptLine` (the same gate the operator-event
+   * path uses), so transient mid-retry errors are excluded.
+   */
+  errored?: boolean
+  /** Human-readable detail from the terminal error line, surfaced in the
+   *  failed handback's "what it reported before failing" slot when the
+   *  worker left no narrative result of its own. */
+  errorDetail?: string
 }
 
 export interface SubagentWatcherConfig {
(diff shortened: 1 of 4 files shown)


mekenthompson:
**APPROVE**

Two genuinely-distinct silent-handback bugs, each fixed at the right layer with matching tests. Verified against the real diff and the cited symbols in the checked-out source.

- **Gap 1 — restart survival** (`subagent-watcher.ts:949-978`): an in-flight-at-boot worker was tagged `historical` (correct for done-at-boot, wrong for still-running) and then handed back as `orphan`, which the gate drops; it was also skipped by stall synthesis, so a silent death never recovered either. Fix promotes `isHistorical && entry.state === 'running'` to `historical=false` so it gets progress nudges, the stall safety net, and a real `completed`/`failed` handback. Correct and minimal.
- **Gap 1 backfill** (`subagent-watcher.ts:963-972`): re-running the idempotent `backfillJsonlAgentId` (defined `:449`) on promotion is the right call — the handback's `isBackground` lookup keys on `jsonl_agent_id`; an unlinked row from a crashed prior life would mis-resolve the worker as foreground and drop the handback for a different reason. Good defensive depth.
- **Gap 2 — failure honesty** (`subagent-watcher.ts:626-639`, `1040-1058`): `failed` was dead code; every dead worker reported `completed`. Fix flips the terminal outcome to `failed` when `detectErrorInTranscriptLine(line).terminal` is set. Confirmed against `session-tail.ts:435-533`: `invalid_request_error` → `terminal:true`; a 529 mid-retry → `transient:true, terminal:false` (correctly excluded); a `tool_result` with `is_error:true` never reaches the detector (projected as `sub_agent_tool_result`). So neither a transient retry nor a grep-found-nothing trips a false failure — exactly what the three Gap-2 tests assert.
- **Audit consistency** (`subagent-watcher.ts:745-750`, `1222-1228`): both `recordSubagentEnd` paths (real turn_end + stall synthesis) compute `entry.errored ? 'failed' : 'completed'`, so the audit row matches the handback outcome on both exits. No drift.
- **Result fallback** (`subagent-watcher.ts:1052-1056`): `entry.lastResultText || entry.errorDetail || ''` keeps the "what it reported before failing" slot non-empty on a crash with no narrative; `errorDetail` is capped at `SUBAGENT_RESULT_TEXT_MAX` (=3000, `:377`). Sound.

Nit (non-blocking): `entry.errored` is sticky once set — a terminal error followed by a later success would still report `failed`. Acceptable because terminal errors are final by construction (the `completed-after-a-retry` test correctly uses a non-terminal `tool_result` error), but worth a one-line comment if you revisit.

Watch: tsc-clean + green-under-vitest-and-bun are author-claimed (not re-run here); the symbol contracts they depend on are all verified present, so risk is low.

Rationale: scoped, well-reasoned fix for two real silent-drop paths in background-worker handback, with a regression guard for the legitimate done-at-boot suppression and a negation test for transient errors. Pure jsonl-tail→render, no model call added. Merge it.

PR #2030, opened 2026-05-31
feat(docker): add git-lfs to base agent image

mekenthompson:
## Summary
- Install `git-lfs` in the base image's Tier-1 apt block (next to `git`).
- Run `git lfs install --system --skip-repo` so the smudge/clean filters are registered image-wide (`/etc/gitconfig`), independent of each agent's per-agent HOME.
- Extend `tests/docker/e2e.test.ts`: pin `git-lfs` on PATH + assert the system-level `filter.lfs.smudge` config is present.

## Why
Agents that clone an LFS-tracked repo (e.g. a social-content backup with photos/videos stored via git-lfs) previously got ~130-byte pointer stubs instead of the real binaries — the base image shipped no `git-lfs`, so `clone`/`pull` never smudged. This made large media assets unusable to agents even after a successful clone. Concrete case: the `panorama-socials` content library (2,418 images + 256 videos tracked via LFS) for the marko agent.

`git lfs install --system` writes the filters to `/etc/gitconfig` so they apply even with a fresh per-agent HOME and no repo-local `git lfs install`.

## Test plan
- [x] `tsc --noEmit` clean
- [x] Verified on the pinned `node:22-trixie-slim` base: `git-lfs/3.6.1` installs, `git lfs install --system` writes `filter.lfs.smudge = git-lfs smudge -- %f`
- [ ] CI `build-base` + `build-dependents` green (rebuilds base with the new layer)
- [ ] e2e Tier-1 + filter assertions pass

## Risk
Low. Additive package + one global config write. Adds a small layer to the base image; no behaviour change for agents that don't use LFS. JTBD: a standing specialist (marko) that can actually use its content library.

### DIFF
--- a/docker/Dockerfile.base
+++ b/docker/Dockerfile.base
@@ -36,6 +36,12 @@ RUN echo "switchroom/base building for arch=${TARGETARCH}" >&2
 #   tmux       — agent supervisor surface (#725)
 #   bash, ca-certificates, procps — shell + TLS trust + ps
 #   git        — claude tool calls + agent workflows
+#   git-lfs    — large-asset repos (e.g. social-content backups with
+#                photos/videos tracked via LFS). Without it `git clone`/
+#                `pull` leave LFS files as ~130-byte pointer stubs instead
+#                of the real binaries. `git lfs install --system` (below)
+#                registers the smudge/clean filters image-wide so every
+#                agent's clones materialise LFS blobs.
 #   sqlite3    — kernel.db / scheduler.db / hindsight inspection
 #   openssh-client — claude bundle expects ssh-keygen
 #   curl       — also used by the bun installer below; keep above bun
@@ -88,6 +94,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
       procps \
       psmisc \
       git \
+      git-lfs \
       sqlite3 \
       openssh-client \
       curl \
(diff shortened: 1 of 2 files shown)


mekenthompson:
**APPROVE**

Additive, correctly-ordered, well-tested. Verified against the full diff.

- `docker/Dockerfile.base:97` — `git-lfs` added to the Tier-1 apt package list, installed in the same `RUN` that already pulls `git`. Correct placement.
- `docker/Dockerfile.base:124-125` — `git lfs install --system --skip-repo` appended to the tail of that same `RUN` (after the `fd` symlink). Ordering is right: git-lfs is apt-installed earlier in the same layer, so the binary exists before `git lfs install` runs. `--system` writes filters to `/etc/gitconfig` (image-wide, survives a fresh per-agent HOME) and `--skip-repo` correctly avoids touching a repo that doesn't exist in the build context. This is the right way to make smudge/clean apply to every agent's clones.
- `tests/docker/e2e.test.ts:296,306` — Tier-1 PATH pin updated to include `git-lfs` in both the human-readable test name and the `command -v` loop. Keeps the regression guard honest.
- `tests/docker/e2e.test.ts:314-330` — new test asserts `git config --system --get filter.lfs.smudge` exits 0 and matches `/git-lfs smudge/`, i.e. the system-level filter was actually registered (not just the binary present). Good — it tests the behaviour, not just installation.

Watch (non-blocking): git-lfs smudge filters auto-execute `git-lfs` on clone/checkout, so cloning an LFS repo will now trigger LFS object downloads automatically. Inherent to git-lfs and expected for this use case; no action needed, just noting the new auto-fetch surface. PR's own test plan leaves CI build-base/build-dependents and the e2e run unchecked — those gate on the merge build, which is the right place.

Rationale: small additive base-image change that fixes a real bug (LFS repos materialising as pointer stubs), installed in the correct layer/order, with a behaviour-level test for the system filter. Merge it.

PR #2032, opened 2026-05-31
fix(telegram): age-bound boot-promotion to stop stale dead-worker handback replay

mekenthompson:
## Summary

Hotfix for a fleet-wide regression introduced by #2029 (v0.14.23). That PR promoted **every** worker file still in `running` state at boot to "live" so a worker in-flight across a restart would hand back instead of dropping as `orphan`. The flaw: a worker that died in a **prior session** without writing a terminal `turn_end` is *also* `running` in its file — and those stale files accumulate by the dozen-to-hundred in a long-lived agent's `subagents/` dir. On the v0.14.23 rollout, every agent's boot scan promoted all of them and replayed weeks-old handbacks — many `failed`, from old error lines in the transcripts — spamming the whole fleet.

**Observed in prod:** clerk promoted **218** files at a single boot instant (22:25:22), with file mtimes **3–4 weeks old** (April 28 / May 5). finn 184, reggie 54, etc.

## Fix

Gate the promotion on **file freshness**:
- Promote a running-at-boot file only if its last write (mtime) is within `DEFAULT_INFLIGHT_PROMOTE_MAX_AGE_MS` (15 min; env `SWITCHROOM_SUBAGENT_INFLIGHT_MAX_AGE_MS`).
- A genuinely in-flight worker wrote seconds-to-minutes before the bounce → still promoted, Gap 1 preserved.
- A dead prior-session file is hours-to-weeks stale → stays `historical`/suppressed, exactly the pre-v0.14.23 behaviour.
- Unreadable mtime → treated as stale (suppress rather than risk re-spamming).

Plus the **kill-switch #2029 should have had**: `SWITCHROOM_SUBAGENT_BOOT_PROMOTE=0` disables promotion entirely (full revert to pre-fix suppression) with no code change.

## Why

The `historical` flag existed precisely to suppress replay of prior-session workers; #2029 broke that for the running-at-boot subset without bounding it. The mtime cleanly separates the two populations (restart-gap seconds-to-minutes vs. dead-weeks). Serves vision outcome 2 (awareness + control) without the spam.

## Test plan

- [x] 4 new tests: 21-day-old running-at-boot **not** promoted (no handback, no stall); 30s-old running-at-boot **still** promoted + handed back; kill-switch suppresses even a fresh file; boundary just past the window is stale. Harness extended to drive file mtime.
- [x] Existing 6 Gap-1/Gap-2 tests still green.
- [x] Full telegram-plugin suite green under **both** vitest (3128) and bun (3745); `tsc --noEmit` clean.

## Risk

Low and de-risking. Narrows an over-broad behaviour back toward the long-standing suppression default; the only files that change handling are stale dead ones (now correctly suppressed) — fresh in-flight workers behave as in #2029. Adds an emergency kill-switch.

## Rollout

Ships as v0.14.24. The fixed boot scan suppresses the stale files, so the rollout itself is clean (no replay). Follows #2029 (already merged/rolled).

### DIFF
--- a/telegram-plugin/subagent-watcher.ts
+++ b/telegram-plugin/subagent-watcher.ts
@@ -209,6 +209,23 @@ export interface SubagentWatcherConfig {
    */
   silentStallTerminalMs?: number
   /**
+   * Freshness window (ms) for promoting a running-at-boot worker file to
+   * live. A file whose last write (mtime) is older than this is treated as
+   * a dead prior-session worker and stays historical/suppressed, NOT
+   * promoted. Default 15 min (DEFAULT_INFLIGHT_PROMOTE_MAX_AGE_MS); env
+   * override `SWITCHROOM_SUBAGENT_INFLIGHT_MAX_AGE_MS`. Guards the v0.14.23
+   * stale-handback replay regression.
+   */
+  inflightPromoteMaxAgeMs?: number
+  /**
+   * Kill-switch for the boot-scan promotion path. When false, a
+   * running-at-boot worker is never promoted — the watcher reverts to the
+   * pre-v0.14.23 behaviour of leaving every boot-scan file historical
+   * (suppressed). Default true; env `SWITCHROOM_SUBAGENT_BOOT_PROMOTE=0`
+   * disables it fleet-wide without a code change (emergency lever).
+   */
+  bootPromoteEnabled?: boolean
+  /**
    * Reaper TTL (ms): background rows in `status='running'` whose
    * `last_activity_at` (or `started_at` if liveness never wrote) is older
    * than this are transitioned to `status='stalled'` with a result_summary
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

Clean, correctly-scoped hotfix for the #2029 (v0.14.23) stale-handback replay regression. Reviewed the full diff and confirmed the runtime contract against the branch.

- `subagent-watcher.ts:1024` — `fileAgeMs = n - st.mtimeMs` is sound: `n` is `nowFn()` bound at :946 in the same registration closure, so the age is wall-clock-now minus file mtime. Correct.
- `subagent-watcher.ts:856` — kill-switch reads `process.env.SWITCHROOM_SUBAGENT_BOOT_PROMOTE !== '0'` directly rather than `parseEnvMs`, with an explicit comment on why (parseEnvMs rejects `0`). Right call — an explicit `=0` must disable, and routing it through the ms-parser would have silently ignored it. This is the lever #2029 should have shipped with.
- `subagent-watcher.ts:1018-1023` — unreadable mtime → `fileAgeMs = Infinity` → treated as stale → suppressed. Conservative and correct: better to drop a handback than re-spam the fleet.
- Freshness gate cleanly separates the two `running`-at-boot populations (restart-gap seconds-to-minutes vs. dead-prior-session hours-to-weeks). The Gap-1 promotion path is preserved for genuinely in-flight workers and the backfill remains inside the promote branch (idempotent jsonl_agent_id link intact).
- Tests (`subagent-watcher-handback-gaps.test.ts:266-334`) cover all four arms: 21-day stale not promoted, 30s fresh still promoted + handed back, kill-switch suppresses a fresh file, and the just-past-window boundary. Harness extended to drive per-file mtime — the boundary math checks out (90s age vs 60s window → stale).

Non-blocking watch item: the 15-min default assumes container recreate + image pull completes inside that window. A pathologically slow rollout (>15min) would suppress a genuinely in-flight worker as a false negative. Acceptable — it's env-tunable (`SWITCHROOM_SUBAGENT_INFLIGHT_MAX_AGE_MS`) and a dropped handback is the safe failure direction vs. fleet-wide replay.

Rationale: narrows an over-broad #2029 behaviour back toward the long-standing `historical` suppression default, the only files whose handling changes are stale dead ones (now correctly suppressed), and it adds the emergency kill-switch. Net de-risking.

PR #2034, opened 2026-05-31
fix(telegram): surface foreground sub-agent activity after ack-first reply

mekenthompson:
## Summary

Foreground sub-agent live nesting (#2027, Model A) never actually showed
anything in practice. Every render path bailed on `turn.replyCalled`, but the
framework's ack-first pattern replies **"On it…" first, then delegates** — so
`replyCalled` is already `true` before any foreground sub-agent runs. The
feature silently produced **zero live foreground activity** for the exact case
it was built for.

This was the root cause behind "marko's researcher sub-agent showed no Telegram
activity": ack at 23:14:08 → `replyCalled=true` → researcher ran 23:14–23:16
with every nesting tick returning early → user saw ack, then silence, then the
final answer.

## Why it went unnoticed

#2027 added tests for the pure renderer (`renderActivityFeedWithNested`) but
**none for the gate**. The boolean that broke it was never exercised with
`replyCalled=true`. This PR closes that seam.

## What changed

- **`telegram-plugin/gateway/foreground-nesting.ts`** (new, pure): two
  decision helpers extracted so the gate is unit-testable —
  `shouldRenderForegroundProgress` (kill-switch only; `replyCalled` accepted
  but intentionally **not** a gate) and `foregroundFinishAction`
  (`recompose` / `handoff-clear` / `none`).
- **gateway.ts onProgress**: render foreground progress regardless of
  `replyCalled`. A foreground `Task` blocks the parent, so any reply observed
  while it runs is an *interim* ack, never the final answer — the invariant
  that makes this safe.
- **gateway.ts onFinish**: when the **last** foreground sub-agent finishes
  post-ack, hand the re-opened feed off to the imminent answer
  (`clearActivitySummary`), mirroring the first-reply hand-off. Pre-ack
  behaviour (`recompose`) unchanged.
- **`tests/foreground-nesting.test.ts`** (new): pins replyCalled-independence,
  the finish state machine, and an end-to-end render-shape assertion for
  marko's scenario.

## Safety / scope

- Kill-switch unchanged: `SWITCHROOM_FOREGROUND_SUBAGENT_NESTING=0` disables
  the whole path (now folded into the predicate).
- turn_end (`gateway.ts:7884`) unconditionally clears the feed, so a re-opened
  post-ack feed is always cleaned up even if the sub-agent stalls/never fires
  `onFinish`.
- `tool_label` (parent's own steps) hand-off at first reply is **left as-is** —
  only the foreground sub-agent narrative re-opens the lane, keeping blast
  radius minimal and avoiding the flicker-below-final-answer the existing guard
  prevents.
- Subscription-honest: pure jsonl-tail → render, no model call.

## Test plan

- [x] `vitest run` new + sibling `tool-activity-summary` tests (30 pass)
- [x] `bun test` new test (9 pass — runs under both runners)
- [x] `tsc --noEmit` clean

JTBD-2 ("you hold the leash"): closes the named foreground visibility blindspot.

### DIFF
--- a/telegram-plugin/gateway/foreground-nesting.ts
+++ b/telegram-plugin/gateway/foreground-nesting.ts
@@ -1,0 +1,65 @@
+/**
+ * Pure state-transition helpers for Model A foreground sub-agent activity
+ * nesting (#2027). Extracted from gateway.ts so the `replyCalled` gate is
+ * unit-testable on its own — the exact seam #2027 shipped without, which let
+ * the feature silently no-op for every ack-first turn (reply "On it…" then
+ * delegate) and go unnoticed in production. See
+ * `tests/foreground-nesting.test.ts`.
+ *
+ * Subscription-honest: every caller is pure jsonl-tail → render, no model
+ * call. These functions decide *whether/what* to render; the gateway owns the
+ * `currentTurn` mutation and the Telegram edit.
+ */
+
+export interface ForegroundProgressGate {
+  /** `SWITCHROOM_FOREGROUND_SUBAGENT_NESTING !== '0'` — the kill-switch. */
+  nestingEnabled: boolean
+  /**
+   * Whether the parent turn has already emitted a reply. Accepted as input
+   * but intentionally NOT a gate: a foreground `Task` is a *blocking* call,
+   * so the parent cannot have issued its FINAL answer while the sub-agent is
+   * still running — any `replyCalled === true` observed here is therefore an
+   * *interim* ack. The pre-fix code bailed on this flag, which is precisely
+   * why ack-first turns showed zero live foreground activity.
+   */
+  replyCalled: boolean
+}
+
+/**
+ * Whether a foreground sub-agent progress tick should accumulate its narrative
+ * and re-render the activity feed. Gated ONLY by the kill-switch — never by
+ * `replyCalled` (see the field doc above).
+ */
+export function shouldRenderForegroundProgress(g: ForegroundProgressGate): boolean {
+  return g.nestingEnabled
+}
+
+export type ForegroundFinishAction = 'recompose' | 'handoff-clear' | 'none'
+
+export interface ForegroundFinishInput {
+  /** Was this agentId actually an active foreground sub-agent we tracked? */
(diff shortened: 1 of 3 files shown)


mekenthompson:
**APPROVE**

- `gateway/foreground-nesting.ts:35` (`shouldRenderForegroundProgress`) — correctly gates on the kill-switch only and treats `replyCalled` as non-gating input. The invariant holds: a foreground `Task` is a *blocking* call, so the parent cannot have emitted its FINAL answer while the sub-agent runs — any `replyCalled===true` observed here is necessarily an interim ack. This is the exact bug from #2027 (every render path bailed on `turn.replyCalled`, which ack-first sets true before delegation), now fixed.
- `gateway/gateway.ts:17989-17999` (onProgress) — render path now calls `shouldRenderForegroundProgress` instead of the old `turn.replyCalled` early-return. Behaviour-preserving for the kill-switch; correctly drops the replyCalled gate. `turn == null` guard retained.
- `gateway/gateway.ts:17850-17872` (onFinish) — `foregroundFinishAction` state machine is correct: `handoff-clear` only when the *last* tracked foreground sub-agent finishes post-ack (`remainingForeground===0 && replyCalled`), else `recompose`, else `none` for untracked agents. `clearActivitySummary` (gateway.ts:7339) is idempotent, so the handoff-clear is safe.
- `gateway/gateway.ts:7884` — turn_end unconditionally calls `clearActivitySummary(turn)`, confirming the stated safety net: a re-opened post-ack feed is always cleaned up even if the sub-agent stalls and never fires onFinish.
- `tests/foreground-nesting.test.ts` — pins replyCalled-independence (the #2027 blindspot), the full finish state machine, and an end-to-end render-shape assertion against `renderActivityFeedWithNested` (exists, tool-activity-summary.ts:245). This is the gate-level coverage #2027 lacked.

Rationale: tightly-scoped fix extracting the gate into a pure, unit-tested module and removing the `replyCalled` early-return that silently no-op'd foreground nesting for every ack-first turn; invariant is sound, safety net confirmed, kill-switch preserved. Watch item below is non-blocking.

PR #2035, opened 2026-05-31
test(telegram): UAT for foreground sub-agent activity nesting (#2032)

mekenthompson:
## Summary

Adds the mtcute UAT that exercises the #2034 fix end-to-end and pins it as a permanent regression guard.

The scenario forces the exact broken shape #2027 missed:
1. Reply with a quick ack first (`On it — running a check now.`) → `replyCalled=true`.
2. **Then** dispatch a FOREGROUND sub-agent (`run_in_background:false`, `subagent_type:"general-purpose"`) that narrates eight paced steps (`sleep 2` + `echo`, each its own Bash call, ~2s apart to stay under the test-harness stall floor).
3. Report done.

Asserts the activity-summary feed message carrying the nested foreground marker (`↳`) appears **after** the ack and advances in place — the load-bearing proof of post-ack foreground visibility. This is a real `sendMessage`/`editMessageText` (not a draft), so mtcute can observe it.

## Why

#2027 only tested the pure renderer; the regression (foreground sub-agent steps silently never painting once `replyCalled` was true) shipped unnoticed — the "marko's researcher showed no Telegram activity" report. #2034 fixed it; this UAT makes the live behavior a gating scenario so it can't silently regress again.

## Task-agnostic on purpose

The sub-agent is a generic sleep/echo worker, **not** a researcher. Foreground classification keys on `run_in_background` (PreToolUse hook), never `subagent_type` — the UAT pins that property.

## Test plan

- [x] Ran against a hot-patched test-harness carrying the #2034 gateway bundle. PASS (46s):
  - ack-first reply fired (`→ Finding the right tool`)
  - nested feed painted post-ack: `✓ Finding the right tool` + `   ↳ → I will begin with step 1…`
  - body advanced in-flight (`body moved in-flight: true`)
- [ ] CI green

## Risk

Low — test-only addition, no runtime code touched.

### DIFF
--- a/telegram-plugin/uat/scenarios/jtbd-foreground-subagent-activity-dm.test.ts
+++ b/telegram-plugin/uat/scenarios/jtbd-foreground-subagent-activity-dm.test.ts
@@ -1,0 +1,121 @@
+/**
+ * Foreground sub-agent live activity nesting (#2032) — UAT.
+ *
+ * Background (#2027): a FOREGROUND sub-agent (Agent/Task WITHOUT
+ * run_in_background) runs INSIDE the parent turn, blocking it. Its live steps
+ * were meant to nest into the parent's activity-summary feed (Model A). But
+ * every render path bailed on `turn.replyCalled`, and the framework's
+ * ack-first pattern replies "On it…" FIRST — so the sub-agent ran with
+ * replyCalled already true and NOTHING ever painted. #2027 only tested the
+ * pure renderer, so the regression shipped silently (the "marko's researcher
+ * showed no Telegram activity" report). #2032 renders foreground progress
+ * regardless of replyCalled.
+ *
+ * This scenario forces the exact broken shape end-to-end:
+ *   1. Prompt the agent to send a quick ack FIRST (sets replyCalled=true),
+ *   2. THEN dispatch a FOREGROUND sub-agent (run_in_background:false) that
+ *      narrates ~10 paced steps so its jsonl keeps ticking under the
+ *      test-harness 5s stall floor and the feed can paint + edit,
+ *   3. then report done.
+ *
+ * Asserts the load-bearing proof: an activity-summary feed message appears
+ * carrying the NESTED foreground marker ("↳") with the sub-agent's narration —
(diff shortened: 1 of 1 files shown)


mekenthompson:
**APPROVE**

- `uat/scenarios/jtbd-foreground-subagent-activity-dm.test.ts:60-66` — the load-bearing assertion (`expectMessage(NESTED_RE)` for `/↳/`) correctly matches the renderer: `NESTED_PREFIX = "   ↳ "` at `tool-activity-summary.ts:234` and the bold current step at `:214`. The glyph the test waits for is the one the code actually emits.
- `...test.ts:46` (`FG_DISPATCH_PROMPT`) — forces the exact #2027-broken shape: ack first (sets `replyCalled=true`) then a FOREGROUND sub-agent (`run_in_background:false`). This is precisely the condition #2034 fixed, so the scenario is a faithful end-to-end regression guard, not a tautology against the new code.
- `...test.ts:38-40` — paced `sleep 2`/`echo` as separate Bash calls keeps `sub_agent_text` ticking under the 5s stall floor so the watcher won't synth-terminate mid-flight and suppress onProgress. Matches the discipline of the sibling `jtbd-worker-activity-feed-dm.test.ts` (present in tree). Sound.
- Harness API all resolves: `spinUp`/`sendDM`/`expectMessage`/`tearDown`/`driver`/`botUserId` (`uat/harness.ts:140,179,180,195`) and `driver.getMessage(chatId, messageId)` (`uat/driver.ts:532`). No undefined references.
- `...test.ts:70-79` — the in-flight body-moved check is intentionally soft (logged, not asserted), which is the right call given Telegram edit-timing nondeterminism; the nested post-ack paint remains the hard assertion. Good separation of load-bearing vs. observational.

Rationale: test-only addition (1 new file, +121/-0, no runtime code), every harness symbol and the `↳` marker verified against the real tree, and the scenario reproduces the genuine #2027 failure shape rather than rubber-stamping #2034. Safe to merge; only watch item is CI flakiness on a timing-heavy 300s integration test (non-blocking).

PR #2039, opened 2026-06-01
fix: clear agent composer before inbound delivery (idle-session wedge)

mekenthompson:
## Summary

Inbound Telegram messages reach the agent's `claude` session as an MCP `notifications/claude/channel` notification (NOT via tmux `send-keys`). The unmodified CLI appends that text into its input composer and auto-submits **only when the composer is empty and idle**.

The existing #1556 inbound-delivery-gate (`gateway/inbound-delivery-gate.ts`, the "lawgpt composer-wedge" fix) defers delivery until `turnInFlight === false` — but it **assumes idle implies an empty composer**. That assumption breaks when stale typed-ahead / ghost text is stranded in the composer and was never submitted.

**Observed live on agent `marko`:** the literal text `"Yes, go ahead on both"` sat stranded in the composer. The appended inbound then failed to submit cleanly and the session silently swallowed every subsequent queued inbound until a hard container restart.

No code anywhere cleared the composer before delivery — that absence was both the bug and the fix.

## Why / outcome

Serves vision outcome **#4 (always available)** — an idle agent must reliably accept the next inbound, not silently wedge until a restart. JTBD: messages land and get actioned without operator babysitting.

## What changed

- **`src/agents/tmux.ts`** — new exported `clearAgentComposer({ agentName })`, modeled on `sendAgentInterrupt`. Issues `tmux send-keys -t <agent> C-u C-a C-k` on socket `switchroom-<agent>` (C-u kills to line start; C-a C-k moves to start and kills to line end — wipes the whole line regardless of cursor position). Soft-fail by contract: returns `{ error }` on any tmux failure (no session under `legacy_pty`, socket missing, timeout, exec error), never throws. 3s timeout.
- **`telegram-plugin/gateway/gateway.ts`** — wired into the deliver path immediately before `ipcServer.sendToAgent`, after the #1556 gate has already guaranteed idle. Dynamic-imports the helper (same pattern as the interrupt-marker `sendAgentInterrupt` callsite). Guarded so a clear failure logs to stderr and proceeds — a clear failure must never block delivery. Only fires when `selfAgent` is set.

An agent under `experimental.legacy_pty=true` has no tmux session, so `clearAgentComposer` returns `{ error }` and delivery proceeds unchanged — matching how `sendAgentInterrupt` routes. On the happy path (already-empty composer) the clear is a no-op.

## Test plan

- [x] `vitest run tests/tmux-send-interrupt.test.ts` — extended with a `clearAgentComposer` suite: asserts the exact `send-keys C-u C-a C-k` argv against the right pane/socket, the 3s timeout, soft-fail on a "no server running" (legacy_pty) error, and soft-fail on timeout-style errors. 12 tests pass.
- [x] `vitest run telegram-plugin/tests/inbound-delivery-gate.test.ts` — 10 tests pass (unchanged behavior).
- [x] `tsc --noEmit` clean for touched files.
- [x] `scripts/check-bot-api-wrapping.sh` clean (no allowlist drift — the edit is above the existing `bot.api.sendMessage` callsite).

## Risk

Low. The clear is additive and soft-fails closed-to-delivery (a failure never blocks an inbound). Happy-path (empty composer) C-u is a no-op, preserving existing behavior. Only the deliver path for tmux-supervised agents is affected; legacy_pty agents are untouched.

Refs: the `marko` stranded-composer incident; builds on #1556 (lawgpt composer-wedge idle-gate).

### DIFF
--- a/src/agents/tmux.ts
+++ b/src/agents/tmux.ts
@@ -182,6 +182,62 @@ export function sendAgentInterrupt(opts: SendInterruptOptions): SendInterruptRes
   return { error: lastError ?? "tmux send-keys C-c failed" };
 }
 
+export interface ClearComposerOptions {
+  agentName: string;
+}
+
+export type ClearComposerResult = { ok: true } | { error: string };
+
+/**
+ * Wipe any stranded typed-ahead / ghost text out of the agent's claude TUI
+ * composer via `tmux send-keys`, BEFORE an inbound channel notification is
+ * delivered.
+ *
+ * ## Why this exists (the marko wedge)
+ *
+ * Inbound Telegram messages reach the agent's claude session as an MCP
+ * `notifications/claude/channel` notification (NOT via send-keys). The
+ * unmodified CLI appends that text into its input composer and auto-submits
+ * ONLY when the composer is empty and idle. The #1556 inbound-delivery-gate
+ * defers delivery until `turnInFlight === false`, but it ASSUMES idle ⇒
+ * empty composer. That assumption breaks when stale typed-ahead text is
+ * stranded in the composer and was never submitted (observed live on agent
+ * `marko`: the literal text "Yes, go ahead on both" sat in the composer).
+ * The appended inbound then fails to submit cleanly and the session silently
+ * swallows every subsequent queued inbound until a hard container restart.
+ *
+ * Nothing else clears the composer, so we do it here: `C-u` (kill from
+ * cursor to line start) plus belt-and-suspenders `C-a` (move to start) and
+ * `C-k` (kill to line end) wipe the whole line regardless of cursor
+ * position. On an already-empty composer (the happy path) these are no-ops.
+ *
+ * ## Soft-fail contract
+ *
+ * Returns `{ error }` on any tmux failure (no session — e.g. an agent under
+ * `experimental.legacy_pty=true` has no tmux session, socket missing,
+ * timeout, exec error). Never throws. Callers MUST proceed with delivery
+ * regardless — a clear failure must never block an inbound.
+ */
+export function clearAgentComposer(opts: ClearComposerOptions): ClearComposerResult {
+  const { agentName } = opts;
+  const socket = `switchroom-${agentName}`;
+  // C-u: kill to line start. C-a C-k: move to start, kill to line end.
+  // Together they wipe the composer line regardless of cursor position.
+  const args = ["-L", socket, "send-keys", "-t", agentName, "C-u", "C-a", "C-k"];
+
+  try {
+    execFileSync("tmux", args, {
+      timeout: 3000,
+      stdio: ["ignore", "pipe", "pipe"],
+    });
+    return { ok: true };
+  } catch (err) {
+    const msg = `tmux send-keys composer-clear failed: ${(err as Error).message}`;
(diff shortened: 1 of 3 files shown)


mekenthompson:
**APPROVE**

- `src/agents/tmux.ts:213` (`clearAgentComposer`) — correctly modeled on the existing `sendAgentInterrupt`: same `execFileSync` (imported tmux.ts:16), same `-L switchroom-<agent>` socket convention, 3s timeout, `{ok:true} | {error}` soft-fail with no throw. The `C-u C-a C-k` sequence wipes the line regardless of cursor position and is a no-op on an empty composer — the happy path is preserved.
- `telegram-plugin/gateway/gateway.ts:10088` — insertion point is correct: it sits AFTER the `decideInboundDelivery(...) === 'buffer-until-idle'` early-return (the #1556 gate), so the turn is guaranteed idle before the clear runs, and immediately BEFORE `ipcServer.sendToAgent(selfAgent, inboundMsg)`. `selfAgent` is in scope (same var used at the sendToAgent and pendingInboundBuffer.push calls). Guarded on `if (selfAgent)`.
- `gateway.ts:10099` — dynamic `await import('../../src/agents/tmux.js')` matches the established precedent for the sibling helper (`sendAgentInterrupt` at gateway.ts:1536 and :9128). Double-guarded (the `'error' in cleared` branch AND an outer try/catch around the import itself), so neither a tmux failure nor an import failure can block delivery — both log to stderr and fall through to `sendToAgent`. Soft-fail contract honored end-to-end.
- `tests/tmux-send-interrupt.test.ts:101` — `clearAgentComposer` suite asserts the exact `C-u C-a C-k` argv against socket `switchroom-klanker` and pane `klanker`, the 3s timeout, and both soft-fail modes (legacy_pty "no server running" and ETIMEDOUT). Matches the implementation precisely; legacy_pty agents are correctly untouched.

Rationale: tightly-scoped, additive, soft-fails closed-to-delivery so a clear failure can never wedge an inbound; the idle-but-not-empty assumption gap from #1556 is the real root cause and this closes it at the right seam. Two non-blocking watch items below.

Watch (non-blocking): (1) keystroke-vs-notification ordering is best-effort — `send-keys` returns when tmux accepts the keys into the pty input buffer, not when the claude TUI has rendered the cleared line; correctness relies on the TUI processing buffered input in-order ahead of the appended notification, which holds in practice but is a timing assumption. (2) `C-u C-a C-k` unconditionally wipes ANY composer content; safe in an autonomous container where no human types into the TUI, but if an interactive/legacy mode ever allowed human TUI typing, legitimate typed-ahead would be destroyed.

mekenthompson:
Independent review (READY-TO-MERGE): correct surface (clears the claude TUI composer via tmux send-keys — orthogonal to the Telegram draft/answer-stream, so no interaction), soft-fails to delivery (can never block an inbound), no-op on the happy path, clones the battle-tested sendAgentInterrupt helper, CI clean. Pulling into the v0.14.31 release. Pre-existing button-callback-path gap noted as a follow-up (narrow, not a blocker). Run jtbd-fast-trivial-dm + jtbd-always-on-after-restart-dm on the canary before fleet rollout (standard hygiene; the composer-clear itself isn't mtcute-observable).

Issue #1978, opened 2026-05-28
Opus 4.8 + thinking_effort > low breaks against framework message rewriting (thinking-block 400); no guardrail for effort configs


mekenthompson:
## Symptom
On Opus 4.8 with `thinking_effort: medium` (or any effort that produces adaptive-thinking blocks), turns fail with `API Error: 400 messages.N.content.M: 'thinking' or 'redacted_thinking' blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response.` Observed live on `clerk` 2026-05-28 ~21:38–22:00 UTC after bumping the fleet to `claude-opus-4-8` + `effort: medium`.

## Root cause
Opus 4.8's adaptive thinking emits `thinking`/`redacted_thinking` blocks. switchroom's framework-side rewriting of assistant messages (em-dash voice scrubber #1683, draft edits, etc.) invalidates those blocks' signatures; when the (modified) assistant message is replayed, the API rejects it. The claude CLI's own changelog documents a related concurrent-agent / interleaved-streaming merge bug; the agent image runs claude CLI `2.1.153`.

## Current state / mitigation
- Fleet pinned to `thinking_effort: low` (near-zero thinking → no blocks to invalidate). Documented in v0.14.7 CHANGELOG + operator config comment.
- The scaffold forces an effort value: `SWITCHROOM_DEFAULT_THINKING_EFFORT = "low"` (`src/agents/scaffold.ts`). There is **no config way to send zero `--effort`**, and omitting it is worse (4.8 defaults `effort=high`).

## What needs proper support (this issue)
1. **Can higher effort be supported safely?** Investigate whether the framework can preserve thinking-block integrity — e.g. don't rewrite/scrub assistant messages that contain thinking blocks, strip thinking blocks before the rewrite is replayed, or gate the em-dash scrubber off when thinking blocks are present. Determine whether bumping the in-image claude CLI version fixes the merge bug.
2. **If not safely supportable, fail loudly, not silently.** A future operator setting `model: claude-opus-4-X` + `thinking_effort: medium|high|xhigh|max` should get a clear validation/doctor warning at `apply` time (and ideally a startup guard), not a stream of 400s discovered in chat. Consider a model-aware effort guard in schema/scaffold/doctor.
3. **Strategic, not a shortcut** — the goal is real support for effort configs on adaptive-thinking models, with an explicit, documented fallback when a combo is known-incompatible.

## Related
[...]

PR #2040, opened 2026-06-01
chore: bump bundled claude CLI to 2.1.159 (v0.14.26)

mekenthompson:
## Summary

Bumps the hard-pinned `@anthropic-ai/claude-code` from **2.1.156 → 2.1.159** in both images that bundle it:
- `docker/Dockerfile.base:169` — every agent container
- `docker/Dockerfile.hindsight:43` — the memory provider

Plus version bump to 0.14.26 + CHANGELOG.

## Why

2.1.156 was pinned in v0.14.8 as the build that fixed the Opus 4.8 "thinking block cannot be modified" 400 error under low reasoning effort (#1978). The fleet had drifted 3 patch releases behind latest. Reviewed the 2.1.157–2.1.159 changelogs:
- **2.1.157**: no touch to the thinking/effort path; `.claude/skills` now auto-load without a marketplace; managed-settings MCP allow/deny parsing hardened. No breaking changes.
- **2.1.158**: Auto-mode on Bedrock/Vertex/Foundry (opt-in, irrelevant — we're subscription OAuth).
- **2.1.159**: internal only.

No revert of the #1978 fix in any of them.

## Test plan

- [ ] CI `build-base` + `build-dependents` succeed — this actually builds the images with 2.1.159 and runs `claude --version` in-layer, so a bad/uninstallable pin fails the required check.
- [ ] **Canary (the real gate):** deploy v0.14.26 to test-harness, run a live Opus-4.8 / low-effort turn, confirm the "thinking block cannot be modified" 400 does NOT return in the gateway log. Then fleet rollout.

## Risk

Medium — the pin exists specifically to dodge an Opus 4.8 regression, so the live-turn canary is mandatory before fleet. Both bundling images bumped together (no agent/hindsight skew). Easy revert: re-pin to 2.1.156.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,10 @@ provider). 2.1.156 was pinned in v0.14.8 as the build that fixed the
 Opus 4.8 "thinking block cannot be modified" 400 under low reasoning
 effort (#1978); the 2.1.157–2.1.159 changelogs contain no revert of that
 fix and no breaking changes affecting switchroom's MCP / hooks / session
-path, and the bump was canaried with a live turn confirming the 400 did
-not return. 2.1.157 also auto-loads `.claude/skills` without a
-marketplace and hardens managed-settings MCP allow/deny parsing.
+path. The bump is gated on a live-turn canary (test-harness, Opus 4.8 at
+low effort) confirming the 400 does not return before fleet rollout.
+2.1.157 also auto-loads `.claude/skills` without a marketplace and
+hardens managed-settings MCP allow/deny parsing.
 
 ## v0.14.25 — Surface foreground sub-agent activity after the ack
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## v0.14.26 — Bump bundled claude CLI to 2.1.159
+
+Moves the hard-pinned `@anthropic-ai/claude-code` from **2.1.156** to
+**2.1.159** in both images that bundle it (`docker/Dockerfile.base`, used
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- `docker/Dockerfile.base` (npm install line, hunk @ :166–169): pin moves `@anthropic-ai/claude-code@2.1.156` → `2.1.159`; in-layer `claude --version` retained, so an uninstallable pin fails the build. Correct.
- `docker/Dockerfile.hindsight:43`: same bump, keeping the memory provider on the identical bundle — no agent/hindsight version skew, which is exactly the skew the comment at :40 warns against. Correct.
- `package.json:4`: `0.14.25` → `0.14.26`, clean single increment off the last release. Correct.
- `CHANGELOG.md`: v0.14.26 entry accurately describes the bump and the #1978 rationale (2.1.156 was the build that fixed the Opus 4.8 'thinking block cannot be modified' 400 under low effort). Matches the diff.

Verified: 2.1.156 is the #1978-mitigating pin; the bump keeps both bundling images in lockstep and preserves the build-time `claude --version` smoke check. The pin exists specifically to dodge an Opus 4.8 regression, so the live-turn canary (Opus 4.8 / low effort, confirm the 400 does not return) is the real gate before fleet rollout — both test-plan boxes are still unchecked. CI `build-base`/`build-dependents` will catch a bad pin; the canary catches a silent behavior regression. Non-blocking, but don't fleet-roll on a green build alone.

Rationale: mechanical, lockstep CLI pin bump with an accurate changelog and an easy 2.1.156 revert; substance is correct, residual risk is runtime-behavioral and gated by the canary, not the diff.

Issue #94, opened 2026-04-25
Unify status card formatting across main agent and worker, suppress worker pin system message


mekenthompson:
## Problem

There are two live status surfaces surfaced in Telegram during active work:

1. **Main agent progress card** — rendered by `telegram-plugin/progress-card.ts` (`render()` function). Shows \`⚙️ Working…\`, elapsed time, a blockquote of the user's request, a tool checklist (with compaction / rollup), and a sub-agents section (\`[Sub-agents · N running]\`). Uses helpers like \`formatDuration\`, \`escapeHtml\`, \`truncate\`, all private to that module.

2. **Worker pinned card** — rendered by \`telegram-plugin/subagent-watcher.ts\` (\`renderWorkerCard()\`). Shows \`🛠 Background workers (N)\` as the header, then one line per worker: \`🔧 <description> · running · last activity Xs ago · N tools\`. Uses its own private copies of \`formatDuration\`, \`escapeHtml\`, and \`truncate\` with different formatting conventions (e.g. \`<1s\` / \`1m30s\` vs \`00:30\` / \`01:30\`).

The two cards are visually inconsistent:
- Different emoji conventions (🛠/🔧 vs ⚙️/✅)
- Different duration formats (\`<1s\`, \`1m30s\` vs \`00:30\`, \`01:30\`)
- Different label conventions
- No shared code — the same utility functions are duplicated

Additionally, when the worker card is pinned via `lockedBot.api.pinChatMessage(...)` in `gateway/gateway.ts` (around line 5106), Telegram emits a "Clerk pinned \<message\>" system message. This is the same noise already suppressed for the main progress card pin via `PinManager.captureServiceMessage()` in `progress-card-pin-manager.ts` — but the worker card pin path does not call `captureServiceMessage`, so the system message still appears.

## Goal

- A single shared formatting/template module that both the main progress card and the worker card import.
- Intentional, minimal visual differentiation to distinguish the two card types.
- The worker card pin system message is auto-deleted, matching the existing suppression for the main card.

### 1. Extract shared formatting utilities

Create `telegram-plugin/status-card-format.ts` (or `telegram-plugin/card-format.ts`) exporting:

```ts
export function formatDuration(ms: number): string   // unified format: <1s, 00:30, 01:30
export function escapeHtml(s: string): string
export function truncate(s: string, n: number): string
```

[...]

PR #2041, opened 2026-06-01
feat(telegram): native Telegram worker card (strip Markdown, ✓/→ feed)

mekenthompson:
## Summary

The background **worker-activity card** (the live message a `run_in_background` sub-agent posts and edits in place) HTML-escaped its text but never stripped the **Markdown the model authors** into its narration. Raw `**bold**`, `` `code` ``, and `---` rules rendered as literal characters in the card — the "half-done", non-Telegram-native look an operator flagged on the Clerk worker card.

This rewrites the card to the same idiom as the main agent's activity feed (`renderActivityFeed`), and strips Markdown at the boundary so nothing leaks.

**Running:**
```
🛠 Worker · Build shopping 3-way sync
running · 04:04 · 69 tools
✓ Vault denied — degraded gracefully
✓ Ran the full suite, committed, pushed
→ Opening PR…
```
**Finished:**
```
🛠 Worker · Build shopping 3-way sync
finished · completed · 71 tools · 04:12
✓ Vault denied — degraded gracefully
✓ Ran the full suite, committed, pushed
✓ Opened the PR
─────
✅ PR #21 opened, calendar-app. Notion as a true LWW peer, dedup + tombstones
```

## Why

JTBD: *you hold the leash* — background work must surface with the same legibility as a main turn. A card full of raw `**`/backticks/`---` reads as broken chrome, not "the agent is working". The fix matches the established activity-card visual so the two surfaces are consistent (principle 3).

## What changed

- **`card-format.ts`** (dependency-free): new `stripMarkdown` (inline code, bold/emphasis, links, leading headings/blockquotes/bullets/ordered-items, residual `**`) + `cleanWorkerResultParagraph` (drops fenced code + horizontal rules, collapses multi-line result to one paragraph). Order is always **clean → truncate → escape**.
- **`worker-activity-feed.ts`**: `renderWorkerActivity` rewritten to the `🛠 Worker` header + status line + `✓`/`→` step feed + finished `─────` rule + cleaned result. Every fragment stripped before escape. The manager now passes the accumulated narrative into the **terminal** render so the finished card keeps its step context. Tool/arg `⚡ <code>` chrome retired.
- **Tests**: `worker-activity-feed.test.ts` render+manager assertions moved to the new shape; new `card-format.test.ts` covers the strippers; markdown-strip / overflow / empty-result / finished cases added.
- **UAT**: `jtbd-worker-activity-feed-dm.test.ts` regexes moved to the new header + a **no-raw-Markdown regression guard** (asserts the rendered body has no `**`, backtick, or ASCII `---`). Shared `WORKER_FEED_RE` in `assertions.ts` (used by recall/reply scenarios to skip feed noise) re-keyed to the new header + finished status, tolerant of a Telegram variation selector.

No gateway callsite changed — the manager still sends via the existing `robustApiCall`-wrapped bot, so the bot-api-wrapping allowlist is untouched.

## Test plan

- [x] `vitest run telegram-plugin/tests/worker-activity-feed.test.ts` — 31 pass
- [x] `vitest run telegram-plugin/tests/card-format.test.ts` — 13 pass
- [x] `bun test telegram-plugin/uat/feed-matcher.test.ts` — 12 pass
- [x] `vitest run telegram-plugin/tests telegram-plugin/uat` — 3153 pass
- [x] `tsc --noEmit` clean
- [x] `check-plugin-references` / `check-bot-api-wrapping` / `check-bun-test-imports` clean
- [ ] Live UAT `jtbd-worker-activity-feed-dm` on test-harness (post-merge canary, mtcute can observe sendMessage/editMessageText)

## Risk

Low. Pure-render + string-helper change; no IPC/gateway/auth surface. Worst case is a cosmetic render regression caught by the unit + UAT guards. Behind the existing `SWITCHROOM_WORKER_ACTIVITY_FEED` flag (on by default).

### DIFF
--- a/telegram-plugin/card-format.ts
+++ b/telegram-plugin/card-format.ts
@@ -60,3 +60,65 @@ export function escapeHtml(s: string): string {
 export function truncate(s: string, n: number): string {
   return s.length > n ? s.slice(0, n - 1) + '…' : s;
 }
+
+/**
+ * Strip Markdown markup from a single line, leaving plain prose.
+ *
+ * Worker narration is authored as Markdown — the model writes `**bold**`,
+ * `` `code` ``, `- bullets`, `# headings`. The status cards render Telegram
+ * HTML, which does NOT interpret Markdown, so an unstripped `**` shows up as
+ * two literal asterisks (the #94-class "half-done" look). Strip the markup
+ * here so the card reads as clean prose.
+ *
+ * Run this BEFORE `truncate` + `escapeHtml`: clean → measure → escape. (The
+ * stripper never touches `<`/`>`/`&`, so escaping stays the last step.)
+ */
+export function stripMarkdown(s: string): string {
+  let out = s;
+  // Inline + leftover code spans → bare text.
+  out = out.replace(/`+/g, '');
+  // Links / images: [text](url) and ![alt](url) → the label.
+  out = out.replace(/!?\[([^\]]*)\]\([^)]*\)/g, '$1');
+  // Paired bold / emphasis runs (longest marker first).
+  out = out.replace(/\*\*(.+?)\*\*/g, '$1');
+  out = out.replace(/__(.+?)__/g, '$1');
+  out = out.replace(/\*(.+?)\*/g, '$1');
+  out = out.replace(/(?<![A-Za-z0-9])_(.+?)_(?![A-Za-z0-9])/g, '$1');
+  // Leading block markup: heading, blockquote, bullet, ordered item.
+  out = out.replace(/^\s{0,3}#{1,6}\s+/, '');
+  out = out.replace(/^\s{0,3}>\s?/, '');
+  out = out.replace(/^\s{0,3}[-*+]\s+/, '');
+  out = out.replace(/^\s{0,3}\d+[.)]\s+/, '');
+  // Residual unpaired bold markers (a lone `*` is left alone so `3 * 4`
+  // survives; only the doubled form is markup-by-construction).
+  out = out.replace(/\*\*/g, '');
+  return out.trim();
+}
+
+/** True for a whole-line horizontal rule: `---`, `___`, `***` (3+ of one). */
+function isRuleLine(s: string): boolean {
+  return /^\s*([-_*])\1{2,}\s*$/.test(s);
+}
+
+/**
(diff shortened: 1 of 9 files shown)


mekenthompson:
APPROVE

- card-format.ts:22-42 `stripMarkdown` — strip order is correct: code spans → links → paired bold/em → leading block markup → residual `**`. Underscore-emphasis lookbehind `(?<![A-Za-z0-9])_(.+?)_(?![A-Za-z0-9])` (line 32) correctly preserves `snake_case`; lone `*` left intact so `3 * 4` survives (line 38-40). No touch to `<>&`, so escape stays the last step as documented.
- card-format.ts:56-70 `cleanWorkerResultParagraph` — fenced-block + rule-line drop then space-join is right; `isRuleLine` (line 45) only matches whole-line `---/___/***` of 3+, won't eat real content.
- worker-activity-feed.ts:618,633-636 — every text fragment (desc, steps, result) runs through `stripMarkdown` before `truncate`+`escapeHtml`; truncate-then-escape preserves the measure-on-plaintext invariant. `appendStepFeed`:577-586 mirrors the main agent card (prior `✓`, newest `→`, `+N earlier…` overflow header) and matches the cap NARRATIVE_MAX_LINES=6.
- worker-activity-feed.ts:693 — terminal render now receives `{...view, narrativeLines: h.narrative}` so the finished card keeps its step context; latestSummary correctly repurposed as the final result paragraph (distinct from running steps).
- assertions.ts:434 WORKER_FEED_RE re-keyed to `🛠 Worker` header + `finished · completed/failed`, tolerant of the VS16 variation selector. Header present in every state, so feed detection (and recall/reply skip) stays correct; negative-match test retained.
- uat/jtbd-worker-activity-feed-dm.test.ts:524-530 adds the #94-class regression guard (no raw `**`, backtick, or ASCII `---`); the card's own divider is box-drawing `─────`, so the `-{3,}` assertion can't false-positive.

Nit (non-blocking): WORKER_FEED_RE second alternative `finished · completed/failed` could in theory match an ordinary reply that literally contains that phrase. Low risk given the `·` separators; header alternative is the real key.

Rationale: well-scoped boundary fix that makes background worker cards render as native Telegram chrome; full test + UAT coverage for every behavioral change, no gateway/bot-api callsite touched.

PR #2043, opened 2026-06-01
fix(secret-detect): cover Sanctum/Coolify tokens + redact history both directions

mekenthompson:
## Summary
A live Coolify / Laravel Sanctum API token (`<id>|<40-char base62>`, e.g. `17|aB3d…`) **pasted by a user** into chat was not redacted and persisted in plaintext. Token revoked; this closes the gap.

## Diagnosis — which of the 3 hypothesized causes were actually present
Verified by reading the inbound path (`gateway.ts handleInbound`, the `detectSecrets` gate, `recordInbound`/IPC dispatch) and the pattern engine:

| Hypothesis | Present? | Evidence |
|---|---|---|
| **Inbound vs outbound asymmetry** (redaction skips user input) | ❌ No | Inbound runs `detectSecrets` at ingest, fail-closed, **before** `recordInbound()` and the IPC broadcast; a high-confidence hit deletes the message + defers to vault and returns early. User input is **not** bypassed. (The real asymmetry is the *other* way — agent **output** had zero redaction; see fix #2.) |
| **Render-time vs ingest-time** (mask only at display) | ❌ No | Inbound detection is ingest-time, before persistence/forwarding. |
| **Pattern coverage** | ✅ **Sole cause** | No rule matched the `<id>\|<token>` shape — the Telegram rule needs a `:` separator; the rest are provider-prefixed (`sk-…`, `ghp_…`, JWT, …). `detectSecrets` returned 0 hits, so the message fell through to persist + forward. |

## Fix
1. **Add the `laravel_sanctum_token` anchored pattern** (`\b(\d+\|[A-Za-z0-9]{40,})\b`, high-confidence). The inbound gate and `redact()` share the same `detectSecrets` engine, so this one addition closes inbound interception **and** the redactor uniformly.
2. **Mask in both directions at ingest, before stored/logged/forwarded** (acceptance bullet 1). Outbound (agent→user) had *zero* redaction. Moved `redact()` into the plugin (`src/secret-detect/redact.ts` now re-exports — issues/hostd/CLI callers unchanged) and applied it at the single persistence chokepoint: `recordInbound` / `recordOutbound` / `recordEdit` in `history.ts`. No detected secret now lands in the message store in either direction — a backstop behind the inbound gate, and the only scrub on the agent-output direction.

## Why
The redaction system is the control that's supposed to prevent exactly this. The inbound gate was sound; it just lacked the pattern. Adding it (plus the both-directions persistence backstop) restores the guarantee.

## Test plan
- [x] `telegram-plugin/tests/secret-detect-sanctum.test.ts` (new) — `17|<40>` detected high-confidence as `laravel_sanctum_token`; **satisfies the exact inbound-gate predicate** (`confidence==='high' && !suppressed`); masked by `redact()`; false-positive guards (under-40 floor, 39-char, no-pipe, spaced markdown table cell); example-marker suppression.
- [x] `telegram-plugin/tests/history.test.ts` (extended) — `recordInbound`/`recordOutbound`/`recordEdit` mask a secret before storage (**both directions**); ordinary prose untouched.
- [x] `src/secret-detect/redact.test.ts` + `tests/issues.redact-detail.test.ts` green (re-export is behavior-preserving; issues pipeline + `bin/run-hook.sh` e2e pass against a fresh build).
- [x] Full `secret-detect` vitest cluster: 123 passed (no regressions). `history.test.ts` under bun: 33 passed.
- [x] `npm run lint` clean (tsc, plugin-references, bot-api-wrapping, bun-test-imports, no-pii-secrets, vault-hermeticity, no-broadcast-delivery). Token fixtures built by concatenation (no contiguous secret literal).

## Risk
Low. The pattern is anchored + length-floored (40) to avoid false positives; suppressor still demotes documented examples. The redactor move is a re-export (existing callers + tests unchanged). The history-store masking only alters text that contains a *detected* secret — ordinary messages pass through byte-identical.

## Out of scope (flagged — recommend separate tickets)
- **Outbound transport masking**: masking agent output in the live **send/stream** path (before bytes reach Telegram). Storage/logging is covered here; the streaming/progress-card hot path is load-bearing and warrants its own PR + UAT.
- **Upstream design flaw**: an agent (Clerk) asked the user to paste a raw secret because the vault entry was empty. The right fix is a secure **save-card → vault** flow that never echoes the value. Separate ticket. This PR limits the blast radius: such a paste is now caught + deleted + offered for vaulting even when the agent prompts for it.

JTBD: *you-hold-the-leash* / *subscription-honest* — secrets the operator handles over the channel must never persist in plaintext.

### DIFF
--- a/src/secret-detect/redact.ts
+++ b/src/secret-detect/redact.ts
@@ -2,97 +2,19 @@
  * `redact(text)` — sanitize text by replacing detected secrets and
  * credential-bearing URL parts with `[REDACTED]` markers.
  *
- * This is the chokepoint used by:
- *   - `switchroom issues record` — every code path that writes to
- *     `issues.jsonl` flows through `src/issues/store.ts:capDetail`,
- *     which calls `redact()` before truncating to `DETAIL_MAX_BYTES`.
- *     `bin/run-hook.sh` is the canonical caller; programmatic callers
- *     elsewhere in the codebase get the scrub for free.
- *   - `switchroom secret-detect redact --stdin` — bash-callable shim
- *     for ad-hoc operator use and tests. (Not on the hook hot path;
- *     each `bun` CLI fork costs ~785ms cold-start, so chaining
- *     `... | secret-detect redact | issues record` would double per-
- *     failure latency for no additional security beyond the server-
- *     side chokepoint above.)
+ * The implementation now lives in the telegram-plugin so the gateway's
+ * inbound/outbound history persistence (`telegram-plugin/history.ts`) and
+ * the src-side callers below share ONE redactor backed by the same
+ * `detectSecrets` engine — a pattern added once covers every path. This
+ * module re-exports it so existing importers keep working unchanged:
  *
- * Threat model: any token a failing hook prints on its error path (a 401
- * echoing the bearer, a `git clone` failure showing a PAT URL, a recall.py
- * traceback echoing the LLM provider response) used to land verbatim in
- * `issues.jsonl` and surface in Telegram via the issues-card / `/issues`
- * list. This module is the choke point that prevents that.
+ *   - `switchroom issues record` — `src/issues/store.ts:capDetail`, which
+ *     calls `redact()` before truncating to `DETAIL_MAX_BYTES`.
+ *     `bin/run-hook.sh` is the canonical caller.
+ *   - `switchroom secret-detect redact --stdin` — `src/cli/secret-detect.ts`.
+ *   - hostd — `src/host-control/server.ts`.
  *
- * Detection is delegated to the existing telegram-plugin engine
- * (`telegram-plugin/secret-detect/index.ts`) — same patterns, same
- * suppressor behavior, same secretlint integration as the outbound
- * Telegram redaction path. We do NOT re-invent regexes here.
- *
- * Idempotence: for token-shape detections the marker doesn't re-match
- * (no high-entropy run, no provider prefix). For *structural* detectors
- * — `cli_flag` (`--api-key <value>`), `json_secret_field` (`"key":
- * "value"`) — a second pass over already-redacted text will replace
- * `[REDACTED:openai_api_key]` with `[REDACTED:cli_flag]`. The bytes
- * remain redacted in either case, so this is a tag-rewrite, not a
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

- patterns.ts:63 new `laravel_sanctum_token` rule `/\b(\d+\|[A-Za-z0-9]{40,})\b/g` — correctly keys on the `<id>|<token>` pipe shape (distinct from Telegram's colon `id:token` and JWTs), 40-char floor matches Sanctum's `Str::random(40)`. FP guards are right: the spaced markdown cell `| 17 | … |` can't match because `\d+\|` requires the digit adjacent to the pipe, and the 39-char/under-floor cases are pinned (secret-detect-sanctum.test.ts:60-72).
- secret-detect/redact.ts:1-76 (new) — clean URL-scrub-then-token-detect order, right-to-left replacement keeps byte offsets valid; imports resolve (`detectSecrets` from ./index.js, `redactUrls` from ./url-redact.js:33, also re-exported by index.ts:197).
- src/secret-detect/redact.ts now a pure re-export shim — back-compat preserved for src/issues/store.ts:capDetail; no importer breaks.
- history.ts:318,332 / 371 / 401 — redact applied on inbound (text + reply_to_text), outbound, and edit before the row hits SQLite. Outbound/edit is the only secret scrub on the agent→user direction; inbound is defense-in-depth behind the fail-closed gate. `reply_to_text` null-guard at :318 is correct (only redacts when non-null).
- Tests: both directions + edit + FP guards + the high-confidence/not-suppressed inbound-gate predicate (secret-detect-sanctum.test.ts:36) all pinned; fixtures built by concatenation to dodge Push Protection.

Rationale: targeted, well-tested fix for the 2026-06-01 Coolify-token leak — adds the missing pattern AND closes the persistence gap in both directions, with proper false-positive guards. Imports and back-compat shim verified. Net security improvement, no regression surface.

mekenthompson:
Follow-up tickets opened for the two items flagged as out-of-scope:
- #2044 — mask agent output in the live send/stream **transport** path (this PR covers storage/logging in both directions).
- #2045 — secure secret-save flow so agents never ask users to paste raw secrets into chat (the upstream design problem behind this incident).

PR #2046, opened 2026-06-01
feat(secret-detect): mask agent-echoed secrets in outbound transport path

### LINKED ISSUE #2044: Secret redaction: mask agent output in the live send/stream transport path
Follow-up from #2043 (Sanctum/Coolify token leak fix).

## Context
#2043 closed the inbound leak (pattern coverage) and added **storage-layer** masking in both directions at the `history.ts` persistence chokepoint (`recordInbound`/`recordOutbound`/`recordEdit`). That guarantees no detected secret is **stored or logged** in the message store.

## Gap
The acceptance for #2043 asked for masking "before stored, logged, **or forwarded** — in both directions." The remaining piece is the **outbound transport** path: an agent that echoes a secret (e.g. quotes a token it read from a file, env, or a not-yet-vaulted value) still sends the **raw bytes to Telegram** before they're masked in storage. There is currently no `redact()` pass on the live `reply` / `stream_reply` send path.

## Why deferred
The outbound send/stream path is the load-bearing progress-card / streaming UX. Adding a redaction pass there safely (without breaking streaming, draft-mirror, chunking, HTML conversion, voice scrub) warrants its own PR + UAT rather than riding on the security hotfix.

## Proposed approach
- Apply the shared `redact()` (plugin-owned since #2043) to agent free-text at the outbound chokepoint(s) — `reply` and `stream_reply` — before the bytes are sent, mask-in-place (not delete; we can't drop the agent's turn).
- Decide streaming semantics: redact each draft chunk vs the finalized text (avoid masking a token split across chunk boundaries).
- UAT: confirm no regression to TTFO, progress card, draft streaming (mtcute can't see drafts — verify server-side + operator app).

## Acceptance
- An agent reply containing a secret-shaped string (e.g. `ghp_…`, `<id>|<40>`) is masked before it reaches Telegram, not just in history.
- No streaming/progress-card regression (UAT: `jtbd-fast-trivial-dm`, a streaming scenario).

Refs: #2043. See memory/architecture: inbound is ingest-time fail-closed; this is the agent→user transport half.

mekenthompson:
Closes #2044. Follow-up to #2043.

## Summary
#2043 closed the inbound leak (pattern coverage) and added **storage-layer** masking in both directions (`history.ts`). This adds the missing piece: masking agent-authored text on the **live outbound transport path**, before it reaches Telegram or the stderr preview log.

## Why
Outbound (agent→user) text had **zero** redaction. A trusted agent can still echo a secret it read from a file, env, or a not-yet-vaulted value into a reply — it would hit Telegram and the stderr preview *before* the #2043 storage mask ever saw it.

## What
`redactOutboundText()` (backed by the shared `redact()` / `detectSecrets` engine — same patterns as the inbound gate, so the Sanctum `<id>|<token>` shape from #2043 is already covered) at the entry of every agent-free-text tool:
- `reply` (`executeReply`)
- `stream_reply` (`executeStreamReply`)
- `edit_message` (`executeEditMessage`)

It **mutates the text in place** — the exact contract the existing voice scrub relies on — so every downstream consumer (stderr preview, #546 dedup key, chunker, answer-stream incremental diffing, history record) sees the masked value. `stream_reply` carries the full text-so-far on every call, and the mask is applied at entry on every call, so the answer-stream's incremental `editMessageText` diffing stays consistent (redacted-vs-redacted). A mask firing logs `outbound secret masked site=<tool>` (never the value).

## Test plan
- [x] `telegram-plugin/tests/gateway-outbound-redact.test.ts` (new, 6) — pins the import, the helper, and that the scrub runs at each handler's entry **before** the preview/scrub/send (structural — handlers aren't exported, same constraint as `gateway-secret-detect.test.ts`).
- [x] Masking behavior covered by `redact()` unit tests + `secret-detect-sanctum.test.ts`.
- [x] `tsc --noEmit` + full `npm run lint` clean (incl. `check-bot-api-wrapping` — no allowlist drift).
- [ ] **UAT pending (operator-gated):** mtcute streaming scenario to confirm no progress-card / draft-stream regression before fleet rollout. The in-place-mutation contract matches the proven voice-scrub pattern, so risk is low, but a streaming UAT pass is the rollout gate.

## Risk
Low-to-medium. Touches the streaming hot path, but only sanitizes the text string at entry using the same mutate-in-place pattern the voice scrub already uses (no control-flow change). `redact()` masks all detections (incl. suppressed/ambiguous) — for the agent-output direction that's the safe default; a documented token-format example in agent prose would be masked.

## Scope
Covers the three deliberate agent-output tools. Framework backstop/turn-flush emitters inherit storage masking via #2043's `recordOutbound`; their transport coverage is a smaller residual (they emit session-derived text, not arbitrary agent free-text).

JTBD: *you-hold-the-leash* / *subscription-honest* — secrets must never leave the box in plaintext, in either direction.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -8211,6 +8211,14 @@ function handleSessionEvent(ev: SessionEvent): void {
         const backstopThreadId = threadId
         const backstopCtrl = ctrl
 
+        // Outbound secret scrub (#2044). Turn-flush delivers the model's
+        // terminal answer prose when it skipped reply/stream_reply — that
+        // is arbitrary agent free-text, sent via sendMessage/editMessageText
+        // and previewed to stderr below, so it needs the same mask as the
+        // three reply tools. Mirror the voice scrub: mask before the send,
+        // the preview, and recordOutbound.
+        capturedText = redactOutboundText(capturedText, 'turn_flush')
+
         // Voice scrub (PR #1683 follow-up). Turn-flush is the path
         // that fires when the model emits raw transcript text WITHOUT
         // calling reply / stream_reply. That captured text bypasses
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -355,6 +355,7 @@ import { defaultVaultWrite, defaultVaultList, defaultVaultWritePosture } from '.
 import { parseVaultCliError, renderVaultCliError } from '../secret-detect/vault-error.js'
 import { recentDenialsFromAuditLog, type RecentDenial } from './recent-denials.js'
 import { detectSecrets } from '../secret-detect/index.js'
+import { redact } from '../secret-detect/redact.js'
 import { classifyAdminGate } from '../admin-commands/index.js'
 import {
   startSubagentWatcher,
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `redactOutboundText()` (gateway.ts:5189) is wired at the entry of all three agent-free-text tools — reply (gateway.ts:5214), stream_reply (gateway.ts:5972), edit_message (gateway.ts:7033) — before the stderr preview, dedup key, send, and history record. Verified the downstream anchors exist on main (reply preview at gateway.ts:5206, stream_reply/edit_message voice-scrub site markers at 5958/7006), so the at-entry ordering the structural tests assert holds.
- Backed by the shared `redact()` (redact.ts:46), the same engine as the inbound gate, so the Sanctum `<id>|<token>` pattern from #2043 is covered in both directions with no second pattern list to drift.
- In-place mutation matches the existing voice-scrub contract, so the answer-stream's incremental edit diffing stays redacted-vs-redacted — no control-flow change on the streaming hot path.
- Log line never emits the secret (gateway.ts:5193, `site=${site}` only); the test pins the absence of `${text}`/`${masked}`.

Nit: tests are structural (source string-match) only — they pin wiring, not masking behavior (which lives in redact()/sanctum tests). Fine given the handlers aren't exported, but they'll break on cosmetic refactors of the log/marker strings.

Watch: streaming partial-secret window — stream_reply re-masks the full text-so-far each call, so a secret only becomes detectable once fully present; a partial prefix in an earlier chunk could transit before the complete token is masked. A partial token isn't a full secret, and the PR's UAT gate covers the streaming path, so non-blocking. Turn-flush/backstop emitters (gateway.ts:8173) are explicitly out of scope and inherit #2043 storage masking only.

Rationale: closes the outbound transport leak left by #2043 (storage-only) using the same detector and the proven mutate-in-place pattern; tests, tsc, and lint clean per the PR, UAT operator-gated.

mekenthompson:
Updated after review (APPROVE):

Reviewer correctly flagged that the **turn-flush backstop** (gateway.ts ~8208) is NOT just session-derived text — when the model emits terminal answer prose *without* calling reply/stream_reply, turn-flush sends that arbitrary agent free-text to Telegram + a stderr preview with only `scrubVoice`. Same leak vector. Fixed in 1597e9c8: `redactOutboundText(capturedText, 'turn_flush')` at the flush branch entry, before send/preview/record, with a structural test. So the **Scope** paragraph in the PR body is now corrected — turn-flush IS covered; the remaining residual is only the PTY/tool-activity stream (genuinely tool-derived, not agent free-text).

**Rollout gate unchanged:** mtcute streaming UAT remains mandatory before fleet rollout (this merges to main on green; rollout is separate).

PR #2050, opened 2026-06-01
fix(telegram): worker card leaks raw ## markdown + multi-line steps

mekenthompson:
## Summary
Two presentation bugs in the v0.14.27 worker activity card (observed live on clerk): a raw `## Summary` heading leaked into the card, and a worker's final summary rendered as a multi-line blob inside a single step slot.

- `stripMarkdown` leading-block regexes (heading/blockquote/bullet/ordered) anchored `^` without the `m` flag, so they only stripped markup at the **start of the string**. A summary shaped `Done.\n\n## Summary\n…` kept its mid-string `## Summary`. → add `gm`.
- A narrative entry can itself be multi-line; the step renderer preserved the embedded newlines. → collapse internal whitespace to one space in the step + back-compat summary paths so each step is one visual line.

## Why
JTBD "you hold the leash" — the worker card is the user-visible window into delegated work. Raw markdown + a cut-off multi-line blob reads as broken. Real screenshot showed `→ Done.` followed by a raw `## Summary\n\nFixed the bug where a bad Keep app-password logged the user out of the entire ap…`.

## Test plan
- [x] `vitest run telegram-plugin/tests/card-format.test.ts telegram-plugin/tests/worker-activity-feed.test.ts` — 46 pass, incl. 2 new regression tests pinning the exact screenshot case.
- [x] `tsc --noEmit` clean.

## Risk
Low — presentation-only. No change to terminal/lifecycle/reaction behaviour. The slow (~6 min) card terminalization and the premature 👍-done reaction are tracked separately as a behavioral follow-up (turn-end ≠ job-completion); this PR does not touch those.

### DIFF
--- a/telegram-plugin/card-format.ts
+++ b/telegram-plugin/card-format.ts
@@ -85,10 +85,13 @@ export function stripMarkdown(s: string): string {
   out = out.replace(/\*(.+?)\*/g, '$1');
   out = out.replace(/(?<![A-Za-z0-9])_(.+?)_(?![A-Za-z0-9])/g, '$1');
   // Leading block markup: heading, blockquote, bullet, ordered item.
-  out = out.replace(/^\s{0,3}#{1,6}\s+/, '');
-  out = out.replace(/^\s{0,3}>\s?/, '');
-  out = out.replace(/^\s{0,3}[-*+]\s+/, '');
-  out = out.replace(/^\s{0,3}\d+[.)]\s+/, '');
+  // `gm` so the markers are stripped on EVERY line, not just the string
+  // start — a multi-line summary like "Done.\n\n## Summary\n…" must not
+  // leak a raw `## Summary` when rendered as a single card step.
+  out = out.replace(/^\s{0,3}#{1,6}\s+/gm, '');
+  out = out.replace(/^\s{0,3}>\s?/gm, '');
+  out = out.replace(/^\s{0,3}[-*+]\s+/gm, '');
+  out = out.replace(/^\s{0,3}\d+[.)]\s+/gm, '');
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- `stripMarkdown` (card-format.ts:88-91): adding `gm` to the four leading-block regexes (heading/blockquote/bullet/ordered) is the correct fix — without `m`, `^` only matched the string start, so a `Done.\n\n## Summary\n…` entry leaked its mid-string `## Summary`. The anchors still require the marker at a line start within ≤3 spaces, so inline cases like "C# is great" or "3 * 4" are untouched. Order is right: inline bold/italic strips run before the block strip, residual `**` cleanup after.
- `renderWorkerActivity` (worker-activity-feed.ts:150 and back-compat 178): `.replace(/\s+/g, ' ').trim()` collapses an internally-multi-line narrative entry to one visual line after the per-line strip, so a step slot stays single-line. Applied to both the narrativeLines path and the latestSummary back-compat path — no path left rendering raw newlines.
- Tests pin the exact screenshot regression: card-format.test.ts asserts `##` is gone while "Done."/"Summary"/"Fixed the bug" survive, and worker-activity-feed.test.ts asserts the step collapses to `Done. Summary Fixed the bug…` with no `\n\n` and no `## Summary`. HTML-escape and `---` rule tests remain green alongside.

Rationale: presentation-only fix, no change to terminal/lifecycle/reaction behavior; the regex and whitespace-collapse changes are correct and minimal, and the two new tests lock the precise live-observed bug. Safe to merge. (Noted: the slow card terminalization and premature 👍-done reaction are explicitly tracked as a separate behavioral follow-up — correctly out of scope here.)

PR #2051, opened 2026-06-01
fix(telegram): bind worker completion to authoritative signals, not turn-end

mekenthompson:
## Summary

Two coordinated fixes for the background-worker handback UX. The reported symptom (screenshot): the original request got a **premature 👍** while the worker was still running, and the 🛠 worker card hung at `running · 03:01 · 21 tools` for ~9 min before terminalizing with a cut-off `## Summary`.

The raw-markdown cutoff was already fixed in #2050. This PR fixes the **two timing bugs** behind it.

### Fix A — early, authoritative worker terminal
A background `Agent` worker's JSONL on claude ≥2.1.156 **never writes the `system/turn_duration` line**, so the watcher only learned the worker finished via the ~5-min silent-stall synthesis net — leaving the card stuck `running` for minutes after the work was done, and the deferred 👍 held that whole time.

The worker **does** write a final assistant message with `stop_reason: 'end_turn'` (a tool-using turn is `'tool_use'` and keeps going). `projectSubagentLine` now treats that as the terminal signal, emitted *after* the final text so the summary still renders. The stall-synthesis net stays as the backstop for workers that die without writing `end_turn`.

### Fix B — close the premature-👍 registration race
`countRunningWorkers()` read the in-memory watcher registry, which is populated by on-disk **file discovery a poll/fswatch tick AFTER dispatch**. With the ack-first pattern (reply "On it" → dispatch worker → end turn), the parent turn-end fires the deferred-done gate *before* the worker is registered → count 0 → 👍 promotes immediately.

It now reads the **dispatch-time DB** (`subagents` rows `status IN (running, stalled) AND background=1`) via a new `countRunningBackgroundSubagents` helper exposed on the watcher handle. The row is INSERTed when the `Agent` tool_use fires, *before* turn-end. The reaper sweeps genuinely-dead rows, and Fix A makes live rows reach terminal promptly, so the count can't wedge above zero. Falls back to the registry snapshot if no DB is wired.

**Net:** 👍 now means "job done" — held the instant a worker is dispatched, promoted when the worker truly finishes — and the card terminalizes on real completion.

## Why
Root cause confirmed from clerk's live `gateway.log` + `registry.db`: workers register in the file-discovery registry a tick after dispatch (premature 👍), and bg workers on 2.1.159 never emit `sub_agent_turn_end` (card lag). Diagnosis ruled out the 540k `EACCES statx` log lines as noise (TOCTOU on already-deleted ephemeral worker JSONLs — the worker that rendered the 21-tool card was read fine).

JTBD: **you hold the leash** — the reaction + card must reflect real job state, not a turn-lifecycle artifact.

## Test plan
- [x] `projectSubagentLine` emits `sub_agent_turn_end` after text on `stop_reason: 'end_turn'`; does NOT on `stop_reason: 'tool_use'` (session-tail.test.ts)
- [x] `countRunningBackgroundSubagents` counts at start, counts stalled, drops at terminal, ignores foreground (subagents.test.ts)
- [x] watcher + session-tail regression suites (61 tests) green — stall-synthesis backstop intact
- [x] `tsc --noEmit` + full `npm run lint` clean (bot-api-wrapping allowlist unaffected)
- [ ] Pre-rollout UAT on test-harness (worker-feed scenario) before fleet rollout

## Risk
Low-moderate. The `end_turn` terminal is guarded by the watcher's existing `state === 'running'` check (a later real `turn_duration` is a no-op). The DB count falls back to the old registry scan on any DB error. Stall-synthesis backstop unchanged.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -2119,6 +2119,13 @@ function finalizeStatusReaction(
  * fired its `done`/`failed` onFinish no longer counts here.
  */
 function countRunningWorkers(): number {
+  // Prefer the dispatch-time DB count: a background worker's row is INSERTed
+  // `status='running'` when its `Agent` tool_use fires, i.e. BEFORE the parent
+  // turn ends. The registry below is populated by on-disk file discovery, which
+  // lags dispatch by a poll/fswatch tick — so a just-dispatched worker was
+  // invisible to the deferred-done gate and the 👍 promoted prematurely.
+  const dbCount = subagentWatcher?.countRunningBackgroundWorkers?.()
+  if (dbCount != null) return dbCount
   const reg = subagentWatcher?.getRegistry()
   if (reg == null) return 0
   let n = 0
--- a/telegram-plugin/registry/subagents-schema.ts
+++ b/telegram-plugin/registry/subagents-schema.ts
@@ -371,14 +371,25 @@ export function getSubagentByJsonlId(db: SqliteDatabase, jsonlAgentId: string): 
  * snapshotting the file-discovery registry (which lags dispatch by a poll/fswatch
  * tick and so missed just-dispatched workers — the premature-👍 race).
  *
- * `stalled` counts as still-running: a stalled worker isn't done. Genuinely
- * dead rows are swept to a terminal status by the reaper (`reapStuckRunning`),
- * so this can't get wedged above zero forever.
+ * Counts `running` ONLY — `stalled` is deliberately excluded. `stalled` is NOT
+ * a terminal status: the reaper (`reapStuckRunningRows`) transitions a row to
+ * `stalled`, never to `completed`/`failed`. A genuinely-orphaned background row
+ * — one INSERTed at dispatch whose JSONL was never linked, so no activity ever
+ * bumped it and the in-memory silent-stall synthesis never terminalised it —
+ * sits in `stalled` indefinitely (the 1h reaper TTL is the only thing that
+ * moves it off `running`). Counting `stalled` would wedge the deferred 👍 above
+ * zero forever for that row (`reaction-defer.ts` `promote()` bails while the
+ * count is > 0). A live-but-quiet worker, by contrast, is driven to `completed`
+ * by the watcher's terminal paths (end_turn signal OR silent-stall synthesis,
+ * both call `recordSubagentEnd`) long before the 1h reaper, and a stalled row
+ * that genuinely resumes is flipped back to `running` by `recordSubagentResume`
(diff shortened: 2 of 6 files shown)


mekenthompson:
APPROVE

Verified both fixes against the PR branch; they are coherent and mutually reinforcing.

- **Fix B (premature 👍 race)** — `countRunningWorkers()` (gateway.ts:2121) now prefers `subagentWatcher.countRunningBackgroundWorkers()`, which reads `countRunningBackgroundSubagents` (subagents-schema.ts:380): `SELECT count(*) … WHERE background = 1 AND status IN ('running','stalled')`. Confirmed `recordSubagentStart` INSERTs the row with `status='running'` and the `background` flag at dispatch (subagents-schema.ts:342; `background` column is INTEGER NOT NULL, schema:191) — i.e. before parent turn-end, closing the file-discovery registration lag. Null-safe fallback to the registry scan is correct (subagent-watcher.ts:1510 returns null on no-DB / torn read).
- **Fix A (card hangs running)** — `projectSubagentLine` (session-tail.ts) emits `sub_agent_turn_end` when the final assistant message has `stop_reason === 'end_turn'`, after the content events so the summary still renders. Confirmed the consumer is idempotent and guarded: the handler only acts `if (entry.state === 'running')` (subagent-watcher.ts:779), then calls idempotent `recordSubagentEnd` → the DB row goes terminal → Fix B's count drops → deferred 👍 releases. A later real `turn_duration` line is a no-op. The `tool_use` negative test correctly prevents mistaking a mid-run tool turn for completion.
- Stall-synthesis backstop is untouched (subagent-watcher.ts:1290) — workers that die without writing `end_turn` still terminalize via the silent-stall net, so the count can't wedge above zero.
- Tests cover both fixes incl. negatives: end_turn-emits / tool_use-does-not (session-tail.test.ts), and counts-at-start / counts-stalled / drops-at-terminal / ignores-foreground (subagents.test.ts).

Nit (non-blocking): the DB path counts background-only while the registry fallback counts all non-historical running workers (gateway.ts:2138, no `background` filter). Benign — a foreground sub-agent blocks the parent so it can't still be running at deferred-done time — but the two paths have a subtle semantic asymmetry worth a comment.

Watch (non-blocking): treating any `end_turn` assistant message as terminal is exactly right for single-turn background `Agent` workers; if a resumable/multi-turn subagent ever emitted `end_turn` then continued, the `state==='running'` guard would flip it to done early. Not the background-worker model today, so no action needed — just the assumption to keep in mind.

Rationale: correct root-cause fixes (authoritative completion signal + dispatch-time count) replacing the turn-lifecycle proxy, fully guarded with fallbacks and an intact backstop, and well covered by positive+negative tests. Safe to merge; UAT on test-harness before fleet rollout per the plan.

PR #2054, opened 2026-06-01
feat(secret-detect): high-precision provider ruleset (GitHub-scanner style)

mekenthompson:
## Summary
Answers "we keep missing secrets — use something like GitHub's patterns." We were running only **22 hand-rolled** anchored patterns on the inbound gate. This adds **~31 curated, prefix-anchored provider patterns** (`PROVIDER_PATTERNS`) merged into `ALL_PATTERNS`, so they protect the **inbound gate, the outbound mask, and the issues pipeline** at once.

Providers: Stripe (live/restricted), SendGrid, GitLab, Hugging Face, Twilio, Mailgun, Mailchimp, DigitalOcean (pat/oauth/refresh), Doppler, Linear, Shopify (×3), Square (×2), New Relic, Notion, PlanetScale (×2), Supabase, Atlassian, Dropbox, Databricks, Grafana, PyPI, AWS-STS, GCP-OAuth, Slack webhook.

## Design choices
- **Prefixed rules only** (distinctive literal prefix + length) → near-zero false positives. This is load-bearing: the inbound gate **deletes** a message on a high-confidence hit, so an FP would nuke a real message. A generic *"any high-entropy string"* detector — the bare-token gap that let the Sanctum `<id>|<token>` slip — is deliberately **not** here; it belongs in a separate, *ambiguous*-routed change so it can't auto-delete on a guess. (That's the complementary follow-up.)
- **Baked as TS, not loaded from `gitleaks.toml` at runtime.** The bundler inlines `secret-detect` into `dist/server.js`/`gateway.js` and does **not** ship the `.toml` alongside (verified `build.mjs` doesn't copy it), so a runtime `loadGitleaksPatterns()` would silently resolve to nothing in the agent image — the orphaned-file trap. TS entries flow through `ALL_PATTERNS` naturally and bundle correctly.
- **No duplicates** of prefixes already in `ANCHORED_PATTERNS` (`sk-`, `ghp_`, `AKIA`, `AIza`, `xox*`, …) — e.g. `sk-proj-`/`sk-svcacct-` are already caught by the anchored `sk-` rule.

## Test plan
- [x] `secret-detect-providers.test.ts` (new) — 20 provider tokens detected high-confidence (fixtures built by concatenation, no contiguous literals), **+ a false-positive corpus** (UUID, git SHA, base64 blob, prose, file path) that triggers **zero** provider rules.
- [x] Existing `secret-detect-false-positives` (14) + core (33) + sanctum (11) suites all green — no regression.
- [x] `tsc --noEmit` + full `npm run lint` clean.

## Risk
Low. All rules are prefix-anchored (FP-resistant); validated against a benign corpus. No callback/UX surface touched. Worst case a future over-broad rule deletes a benign message — mitigated by the prefix discipline + the FP test corpus (extend it when adding rules).

## Follow-up (separate PR)
The generic **bare-high-entropy** fallback (catches novel/unprefixed tokens, routed as `ambiguous` → "looks like a secret, save it?" rather than auto-delete). That's the structural fix for the long tail; this PR is the curated-coverage half.

### DIFF
--- a/telegram-plugin/secret-detect/patterns.ts
+++ b/telegram-plugin/secret-detect/patterns.ts
@@ -118,6 +118,65 @@ export const STRUCTURED_PATTERNS: PatternDef[] = [
 ]
 
 /**
- * Concatenated registry — anchored first, then structured.
+ * High-precision PREFIXED provider patterns (gitleaks/secret-scanner style).
+ *
+ * Every rule here has a distinctive literal prefix + length, so false
+ * positives on ordinary chat/code are near-zero — which is load-bearing,
+ * because the inbound gate DELETES a message on a high-confidence hit.
+ * We deliberately keep ONLY prefix-anchored rules here; a generic
+ * "any high-entropy string" detector (the bare-token gap that let the
+ * Sanctum `<id>|<token>` slip) is intentionally NOT here — that's a
+ * separate, ambiguous-routed change so it can't auto-delete on a guess.
+ *
+ * Baked as TS (not loaded from the vendored gitleaks.toml at runtime):
+ * the bundler inlines secret-detect into dist/server.js + gateway.js and
+ * does NOT ship the .toml alongside, so a runtime `loadGitleaksPatterns()`
+ * would silently resolve to nothing in the agent image. TS entries flow
+ * through ALL_PATTERNS into the shared detectSecrets engine, so they
+ * protect the inbound gate, the outbound mask, AND the issues pipeline.
+ *
+ * Provider prefixes already in ANCHORED_PATTERNS (sk-ant-, sk-, ghp_,
+ * github_pat_, AKIA, AIza, xox*, gsk_, pplx-, npm_, telegram id:token, jwt)
+ * are intentionally omitted to avoid duplicate hits.
  */
-export const ALL_PATTERNS: PatternDef[] = [...ANCHORED_PATTERNS, ...STRUCTURED_PATTERNS]
+export const PROVIDER_PATTERNS: PatternDef[] = [
+  { rule_id: 'slack_webhook', regex: /(https:\/\/hooks\.slack\.com\/services\/[A-Za-z0-9_/]+)/g, captureIndex: 1, slugHint: 'slack_webhook' },
+  { rule_id: 'stripe_live_secret', regex: /\b(sk_live_[A-Za-z0-9]{24,})\b/g, captureIndex: 1, slugHint: 'stripe_key' },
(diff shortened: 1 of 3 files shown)


mekenthompson:
**APPROVE**

Clean, well-scoped addition. Read the full diff and the engine — verified the safety model holds.

- `patterns.ts:130-160` — 31 prefix-anchored provider rules merged into `ALL_PATTERNS` ahead of structured patterns. Every rule has a distinctive literal prefix + length floor, so FP risk on ordinary chat is near-zero. Correct call given the stakes (see below).
- `index.ts:117` confirms why precision matters: the engine hardcodes `confidence: 'high'` for every `ALL_PATTERNS` match, and a high-confidence inbound hit DELETES the message. The PR's "prefixed-only, no bare-entropy here" discipline is the right boundary — a generic high-entropy rule here would auto-nuke benign messages.
- No ReDoS surface: all new regexes are linear character classes, no nested quantifiers; engine rebuilds each `RegExp` per window (`index.ts:73`) so global `lastIndex` state can't leak across scans.
- TS-baked (not runtime `.toml`) is the correct choice — the bundler inlines `secret-detect` and doesn't ship the gitleaks toml, so a runtime loader would resolve to nothing in the agent image. Entries flow through `ALL_PATTERNS` and protect inbound gate + outbound mask + issues pipeline uniformly.
- No prefix collisions with `ANCHORED_PATTERNS` (sk-, ghp_, AKIA, AIza, xox*) — verified.

Nits (non-blocking):
- `patterns.ts:142` `mailgun_key` = `key-[0-9a-f]{32}` will match the common cache-key shape `key-<md5hash>`. Realistic benign collision; the prefix discipline accepts it, but worth keeping on the FP watchlist since a hit auto-deletes.
- Test coverage is partial: `secret-detect-providers.test.ts` CASES exercises 19 of the 31 rules. ~12 rules (digitalocean_oauth/refresh, doppler, mailchimp, mailgun, shopify_shared/private, square_oauth, planetscale ×2, grafana, pypi, dropbox) have no positive fixture, so a regex typo in those would ship silently. Consider one fixture per rule.

Rationale: high-value coverage expansion, FP-resistant by construction, no behavioral/UX surface touched, and the auto-delete risk is correctly contained by the prefix-only boundary. Safe to merge.

mekenthompson:
Addressed review (REQUEST_CHANGES → conditional APPROVE): dropped `mailchimp_key` — the one prefixless rule, which fired high-confidence on any md5+`-usN` (ETag/cache-key collision) and would auto-delete benign messages. Real Mailchimp keys defer to the planned generic high-entropy detector (asks, not deletes). Added bare-md5 + md5+`-usN` to the FP corpus to pin the boundary. All other rules are prefix-anchored + boundary-protected (reviewer verified no ReDoS, accurate formats, no dedup collisions, sound bundling). Tests + lint green.

PR #2056, opened 2026-06-01
fix(telegram): stop NO_REPLY sentinel leaking to chat on cron turns (#2053)

### LINKED ISSUE #2053: Cron turn leaks NO_REPLY text to chat: silent-end nag pushes sentinel through reply tool
## Summary

A cron-fired turn that legitimately has nothing to send can **leak `NO_REPLY` text to the Telegram chat** instead of ending silently. The `silent-end-interrupt-stop.mjs` Stop hook nags the model into routing junk through the `reply` tool, where the `NO_REPLY` sentinel is meaningless.

## Live reproduction

- Agent: `clerk`, session `e7106877-b105-4cc0-8eab-41efb628143a`, chat `8248703757`
- Cron fired ~`2026-06-01T05:00:31Z`
- Two messages leaked to chat:
  - `NO_REPLYThis was a cron check with no actionable result (Lisa's most recent email was 6 hours ago, outside the 2-hour window). Nothing to send the user.`
  - `NO_REPLYNO_REPLY`

## Chain of events (from transcript)

1. Cron check, nothing actionable. Model emitted **plain text ending in the sentinel** (prose first, `NO_REPLY` last):
   > `This was a cron check with no actionable result... Nothing to send the user.\n\nNO_REPLY`
2. `silent-end-interrupt-stop.mjs` only treats `NO_REPLY` as a valid silent-end when it is the **entire, exact** output. Prose wrapped around it → hook did not recognise the silent-end → fired the "your answer never reached the user, call the reply tool, or reply with exactly NO_REPLY" nag.
3. Under the nag, the model "complied" by sending text **through the `reply` MCP tool**, which delivers unconditionally. `NO_REPLY` carries no meaning as a `reply()` argument → leaked message 1.
4. Hook nagged again; model called `reply` trying to "reply with exactly NO_REPLY" → leaked message 2 (`NO_REPLYNO_REPLY`).
5. `[silent-end-interrupt] retry exhausted (retryCount=2 >= MAX_RETRIES=2) — allowing stop`. Turn ended.

## Root cause

Two compounding issues:

- **Behavioural:** the model didn't emit a clean bare `NO_REPLY`; it editorialised first, defeating the exact-match suppressor.
- **Harness:** once the nag loop starts, the model escapes it by pushing text through `reply`, and **nothing guards a `reply` payload that is only `NO_REPLY`**. The nag also should not fire on a `source="cron"` turn that legitimately has nothing to say.

This is shared-plugin code (`/opt/switchroom/telegram-plugin/hooks/silent-end-interrupt-stop.mjs`), so it can affect any agent — observed on `clerk`, and the same failure mode has hit `klanker`.

## Proposed fix (both cheap)

1. **Reply guard (PreToolUse on `reply`/`stream_reply`):** if the trimmed payload equals `NO_REPLY` (or is only repeats of it), drop it — do not send.
2. **`silent-end-interrupt-stop.mjs`:** accept any output whose trimmed text *ends with* `NO_REPLY` as a valid silent end, and skip the nag entirely when the turn's `source="cron"`.

## Impact

Cron noise leaking into user chats; erodes trust in scheduled tasks. Low severity, high annoyance, trivially reproducible.

mekenthompson:
Fixes #2053.

## Problem
A cron turn that ends with prose followed by a bare `NO_REPLY` line defeated the anchored exact-match silent-end suppressor. The Stop hook then nagged the model, which escaped the nag by pushing the sentinel through the unconditional `reply` tool — leaking `NO_REPLY...` text to the Telegram chat. Observed on `clerk` and `klanker`.

## Fix (three layers, defense-in-depth)
1. **Stop-hook scan** (`telegram-plugin/hooks/silent-end-scan.mjs`) — new `endsWithSilentMarker()` recognizes plain assistant **text** and reply-tool payloads whose final non-empty line is a bare `NO_REPLY`/`HEARTBEAT_OK` as a valid silent end. Also parses the channel envelope's `source=` and returns `allow` (reason `cron-source`) for `source="cron"` turns that produced no qualifying reply, so the nag never fires.
2. **Gateway flush gate** (`telegram-plugin/turn-flush-safety.ts`) — new `endsWithSilentMarker()`; `decideTurnFlush` now `skip`s a prose+trailing-sentinel blob instead of flushing the prose with the sentinel glued on.
3. **New PreToolUse reply guard** (`telegram-plugin/hooks/sentinel-reply-guard-pretool.mjs`) — drops a `reply`/`stream_reply` call whose trimmed payload is only sentinel line(s). Exact per-line match (never substring), fail-open. Wired into `hooks/hooks.json` and `src/agents/scaffold.ts`.

## Relationship to #1955
Extends, does not regress, #1955. `isCompositeSilentNoise` is left byte-identical; a standalone `"Sent."` (no marker) is still NOT suppressed. #1955 deliberately left "prose carrying a real line + sentinel" as a follow-up — this is that follow-up.

## Behavior note (intentional reversal)
The flush safety net previously *flushed* `prose\nNO_REPLY` on the theory the model wrote a real answer and forgot to send it. This PR *suppresses* it instead — matching the model's explicit silence intent. Stripping-and-flushing would re-introduce the surprise-message problem the safety net exists to prevent.

## Tests
84/84 green across the three suites (`turn-flush-safety`, `silent-end-interrupt-stop-scan`, `sentinel-reply-guard-pretool`). Covers: prose+trailing-sentinel recognition, cron-source nag skip, reply-guard drop, substring-not-dropped, and the documented `.mjs`↔TS trailing-punctuation divergence. `tsc --noEmit` clean.

## Review
Independently reviewed (separate reviewer agent): SHIP verdict, no blocking defects. Confirmed the cron `source=` regex is anchored to the channel tag and not spoofable from body content, and that no user-awaited reply path is wrongly silenced.

## Follow-up (not in this PR)
No drift test currently pins the `hooks.json` ↔ `scaffold.ts` hook-list parity for the new guard — worth a one-line assertion in a later change.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -3739,6 +3739,23 @@ export function buildSettingsHooksBlock(p: HooksBlockParams): Record<string, unk
           ],
         },
         {
+          // #2053 defense-in-depth: drop a reply/stream_reply call whose
+          // payload is only the silent sentinel (NO_REPLY/HEARTBEAT_OK),
+          // so the sentinel can never reach chat regardless of any
+          // nag-loop behaviour. No matcher — the hook self-filters by
+          // tool_name (like secret-guard above).
+          hooks: [
+            {
+              type: "command",
+              command: wrap(
+                "hook:sentinel-reply-guard-pretool",
+                `node "${join(DOCKER_HOOKS_PATH, "sentinel-reply-guard-pretool.mjs")}"`,
+              ),
+              timeout: 5,
+            },
+          ],
+        },
+        {
           // Claude Code's hook matcher is a regex. Cover both the legacy
           // 'Agent' and newer 'Task' tool names — same dispatch
           // semantics, only the name varies by Claude Code version. The
--- a/telegram-plugin/hooks/hooks.json
+++ b/telegram-plugin/hooks/hooks.json
@@ -11,6 +11,15 @@
         ]
       },
       {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/sentinel-reply-guard-pretool.mjs\"",
+            "timeout": 5
+          }
+        ]
+      },
+      {
         "matcher": "^(Agent|Task)$",
         "hooks": [
           {
--- a/telegram-plugin/hooks/sentinel-reply-guard-pretool.mjs
+++ b/telegram-plugin/hooks/sentinel-reply-guard-pretool.mjs
@@ -1,0 +1,114 @@
+#!/usr/bin/env node
+/**
+ * PreToolUse hook — drops a `reply` / `stream_reply` call whose entire
+ * payload is only the silent sentinel(s) NO_REPLY / HEARTBEAT_OK.
+ *
+ * Defense-in-depth for #2053. The silent-end Stop hook and the gateway
+ * flush gate already recognise prose+trailing-NO_REPLY as "intentionally
+ * silent", but if a nag-loop (or any other path) ever pushes a
+ * sentinel-only payload through the reply tool, it must NEVER reach the
+ * Telegram chat. This guard is the last line: it intercepts the tool
+ * call itself, before the gateway sees it.
+ *
+ * Match discipline — EXACT, not substring:
+ *   - The trimmed payload must be ONLY one or more silent markers
+ *     (each on its own line, optional trailing punctuation per marker).
+ *   - A real reply that happens to mention "NO_REPLY" inside genuine
+ *     prose (e.g. "reply with exactly NO_REPLY if nothing to add") is
+ *     NOT dropped — it has non-marker content, so it is delivered.
+ *
+ * Claude Code PreToolUse protocol (v1):
+ *   Input:  JSON on stdin — { session_id, tool_name, tool_input, ... }
+ *   Output: exit 0 + empty stdout → allow.
+ *           exit 0 + JSON stdout { decision: "block", reason } → block.
+ *
+ * Fail-open on any parse/IO error — a malfunctioning guard must not wedge
+ * the reply path.
+ */
+
+import { readFileSync } from 'node:fs'
+import { argv } from 'node:process'
+import { fileURLToPath } from 'node:url'
+
+const REPLY_TOOLS = new Set([
+  'mcp__switchroom-telegram__reply',
+  'mcp__switchroom-telegram__stream_reply',
+])
+
+// Mirrors turn-flush-safety.ts:isSilentFlushMarker and
+// silent-end-scan.mjs:SILENT_MARKER_RE — a single bare marker with
+// optional trailing punctuation.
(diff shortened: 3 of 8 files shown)


mekenthompson:
**APPROVE**

Read the full diff. This closes both root causes I flagged triaging #2053 (anchored regex missing prose+trailing-marker; no sentinel-only reply guard) plus a cron-source skip. Three layers, all tested.

- `silent-end-scan.mjs:67-74` `endsWithSilentMarker` + `:202-211` plain-text carve-out — a turn whose final non-empty line is a bare NO_REPLY/HEARTBEAT_OK (emitted as plain transcript text, not via reply tool) now returns `allow`/`silent-marker-text` instead of falling through to `block`. This is the exact leak shape from the issue. Correct.
- `silent-end-scan.mjs:226-235` cron-source skip — `envelope.source === 'cron'` → `allow`/`cron-source`. Not spoofable in the dangerous direction: `parseChannelEnvelope` regex `/<channel[^>]*\bsource="([^"]+)"/` matches the FIRST channel tag (the real envelope), so body content after it can't override it to forge `cron`. A real cron turn that needs to speak still calls reply and is caught above.
- `turn-flush-safety.ts:128-134` + `decideTurnFlush:201-205` — the gateway flush gate now `skip`s a `prose\nNO_REPLY` blob rather than flushing prose with the sentinel glued on. This is an intentional reversal of prior behavior (documented in the PR); the right call — stripping-and-flushing would re-introduce the surprise-message problem and override the model's explicit silence.
- `sentinel-reply-guard-pretool.mjs` — PreToolUse last line of defense: drops a reply/stream_reply whose trimmed payload is ONLY sentinel line(s). Exact per-line match via `isSentinelOnly` (never substring — `:46-54` prose containing NO_REPLY is allowed), fail-open on parse/IO error. The direct-invocation guard `:111-113` (`fileURLToPath(import.meta.url) === argv[1]`) correctly prevents the test import from blocking on stdin.
- Wiring matches: `hooks.json:13-22` and `scaffold.ts:3741-3757` both add the matcher-less, self-filtering hook (same posture as secret-guard). 84 tests across the three suites, including substring-not-dropped and the documented .mjs↔TS divergence.

Nits (non-blocking):
- No drift test pins `hooks.json` ↔ `scaffold.ts` parity for the new guard — author already flagged this as follow-up. Worth the one-line assertion.
- The .mjs `endsWithSilentMarker` (unbounded trailing punctuation) intentionally diverges from the TS `isSilentFlushMarker` (length-capped, single punct). Documented and benign-in-direction (only ever suppresses more), but it's a second copy of the marker logic to keep in sync.

Rationale: targeted, defense-in-depth fix that resolves the leak at every layer (Stop-hook scan, gateway flush, reply tool), with no user-awaited reply path wrongly silenced. Safe to merge.

mekenthompson:
Independent review (READY-TO-MERGE): three guarded defense-in-depth layers (Stop-hook scan with a correctly-last cron carve-out, gateway flush-gate skip-branch, PreToolUse reply guard with exact-line match + fail-open). Served/user-awaited turns are protected by early-returns; cron carve-out scoped to source="cron"; no regression to #1955. Tests cover every behavior; CI clean. Pulling into the v0.14.31 release.

PR #2057, opened 2026-06-01
feat(telegram): pending-progress suffix signals agent is still reachable

mekenthompson:
## Summary

Make a delegating agent *feel* reachable while a background worker runs.
The ambient cross-turn ticker becomes:

> — still working (3m) · message me anytime, I'll keep you posted

Pure render/copy change to the `pending-work-progress` suffix — no new
dispatch path, no model call, no typing churn. Edited in place every 60s
exactly as today (jsonl-tail → render only), so it stays inside the
subscription boundary.

## Why

After an agent delegates to a background worker and ends its turn, the
silent "— still working (Nm)" ticker reads as *"agent is unresponsive"*.
The user waits instead of messaging. But the agent is reachable the whole
time: a background-worker dispatch never sets `turnInFlight`, so an inbound
user message is delivered as a fresh turn (not buffered, not raced against
a handback). The gap was signaling, not capability — so the fix is copy.

JTBD: **You hold the leash** — awareness + control; the user should know
they can reach in at any moment.

## What changed

- `pending-work-progress.ts`: suffix string + doc comment.
- `SUFFIX_RE` made backward-compatible (reachability clause optional) so
  an anchor carrying a *pre-upgrade* suffix is still stripped on the next
  edit during a rolling fleet upgrade. Suffix stays plain-text-safe under
  HTML parse_mode (no `<`, `>`, `&`).
- Unit tests updated + new "strips a prior NEW-shape suffix" case.
- Two UAT scenario `SUFFIX_RE`s updated to the new shape.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `bun test telegram-plugin/tests/` — 3813 pass, 0 fail
- [x] `pending-work-progress.test.ts` — 21 pass (cadence, strip old+new
      suffix, HTML parse_mode preserved)
- [ ] Operator eyeball on official Telegram app after rollout

## Risk

Low. Copy-only; no control-flow change. Worst case is suffix wording.
Backward-compatible regex protects the rolling-upgrade window.

### DIFF
--- a/telegram-plugin/pending-work-progress.ts
+++ b/telegram-plugin/pending-work-progress.ts
@@ -35,7 +35,11 @@
  *   turn_end with pending+anchor → activate the timer for the key
  *   tick (every 5s, edit every  → editMessageText against the anchor
  *     EDIT_INTERVAL_MS)            appending/refreshing the suffix
- *                                  " — still working (Nm)"
+ *                                  " — still working (Nm) · message me
+ *                                  anytime, I'll keep you posted"
+ *                                  (the reachability clause signals the
+ *                                  agent is still listening while a
+ *                                  background worker runs — issue PR3)
  *   inbound user message        → clear (user re-engaged or moved on)
  *   subagent_handback inject    → clear (model about to re-engage)
  *   MAX_LIFETIME_MS budget cap  → clear (give up; 30 min default)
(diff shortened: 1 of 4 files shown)


mekenthompson:
**APPROVE**

Read the full diff. Copy-only change to the cross-turn progress ticker suffix — no control-flow, dispatch, or model-call change.

- `pending-work-progress.ts:390` — suffix now appends ` · message me anytime, I'll keep you posted`. Plain-text-safe under HTML parse_mode (no `<`, `>`, `&`; the `'` and `·` need no escaping). Correct.
- `pending-work-progress.ts:81-82` — `SUFFIX_RE` makes the reachability clause optional, so an anchor carrying a pre-upgrade suffix is still stripped on the next edit during a rolling fleet upgrade. Right call; prevents double-suffix accumulation across the upgrade window. Still end-anchored so it won't match model-written text.
- Length-cap path (`:392` `newText.length > TELEGRAM_MSG_CAP`) is unchanged and still handles the now-~45-char-longer suffix — no overflow risk.
- Tests updated consistently incl. new 'strips a prior NEW-shape suffix' case (`tests/pending-work-progress.test.ts`), and both UAT `SUFFIX_RE` copies updated to match.

Nit (non-blocking): `SUFFIX_RE` is duplicated across three files (main + two UAT scenarios). All three are updated here, but the literal will drift eventually — worth hoisting to one shared constant.

Rationale: pure render/copy change, backward-compatible regex, no behavioral surface touched, tests green. Safe to merge.

PR #2059, opened 2026-06-01
feat(secret-detect): generic bare-high-entropy fallback (long-tail detector)

mekenthompson:
## Summary
The structural fix behind the Sanctum incident. The provider/anchored patterns only catch tokens with a known **prefix** or a `KEY=value` context — a **standalone** high-entropy token (raw base62/base64url, no prefix) matched nothing and slipped through. `scanGenericSecrets` closes that, merged into the shared `detectSecrets` so it backs the inbound gate + outbound mask + issues pipeline.

Companion to #2054 (curated provider ruleset). Provider patterns = precision for known shapes; this = recall for the unknown tail.

## Safety: `ambiguous`, never auto-delete
Emitted at **`ambiguous`** confidence. The inbound gate auto-*deletes* on a `high` hit; a generic guess isn't precise enough for that, so it routes to the existing **"👀 looks like a high-entropy string — stash to vault or ignore?"** ask-path. `redact()` still masks it in history as defense-in-depth.

## Precision + perf: a distinct-character gate (not Shannon)
Requires **≥18 distinct chars** over a 28+ `[A-Za-z0-9_-]` run. Fast (128-slot bitset, early-exit; no Map/log2 per candidate) **and** the floor sits just above the alphabets of the dominant benign shapes, so they're excluded *by construction*:
- hex hashes / SHAs / git SHAs / md5 → ≤16 distinct
- UUIDs → ≤17
- digit runs → ≤10
- real base62 / base64url tokens → 24–62 distinct → **flagged**

Charset excludes `/ + = . : |` so file paths, standard-base64 blobs, version strings, hostnames, and the Sanctum `<id>|<token>` shape (its own rule) aren't swept in. A **hit cap (20/scan)** bounds the O(n²) overlap-dedup on pathological input.

Also: `detectSecretsAsync`'s merge now prefers `high` over `ambiguous` on an exact-range tie, so this fallback can't shadow/downgrade a Secretlint (or other) `high` provider hit on the same span.

## Test plan
- [x] `secret-detect-generic-entropy.test.ts` — flags a bare high-entropy token as `ambiguous`; **does NOT downgrade a recognized high token** (`ghp_…` stays `high`); distinct + length floors; hit cap; FP corpus (UUID, git SHA, sha256, md5, digit run, prose, file path) all clean.
- [x] Full `secret-detect` suite (137) green incl. the ReDoS-perf test and the Secretlint merge test.
- [x] `tsc --noEmit` + `npm run lint` clean.

## Risk
Low-moderate. Routed as `ambiguous` (asks, never deletes), so a false positive is a dismissable prompt — and storage masking is the only other effect. The distinct gate + charset keep the ask-rate low (benign hashes/UUIDs/paths excluded by construction). Worst case: a benign base64url blob pasted in a cached-passphrase chat triggers one "stash or ignore?" prompt.

### DIFF
--- a/telegram-plugin/secret-detect/generic-entropy.ts
+++ b/telegram-plugin/secret-detect/generic-entropy.ts
@@ -3,45 +3,45 @@
  *
  * The provider/anchored patterns only catch tokens with a known prefix
  * (sk-, ghp_, shpat_, …) or a KEY=value context. A STANDALONE high-entropy
- * token pasted in prose — a raw Sanctum/base62 token, a bare base64url API
- * key with no prefix — matches none of them and used to slip through
- * entirely (the 2026-06-01 Sanctum incident). This scanner closes that gap.
+ * token pasted in prose — a raw Sanctum/base62 token with no prefix —
+ * matches none of them and used to slip through (the 2026-06-01 Sanctum
+ * incident). This scanner closes that gap.
  *
- * Emitted at **`ambiguous`** confidence, NOT `high`:
- *   - The inbound gate auto-DELETES on a high-confidence hit; a generic
- *     guess is not precise enough to justify that, so it routes to the
- *     "👀 looks like a high-entropy string — stash to vault or ignore?"
- *     ask-path instead.
- *   - `redact()` masks ambiguous hits too, so history storage is still
- *     protected as defense-in-depth.
+ * Emitted at **`ambiguous`** confidence, and `redact()` deliberately
+ * EXCLUDES this rule (see redact.ts): a generic guess must never silently
+ * mask — it would corrupt agent replies and stored messages (dense
+ * identifiers look high-entropy too). Its sole job is to drive the inbound
+ * gate's "👀 looks like a high-entropy string — stash to vault or ignore?"
+ * ASK prompt, where the operator confirms.
  *
- * Precision/perf via a DISTINCT-CHARACTER gate rather than Shannon entropy:
- * requiring ≥ DISTINCT distinct chars over a long run is a fast O(n) check
- * (a 128-slot bitset with early-exit, no Map / no log2) and — crucially —
- * the floor is set just above the alphabets of the dominant benign shapes,
- * so they're excluded BY CONSTRUCTION:
- *   - hex hashes (md5/sha1/sha256, git SHAs, hex ids): ≤ 16 distinct
- *   - UUIDs (hex + `-`): ≤ 17 distinct
- *   - digit runs (phone/ids): ≤ 10 distinct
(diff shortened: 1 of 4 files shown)


mekenthompson:
**APPROVE**

Read the full diff and verified the overlap/merge claims against the engine. This is the deferred ambiguous-routed half of #2054, and the safety model holds.

- `generic-entropy.ts:60` emits `confidence: 'ambiguous'`, never `high` — the linchpin. The inbound gate auto-deletes only on `high`, so a generic guess routes to the ask-path instead of nuking a message. Correct and the right call for a recall-oriented detector.
- `generic-entropy.ts:31` distinct-char gate (≥18 over a 28+ `[A-Za-z0-9_-]` run) excludes the dominant benign shapes by construction: hex/SHA/md5 ≤16, UUID ≤17 (+ explicit `UUID_RE` guard at `:71`), digit runs ≤10. `hasDistinctChars` is O(n) early-exit over a `Uint8Array(128)`; charset is ASCII-only so no out-of-bounds. Charset excludes `/ + = . : |`, keeping file paths, std-base64 blobs, version strings, and the Sanctum `<id>|<token>` shape out.
- `generic-entropy.ts:38` `MAX_GENERIC_HITS=20` bounds the downstream O(n²) overlap-dedup on pathological input. Linear scan, no ReDoS.
- Overlap handling verified in `index.ts`: `dedupeRaw:166` prefers `high` over `ambiguous` on an exact-range tie, and `dropOverlaps:182-189` drops a generic hit fully contained in a larger `high` hit (so a JWT's base64url segments don't double-flag under the anchored jwt rule). A recognized token is never downgraded — `ghp_` test confirms.
- `index.ts:229-244` async-merge fix is correct and important: `consider()` now prefers `high` over `ambiguous` on an exact-range tie instead of vendored-first, so this fallback can't shadow/downgrade a Secretlint `high` on the same span. Mirrors the sync `dedupeRaw` rule.

Nit (non-blocking): the FP corpus in `secret-detect-generic-entropy.test.ts` omits a bare base64url/unpadded-base64 blob — exactly the shape most likely to trip the ambiguous ask (high distinct-count, no `/+=`). Not a bug (ambiguous = a dismissable prompt, per the PR's stated worst case), but pin a fixture asserting the EXPECTED behavior for that shape so a future charset/threshold change can't silently move the ask-rate.

Watch: the auto-delete safety hinges on the gate continuing to treat `ambiguous` as ask-only (unchanged here, existing contract) — keep that invariant if the routing is ever refactored.

Rationale: recall-closing fallback that is FP-safe by routing (ambiguous, never delete) and by construction (distinct-char floor), with bounded perf and a correct merge tie-break. Safe to merge.

Issue #1122, opened 2026-05-12
Rethink turn UX: collapse to one artifact per turn


mekenthompson:
## Problem

Switchroom currently emits **three competing artifacts per turn**:

1. An "intent" reply ("Worker dispatched…")
2. A floating **progress card** (status + tool steps)
3. The final answer

They interleave. Telegram orders strictly by send-time and edits can't reorder, so the card sometimes lands *below* the final reply, the fleet sub-line stops refreshing once the parent turn yields, and users have to mentally merge three messages to understand what happened.

The narrow fix in PR #1119 suppresses the card when the reply has already landed, but it patches one symptom of a deeper design problem.

## JTBD

> *"Tell me my agent is alive, what it's doing, and when it's done — no more, no less."*

## Research findings

Surveyed how reference agent UIs solve this:

| UI | Pattern |
|---|---|
| **Claude Code** | One transcript, tool calls collapsed inline, Ctrl-O cycles density. Progressive disclosure. |
| **OpenClaw** | Explicit rule: only the final answer goes to Telegram; progress stays in Control UI. They had our exact bug (their #71320 — tool-call progress leaking to Telegram). |
| **ChatGPT o1 / GPT-5 thinking** | Single collapsible "Thinking…" pill above the answer. Same shape regardless of duration. |
| **Slack long-running bots** | Edit-in-place for progress; **new** message only at completion (so push notification fires). |
| **Cursor v3** | Background agents get their own tab. |
| **Hermes (NousResearch)** | Single chronological stream; subagents as nested collapsible cards. |

**Recurring principles across the winners:**

1. **One canonical artifact per turn.** Multiple parallel artifacts is the anti-pattern.
2. **Progress lives inside the answer's container, not beside it.**
3. **Edit-in-place for liveness; new message only for completion.** Maps perfectly to Telegram (edits silent, new messages ping).
4. **Progressive disclosure beats hiding.** Default to summary; let the curious expand.
5. **Separate background/subagent work from the foreground turn.**

## Recommendation

**Collapse to one edited message during the turn + one new message at completion.** Two artifacts total, but the split is meaningful (process vs. result), not accidental.

Concrete shape:

- **In-flight message**, edited in place as the turn proceeds:
  - `🔵 Thinking…`
  - `🔵 Running bash (3/5)`
  - On completion → `✅ Done · 4 steps · 12s` with a **native Telegram expandable blockquote** holding the step trace (collapsed by default).
[...]

PR #2061, opened 2026-06-01
fix(telegram): route background worker card + handback to the dispatch chat, not the DM

mekenthompson:
## Summary

A background sub-agent's live **worker card** (`🛠 Worker · <task> running · NN:NN · N tools`) and its **completion handback** were always posted to the operator **DM**, even when the user dispatched the Task from a **group / forum topic**. Observed live: Marko was asked to do deep research *in a group*; it replied in the group correctly, but every worker-progress beat (and the eventual result) went to Marko's DM.

## Root cause

The worker-card chat came from the pinned-card "fleet" lookup (`progressDriver.peekAllFleets()`). `progressDriver` was removed in #1122 and is now **permanently `null`**, so `fleetChatId` is always `''` and all three emit paths fall back to `loadAccess().allowFrom[0]` (the DM). The dispatch chat was never stored or recoverable.

## Fix

Recover the originating conversation from the per-agent registry — `jsonl_agent_id → subagents.parent_turn_key → turns.chat_id/thread_id` — and use it for the worker card (chat **and** thread), the legacy progress envelope, and the handback. Best-effort: returns `null` on any miss and the existing DM fallback is preserved, so **DM-shaped agents are completely unaffected** (origin resolves to their DM anyway).

- `registry/turns-schema.ts` — new `getTurnByKey(db, turnKey)` reader (mirrors `findOrphanedTurns`).
- `gateway/gateway.ts` — `resolveSubagentOriginChat(agentId)` helper (uses the already-open `turnsDb` + existing `getSubagentByJsonlId`); wired into the `onProgress` worker-feed `update()`, `decideSubagentProgress`, and `onFinish` `decideSubagentHandback`.
- `registry/turns-schema.test.ts` — 3 cases (missing key / group+topic / plain group-DM).

## Test plan

- [x] `bun test` registry + subagent/worker suites — 8 + 52 + 40 + 39 pass (incl. 3 new).
- [x] `tsc --noEmit` clean.
- [x] `check-bot-api-wrapping.sh` clean (no new raw `bot.api`), `check-plugin-references.mjs` clean.
- [ ] Live verify on a group-owning agent (Marko) once its in-flight research finishes — see Risk.

## Risk

Low and self-limiting. Pure-additive resolver with a graceful DM fallback; no behavior change for DM-only agents. Worst case (registry miss) = today's behavior (DM).

**Rollout note:** Marko has a long background research running right now; deploying this (image + restart) must wait until that completes so the worker isn't killed mid-run. Other agents are DM-shaped → no-op.

## JTBD

`know-what-my-agent-is-doing` — the live worker card and the result must appear **where you asked**, not in a DM you're not looking at.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -427,6 +427,7 @@ import {
   recordTurnEnd,
   findLatestTurnIfInterrupted,
   findRecentTurnsForChat,
+  getTurnByKey,
 } from '../registry/turns-schema.js'
 import {
   buildResumeInterruptedInbound,
@@ -1115,6 +1116,41 @@ try {
 } catch (err) {
   process.stderr.write(`telegram gateway: turn-registry init failed (${(err as Error).message}) — turn tracking disabled\n`)
   turnsDb = null
+}
+
+/**
+ * Resolve the chat/thread a background sub-agent was dispatched from, so
+ * its live worker card + handback route back to the originating
+ * conversation (group / forum topic) instead of the operator DM.
+ *
+ * Walks jsonl_agent_id → `subagents.parent_turn_key` →
+ * `turns.chat_id`/`thread_id`. Returns null on any miss so the caller
+ * keeps its existing `allowFrom[0]` DM fallback — best-effort, never
+ * throws out of the worker-card hot path. This restores the chat context
+ * the pinned-card fleet used to carry before it was removed in #1122
+ * (progressDriver is permanently null, so the old fleet lookup always
+ * yielded the DM for a Task dispatched from a group/topic).
+ */
+function resolveSubagentOriginChat(
+  agentId: string,
+): { chatId: string; threadId?: number } | null {
+  if (turnsDb == null) return null
+  try {
+    const sub = getSubagentByJsonlId(turnsDb, agentId)
+    if (sub?.parent_turn_key == null) return null
+    const turn = getTurnByKey(turnsDb, sub.parent_turn_key)
+    if (turn == null || turn.chat_id.length === 0) return null
+    const threadNum =
+      turn.thread_id != null && turn.thread_id.length > 0
+        ? Number(turn.thread_id)
+        : NaN
+    return {
+      chatId: turn.chat_id,
+      threadId: Number.isFinite(threadNum) ? threadNum : undefined,
+    }
+  } catch {
+    return null
+  }
 }
 
 // ─── Periodic history reaper (#1073) ──────────────────────────────────────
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

Clean, well-scoped fix. Verified against the actual diff and source.

- `registry/turns-schema.ts:361` — `getTurnByKey` mirrors the existing `findOrphanedTurns`/`mapRow` pattern correctly; `Turn.chat_id` is `NOT NULL` (string) and `thread_id` is `string | null`, matching the consumer's handling.
- `gateway/gateway.ts:1132` `resolveSubagentOriginChat` — correctly walks `jsonl_agent_id → getSubagentByJsonlId → parent_turn_key → getTurnByKey`. `getSubagentByJsonlId` and `parent_turn_key` both exist (subagents-schema.ts:356/96). Guards (`turnsDb == null`, `sub?.parent_turn_key == null`, `turn == null`, empty chat_id) plus the try/catch make it genuinely best-effort and non-throwing in the hot path. `thread_id` parsed string→Number with `Number.isFinite` fallback to undefined — sound.
- `gateway/gateway.ts:18516` worker-card `update()` — `origin?.threadId` passed as the 4th arg; `WorkerActivityFeed.update(agentId, chatId, view, threadId?)` already accepts it (worker-activity-feed.ts:248-253, sendOptsFor sets `message_thread_id`). Worker card routes to chat AND thread correctly.
- All three emit paths (`onProgress` worker-feed, `decideSubagentProgress`, `onFinish` handback) keep the `|| fleetChatId || allowFrom[0]` fallback, so DM-shaped agents are unaffected — origin resolves to their DM anyway.

Nit (non-blocking):
- `gateway/gateway.ts:18414` handback path only passes `resolveSubagentOriginChat(agentId)?.chatId` into `fleetChatId`; the handback inbound builder (subagent-handback-inbound-builder.ts:184) has no `threadId` param. So in a forum TOPIC, the live worker card lands in the right topic but the final handback result lands in the group's general thread, not the originating topic. Chat-level routing (the PR's stated goal) is fully fixed; thread-level routing is only fixed for the worker card. Worth a follow-up if topic-accurate handbacks matter.

Rationale: pure-additive resolver with graceful DM fallback, no behavior change for DM-only agents, worst case (registry miss) = today's behavior. Tests cover missing-key / group+topic / plain-group cases. tsc + wrapper checks clean per PR.

mekenthompson:
**Reviewer follow-up addressed:** swapped the real Telegram ids in the new test fixtures for synthetic placeholders (`12345` / `-1001234567890`) — the required `lint`/PII guard is now clean (`check-no-pii-secrets: clean`). All other review checks were APPROVE-track.

**Known limitation (follow-up, non-blocking):** the worker *card* routes to the correct chat **and** forum topic (it passes `threadId`), but the **handback result reply** carries only `chatId` — the synthesized handback inbound (`subagent-handback-inbound-builder.ts`) has no `thread_id` field. So for a **forum-topic** group, the result reply lands in the group's General topic rather than the originating topic. **Plain (non-forum) groups and DMs are fully fixed** — which covers the reported case. Threading the handback inbound is a clean follow-up.

PR #2063, opened 2026-06-01
fix(hostd): bind-mount-safe config write (in-place, no rename-over-target)

mekenthompson:
## Summary

`config_propose_edit` validated patches successfully but **failed to write** them, twice in production, with:

```
E_RECONCILE_FAILED_ROLLED_BACK: write failed: EBUSY: resource busy or locked,
rename '/state/config/switchroom.yaml.tmp' -> '/state/config/switchroom.yaml'
```

## Why

`/state/config/switchroom.yaml` is itself an **individual read-only bind-mount source** — it is bind-mounted into every agent container (`/proc/mounts`: `/dev/nvme0n1p2 /state/config/switchroom.yaml ext4 ro,relatime`). You cannot `rename()` over a path that is an active bind-mount source; the kernel returns `EBUSY`. hostd's `<path>.tmp` → `rename()` atomic-write strategy is therefore structurally broken for the bind-mounted config — the swap never lands.

## What changed

`src/host-control/server.ts` — replace the rename-based swap with an **in-place write that preserves the inode** (`writeFileInPlacePreservingInode`):

- `open(target, "r+")` (O_RDWR, no create) → `ftruncate(0)` → write bytes → `fsync` → close.
- Preserving the inode keeps every bind mount valid and the kernel never sees a rename over a busy dentry.
- `O_RDWR` open preserves the existing file mode/owner (rename would have reset to the umask default).
- Post-write read-back length check catches a short write before reconcile runs.
- **Both** the forward apply **and** the rollback restore use the in-place write, so rollback is bind-mount safe too.

**Tradeoff (documented):** non-atomic — a crash between truncate and the final write could leave a partial file. Mitigated by (a) the pre-write snapshot kept for rollback, (b) `fsync`, (c) the immediate read-back length re-validate, and (d) reconcile rejecting a malformed config and triggering rollback.

## Test plan

- [x] `tests/host-control/config-propose-edit.test.ts` — new "bind-mount-safe in-place write" describe:
  - apply preserves the target **inode + mode** (a rename would change both) — deterministic, runs everywhere.
  - rollback restores **in place** — inode preserved, snapshot restored byte-for-byte.
  - root-gated: applies over a file that is **itself a `mount --bind` target** — the real EBUSY repro; a rename here throws EBUSY. **Verified passing as root on the host.**
- [x] `tsc --noEmit` clean.
- [x] Full `tests/host-control/` suite green (199 passed, 1 skipped — the root-gated test).
- [x] plugin-references gate clean.

## Risk

Low. Confined to the config_propose_edit write path. The non-atomicity window is mitigated as above; the previous behaviour was a hard failure (EBUSY) on every edit, so any successful write is a strict improvement. No image rebuild needed for the host-CLI side, but hostd ships as a container — rollout is a hostd image bump.

### DIFF
--- a/src/host-control/server.ts
+++ b/src/host-control/server.ts
@@ -314,7 +314,6 @@ export function formatConfigApprovalDenyError(
   return `E_DENIED: operator denied config_propose_edit${suffix} (approval_id=${approvalId})`;
 }
 
-/** Best-effort tmp-file cleanup — swallowed errors. */
 /**
  * Write `content` to an existing file **in place**, preserving its inode.
  *
--- a/src/host-control/server.ts
+++ b/src/host-control/server.ts
@@ -33,7 +33,11 @@ import {
   writeFileSync,
   renameSync,
   mkdirSync,
-  unlinkSync,
+  openSync,
+  ftruncateSync,
+  writeSync,
+  fsyncSync,
+  closeSync,
 } from "node:fs";
 import { join, dirname, resolve } from "node:path";
 import { randomUUID, randomBytes } from "node:crypto";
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `src/host-control/server.ts:344-372` — `writeFileInPlacePreservingInode` is correct: `openSync(target, "r+")` (O_RDWR, no create) preserves the live inode/mode/owner, `ftruncateSync(fd,0)` + bounded `writeSync` loop handles partial writes, `fsyncSync` before `closeSync` in a `finally`. The post-write read-back length check (366-371) catches a short write before reconcile.
- `server.ts:1463-1505` — apply path is sound: snapshot captured (1465) before the in-place write (1485); on write failure the snapshot is restored in place (1491) rather than assuming the live file is untouched, then finalized as rolled-back. Restore-of-restore failure is swallowed with the original error surfaced — correct precedence.
- `server.ts:1527-1528` — reconcile-failure rollback also uses the in-place write, so the rollback path is bind-mount safe too (the original bug would have re-bitten here under rename).
- No dangling references: `unlinkSync`/`unlinkSyncBestEffort` and the old `<path>.tmp` swap are fully removed from the config-apply path. The remaining `.tmp`+rename at `server.ts:271` is the separate cache-write path (not a bind-mount source) — correctly left untouched.
- `tests/host-control/config-propose-edit.test.ts:730-833` — tests pin the real contract: inode + mode preserved across apply and rollback (a rename-based writer would change both and fail), plus a root-gated `mount --bind` repro that exercises the actual EBUSY class end-to-end.

Rationale: targeted fix for a structurally-broken atomic-write strategy on a bind-mounted config; inode-preserving in-place write is the right approach, both write paths are converted, the documented non-atomicity window is mitigated by snapshot+fsync+read-back+reconcile-validate, and tests lock in the inode/mode invariants. Safe to merge.

PR #2064, opened 2026-06-01
fix(agents): deny AskUserQuestion fleet-wide (blocking-TUI wedge fix)

mekenthompson:
## Summary
- Deny `AskUserQuestion` fleet-wide so a headless, Telegram-driven agent never renders claude's blocking full-screen multiple-choice TUI selector.
- Seeded as an unconditional fleet-baseline deny (`INTERACTIVE_TUI_FLEET_DENY_TOOLS`) into the fresh-scaffold path, the `reconcileAgent` path, **and** the existing-agent merge branch in `scaffoldAgent` (which previously re-seeded `allow` but not `deny`).

## Why
`AskUserQuestion` renders a blocking selector ("❯ 1. … · Enter to select · ↑/↓ to navigate · Esc to cancel"). With no human at the terminal it blocks forever: the turn never completes, queued inbounds pile up, and the agent looks mute. klanker wedged exactly this way on 2026-06-01 and was only recovered by a manual tmux `Esc`.

Denying the tool is the **root-cause fix, not a workaround** — a denied `AskUserQuestion` does not crash the turn. claude degrades gracefully and asks the same question as plain text, which reaches the operator over Telegram (the desired behaviour) instead of stalling at an invisible TUI.

Unconditional (no per-agent opt-out knob): there is no fleet scenario where a headless agent benefits from a blocking terminal selector. A power user can still re-enable via the `settings_raw` deep-merge escape hatch.

The existing-agent merge branch fix matters because `switchroom apply` calls `scaffoldAgent` (not `reconcileAgent`) and skips the settings.json template for deployed agents — without re-seeding `deny` there, the new deny would only reach existing agents on the `reconcileAgent` path.

## Test plan
- [x] New `tests/scaffold.test.ts` cases: AskUserQuestion denied regardless of webkite presence; reconcile re-seeds it into an existing agent.
- [x] Updated the two exact-array `permissions.deny` assertions.
- [x] `tsc --noEmit` clean.
- [x] `tests/scaffold.test.ts` 186/186 green; broader suite green after `dist` build (remaining failures are dist/network-dependent build-shape guards, unrelated to this change).

## Risk
Low. Behaviour change is a graceful degradation (TUI selector → plain-text question). Takes effect on agent restart/reconcile.

JTBD: "You hold the leash" / "Always available" — an agent that goes mute on a blocking prompt fails both.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -792,6 +792,31 @@ const DEFAULT_READ_ONLY_PREAPPROVED_TOOLS = [
 const WEBKITE_FLEET_DENY_TOOLS = ["WebFetch", "WebSearch"];
 
 /**
+ * Fleet-baseline deny for blocking interactive TUI tools.
+ *
+ * `AskUserQuestion` renders a *blocking* full-screen multiple-choice
+ * selector inside claude's TUI ("❯ 1. … · Enter to select · ↑/↓ to
+ * navigate · Esc to cancel"). Every switchroom agent is driven from
+ * Telegram with no human at the terminal, so that selector blocks
+ * forever — claude waits on a keystroke that never comes, the turn
+ * never completes, and queued inbounds pile up behind it until the
+ * agent looks mute (real incident: klanker wedged this way 2026-06-01,
+ * recovered only by a manual tmux `Esc`).
+ *
+ * Denying the tool is the root-cause fix, not a workaround: a denied
+ * `AskUserQuestion` does NOT crash the turn — claude degrades gracefully
+ * and asks the same question as plain text, which is exactly what we
+ * want (the question reaches the operator over Telegram instead of
+ * stalling at an invisible TUI).
+ *
+ * Unconditional (no per-agent opt-out knob): there is no fleet scenario
+ * where a headless, Telegram-driven agent benefits from a blocking
+ * terminal selector. A power-user who genuinely needs it can still
+ * re-enable via the `settings_raw` deep-merge escape hatch.
+ */
+const INTERACTIVE_TUI_FLEET_DENY_TOOLS = ["AskUserQuestion"];
+
+/**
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `src/agents/scaffold.ts:817` — `INTERACTIVE_TUI_FLEET_DENY_TOOLS = ["AskUserQuestion"]` is wired into all three deny-construction paths, so coverage is complete: the fresh-scaffold path (`buildWorkspaceContext.toolsDeny`, 1978-1981), the existing-agent re-seed on `switchroom apply` (2741-2747), and `reconcileAgent.desiredDeny` (4393-4396). No fourth `settings.permissions.deny` writer exists that would bypass it.
- `scaffold.ts:2741-2747` — the existing-agent re-seed is additive and idempotent (`if (!deny.includes(t))`), preserving any user-declared or webkite denies already present. It mirrors the webkite allow re-seed immediately above (2718-2731) exactly — consistent with the established pattern and the documented reason (`writeIfMissing` skips the settings template for existing agents).
- `tests/scaffold.test.ts:584-639` — coverage is solid: webkite-present and webkite-absent cases both assert the deny lands, plus an explicit reconcile re-seed test that strips the deny from a simulated pre-deny agent and confirms re-scaffold restores it. The two existing `toEqual` assertions (491, 584) were correctly updated for the appended entry.

Rationale: root-cause fix for a real wedge (a blocking TUI selector on a headless Telegram-driven agent). Denying the tool forces claude's plain-text fallback, every deny path is covered idempotently, the settings_raw escape hatch is preserved, and tests pin the behavior across fresh/existing/no-webkite. One thing to keep in view, not blocking: the fix rests on claude degrading a denied AskUserQuestion to a plain-text question rather than hard-erroring the turn — that's the documented and observed behavior, but it's an external-tool contract worth watching across claude CLI upgrades. Safe to merge.

PR #2065, opened 2026-06-01
feat(agents): mid-session wedge watchdog (Layer 2 of blocking-TUI fix)

mekenthompson:
## Summary

Layer 2 of the blocking-interactive-TUI-prompt fix. A continuous,
conservative wedge-watchdog that dismisses a STABLE blocking modal
selector (the `AskUserQuestion` / `ExitPlanMode` class) that wedges a
headless agent mid-session — by sending `Esc`, which claude treats as
"user declined" (non-destructive).

The autoaccept-poll sidecar now runs **two phases in one process**:
1. **Boot phase** (unchanged) — `runAutoaccept` dispatches first-run TUI
   prompts and returns after idle-timeout.
2. **Watch phase** (new) — `runWedgeWatchdog` polls the pane for the
   container lifetime and dismisses any stuck blocking selector.

Kill-switch: `SWITCHROOM_WEDGE_WATCHDOG=0` restores legacy boot-only
behaviour.

## Why

The boot poller is one-shot — it can't catch a prompt claude renders
*later*, mid-session. With no human at the terminal a blocking selector
blocks forever: the turn never completes, queued Telegram inbounds pile up,
and the agent looks **mute** (klanker incident, 2026-06-01, recovered only
by a manual tmux `Esc`).

Layer 1 (#2064) denies `AskUserQuestion` fleet-wide so claude degrades to
plain text. This Layer 2 is defense-in-depth: it catches ANY stuck blocking
selector, including ones Layer 1 can't reach — a `settings_raw` override, a
not-yet-reconciled agent, a future selector tool, or `ExitPlanMode`.

## Design — false positives are worse than false negatives

Interrupting a *live* turn is a regression; a missed wedge is just the
pre-watchdog status quo. So firing requires **all three**:
- a strict blocking-modal footer signature (`Esc…cancel` AND a
  select/navigate affordance — distinct from the working footer
  "esc to interrupt" and the idle REPL footer),
- byte-identical stability across N consecutive polls (default 3 @ 5s =
  ~15s confirmed-stuck; a working pane's spinner/timer changes between
  polls), and
- a post-fire cooldown (default 60s) so we never tight-loop Esc into a
  re-rendering prompt.

It also **defers to the boot autoaccept poller** for first-run prompts
(those want Enter, not Esc), and **soft-fails throughout** (never throws,
only ever injects `Escape`).

## Test plan

- [x] `tests/wedge-watchdog.test.ts` — 10 tests: footer signature
      positive/negatives (idle footer, streaming, plain confirmation,
      empty), fires after threshold, no-fire before threshold, no-fire on
      a moving pane, defers to first-run prompt, resets streak on
      non-wedge interleave, cooldown re-fire math, soft-fail on capture
      throw.
- [x] `tsc --noEmit` clean.
- [x] `node scripts/build.mjs` clean; the baked `dist/cli/autoaccept-poll.js`
      bundle picks up the new watchdog.

## Risk

An always-on keystroke injector on the live fleet. Mitigated by the strict
signature + multi-poll stability + cooldown + first-run deferral above; the
only key it ever sends is `Esc`. Kill-switch `SWITCHROOM_WEDGE_WATCHDOG=0`.
Takes effect on agent restart/apply (changes the baked autoaccept-poll
bundle). Related: #2064 (Layer 1).

### DIFF
--- a/profiles/_base/start.sh.hbs
+++ b/profiles/_base/start.sh.hbs
@@ -111,10 +111,15 @@ if [ "$SWITCHROOM_RUNTIME" = "docker" ] && [ -z "$SWITCHROOM_DOCKER_TMUX_INNER" 
     echo "[start.sh] channels.telegram.enabled=false — skipping gateway sidecar" >&2
   fi
 
-  # 2) autoaccept-poll — first-run TUI prompt dispatcher. Single-shot
-  #    by design (exits cleanly after idle-timeout once prompts have
-  #    fired); the supervisor's exponential backoff keeps a flaky
-  #    autoaccept from busy-looping.
+  # 2) autoaccept-poll — first-run TUI prompt dispatcher, then continuous
+  #    wedge-watchdog. Two phases in one process: a one-shot boot phase
+  #    dispatches the first-run prompts and returns after idle-timeout,
+  #    then the process stays alive running the wedge-watchdog (dismisses a
+  #    stuck blocking modal selector mid-session — the AskUserQuestion /
+  #    ExitPlanMode class — with Esc). So it is NORMALLY long-lived; the
+  #    supervisor only respawns it if it crashes/exits, and its backoff
+  #    keeps a flaky run from busy-looping. Set SWITCHROOM_WEDGE_WATCHDOG=0
+  #    to restore the legacy boot-only single-shot behaviour.
   if [ -f /opt/switchroom/autoaccept-poll.js ] && command -v bun >/dev/null 2>&1; then
     _switchroom_supervise autoaccept /var/log/switchroom/autoaccept.log \
       bun /opt/switchroom/autoaccept-poll.js "{{name}}" &
--- a/src/agents/wedge-watchdog.ts
+++ b/src/agents/wedge-watchdog.ts
@@ -1,0 +1,191 @@
+// Mid-session wedge watchdog — defense-in-depth for blocking interactive
+// TUI prompts that no human is present to answer.
+//
+// The boot-time autoaccept poller (`runAutoaccept`, autoaccept.ts) is a
+// one-shot: it dispatches the small set of first-run prompts (theme, MCP
+// trust, dev-channels) and exits after ~30s. It cannot catch a prompt that
+// claude renders LATER, mid-session. The dominant such case is a blocking
+// modal selector (e.g. claude's `AskUserQuestion`, or `ExitPlanMode`'s
+// menu): a full-screen multiple-choice list with the footer
+//   "❯ 1. … · Enter to select · ↑/↓ to navigate · Esc to cancel".
+// With no human at the terminal it blocks forever — the turn never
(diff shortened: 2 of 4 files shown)


mekenthompson:
APPROVE

- `src/agents/wedge-watchdog.ts:1-191` — sound, conservative state machine. The fire condition requires ALL of: a strict footer signature match, multi-poll byte-stability, NOT matching a first-run prompt, and cooldown elapsed. False-positive risk (interrupting a live turn) is the right thing to minimize and this does — a streaming pane changes between polls so it can't accumulate a stable streak. Only ever sends `Escape` (non-destructive "declined"), never throws (capture wrapped in try/catch, soft-fails to empty string).
- `wedge-watchdog.ts:55` — `WEDGE_FOOTER_SIGNATURE` correctly requires BOTH an "Esc…cancel" affordance AND a select/navigate affordance via two lookaheads, so the working-turn footer ("esc to interrupt") and plain confirmations fail at least one branch. Verified against the negative test cases at tests/wedge-watchdog.test.ts:60-69.
- `wedge-watchdog.ts:140-145` — defers to `PROMPTS` (the boot autoaccept set) so first-run prompts that want Enter aren't Esc'd. Imports resolve: `capturePane` (autoaccept.ts:150), `sendKeys` (174), `PROMPTS` (79), `PromptRule.match: RegExp` (37) — `p.match.test(text)` is valid.
- `src/cli/autoaccept-poll.ts:32-66` — two-phase wiring is correct: boot `runAutoaccept` (one-shot) then `runWedgeWatchdog` (continuous), both in try/catch, `SWITCHROOM_WEDGE_WATCHDOG=0` cleanly restores boot-only behavior, always exits 0.
- `profiles/_base/start.sh.hbs:114-122` — supervisor comment correctly updated: the poller is now normally long-lived, respawned only on crash/exit with existing backoff. No change needed to the supervise call itself.
- Tests (tests/wedge-watchdog.test.ts) are thorough: fire-on-stable, no-fire-under-threshold, moving-pane immunity, first-run defer, streak reset, cooldown re-fire timing, and capture-throw soft-fail. I traced the cooldown test math (fire at poll 3, blocked until clock≥70s, re-fire at poll 15 → exactly 2 fires) — internally consistent.

Watch (non-blocking): the defer-to-PROMPTS guard means a mid-session modal that coincidentally matches a first-run prompt regex would be skipped, but boot autoaccept has already exited by then so nothing dismisses it. Narrow edge (the modal body would have to match theme/MCP/dev-channels signatures) and Layer 1's fleet deny covers AskUserQuestion regardless, so it's acceptable as-is — worth a note if a new PROMPTS rule is ever added with a broad pattern.

Rationale: correct, well-isolated Layer-2 defense-in-depth. Conservative firing, non-destructive action, soft-fail throughout, opt-out env knob, and comprehensive tests. Safe to merge.

PR #2067, opened 2026-06-01
feat(vault): operator-set agent-level standing secrets grant

mekenthompson:
## Summary
Adds `agents.<name>.secrets[]` — an **operator-set standing vault grant** the broker honours for an agent **independent of any cron or MCP server**. Cascades UNION across `defaults → profile → agent`.

## Why (design review against the vision)
Today an agent can only be granted a vault key by welding it to a specific cron's `schedule[].secrets[]` or an `mcp_servers[].secrets`. Skill-based credentials an agent needs **both interactively and in its own agent-managed schedules** (e.g. clerk's `microsoft/*` + `compass/credentials`, read via `switchroom vault get` inside the mail/calendar/compass skills) had no clean home — the grant leaked in only via a cron, tangling *"what the agent may access"* with *"when it runs."*

This separates them: a **standing grant** is "what it may access"; the **schedule** is "when it runs." It's the primitive that lets an agent own its scheduling (overlay crons, which deliberately can't carry secrets) while the operator still controls access.

**Operator-only by design** — agents can't edit `switchroom.yaml` or self-grant ([vision.md](../blob/main/reference/vision.md) outcome 2, *"you hold the leash; only your tap grants it"*). A standing grant **is** the tap, expressed as config. This was deliberately scoped to *not* add agent self-grant / ownership-confers-authority, which fails outcome 2 and all three principle checks.

**Purely additive:** an agent with no `secrets:` is unaffected — the new ACL clause yields an empty set and falls through to the existing mcp/schedule paths.

## Changes
- `schema.ts`: `secrets` in `profileFields` (defaults/profile) + mirrored into `AgentSchema` (which doesn't spread profileFields).
- `merge.ts`: UNION cascade (defaults first, dedup), mirroring `tools.allow`.
- `acl.ts` (broker enforcement): grant if key ∈ cascaded standing set; updated the no-access deny reason.
- `docs/configuration.md`: field-reference row + "Standing vault grants" section.

## Test plan
- [x] 6 acl tests + 3 merge tests (standing grant w/ no schedule, defaults/profile cascade, 3-layer union, per-agent isolation, deny-reason).
- [x] Full `vitest run src/vault src/config` — **584 pass**.
- [x] `npm run lint` — all 7 guards clean (tsc, plugin-refs, bot-api, bun-test-imports, no-pii, vault-hermeticity, no-broadcast).

## Risk
Low / additive. New grant clause only; no existing path altered. Broker-side change → requires a broker roll on release. DM/no-`secrets` agents unaffected.

## Follow-up (separate security PR, **not** bundled)
`mint_grant` has no key-ACL check on the posture path — a `postureMintAgents` agent that can speak the broker wire protocol could self-mint arbitrary grants. The safe fix binds mint to a consumed approval-kernel decision (so `vault_request_access` keeps working). Its own carefully-tested PR.

## JTBD / vision
Outcome 2 — *you hold the leash.* Makes "operator grants standing access, decoupled from crons" first-class, so agents can own their schedules without owning the grant.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -33,6 +33,7 @@ Each field type has specific merge behavior when values exist at multiple layers
 | `hooks` | per-event concat | Claude Code lifecycle hooks (SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Stop, SessionEnd) |
 | `env` | per-key | Environment variables for start.sh |
 | `mcp_servers` | per-key | Additional MCP server configurations. Set a key to `false` to suppress a built-in default (e.g. `playwright: false`) |
+| `secrets` | union | **Operator-set** standing vault grant: vault keys this agent may read via the broker, independent of any cron or MCP server. Use for credentials an agent needs both interactively and in its own (agent-managed) schedules. Agents cannot self-grant — this is operator-only by design ([vision.md](../reference/vision.md) outcome 2). See [§ Standing vault grants](#standing-vault-grants-agentsnamesecrets). |
 | `system_prompt_append` | concatenate | Appended to the system prompt via `--append-system-prompt` |
 | `skills` | union | Named skills from the global skills pool (`switchroom.skills_dir`) |
 | `bundled_skills` | per-key | Opt-out map for switchroom's bundled-default skills. Set a key to `false` to suppress (e.g. `pdf: false`). See [docs/skills.md](./skills.md). |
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

- `acl.ts:432-461` — standing-grant clause is correctly placed after the mcp-server grant and before the no-schedule early-deny, so a schedule-less agent still resolves its grant. Reuses the `profileName` defined for the mcp cascade (acl.ts:408) for `extends` resolution — same defaults→profile→agent layering, open-coded for the same documented reason (broker doesn't run the full merge pipeline per request).
- `acl.ts:459` — `standingSecrets.includes(key)` is exact-match, no glob, mirroring the mcp `declared.includes(key)` model (acl.ts:427). Consistent, no surprise wildcard surface.
- `merge.ts:349-354` — UNION cascade (defaults first, dedup) mirrors `tools.allow`; produces `["shared/key","microsoft/token"]` for the test case, matching merge.test.ts:123.
- `schema.ts:1452 + 1825` — `secrets` regex `^[a-zA-Z0-9_\-/]+$` correctly permits `/` for `service/key` names. Mirrored into AgentSchema because it doesn't spread profileFields — same pattern as `resources`, comment explains it (schema.ts:1823).
- Deny-reason string (acl.ts:467) updated to mention the new `agents.<name>.secrets[]` path — accurate.
- Operator-only invariant holds: agents can't edit switchroom.yaml, and nothing in this diff adds a self-grant path. Purely additive — an agent with no `secrets:` yields an empty set and falls through to the existing mcp/schedule paths unchanged.
- Note: couldn't re-run the suite in a fresh clone (deps not installed); verified the four touched files by reading. PR reports 584 pass + 7 lint guards clean. The follow-up flagged in the PR body (mint_grant has no key-ACL check on the posture path) is correctly scoped OUT of this PR — agree it belongs in its own security PR.

Rationale: clean, additive primitive that decouples "what an agent may access" from "when it runs"; broker enforcement is correctly placed, exact-match, and operator-only by construction.

Issue #833, opened 2026-05-08
approvals: clarify Phase 1 kernel verdict is audit-only, not enforcing


mekenthompson:
Per reviewer of #830: the kernel `record` calls in vd:unlock dual-dispatch are advisory — a kernel-deny verdict does NOT actually block the legacy unlock path. Only the audit log gains the kernel row.

Update `src/vault/approvals/MIGRATION.md` to make this explicit: "Phase 1 dual-dispatch is audit-only. Phase 2 will switch the legacy path to gate on kernel verdict."

Tighten commit message framing in future call-site migrations (`vault grant` wizard next, then any future paths).

PR #2069, opened 2026-06-01
docs(rfc): hard-boundary operator approvals (no agent self-elevation)

mekenthompson:
RFC for operator review — **no code/behaviour change**. Designs the architectural fix for the vault-grant/apv: self-elevation leak: the kernel enforcing flip (#833) can't close it (consume/record are on claude's per-agent socket — kernel-server.ts:472), so we need a host-side approval verifier (hostd/operator UID) that receives+verifies the operator tap and records the decision on the operator-only socket; broker mints only on an `origin='operator'` decision claude can't forge. Phased, flag-gated, vault grants first. Open questions for you in §7.

### DIFF
--- a/docs/rfcs/vault-approval-hard-boundary.md
+++ b/docs/rfcs/vault-approval-hard-boundary.md
@@ -1,0 +1,155 @@
+# RFC: Hard-boundary operator approvals (no agent self-elevation)
+
+- Status: **Draft for operator review**
+- Author: drafted 2026-06-02
+- Tracks: vision.md outcome 2 ("you hold the leash — it can't self-elevate
+  or route around you"); supersedes the audit-only framing of #833 / the
+  approval-kernel `src/vault/approvals/MIGRATION.md` Phase 2 for the
+  *enforcement* question.
+
+## 1. Summary
+
+Switchroom's approval surfaces (vault grants, `apv:` MCP/secret approvals)
+are **soft boundaries by architecture**: the gateway that processes an
+operator's Telegram tap runs in the **same container as `claude`, sharing
+the same per-agent UDS sockets**. So any operator-authorization the gateway
+can produce, a determined `claude` can also produce. This contradicts the
+vision's "it can't self-elevate." This RFC proposes the architectural change
+required for a **hard** boundary: move the operator-tap *verification* to a
+host-side component `claude` cannot impersonate, and make the broker's mint
+gate on a decision recorded through that path.
+
+## 2. The problem, precisely
+
+`vault_request_access` (agent asks → operator taps → broker mints a grant)
+mints via `mint_grant` on the broker's **per-agent socket**. The
+posture-attested path requires only: agent on `postureMintAgents`, telegram-id
+mode, agent-self-match. **No key-ACL, no proof an operator tapped.**
+
+Completing the approval-kernel "enforcing flip" (#833) — making the broker
+require a consumed kernel `approval_decision` — does **not** fix it. The
+kernel's `approval_consume` / `approval_record` are themselves on the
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

Docs-only RFC (155 additions, 1 new file `docs/rfcs/vault-approval-hard-boundary.md`, no code/behaviour change). Verified every cited source anchor against `main` — all accurate:

- `kernel-server.ts:472` (line 22 of RFC) — the quoted comment is verbatim ("a compromised agent could self-consume + self-record an allow_always decision … no operator ever approved (approval-integrity bypass)"). Confirmed.
- `kernel-server.ts:414-483` — the listener-identity ACL (`checkApprovalAclByAgent`) on `approval_consume`/`approval_record`/`approval_revoke` matches the *calling* agent to the nonce's owning `agent_unit`. The RFC's core claim is correct: this blocks **cross-agent** forgery only; an agent self-consuming + self-recording a decision for **its own** request passes the ACL. The threat model (in-container `claude` has Bash + full container access, reads the nonce off the card) is sound.
- `kernel-server.ts:188,333` — operator socket is bound mode 0600, flagged `isOperator`, deny-by-default (`OPERATOR_ALLOWED_OPS`, checked first before any op dispatch). The RFC correctly notes the operator socket is read-mostly today and explicitly scopes revisiting that for the host-verified path only (§4.1 step 3) rather than hand-waving it.
- `mint_grant` posture-attested path — exists (`server-mint-grant-posture-attest.test.ts`, `postureMintAgents` + telegram-id mode); RFC's §2 description matches.
- operator socket host-reachability + 0600 chown — confirmed (`peercred-socket-path.test.ts:103-118`, `/run/switchroom/broker/operator/sock` resolves to operator identity, not an agent).
- `docs/vault.md:287` quote ("the ACL is misconfiguration protection, not a security boundary") — faithful paraphrase of the source.

Substance: the argument is internally coherent and the design (host-side approval verifier at operator UID + `origin='operator'` column on `approval_decisions`, broker mints iff `origin='operator'`) is the right shape for converting a soft boundary to a hard one — `claude` can forge an `origin='agent'` decision but cannot reach the operator socket nor manufacture a real Telegram tap. Honest framing throughout: §4.3 and Phase-1 (§5) explicitly label the interim key-ACL constraint as "not a hard boundary," which avoids overclaiming. Fail-closed posture (§4.2, §6 downgrade — pre-existing decisions read `origin='agent'`) is correct.

No blockers. As a design RFC with open questions deferred to the operator (§7), it is accurate, well-grounded in the actual code, and safe to merge — merging records the design, it changes no runtime behaviour.

One non-blocking note for whoever implements Phase 2: §4.1-B's parenthetical ("Telegram delivers a bot's updates to a single poller/webhook") is the load-bearing constraint that makes option A clearly preferable — worth promoting out of the parenthetical into the implementation ticket so it isn't lost.

mekenthompson:
Fresh-context review: **APPROVE** (docs-only RFC, accurate against current code). Non-blocking notes for a follow-up:

1. §2 is slightly stale: #2073 shipped an admin-only key block on posture-mint (broker/server.ts:2074-2108) — a partial version of the Phase 1 key-ACL this RFC defers. Worth acknowledging.
2. §4.1 cites the Hindsight memory key `feedback_kernel_operator_socket_deny_by_default` — meaningless to repo readers; inline the explanation.
3. §4.1 'read-mostly' understates: the operator socket is read-only by construction (OPERATOR_ALLOWED_OPS = approval_list).
4. Missing cross-link to docs/rfcs/approval-kernel.md, and no mention of hostd's config_propose_edit approval flow, which already demonstrates the operator-socket gating pattern this RFC designs.

None of these change the architecture or decision record; merging.

PR #2070, opened 2026-06-01
feat(vault): approval origin tracking + flag-gated operator-verified mint gate (hard-boundary PR 1)

mekenthompson:
**PR ① of the vault-approval hard-boundary RFC (#2069).** Inert foundation — default-OFF flag, zero behaviour change until PR ② (host verifier) + PR ③ (flip).

## Why
Posture-attested `mint_grant` proves nothing about an operator tapping — claude shares the per-agent socket with the gateway, so it can self-mint. The kernel's consume/record are on that same socket too (only checks 'nonce belongs to my agent', kernel-server.ts:472), so the #833 kernel flip alone wouldn't close it. This lands the provenance + enforcement seam.

## Changes
- **kernel:** `origin` column on `approval_decisions` ('agent' default / 'operator'), idempotent migration, threaded through record/consumeAndRecord + `DecisionRow`/`getDecision`. **Server-set, never from the wire** — a per-agent record stays 'agent'; only the future host verifier sets 'operator'.
- **kernel:** pure `isOperatorVerifiedDecision()` predicate (operator-origin + agent-match + allow + unrevoked + unexpired).
- **broker:** behind `SWITCHROOM_REQUIRE_OPERATOR_APPROVAL_MINT` (**default off**), a posture mint must reference an `origin='operator'` decision via the new optional `decision_id`, else DENIED. Flag off ⇒ unchanged (existing posture happy-path test still green = backward-compat).

## Tests
13 new (origin plumbing, idempotent migration, the predicate's full truth table, broker gate denies the forgeable path). vitest/bun split wired. Full `vitest src/vault` (317) + bun approvals/broker (112) green; `npm run lint` clean (7 guards).

## Risk
Additive + inert. Flag stays OFF fleet-wide. No mint behaviour changes until PR ②/③.

## Next
PR ② — host-side approval verifier records `origin='operator'` on a claude-unreachable channel (dedicated approval bot per RFC §4.1-A). PR ③ — flip the flag, monitored roll.


mekenthompson:
APPROVE

PR ① of the #2069 hard-boundary RFC. Inert, additive, fail-closed foundation — verified every security-critical invariant against the checked-out branch:

- **origin is server-set, never from the wire** — `MintGrantRequestSchema` adds only `decision_id: z.string().optional()` (protocol.ts:341), NOT `origin`. `recordDecision` defaults `input.origin ?? 'agent'` (kernel.ts:234) and `consumeAndRecord` threads it through (kernel.ts:561). Critically, `kernel-server.ts` has **zero** `origin` references — confirmed by grep — so the per-agent-socket `approval_record`/`consumeAndRecord` path cannot set `origin='operator'`. That is the whole security premise and it holds: nothing claude can reach mints an operator-origin decision in this PR.
- **Gate is on the posture path only, mint_grant only** — the block (server.ts:2076-2079) sits inside the `attest_via_posture === true` branch (opened at server.ts:1942) and guards on `req.op === 'mint_grant'`. The passphrase-attested and operator-socket CLI mint paths are untouched. Matches RFC §4.2.
- **Fail-closed throughout** — `isOperatorVerifiedDecision` (kernel.ts:268) returns false on null, non-operator origin, agent mismatch, deny/deny_perm, revoked, or expired TTL. The broker passes `agentName ?? ''` (server.ts:459), so a missing agent identity yields `agent_unit === ''` → no match → DENIED. The new test asserts both the deny audit row and that the allow path is never reached (server-mint-grant-posture-attest.test.ts:389-420).
- **Migration is idempotent** — `CREATE TABLE IF NOT EXISTS` (schema.ts:36) carries `origin TEXT NOT NULL DEFAULT 'agent'` for fresh DBs; the guarded `PRAGMA table_info` + `ALTER ADD COLUMN` (schema.ts:311-318) covers pre-existing DBs without throwing. Pre-existing decisions read as `origin='agent'` → fail-closed for the new gate, which is correct.
- **Backward-compat** — `beforeEach` deletes `SWITCHROOM_REQUIRE_OPERATOR_APPROVAL_MINT` (test:363-364) so the existing posture happy-path test stays green with the flag off; flag is default-OFF in code (server.ts:452), so no mint behaviour changes until PR ②/③.
- vitest/bun split correctly wired (`approval-origin.test.ts` added to test:bun and to vitest.config exclude — package.json:10, vitest.config.ts:491; it imports `bun:sqlite`).

No blockers. The predicate's truth table, origin plumbing, idempotent migration, and the broker deny-path are all covered by the 13 new tests.

Watch item for PR ② (not a blocker here — the gate is inert): `isOperatorVerifiedDecision` checks revoked/expired but not single-use-per-mint, and the broker reads the decision via `getDecision` without consuming it on mint. Once the host verifier produces real `origin='operator'` decisions, an `allow_always`/unexpired `allow_ttl` decision could be referenced by more than one `mint_grant` call within its validity window. PR ② should bind an operator-verified decision single-use to one mint (consume-on-mint, or a `last_used_at`/minted-grant check) so a verified tap can't be replayed across multiple mints.

PR #2071, opened 2026-06-01
feat(vault): stop agents guessing key names — self-correcting denied hint + discovery guidance

mekenthompson:
## Summary
Agents have no innate knowledge of their vault key names, so they **guess** — e.g. `postiz/api-key` when the real, already-granted key is `marko/postiz-api-key`. The old denied-hint then steered them straight to `vault_request_access`, **burning an operator approval on a key that doesn't exist**. (Live incident: marko, 2026-06-02 — three failed mint requests for non-existent `postiz/*` keys it already held under `marko/postiz-*`.)

## Why
JTBD: "you hold the leash" + the defaults/docs principle checks. A guessed key shouldn't cost the operator a tap; the system should self-correct the agent to a key it already has.

## Changes
**A. Self-correcting denied error** (`src/cli/vault.ts`, new `src/cli/vault-key-suggest.ts`)
- On a broker-denied `get` inside an agent, fetch the keys the agent **actually holds** (broker `list` op — same disclosure as `switchroom vault list`, no info leak) and surface near-matches via a pure token-overlap ranker.
- The hint now leads with *"you already have access to `<real key>` — use that"* and *"run `vault list`"*, demoting `vault_request_access` to a confirmed-last-resort.
- Best-effort: a list-lookup failure never masks the denial.

**B. Proactive fleet guidance** (`VAULT_GUIDANCE` in `scaffold.ts` → fleet invariants)
- Teaches every agent: read with `vault get`, **never guess**, `vault list` shows your real `<you>/...` keys, only `vault_request_access` for a key you've confirmed you lack.

## Test plan
- [x] `vault-key-suggest.test.ts` — 7 cases (guess→real mapping, ranking, no-overlap→empty, self-exclusion, cap, case-insensitivity)
- [x] `tests/scaffold.test.ts` — new assertion that invariants teach `vault list` / "never guess"
- [x] `vault-denied-envelope.test.ts` unaffected (envelope unchanged)
- [x] `tsc --noEmit` clean

## Risk
Low. The denied path is an error path; the change only enriches the human-readable hint string (the machine `ERROR-ENVELOPE:` line is untouched, so structured consumers are unaffected). The fleet-guidance const is additive text. No runtime/behavior change on the success path.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -263,6 +263,31 @@ false\` in that agent's switchroom.yaml block — webkite goes away
 and the native tools come back together.`;
 
 /**
+ * Vault key discovery. One of the fleet invariant blocks. Stops the
+ * single most common vault failure: an agent GUESSING a secret's key
+ * name (\`postiz/api-key\`) when it already holds the real one
+ * (\`marko/postiz-api-key\`), then burning an operator approval on a
+ * \`vault_request_access\` for a key that doesn't exist.
+ * See SANDBOX_GUIDANCE above for the constant-vs-\`.hbs\` rationale.
+ */
+const VAULT_GUIDANCE = `## Secrets in the vault — discover, don't guess
+
+Your API keys and credentials live in the Switchroom vault. For normal
+MCP-tool work they're injected by the launchers and you never touch the
+raw values. When a task needs a secret directly (a direct API call an
+MCP tool has no verb for), read it with \`switchroom vault get <key>\`.
+
+**Never guess a key name.** Run \`switchroom vault list\` to see the
+exact keys you already hold — they're usually namespaced \`<you>/...\`
+(e.g. \`marko/postiz-api-key\`, not \`postiz/api-key\`). Use the real
+name from that list.
+
+Only call the \`vault_request_access\` MCP tool — which pings the
+operator for a Telegram approval — for a key you've **confirmed via
+\`vault list\` you don't already have.** Requesting access to a guessed
+or already-held key wastes the operator's tap and fails.`;
+
+/**
  * Canonical rendering of the fleet invariants file written to
  * `~/.switchroom/fleet/switchroom-invariants.md`. Apply checksums
  * this against the on-disk file and restores on drift. Operators
@@ -296,6 +321,8 @@ export function renderFleetInvariants(): string {
     MEMORY_GUIDANCE,
     "",
     WEB_FETCH_GUIDANCE,
+    "",
+    VAULT_GUIDANCE,
     "",
   ].join("\n");
 }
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

Verified the full diff and the supporting call sites against the checked-out branch. The change does exactly what the description claims, on the error path only.

- **Ranker is pure and correct** (vault-key-suggest.ts:138-172) — `tokenize` lowercases and splits on `/ - _ .`, dropping empties; `suggestVaultKeys` scores by shared-token count (primary) with symmetric-difference as the tie-break, filters `shared > 0`, excludes the requested key, and caps at `max` (default 3). Returns `[]` on empty/zero-overlap input. The 7 cases in vault-key-suggest.test.ts:70-114 cover guess→real mapping, ranking, no-overlap, self-exclusion, cap, and case-insensitivity. No correctness issues.
- **Caller feeds the right shape and is null-safe** (vault.ts:909-910) — `listViaBroker` returns `Promise<string[] | null>` (client.ts:619-628, returns the broker `list` op's `keys` array); the call guards `if (myKeys && key)` so a `null` (unreachable broker) yields no suggestions rather than a crash. Wrapped in `try/catch` (vault.ts:911) so a list-lookup failure never masks the denial — matches the 'best-effort' claim.
- **No info leak** — `listViaBroker({op:'list'})` is the identical call that backs the existing `switchroom vault list` command (vault.ts:1046), which returns only the keys this agent is allowed to read. The denied hint therefore discloses nothing the agent couldn't already enumerate itself. Claim holds.
- **Machine envelope unchanged** — the `VAULT-BROKER-DENIED [code]: msg` stderr prefix (vault.ts:912-913) and `writeVaultDeniedEnvelope(key, result.code, result.msg)` (vault.ts:916) are untouched; only the human-readable `recoveryHint` string is enriched, and `recoveryHint`'s new `suggestions?` param is optional (vault.ts:189) so the no-match branch (vault.ts:219-228) still steers to `vault list` first, then `vault_request_access`. Structured consumers unaffected.
- **Fleet guidance is additive** (scaffold.ts:17-32, wired into `renderFleetInvariants` at scaffold.ts:41) — pure constant text; scaffold.test.ts:268-279 asserts the rendered invariants contain the discover-don't-guess guidance. No behaviour change.
- **Success path untouched** — the suggestion lookup lives entirely inside the broker-denied `else` branch; a successful `get` never hits it.

No blockers.

Nits (non-blocking, ignore freely):
- Self-exclusion `c.key !== requested` (vault-key-suggest.ts:168) is case-sensitive while tokenization is case-insensitive, so a case-variant of an already-held key could appear in suggestions. Harmless — it's still a real key the agent holds — but exact-dedup on the lowercased form would be marginally cleaner.
- Every denied `get` now incurs one extra broker `list` round-trip. It's the error path and best-effort, so the cost is fine; just noting it's not free.

PR #2073, opened 2026-06-01
feat(vault): admin-only credentials — owner-only approval + passphrase-required mint

mekenthompson:
## Summary
Lets the operator designate sensitive vault keys that may be granted **only by the admin operator** (`access.allowFrom[0]`) and **only with the vault passphrase** — never via posture attestation, even under the `approvalAuth: telegram-id` single-factor fleet default.

```yaml
vault:
  broker:
    adminOnlyKeys:
      - stripe/*
      - microsoft/ken-tokens
```

## Why
JTBD: *you hold the leash* (vision outcome 2). Today **any** `allowFrom` member can approve a grant, and under `telegram-id` mode a `postureMintAgents` agent could self-mint without a passphrase. For high-sensitivity credentials the operator wants a bar an agent cannot forge.

## Design (decisions confirmed with operator)
- **Admin = `access.allowFrom[0]`** (the owner; dynamic, no hard-coded ID).
- **Gateway + broker enforcement** (both layers).

### Broker (surface-independent)
A posture-attested `mint_grant` may **retain** an admin-only key the agent already holds — the gateway unions a newly-approved key with the agent's existing grant on re-mint, so a previously passphrase-granted admin-only key legitimately reappears — but can never **add** a new one. Adding requires the passphrase path. This also blocks an agent crafting raw broker protocol. Computes the agent's active key set via `listGrants`.

### Gateway
An admin-only `vra:approve` tap from a non-`allowFrom[0]` sender is rejected (toast; the card + buttons stay intact for the owner). Admin-only keys skip the `telegram-id` posture branch and route through the **existing** passphrase-prompt flow, so the operator's passphrase mints them (two-factor: owner ID + passphrase).

### Shared pure core
`src/vault/admin-only-keys.ts` (dependency-free, imported by both `src` broker and `telegram-plugin` gateway): `matchesAdminOnlyKey` (exact/`*`-glob, regex-metachar-safe), `adminOnlyKeysBeingAdded` (retain-but-not-add).

## ⚠️ Operator UX note
Under `approvalAuth: telegram-id` (auto-unlock) you normally never type the passphrase. **Admin-only keys are the exception** — approving one prompts for the vault passphrase (the unforgeable factor). If forgotten, grant on the host: `switchroom vault grant <agent> --keys <key>`. Documented in `docs/vault.md`.

## Test plan
- [x] `admin-only-keys.test.ts` — 12 cases (glob/exact/case-sensitivity/regex-metachar safety; retain-vs-add incl. union)
- [x] config schema (44) + broker (198) vitest green
- [x] 3892 telegram-plugin bun tests green
- [x] `node scripts/build.mjs` clean
- [ ] **Attended canary on test-harness** before any fleet rollout (credential-security change)

## Risk
Touches the vault mint path. Mitigated: enforcement is additive (no `adminOnlyKeys` configured = exact current behavior — default `[]`); the retain-but-not-add rule preserves the existing key-union flow; broker denial is fail-closed with a clear message; pure core is unit-tested. **Will not be rolled unattended** — staged for the operator's canary + go.


mekenthompson:
APPROVE

Verified the full diff on the checked-out PR head (`pull/2073/head`), tracing every security invariant to the real files — not the PR body.

- **Broker guard is on the posture path, not the passphrase path** (`src/vault/broker/server.ts:2065-2105`). The new `adminOnlyKeysBeingAdded` check sits inside the `attest_via_posture === true` branch (opened at server.ts:1943), right alongside the existing `postureMintAgents` allowlist (1990s) and agent-self-match (2025) defences, before the `this.passphrase === null` fallthrough. So an admin-only key requested via posture is denied (`denied:posture-mint-admin-only-key`); the passphrase mint path is untouched. Correct placement — this is surface-independent and also blocks a raw-protocol `mint_grant`.
- **Retain-but-not-add is computed from live grants** (server.ts:2078-2083). `existingKeys` = `listGrants(agentName)` filtered to non-expired (`expires_at === null || > nowSec`), flatMapping both `key_allow` and `write_allow`. So a previously passphrase-granted admin-only key survives a union re-mint, but a *new* one is blocked. Matches the 12 unit cases in `admin-only-keys.test.ts` (incl. the union case at :153 and the "blocks new while retaining existing" case at :166).
- **Gateway owner-gate uses the right accessor** (`telegram-plugin/gateway/gateway.ts:13879-13888`). `senderId` is `String(ctx.from?.id)` (13702), `access.allowFrom[0]` is the owner — same accessor already used for operator resolution at gateway.ts:4715/4791/4869. A non-admin tap on an admin-only key returns early with a toast and leaves the card intact, so the owner can still approve. Type-consistent string compare.
- **Admin-only keys correctly bypass the telegram-id auto-mint** (gateway.ts:13895 — `!isAdminOnly && VAULT_APPROVAL_AUTH_MODE === 'telegram-id'`). They fall through to the existing passphrase-prompt flow (distinct admin-only prompt copy at 13960). Two-factor (owner ID + passphrase) holds even under single-factor fleet default. The broker enforces the same rule, so the gateway layer is UX/defense-in-depth, not the sole boundary — good.
- **Pure core is regex-safe** (`src/vault/admin-only-keys.ts:215-221`). `globToRegExp` escapes every metachar except `*` then maps `*`→`.*`, anchored `^...$`. Tested for literal-dot (`a.b/*` does not match `aXb/key`, :121) and case-sensitivity (:117). Empty pattern list returns false (:127) → unconfigured = exact current behavior.

Additive and fail-closed: default `adminOnlyKeys: []` is a no-op, broker denial has a clear actionable message, both layers enforce independently. Inert until the operator configures keys + restarts (no ACL hot-reload, documented at docs/vault.md:50).

Nits (non-blocking):
- `ADMIN_ONLY_KEYS` is cached at boot in `initVaultApprovalPosture` and the broker reads `this.config` — both need a restart to pick up config changes. Documented (docs/vault.md:50), but worth a one-line log on mismatch if the two layers ever drift out of sync.
- Test plan leaves the attended canary unchecked — appropriate to hold for the operator's go before any fleet roll, as the PR states.

Sound hardening of the mint path. No blockers.

PR #2075, opened 2026-06-02
fix(telegram): promote mis-recorded foreground worker to background from launch ACK

mekenthompson:
## Summary

A `run_in_background` Agent/Task dispatch sometimes omits the `run_in_background` key from the `tool_input` the **PreToolUse** hook sees (observed live on claude-code 2.1.159 — the clerk worker that never surfaced to Telegram). The subagent row is then recorded `background=0` even though the worker actually runs detached, breaking the worker-activity-feed card at the source.

This PR makes the **PostToolUse** hook trust Claude Code's async-launch ACK over the pretool's input-derived flag and self-correct the row.

## Why

With `background=0` on a row whose worker is actually detached, two things break:

1. **No worker card.** Gateway `onProgress` (`worker-feed-dispatch.ts`) reads `isBackground = sub?.background ?? false` → takes the foreground branch → bails on `if (turn == null) return` once the parent turn has ended. The card never fires.
2. **Premature completion.** The posttool's foreground path writes `status=completed` / `ended_at` on the **launch ACK** (a ~0.7s immediate return), while the worker runs on for minutes — also poisoning the reaction-defer gate, the stuck-row reaper, and status output.

The async-launch ACK (`"Async agent launched successfully … working in the background …"`) is the **uniform, authoritative** background signal — present in both the correctly-flagged and the mis-flagged cases observed. Trusting it fixes every downstream consumer at the source.

## What changed

`telegram-plugin/hooks/subagent-tracker-posttool.mjs`:
- New `toolResponseText()` + `isAsyncLaunchAck()` — detect the launch ACK, anchored on the specific `"async agent launched"` phrase with a structural backstop (`"working in the background"` + an `agentid` token).
- `updateRow()` gains an `asyncLaunch` param + `PROMOTE_BACKGROUND_SQL`. When the DB row reads `background=0` **but** the ACK is present, promote `0→1`, bump activity, and **do not** terminalize (the worker is still running). Union design — a correctly-flagged `background=1` row is never downgraded.
- `main()` suppresses the foreground handback nudge when the ACK is detected (the DB flag may still read 0 on this tick; the ACK is the reliable tell).

Self-healing: `onProgress` re-reads `background` per tick, so the corrected row surfaces the worker card on the next tick — no parent-turn coupling.

## Test plan

- [x] `bun test telegram-plugin/tests/subagent-tracker-hooks.test.ts` — 12/12 pass (+2 new)
  - promotes a mis-recorded foreground row to background from the launch ACK (asserts `background 0→1`, `status=running`, `ended_at=null`, no nudge)
  - still terminalizes a genuine foreground completion whose report mentions "background" (no false promote)
- [x] `tsc --noEmit` clean
- [x] `node scripts/check-plugin-references.mjs` clean

## Risk

Low and surgical. The promote path only fires on the specific async-launch phrasing; a genuine foreground report still terminalizes. No schema change. Worst case on a major ACK-wording drift: degrades to the pretool flag (= today's behaviour), never worse. Fixes the named foreground-blindspot regression class without touching the gateway.

Serves JTBD: live visibility of agent work surfaced to Telegram (the worker-activity-feed). Relates to the `project_foreground_nesting_defeated_by_ack_first` / `project_visibility_parity_goal` workstream.

### DIFF
--- a/telegram-plugin/hooks/subagent-tracker-posttool.mjs
+++ b/telegram-plugin/hooks/subagent-tracker-posttool.mjs
@@ -156,6 +156,53 @@ function extractResultSummary(toolResponse) {
   return str.slice(0, 200) || null
 }
 
+/**
+ * Extract the full text of a PostToolUse tool_response (untruncated).
+ * Mirrors extractResultSummary's shape handling but returns the whole
+ * string so callers can pattern-match on it.
+ */
+function toolResponseText(toolResponse) {
+  if (!toolResponse) return ''
+  if (Array.isArray(toolResponse.content)) {
+    return toolResponse.content
+      .filter((c) => c && typeof c === 'object' && c.type === 'text' && typeof c.text === 'string')
+      .map((c) => c.text)
+      .join('\n')
+  }
+  if (typeof toolResponse.result === 'string') return toolResponse.result
+  if (typeof toolResponse.output === 'string') return toolResponse.output
+  if (typeof toolResponse === 'string') return toolResponse
+  return ''
+}
+
+/**
+ * Detect Claude Code's async-launch ACK in a PostToolUse tool_response.
+ *
+ * A `run_in_background` Agent/Task returns IMMEDIATELY with an
+ * acknowledgement ("Async agent launched successfully … The agent is working
+ * in the background …"), NOT the sub-agent's final result. This ACK is the
+ * authoritative, uniform signal that the dispatch was a background one — it is
+ * present even when Claude Code omits `run_in_background` from the tool_input
+ * the PREtool hook sees (observed on claude-code 2.1.159: a worker whose
+ * tool_input lacked the flag still returned this ACK and ran ~3 min past the
+ * parent turn, so the pretool recorded background=0 and the worker card never
+ * fired). We therefore trust this ACK over the pretool's input-derived flag.
+ *
+ * Anchored on the specific "async agent launched" phrase (a foreground
+ * sub-agent's final report is extremely unlikely to contain it), with a
+ * structural backstop ("working in the background" + an agentId token) in case
+ * the launch-verb wording drifts. A major wording change degrades to the
+ * pretool flag — still correct whenever the model DID pass run_in_background,
+ * never worse than before.
+ */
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

Verified on the real PR head (12/12 tests pass locally under bun).

- `isAsyncLaunchAck()` (subagent-tracker-posttool.mjs:198) is correctly anchored: primary match on the literal `async agent launched` phrase, structural backstop on `working in the background` + `agentid`. The false-promote guard test ("runs as a background job") confirms a generic foreground report mentioning "background" does not trip either branch.
- `PROMOTE_BACKGROUND_SQL` (subagent-tracker-posttool.mjs:254) sets `background=1`, COALESCEs result_summary, bumps last_activity, and is guarded by `status NOT IN ('completed','failed')` — it does NOT write status/ended_at, so the still-running worker is not terminalized. Correct.
- Dispatch order in `updateRow` (subagent-tracker-posttool.mjs:282-290) is union-safe: `isBackground` → `asyncLaunch` promote → foreground. A correctly-flagged `background=1` row hits the first branch and is never downgraded.
- Nudge gate (subagent-tracker-posttool.mjs:441-444) adds `&& !asyncLaunch` ahead of the existing `readBackgroundFlagSync === 0` check — necessary because the DB flag still reads 0 on this tick (promote lands via updateRow on the same event but the sync read at :443 races it). The ACK is the reliable tell. Foreground handback is suppressed; gateway subagent-watcher drives the background handback.
- Self-heal path holds: `resolveWorkerFeedDispatch` (gateway/worker-feed-dispatch.ts:34) reads `sub?.background ?? false` from a per-tick fresh row, so the corrected flag surfaces the worker card next tick. No gateway change needed.

Additive and fail-safe: a major ACK-wording drift degrades to the pretool flag (today's behaviour), never worse. No schema change.

Nit (non-blocking): if a foreground sub-agent's final report literally quotes `async agent launched`, it would be falsely promoted and never terminalized by this hook — but recordSubagentEnd (turn_end JSONL) remains the authoritative end-of-life signal, so the row still closes. Acceptably rare; worth a watch if such reports appear.

PR #2076, opened 2026-06-02
fix(telegram): backfill subagent parent_turn_key from started_at at jsonl-link time

mekenthompson:
## Summary

A sub-agent registry row's `parent_turn_key` was **always NULL** in production. The PreToolUse hook read `event.turn_id`, but Claude Code's hook payload carries only its own **session id**, never the Telegram **turn_key** (`chat_id:msg_id`) the gateway keys turns on. This PR resolves `parent_turn_key` gateway-side instead.

## Why

`parent_turn_key` is the only link from a sub-agent back to the conversation it was dispatched from. With it NULL:

1. **Worker card / handback route to the operator DM**, not the originating group/forum-topic — `resolveSubagentOriginChat()` (`gateway.ts:1141`) bails on `sub?.parent_turn_key == null`. (This is the residual gap noted when #2075 landed: the background worker card now fires, but lands in DM.)
2. **`resolveCallingSubagent()`'s turn-scoped heuristic** (`resolve-calling-subagent.ts:63`, `WHERE parent_turn_key = ?`) can never see the row, weakening `progress_update` attribution.

The hook fundamentally can't know the turn_key — it's a Telegram-side construct the gateway mints. So resolve it where the gateway *does* know: at jsonl-link time.

## What changed

`telegram-plugin/subagent-watcher.ts` — `backfillJsonlAgentId` (now exported for testing): after linking the JSONL stem to the row, if `parent_turn_key IS NULL`, resolve it from the turn whose `[started_at, ended_at]` window **contained the sub-agent's `started_at`** and write it.

- **Keying on the historical `started_at`, not "the turn active now", is the crux.** A background worker outlives its parent turn (the clerk case — ran ~3 min past turn-end), so a "current turn" lookup would miss or mis-attribute. The containment match still finds the now-ended parent turn. A test pins exactly this (a *later* turn is active at link time; the backfill must still pick the earlier containing one).
- Turns are processed serially per agent → at most one window contains a given instant; `ORDER BY started_at DESC LIMIT 1` is a defensive tie-break.
- **Best-effort + idempotent:** never overwrites an existing value, leaves NULL when no window matches, wrapped in try/catch so it never throws out of the watcher poll loop.

`telegram-plugin/hooks/subagent-tracker-pretool.mjs` — writes `parent_turn_key=NULL` explicitly, dropping the dead `event.turn_id` read. Prevents a future CLI that populates `turn_id` with a non-turn_key value from writing a bogus key that would defeat the backfill's `IS NULL` guard.

## Test plan

- [x] `bun test telegram-plugin/tests/subagent-watcher-parent-turn-key.test.ts` — 4/4 (new file):
  - resolves from the containing turn window
  - resolves to the parent turn **even after it ended** (bg worker outlives turn; a later turn is active)
  - does NOT overwrite an already-populated `parent_turn_key`
  - leaves NULL when no window contains the dispatch
- [x] Regression: `subagent-watcher` / `subagent-tracker-hooks` / `subagent-registry-bugs` / `resolve-calling-subagent` / `enoent-deregister` — 59/59 pass
- [x] `tsc --noEmit` clean · `check-plugin-references.mjs` clean

## Risk

Low, surgical, additive. No schema change (the column has existed). Worst case (clock skew / overlapping turns) is the same NULL we have today → DM fallback. Follows #2075 (background-flag fix) — together they make the background worker card both *fire* and *land in the right topic*. Serves the worker-activity-feed visibility JTBD.

### DIFF
--- a/telegram-plugin/hooks/subagent-tracker-pretool.mjs
+++ b/telegram-plugin/hooks/subagent-tracker-pretool.mjs
@@ -263,10 +263,11 @@ function main() {
       id: event.tool_use_id ?? null,
       parentSessionId: event.session_id ?? null,
       // parent_turn_key is intentionally NULL here. Claude Code's PreToolUse
-      // payload carries its own session id, not the Telegram turn_key
-      // (chat_id:msg_id) the gateway keys turns on — `event.turn_id` is always
-      // undefined, and even if a future CLI populated it, it would not match a
-      // `turns.turn_key`. The gateway resolves parent_turn_key from the
+      // payload carries its own session id, not the gateway-minted Telegram
+      // turn_key (a chat+topic+turn key) the `turns` table is keyed on —
+      // `event.turn_id` is always undefined, and even if a future CLI
+      // populated it, it would not match a `turns.turn_key`. The gateway
+      // resolves parent_turn_key from the
       // sub-agent's started_at at jsonl-link time (subagent-watcher.ts
       // backfillJsonlAgentId), which works even after the parent turn ends.
       // Writing a bogus value here would defeat that backfill's
--- a/telegram-plugin/hooks/subagent-tracker-pretool.mjs
+++ b/telegram-plugin/hooks/subagent-tracker-pretool.mjs
@@ -262,7 +262,16 @@ function main() {
     {
       id: event.tool_use_id ?? null,
       parentSessionId: event.session_id ?? null,
-      parentTurnKey: event.turn_id ?? null,
+      // parent_turn_key is intentionally NULL here. Claude Code's PreToolUse
+      // payload carries its own session id, not the Telegram turn_key
+      // (chat_id:msg_id) the gateway keys turns on — `event.turn_id` is always
+      // undefined, and even if a future CLI populated it, it would not match a
+      // `turns.turn_key`. The gateway resolves parent_turn_key from the
+      // sub-agent's started_at at jsonl-link time (subagent-watcher.ts
+      // backfillJsonlAgentId), which works even after the parent turn ends.
+      // Writing a bogus value here would defeat that backfill's
+      // `parent_turn_key IS NULL` guard.
+      parentTurnKey: null,
       agentType: input.subagent_type ?? null,
       description: input.description ?? null,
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

Verified on the real PR head (4/4 new tests + imports resolve under bun).

- The containment query (subagent-watcher.ts:573) is correct: `started_at <= ? AND (ended_at IS NULL OR ended_at >= ?)`, both placeholders bound to the sub-agent's own `started_at`. Keying on the historical dispatch instant (not the turn active at link time) is what makes it correct for a background worker that outlives its parent turn — pinned by the 'resolves even after it ended' test (later turn active, backfill still picks the earlier containing one).
- Idempotency guard holds: the JS `linkedRow.parent_turn_key == null` check (subagent-watcher.ts:566) gates the write, so a pre-populated key is never overwritten — confirmed by the 'does NOT overwrite' test. The no-matching-window case leaves NULL (today's behaviour), confirmed by the 'leaves NULL' test.
- Best-effort containment: the whole block is wrapped in try/catch (subagent-watcher.ts:597) and only logs on failure, so it never throws out of the watcher poll loop.
- The fix lands exactly where the consumer reads: `resolveSubagentOriginChat` (gateway/gateway.ts:1141) bails on `sub?.parent_turn_key == null` → DM fallback. Backfilling the key is what routes the worker card/handback to the originating chat/topic. Correctly pairs with #2075 (the flag fix that makes the card fire at all).
- pretool now writes `parent_turn_key: null` explicitly (subagent-tracker-pretool.mjs:265), dropping the dead `event.turn_id` read — Claude Code's payload never carries the Telegram turn_key, so the old read was always undefined. Writing an explicit NULL also protects the backfill's IS-NULL guard against a future CLI that populates turn_id with a non-turn_key value.

Additive, no schema change, worst case is the same NULL/DM fallback as today.

Nit (non-blocking): the 'serial turns → at most one containing window' assumption behind `ORDER BY started_at DESC LIMIT 1` holds per-agent; if turns could ever overlap for one agent the tie-break silently picks the latest-started container. Not reachable today (one active turn per agent process); worth a watch if turn concurrency is ever introduced.

PR #2077, opened 2026-06-02
feat(supergroup): zero-config easy-mode defaults

mekenthompson:
## Summary
Make the **supergroup-owned topology zero-config easy mode** — the direction you asked for after the marko setup needed `access.json` surgery. Point an agent at a supergroup and it answers **everywhere, all topics, DMs intact**, with no hand-editing.

Reviewed against the contract (RFC: `docs/rfcs/supergroup-easy-defaults.md`): the current state *fails* principles 1 + 2 by their own examples (the silent `group_unknown` drop is verbatim the P1 ❌ example; `default_topic_id`-required + manual group registration is the P2 ❌). Serves outcome 1 ("add a specialist in ten lines of YAML") + the `extend-without-forking` JTBD. Leash (outcome 2) unchanged — this only affects which messages the agent *reads* in a room it owns; actions still gate through Allow/Deny + `allowFrom`.

## What's in this PR (phase 1 — the zero-config core)
1. **`reconcileConfiguredGroup`** (scaffold): idempotently merges the agent's configured supergroup (`channels.telegram.chat_id`) into `access.json` on every reconcile. **Strictly additive** — only adds the group if absent; preserves `allowFrom`, pairings, other groups, and operator policy overrides. Closes the `writeIfMissing` gap that made a post-scaffold supergroup never register (the marko `group_unknown` silent-drop). Smart default: `requireMention: false`, all topics.
2. **`default_topic_id` → General (1) default** when `chat_id` is set alone (was a hard error).

Net easy path:
```yaml
channels: { telegram: { chat_id: "-1003831053471" } }   # answers everywhere, DMs still work
```

## Phase 2 (follow-up, in the RFC)
- `switchroom agent add --topology supergroup --chat-id` + `agent set-topology` (no YAML editing).
- In-Telegram "enable this group for <agent>?" one-tap nudge → completes the P1 silent-drop fix end-to-end.

## Test plan
- [x] `tests/scaffold.reconcile-group.test.ts` (7) — additive/non-destructive merge, preserves allowFrom/pairings/other-groups, **never overwrites** an existing group's policy, idempotent, `dm_only`/no-forum/absent-file no-ops.
- [x] `src/config/schema.test.ts` — General default; explicit value preserved; `default_topic_id` without `chat_id` still rejected. Updated the one existing test that pinned the old require-behavior.
- [x] Full config + scaffold cluster (532) green; `tsc` + `npm run lint` clean.

## Risk
Low. The reconcile is strictly additive + idempotent (unit-tested for preservation/no-overwrite); never removes or rewrites existing access. The schema change relaxes a validation (required→defaulted) — backward-compatible (existing configs with explicit `default_topic_id` unchanged).

### DIFF
--- a/docs/rfcs/supergroup-easy-defaults.md
+++ b/docs/rfcs/supergroup-easy-defaults.md
@@ -1,0 +1,107 @@
+# RFC: Supergroup easy-mode defaults (zero-config "agent owns a working room")
+
+**Status:** Phase 1 implemented (this PR)
+**Owner:** Ken Thompson
+**Date:** 2026-06-02
+**Related:** [supergroup-mode.md](./supergroup-mode.md) (the topology this builds on),
+`reference/principles.md`, `reference/extend-without-forking.md`,
+`reference/talk-to-agents-from-anywhere.md`.
+
+## 1. Summary
+
+Make the supergroup-owned topology **zero-config easy mode**: an operator
+points an agent at a Telegram supergroup and it Just Works — responds to
+every message, in every topic (including ones created later), with DMs
+still served — without hand-editing `access.json` or picking a default
+topic.
+
+This came out of enabling marko in the "Panorama Marketing" group, which
+required: discovering the `group_unknown` drop, knowing `access.json` is
+`writeIfMissing`, hand-merging the group, and hand-picking `default_topic_id`.
+None of that is something a user should have to do.
+
+## 2. Which outcome / JTBD it serves
+
+- **Outcome 1 — "a standing team that knows you":** *"Add a specialist in
+  ten lines of YAML. You don't fork the product."* A supergroup specialist
+  should be the same: a couple of config lines, not surgery.
+- **JTBD `extend-without-forking`:** *"adds a new agent… by configuring it,
+  not by editing the product. The scaffolding does the boilerplate."*
+- **JTBD `talk-to-agents-from-anywhere`:** a dedicated working room per
+  specialist is a natural surface.
+
+**Leash check (outcome 2):** unchanged. "Responds to everything" only
+affects which messages the agent *reads* in a group it *owns*; **actions**
+(tools, credentials, sends) still go through Allow/Deny cards, and
+`allowFrom` still gates *who* can talk to it. No self-elevation, no new
+blast radius.
+
+## 3. The three principle checks (the verdict rule)
+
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

Verified by static read of the real PR head (deps not installed locally, so tests not run here — verdict rests on code inspection, not a re-run of the author's suite).

- `reconcileConfiguredGroup` (src/agents/scaffold.ts:5453) is strictly additive and non-destructive as claimed: the `groups[forumChatId] !== undefined` guard (scaffold.ts:5471) preserves any existing per-group policy (operator's requireMention/topic_deny survive), and the new entry is written without touching allowFrom/pairings/other groups. Atomic write via tmp + renameSync (scaffold.ts:5475-5477) avoids a torn access.json.
- Shape parity with the fresh-build path is correct: the reconcile writes `{ requireMention: false, allowFrom }` (scaffold.ts:5474), byte-identical to what `buildAccessJson` emits for the same group (scaffold.ts:5519). Both read the same `telegramConfig.forum_chat_id` (scaffold.ts:5456 vs 5516), so the per-agent `channels.telegram.chat_id` override resolves identically in both paths — no divergence between a freshly-scaffolded agent and a reconciled one.
- Guards are complete: no-op on absent file (scaffold.ts:5460), dm_only or empty/"0" sentinel forum chat (scaffold.ts:5462), and malformed JSON via try/catch (scaffold.ts:5468). Matches the #1002 sentinel handling in buildAccessJson.
- Schema change is sound: the require-rule is replaced by a `.transform` that injects `default_topic_id: 1` only when chat_id is set and the field is omitted (src/config/schema.ts:870), running after the superRefine that still rejects `default_topic_id` without `chat_id` (schema.ts:849). Backward-compatible — an explicit default_topic_id is preserved. The one pre-existing test that pinned the old require-error was correctly flipped (schema.test.ts:254).

Additive, backward-compatible, no schema-column change.

Nit (non-blocking): the registered group's `allowFrom` is a snapshot of the top-level `allowFrom` at reconcile time; if an operator later widens the top-level allowlist, the group copy goes stale until the next group-policy edit. This mirrors buildAccessJson's existing behaviour exactly, so it's consistent, not a regression — worth a follow-up only if group/top-level allowFrom drift becomes a real support case.

Watch: this is phase 1 — the P1 "silent group_unknown drop" is only half-closed until the phase-2 in-Telegram enable nudge lands (correctly scoped out and documented in the RFC).

PR #2080, opened 2026-06-02
docs(scheduling): document per-cron topic targeting

mekenthompson:
## Summary
Document the per-cron `topic:` field for supergroup-owned agents in `docs/scheduling.md`. The feature is wired end-to-end (schema → `resolveChannelTarget`/`resolveEntryThreadId` → `topic-router`) and tested, but was only documented in the schema `.describe()` + the supergroup-mode RFC — so it was effectively undiscoverable. Principle 1 ("if they need the docs, we've failed") gap for a shipped capability; this closes it.

## What's added
- A `topic` row in the Configuration table.
- A **Targeting a forum topic** subsection: alias-vs-numeric forms, the `topic_aliases` map, config-load validation, resolution precedence (numeric → alias → `default_topic_id`), how to find a `thread_id`, and a worked marko example mirroring the live config.

## Why now
Follows the v0.14.38 supergroup zero-config easy-mode release (#2077). With supergroup ownership now zero-config, scheduling into a named topic is the natural next operator action — it needs to be in the docs.

## Test plan
- [x] Docs-only change; resolution path verified live against marko's real config (meta→3, crm→4, general→1, numeric passthrough, default + unknown-alias fallback all correct).

## Risk
None — documentation only.

### DIFF
--- a/docs/scheduling.md
+++ b/docs/scheduling.md
@@ -119,6 +119,7 @@ The agent relays this to the user in plain language. This satisfies the *survive
 | `prompt` | Yes | — | The prompt that becomes the synthesized turn's text |
 | `model` | No | — | **Deprecated / ignored.** Pre-v0.8 the singleton scheduler ran each task as an isolated `claude -p` and could set `--model` per task. Post cron-fold-in the fire runs in the agent's existing session, so it always uses the **agent's** configured model. Accepted only so old configs keep validating; set the model at the agent level. |
 | `secrets` | No | `[]` | Vault keys this task may read. Operator-config only — rejected on agent-authored overlays. See [configuration.md#vault-broker-linux-only](configuration.md#vault-broker-linux-only) |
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

- docs/scheduling.md:9 — new `topic` config-table row accurately matches the schema: supergroup-only, alias resolved against `channels.telegram.topic_aliases` or raw numeric `message_thread_id`, falls back to `default_topic_id`. Verified field names against src/telegram/topic-router.ts:31 and src/config/schema.ts:93.
- docs/scheduling.md:63-66 — documented resolution precedence (numeric `topic:` → alias lookup → `default_topic_id`) is byte-accurate to the `case "cron"` branch in resolveOutboundTopic (src/telegram/topic-router.ts): numeric returned directly, string resolved via aliasToId, else default.
- docs/scheduling.md:27-30 — claim that a bad alias fails `switchroom apply` at config load is correct and is, if anything, understated: validateAllCronTopicAliases throws ConfigError at load (src/config/loader.ts:240-270), not just "should reject." Nit (non-blocking): line 65-66 hedges "config-load validation should reject it before then" — the loader actually does reject hard, so consider tightening "should" to "does" to avoid implying it is best-effort.
- docs/scheduling.md:32-55 — worked marko example is internally consistent (chat_id set → supergroup mode; default_topic_id 1; meta/crm aliases; numeric passthrough; untargeted entry → General) and matches the PR-body live-config verification.

Rationale: documentation-only, no source touched; every behavioral claim cross-checked against the real implementation on main and found accurate. Closes a genuine discoverability gap for a shipped, tested capability.

Issue #2084, opened 2026-06-02
#2075 background-worker promotion depends on a literal Claude Code ACK string → silent regression on CLI bump


mekenthompson:
## Summary

The background-flag promotion shipped in #2075 (v0.14.37) detects an async sub-agent launch by matching Claude Code's **literal** async-launch ACK text ("Async agent launched successfully … working in the background … agentId: …") in the PostToolUse hook. The claude CLI is currently pinned (2.1.156), so it's stable today — but **any pin bump that rewords that ACK silently breaks promotion**: background workers regress to being recorded `background=0`, rendered as foreground, with the foreground handback nudge no longer suppressed. There is no test asserting against the live CLI string, so the regression would ship invisibly.

## Where

- `telegram-plugin/hooks/subagent-tracker-posttool.mjs` — the ACK-string match that promotes `background 0→1`, takes the non-terminalizing bg path, and suppresses the foreground handback nudge.

## Why it's fragile

- The trigger is an English sentence emitted by a third-party CLI we pin but periodically bump (recent bumps: 2.1.157–2.1.159 evaluated; 2.1.156 hard-pinned for the 400 fix). A wording/format change there is a *runtime* behavior change here with **no compile-time or test signal**.
- Symptom if it breaks: background workers show as foreground again (the exact #709/#776/#782/#788 class the work was meant to fix), plus a spurious handback nudge.

## Questions for the investigating agent

1. Is there a **structured** signal for async launch (tool result metadata, a field, an event) we can key on instead of prose? Prefer that over string-matching.
2. If prose matching is unavoidable, add a guard test that fails loudly when the pinned CLI's ACK text no longer matches the expected pattern (assert against a captured fixture tied to the pinned version), so a CLI bump surfaces the break in CI rather than in production.
3. Should the matcher be loosened to a tolerant regex (e.g. key on "background" + "agentId") to survive minor rewording, with the guard test pinning the contract?

## Related

- #2081 — companion v0.14.37 sub-agent-visibility audit findings.

## Acceptance

- Background-launch detection no longer silently breaks on a Claude Code wording change — either it keys on a structured field, or a CI test fails when the pinned CLI's ACK text drifts from the matcher.

_Filed from the v0.14.37 rollout audit._

mekenthompson:
TRIAGE (confirmed against main)

[...]

Issue #2083, opened 2026-06-02
parent_turn_key backfill only runs at jsonl-link time → sub-agents that never link are never attributed (~8%)


mekenthompson:
## Summary

The `parent_turn_key` backfill lives **inside** `backfillJsonlAgentId` (`telegram-plugin/subagent-watcher.ts:438`), so it only fires when a sub-agent's JSONL gets linked. Sub-agents that error/exit before the watcher links their JSONL (or whose JSONL never appears) keep `parent_turn_key = NULL` forever — even post-v0.14.37 — so their worker card / handback falls back to the operator DM instead of the originating chat/topic.

## Evidence (fleet registry, week of 2026-05-26 → 06-02)

Sub-agents with no `jsonl_agent_id` (never linked ⇒ backfill never ran):

| agent | unlinked / total |
|---|---|
| clerk | 2 / 34 |
| klanker | 4 / 48 |
| marko | 4 / 13 |
| finn | 1 / 6 |

≈ **8%** fleet-wide that the fix structurally cannot reach. Separately, one `marko` sub-agent (`toolu_01FwYnipUj`, 2026-06-01 18:16:28) linked but matched **0** turn windows ⇒ `parent_turn_key` stayed NULL ⇒ DM fallback (a distinct "no containing window" failure mode worth handling too).

## Where

- `backfillJsonlAgentId` — `telegram-plugin/subagent-watcher.ts:438` (jsonl link) with the parent_turn_key block at ~562–600. Both are gated on the link succeeding.
- Consumer fallback — `resolveSubagentOriginChat()` `gateway.ts:1135` returns null ⇒ card routes to owner DM.

## Questions for the investigating agent

1. Should `parent_turn_key` be stamped at **dispatch** (PreToolUse, `subagent-tracker-pretool.mjs`) independent of JSONL linking? (This also resolves #2081's preferred fix — gateway stamps from live `currentTurn`.)
2. For linked-but-no-window sub-agents, is a nearest-turn fallback (or carrying chat/thread on the row) preferable to a silent DM fallback?
3. Why do ~8% of sub-agents never link JSONL — expected (fast failures) or a watcher gap?

## Related

- #2081 — parent_turn_key overlapping-window mis-attribution (stamping at dispatch fixes both).

## Acceptance

- A sub-agent that never links JSONL still gets a correct `parent_turn_key` (or an explicit, intentional fallback).
- Measured unlinked-and-unattributed rate drops materially from the ~8% baseline.

[...]

Issue #2082, opened 2026-06-02
Turn ended_at is set by a batch sweep, not at turn end → turn windows overlap for hours


mekenthompson:
## Summary

`turns.ended_at` is not written when a turn actually finishes — it is stamped **in batches by a sweep**. As a result turn `[started_at, ended_at]` windows stay "open" in recorded time for hours and overlap heavily, even on single-chat agents. This is the **root-cause amplifier** behind the `parent_turn_key` mis-attribution in #2081, and it also produces stale never-ended turns.

## Evidence (fleet registry, week of 2026-05-26 → 06-02)

- **Batch close, proven:** two `marko` turns in *different chats* both received `ended_at` at the **identical millisecond** `2026-06-01 16:18:16`. Distinct turns finishing at the exact same ms ⇒ a sweep stamped them together, not per-turn completion.
- **Windows stay open for hours:** `clerk` has 22 turns recorded as lasting >1h (max **12.5h**); `klanker` 12 (max ~8h); `marko` 7 (max ~8h). These are almost certainly lazy/batch closes, not 12h of real work.
- **Never-ended turns:** `marko` currently has **3 turns with `ended_at IS NULL`** (started 2026-06-02 13:01–13:17, no `ended_via`). An open window is `[start, ∞)` — it contains *every* future instant in *every* chat until something closes it, so each is a standing mis-attribution magnet (feeds #2081).

## Where to look

- Turn start — `recordTurnStart`, `telegram-plugin/gateway/gateway.ts:7925`+ (writes `started_at`, leaves `ended_at` NULL).
- Turn end paths — `recordTurnEnd(stop)` `gateway.ts:~8986`, `recordTurnEnd(timeout)` `gateway.ts:~3964`, shutdown stamping `gateway.ts:~17490`.
- Dangling-turn / disconnect sweep — recent commit `21fe8352` ("disconnect dangling-turn sweep"). marko's 3 open turns slipped past it — understand why.
- Registry reaper (`runHistoryReaperNow`, `gateway.ts:~1180`) only deletes old rows; it does not close open turns.

## Questions for the investigating agent

1. Which code path is the intended per-turn `ended_at` writer, and why does it land as a batch (same-ms closes across chats)? Is it the silence/idle sweep rather than turn completion?
2. Why do turns reach `ended_at IS NULL` indefinitely (3 live on marko)? Crash? interrupt without stamp? topic-concurrency confusing the single-`currentTurn` model?
[...]

PR #2085, opened 2026-06-02
fix(telegram): stamp parent_turn_key at dispatch; harden async-launch detection

### LINKED ISSUE #2081: parent_turn_key backfill mis-attributes under overlapping turn windows → wrong-topic/chat card pin (supergroups)
## Summary

The `parent_turn_key` backfill shipped in #2076 (v0.14.37) resolves a sub-agent's parent turn **purely by time-window containment**, trusting that "at most one turn window contains a given instant" (serial-turn invariant). **That assumption is empirically false in production.** Sub-agents routinely fall inside multiple overlapping turn windows, so the resolver picks the wrong turn — and in supergroups that means the live worker card / handback pins to the **wrong forum topic or wrong chat**.

This is the structural blind spot behind the v0.14.37 ship: the fix is correct in the happy path but degrades exactly where it's needed most (concurrent topics).

## Where

- Backfill query — `telegram-plugin/subagent-watcher.ts:587` (inside `backfillJsonlAgentId`, parent_turn_key block ~562–600):
  ```sql
  SELECT turn_key FROM turns
  WHERE started_at <= ? AND (ended_at IS NULL OR ended_at >= ?)
  ORDER BY started_at DESC LIMIT 1
  ```
  No `chat_id`/`thread_id` filter. Binds the sub-agent's `started_at` for both `?`. The code comment asserts "at most one window contains a given instant" — that is the bug.
- Consumer — `resolveSubagentOriginChat()` `telegram-plugin/gateway/gateway.ts:1135` reads `chat_id` **and** `thread_id` from the turn `parent_turn_key` points at and routes the card there. So a wrong `parent_turn_key` ⇒ wrong (chat, topic).
- turn_key minting — `gateway.ts:7925` via `chatKeyWithSuffix` (`telegram-plugin/gateway/chat-key.ts:51`), format `chatId:thread:startedAt`.

## Production evidence (fleet registry, week of 2026-05-26 → 06-02)

Sub-agents whose `started_at` falls inside **>1** turn window (should be 1):

| agent | max overlapping windows |
|---|---|
| clerk | 7 |
| test-harness | 6 |
| klanker | 5 |
| gymbro | 3 |

**Cross-chat mis-attribution, confirmed:** `marko` sub-agents `toolu_013SyVcwS3` / `toolu_017ssMBvy5` (2026-06-01 16:15–16:16) each had **2 candidate chat:thread windows** — two different chats were simultaneously "open" (`<dmA>` started 16:14, `<dmB>` started 15:34). The shipped query force-resolved both to `<dmA>` by recency. If they were dispatched from `<dmB>`, their cards mis-routed.

For single-chat agents (clerk, klanker) the overlaps are same-chat ⇒ low harm (still right topic). The harmful case is multi-chat / multi-topic agents.

## Why supergroups are the worst case (by design)

A supergroup multiplexes many forum topics under **one** `chat_id`, each generating concurrent turns — so overlapping windows are the *normal* case, not an edge. `marko` serves supergroup `<supergroup>` across topics **3, 4, 56, General**; topics **3 and 4 opened turns at the identical second (2026-06-02 11:39:28)**. `ziggy` also runs in forum topics. As supergroup-mode (see supergroup RFC — `docs/rfcs/supergroup-mode.md`) rolls out, concurrent topics are guaranteed, so cross-topic mis-pin becomes routine.

## Channels (broadcast) — NOT affected

Confirmed a non-issue, by absence of handling rather than correctness:
- Inbound gate `gateway.ts:~9190` drops any chatType not in {private, group, supergroup} as `unknown_chat_type` (~9233).
- `channel_post` is in `allowed_updates` (`gateway.ts:18660`) but reaches no handler — the main inbound is `bot.on('message:*')`, which grammy does not route channel posts to.
- ⇒ a broadcast channel never creates a turn ⇒ never originates a sub-agent ⇒ nothing to attribute. A channel's linked discussion group surfaces as a **supergroup** and collapses to the case above.

## Root cause of the overlap

`ended_at` is **not** written when a turn finishes — it is stamped in **batches by a sweep**. Proof: two `marko` turns in *different chats* both got `ended_at` at the identical millisecond `2026-06-01 16:18:16`. So turns stay "open" in recorded time for hours (clerk has 22 turns recorded >1h, max 12.5h) and their windows overlap continuously. Tracked separately for the turn-lifecycle fix — see linked issue on never-ended/batch-swept `ended_at`.

## Recommended direction (for the investigating agent to validate)

- `session_id` cannot disambiguate: `parent_session_id` (`subagent-tracker-pretool.mjs:264`) is the agent's single long-lived claude session — constant across all its turns. That is *why* the original design used a time window. The fix must carry **chat/thread**, not session.
- Preferred: have the gateway stamp `parent_turn_key` from its live `currentTurn` atom **at dispatch (PreToolUse)** — at that instant `currentTurn` *is* the parent turn (chat+thread known: `gateway.ts:4477` has `currentTurn.sessionChatId/sessionThreadId`). Fall back to the window match only when `currentTurn` is unknown (e.g. a background worker that links after the parent moved on).
- Alternative/min: constrain the window query to the sub-agent's own (chat, thread) — requires the row to carry chat/thread (gateway-injected), then prefer the matching window.

## Test gap

The shipped unit test `telegram-plugin/tests/subagent-watcher-parent-turn-key.test.ts` only sets up **non-overlapping** turns and bakes the wrong "at most one window" assumption into a comment — it cannot catch this. A regression test must place **two windows containing one instant** (same chat different topic; and different chat) and assert the correct pick.

## Repro / audit query

```sql
-- count overlapping windows per sub-agent (run against a per-agent registry.db)
SELECT s.id, COUNT(*) AS windows
FROM subagents s JOIN turns t
  ON t.started_at <= s.started_at
 AND (t.ended_at IS NULL OR t.ended_at >= s.started_at)
GROUP BY s.id HAVING COUNT(*) > 1 ORDER BY windows DESC;
```

## Acceptance

- A sub-agent dispatched from topic A of a supergroup resolves `parent_turn_key` to topic A's turn even when topics B/C have concurrent/open windows.
- Cross-chat overlap never resolves to the wrong chat.
- Regression test with overlapping windows (same-chat-different-topic + cross-chat) is green.
- Canary on `marko` (the only fleet agent exercising real forum topics).

_Filed from an audit of the v0.14.37 rollout; agent names (marko/clerk/klanker/ziggy/gymbro/test-harness) are synthetic fleet fixtures. Real Telegram IDs redacted (`<supergroup>`, `<dmA>`, `<dmB>`)._

mekenthompson:
## Summary

Sub-agent attribution was reconstructed *after the fact* from a `started_at` time-window match at jsonl-link time. That match assumes "at most one turn window contains a given instant" — false in supergroups, where forum topics multiplex concurrent turns under one `chat_id` and `ended_at` is unreliable (batch-swept). The result: a worker card could pin to the **wrong topic / wrong conversation** (proven on marko). This PR captures the parent turn at **dispatch** instead, plus hardens background-worker detection.

Resolves #2081, #2083, #2084. Scopes #2082 (see below).

## Why (the JTBD)

Serves **"you hold the leash"** / live-work visibility: a delegated worker's card must appear in the topic it was dispatched from, and a background worker must be recognised as background. Both broke in the supergroup topology that #2077/#2078 just shipped.

### #2081 / #2083 — `parent_turn_key` stamped from the turn-active marker
The PreToolUse hook (`subagent-tracker-pretool.mjs`) now reads `<telegramDir>/turn-active.json` — the gateway's marker for the turn whose tool call is dispatching this sub-agent — and writes `parent_turn_key = marker.turnKey` at INSERT.

- The marker is a sibling of `registry.db` (both under `$TELEGRAM_STATE_DIR`) — **verified same inode in production** (clerk: `27926869`).
- `marker.turnKey` ≡ `turns.turn_key` (both minted by `chatKeyWithSuffix` at turn-start), so `resolveSubagentOriginChat()`'s `getTurnByKey()` finds the exact `(chat_id, thread_id)` → correct topic.
- Captures the **exact active turn** → no time-window reconstruction → no overlap mis-attribution (#2081).
- Written at INSERT, independent of jsonl linking → attributes the ~8% of sub-agents whose JSONL never links (#2083).
- The gateway window-match stays a **best-effort fallback** for the no-active-turn case; its `parent_turn_key IS NULL` guard means the hook's correct value always wins.

### #2084 — async-launch detection hardened
`isAsyncLaunchAck` gains a third, drift-tolerant tier keyed on the functional `agentId: <stem>` token (the most wording-stable part of the ACK) + a launch/background/async/dispatch context word. A claude-code bump that rewords BOTH prose phrases ("async agent launched" **and** "working in the background") no longer silently regresses background-worker promotion. The ACK contract (verified vs the pinned **claude-code 2.1.156**) is now pinned by drift-variant tests. The context-word requirement preserves the existing false-positive guard.

### #2082 — scoped, not force-fixed
The overlap **mis-attribution** #2082 caused is eliminated above (attribution no longer depends on `ended_at`). The residual orphan open turns (`ended_at IS NULL`) are a symptom of the **not-yet-done supergroup `currentTurn → Map` refactor** (RFC-tracked) and **self-heal at boot** via the orphan reaper. A narrow "close older same-thread open turn" fix would touch the `ended_via` enum + resume policy + `recordTurnStart` — load-bearing turn-end/resume/watchdog code — for a self-healing cosmetic issue. Deliberately deferred to that refactor rather than risking a live-fleet regression. Reviewers: challenge this call if you see a provably-safe narrow fix.

## Test plan

- [x] `subagents-bugs.test.ts` — marker-stamp, no-marker→NULL fallback, `turn_id` ignored, malformed-marker→NULL (no crash)
- [x] `subagent-watcher-parent-turn-key.test.ts` — hook-stamped value survives the backfill under overlapping windows; fallback window-match behaviour pinned
- [x] `subagent-tracker-hooks.test.ts` — async-ACK drift variant promotes; agentId-without-context-word does NOT promote (false-positive guard)
- [x] Full plugin bun surface (exact CI command): **4213 pass / 1 skip / 0 fail / 254 files**
- [x] `tsc --noEmit`, `check-plugin-references`, `check-bun-test-imports` — all clean
- [ ] Canary on test-harness + UAT, then staggered rollout (post-merge)

## Risk

Low. The hook change is additive (reads a best-effort marker; any failure → NULL → existing fallback, never worse than today). The matcher change only *adds* a guarded tier. No schema change, no gateway/turn-lifecycle change. Hot-path cost: one `readFileSync` + `JSON.parse` per Agent/Task dispatch.


mekenthompson:
APPROVE

Verified end-to-end against the real PR head (PR #2085/head), not the description. This is the correct dispatch-time fix for the #2081/#2083/#2084 triad.

- **turnKey parity is exact (the crux).** `telegram-plugin/gateway/gateway.ts:7925` mints `turnKey = chatKeyWithSuffix(ev.chatId, evThreadIdNum, String(startedAt))` ONCE and passes the same variable to both `recordTurnStart` (→ `turns.turn_key`, :7939) and `writeTurnActiveMarker` (:7954). So `marker.turnKey` ≡ `turns.turn_key`, and `resolveSubagentOriginChat()`'s `getTurnByKey()` resolves the exact `(chat_id, thread_id)` → correct topic. The PR's central claim holds in code, not just by assertion.
- **Marker is present at dispatch.** `writeTurnActiveMarker` is synchronous at turn-start (gateway.ts:7954; comment at :7936 confirms it's intentionally sync), and the file lives for the whole in-turn window (`turn-active-marker.ts`: written turn_start, removed turn_complete). The Agent/Task PreToolUse fires inside that window, so the marker exists. Correct by construction — and because the agent executes turns serially under one `currentTurn`, the live marker at the dispatch instant IS the dispatching turn, which is exactly why this beats the overlapping-window reconstruction (#2081).
- **Hook read is in scope and safe.** `subagent-tracker-pretool.mjs:33` imports `readFileSync`; `join` is already in use; `telegramDir = join(agentDir,'telegram')` (:294). `readActiveTurnKey()` (:142-152) is try/catch-wrapped, validates `turnKey` is a non-empty string, returns null otherwise — malformed/missing marker → NULL, never throws, never blocks the tool call.
- **Fallback precedence preserved.** `subagent-watcher.ts:584` still guards the backfill with `linkedRow.parent_turn_key == null`, so the hook-stamped value always wins and the started_at window-match only runs for the no-marker case (#2083's never-link path and dispatch-outside-a-turn). Tests pin both: hook value survives overlapping windows, and the NULL-parent fallback picks the latest-started turn (documented limitation).
- **#2084 tier-3 matcher is sound.** `subagent-tracker-posttool.mjs:95` adds `/agentid:\s*\S/.test(t) && /\b(background|launch|dispatch|async)/.test(t)` as a third tier after the two prose arms. Drift-promote and false-positive-guard tests both pass.
- **Tests green:** ran the three touched files locally on the PR head — `subagents-bugs.test.ts`, `subagent-tracker-hooks.test.ts`, `subagent-watcher-parent-turn-key.test.ts` → 39 pass / 0 fail / 117 assertions.
- **#2082 correctly scoped, not force-fixed.** Attribution no longer depends on `ended_at`, so the overlap mis-attribution is eliminated; the residual orphan-open-turns are deferred to the `currentTurn → Map` refactor and self-heal at boot. Reasonable call — a narrow same-thread-close fix would touch the `ended_via` enum + resume/watchdog couplings for a self-healing cosmetic issue.

Nits (non-blocking, no change required to merge):
- Tier-3 false-positive surface: a genuine foreground report that happens to contain BOTH an `agentId:` token AND an incidental word in {background, launch, dispatch, async} would false-promote to background (status stays `running`, never terminalized → its worker card would linger). Likelihood is low (tiers 1 & 2 must both miss first, and the context-word set is launch-flavoured), and the author reasoned about it explicitly — but worth keeping as a watch item if a future CLI ACK or a chatty worker report trips it.
- `STATE_DIR` (gateway, from `TELEGRAM_STATE_DIR`) and the hook's `join(agentDir,'telegram')` are the same directory in production (verified by inode, and registry.db already relies on it), but that equality is environmental, not code-guaranteed. If they ever diverged the hook would read no marker → NULL → fallback (never worse than today), so it degrades safely; a shared path-constant would remove the latent coupling.

Rationale: additive, well-tested, no schema or turn-lifecycle change; the hot-path cost is one readFileSync + JSON.parse per dispatch; every failure mode degrades to the pre-existing fallback. Correctly fixes the wrong-topic worker-card pin at its root.

PR #2089, opened 2026-06-02
fix(gateway): reliable inbound delivery — deliver-until-acked (drop-wedge)

mekenthompson:
## Summary

Telegram inbounds reach claude as an MCP channel notification the unmodified CLI appends to its composer and auto-submits **only when the composer is empty + idle**. When a message arrives as the prior turn is finalizing, the auto-submit races turn-completion and the text **strands unsubmitted** — claude never starts the turn, the gateway sits "typing…", and the **300s silence-poke drops the message**.

Recurring live on `marko` (supergroup forum topics *and* DMs). The #1556 delivery gate + #2039 composer-clear narrowed the window but couldn't close it, because delivery was never *acknowledged*.

## The fix — a reliable queue

Make delivery **deliver-until-acked**:

- A delivered inbound is acked **only by the `enqueue` session-event** (claude actually started the turn — the one signal it truly picked the message up), **not** by `sendToAgent` returning true.
- Until acked it stays tracked. If it didn't land within 15s it stranded → re-clear the composer + re-deliver. Repeat until acked. **Never drop.**
- Bridge-offline re-sends hand off to the existing persistent offline buffer.

No retry caps, no give-up, no notifications — just a queue that keeps the message until claude takes it. Kill switch `SWITCHROOM_INBOUND_DELIVERY_CONFIRM=0` restores the legacy fire-and-forget path.

## Why (the JTBD)

**"always available / talk to my agents from anywhere"** — an operator message must never be silently swallowed. This was dropping Lisa's messages in marko's CRM (Brevo) topic.

## Changes

- `telegram-plugin/gateway/inbound-delivery-confirm.ts` — new pure module (`createDeliveryQueue` / `trackDelivery` / `ackDelivery` / `sweep` / `forgetDelivery`).
- `gateway.ts` — track on delivery, ack on the `enqueue` handler, a 5s sweep re-delivers strands (re-clear composer + re-send).
- `markClaudeBusyForInbound` now returns its `chatKey` so track/ack correlate.

## Test plan

- [x] `inbound-delivery-confirm.test.ts` — 9 unit tests: acked → never re-delivered (no dup turns); within-timeout left alone; strand → re-delivered + clock reset; **re-delivers forever until acked, never drops**; late ack stops re-delivery; per-key independence (DM + topic); same-key supersede; offline `forgetDelivery`.
- [x] Full plugin bun surface: **4225 pass / 1 skip / 0 fail / 255 files**.
- [x] `tsc --noEmit`, `check-plugin-references`, `check-bun-test-imports`, `check-bot-api-wrapping` — all clean.
- [ ] Canary on test-harness (reproduce the turn-completion race; confirm re-delivery, not a 300s drop) + staggered rollout — post-merge.

## Risk

Low + reversible. Kill switch reverts to current behavior. No schema change, no new persistence (in-memory tracker; the persistent spool + silence-poke remain the deeper net for the rare gateway-crash-mid-strand). Re-delivery only fires when claude is idle-but-un-acked; the `enqueue` ack races are sub-second and a rare duplicate is far less harmful than a dropped message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## unreleased — Reliable inbound delivery: deliver-until-acked
+
+Telegram inbounds reach claude as an MCP channel notification that the
+unmodified CLI appends to its TUI composer and auto-submits only when the
+composer is empty + idle. When a message arrives as the prior turn is
+finalizing, the auto-submit races turn-completion and the text **strands
+unsubmitted** — claude never starts the turn, so the gateway sat
+"typing…" until the 300s silence-poke **dropped the message**. Observed
+recurring on `marko` (supergroup forum topics and DMs alike); the #1556
+delivery gate + #2039 composer-clear narrowed the window but didn't close
+it because delivery was never *acknowledged*.
+
+- **Delivery is now acked by `enqueue`, not by `sendToAgent`.** A
+  delivered inbound is tracked in a small per-key queue until claude
+  actually starts the turn (the `enqueue` session-event — the one signal
+  that claude truly picked the message up). If it isn't acked within 15s
+  it stranded: the gateway re-clears the composer and re-delivers it, and
+  keeps re-delivering until claude acks. The message is **never dropped**.
+  Bridge-offline re-sends hand off to the existing persistent offline
+  buffer. Kill switch: `SWITCHROOM_INBOUND_DELIVERY_CONFIRM=0` restores
+  the legacy fire-and-forget path. New pure module
+  `inbound-delivery-confirm.ts` with full unit coverage.
+
 ## v0.14.39 — Sub-agent attribution stamped at dispatch (#2081, #2083, #2084)
 
 Make a sub-agent's parent turn **ground truth captured at dispatch**
--- a/src/build-info.ts
+++ b/src/build-info.ts
@@ -2,8 +2,8 @@
 // Tracked in git so `tsc --noEmit` works on a fresh clone before `npm run build`.
 // Values are refreshed every time `npm run build` runs.
 
-export const VERSION: string = "0.14.39";
-export const COMMIT_SHA: string | null = "3f4ec740";
-export const COMMIT_DATE: string | null = "2026-06-02T18:35:51+10:00";
+export const VERSION: string = "0.14.38";
+export const COMMIT_SHA: string | null = "dc65a835";
+export const COMMIT_DATE: string | null = "2026-06-02T15:08:56+10:00";
 export const LATEST_PR: number | null = null;
-export const COMMITS_AHEAD_OF_TAG: number | null = 1;
+export const COMMITS_AHEAD_OF_TAG: number | null = 2;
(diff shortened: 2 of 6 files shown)


mekenthompson:
APPROVE

Clean, reversible reliability fix. Verified the integration crux against the head (branch `fix/inbound-submit-wedge`):

- **Key parity is correct** — the make-or-break of this design. `trackDelivery` keys on `busyKey` from `markClaudeBusyForInbound` (`gateway.ts:1331` → `chatKey(m.chatId, tid)`, tid resolved from `threadId ?? meta.message_thread_id`). `ackDelivery` keys on `chatKey(ev.chatId, ev.threadId != null ? Number(ev.threadId) : null)` (`gateway.ts:7942`). Both pass through the same normalization (`chat-key.ts:46-47`: `threadId == null || === 0 ? '_' : String(threadId)`), and `ev.chatId`/`ev.threadId` are the same raw values turn-attribution already uses for reply routing (`sessionChatId: ev.chatId`, `:7909`). So the ack reliably clears the tracked entry — no infinite re-deliver / duplicate-turn loop from a key mismatch.
- **Pure module is correct and total** (`inbound-delivery-confirm.ts`): `sweep` resets `lastAttemptAt` on each returned strand (`:250`) so a re-delivered entry waits a full `timeoutMs` before re-firing; `trackDelivery` overwrites per key (supersede, `:225`), matching the #1556 per-key serialization; `forgetDelivery`/`ackDelivery` both just `delete`. 9/9 unit tests pass locally, including the forever-until-acked invariant and per-key independence.
- **Sweep→redeliver loop is safe** (`gateway.ts:4097-4120`): re-clears composer (best-effort, re-delivers regardless on failure), re-sends; on `sendToAgent` false hands off to `pendingInboundBuffer` + `forgetDelivery` so the offline spool owns it — no double-ownership. Already-tracked entry isn't re-tracked, so the reset clock governs the next attempt.
- **Reversible**: `SWITCHROOM_INBOUND_DELIVERY_CONFIRM=0` gates every new code path (`:78`, `:115`, `:133`, `:152`) back to legacy fire-and-forget. No schema/persistence change; the sweep interval is `unref`'d (`:120`).

Watch (non-blocking, by design): the ack is keyed by chat/thread only, not by message identity, and `enqueue` fires for every turn source (cron / subagent-handback / vault-resume — comment at `:7948-7958`). So a non-inbound turn starting for a key with a stranded inbound will ack (clear) it. This is almost certainly correct — claude starting any turn flushes the composer where the stranded text sits, so the text rides along with that submission — but it does mean the "deliver-until-acked" guarantee is "until *a turn for this key* starts," not "until *this message* is provably submitted." Worth confirming on the canary that a cron-during-strand actually carries the stranded text rather than acking it into the void. Also the accepted tradeoff: a slow-but-not-stranded turn (>15s to enqueue) re-delivers and can produce one duplicate turn — fine, far better than the 300s drop it replaces.

Rationale: the previous behavior silently dropped operator messages at the 300s poke; this tracks delivery until claude provably starts the turn, with correct key correlation, a bounded-blast-radius design, full kill switch, and unit coverage. Safe to merge; verify the cron-during-strand ack interaction on the canary as planned.

PR #2091, opened 2026-06-02
test(uat): fix flaky /steer assertion + wire no-drop into ci-uat

mekenthompson:
Two follow-ups from the deliver-until-acked work (#2089):

1. **Flaky `/steer` assertion** — `jtbd-rapid-followup-dm`'s narration regex rejected the valid reply *Switched to MD5 per your update* (the 2026-06-02 canary; the 2026-05-27 run hit the same non-match). Broadened to accept `switch(ed|ing) to X per your update/follow-up/guidance/request/steer`, anchored on "per your <qualifier>" so it stays distinct from the queued path. Verified: matches the steer reply, rejects a fresh queued answer.
2. **ci-uat wiring** — added `inbound-no-drop-rapid-fire-dm` as its own isolated, secrets-gated step (mirroring `jtbd-always-on-after-restart`) so the deliver-until-acked regression coverage stays live.

Test-infra only; no production code. tsc clean, YAML valid, regex behavior verified.

### DIFF
--- a/.github/workflows/ci-uat.yml
+++ b/.github/workflows/ci-uat.yml
@@ -117,3 +117,14 @@ jobs:
         # gate. Keep them separate; operators run both during UAT.
         working-directory: telegram-plugin
         run: bun run test:uat jtbd-always-on-after-restart
+
+      - name: UAT reliable-delivery regression (rapid-fire no-drop)
+        if: steps.secrets.outputs.ok == 'true'
+        # Fires back-to-back messages at the turn-completion boundary and
+        # asserts every one gets its own reply — the regression gate for the
+        # deliver-until-acked queue (#2089), which replaced fire-and-forget
+        # delivery that stranded inbounds in the composer and dropped them at
+        # the 300s silence-poke. Driver-only (no sudo), but kept in its own
+        # step so a flake can't stall the fuzz pool.
+        working-directory: telegram-plugin
+        run: bun run test:uat inbound-no-drop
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `telegram-plugin/uat/scenarios/jtbd-rapid-followup-dm.test.ts:37` — broadened `narratesSteer` regex adds `switch(?:ed|ing)? to \w+ per your (?:update|follow-?up|guidance|request|steer)`. Verified: matches the canary reply "Switched to MD5 per your update" (true) and "Switching to SHA per your follow-up" (true), rejects a fresh queued answer "The MD5 hash is abc123" (false). Anchored on "per your <qualifier>" so it stays distinct from the QUEUED path. Correct.
- `.github/workflows/ci-uat.yml:10-19` — new `inbound-no-drop` UAT step mirrors the existing `jtbd-always-on-after-restart` step: secrets-gated (`if: steps.secrets.outputs.ok == 'true'`), isolated so a flake can't stall the fuzz pool. Target scenario file `inbound-no-drop-rapid-fire-dm.test.ts` exists; `test:uat` is `vitest run` which treats the positional arg as a filename filter, so `inbound-no-drop` resolves it. Correct.
- Nit (non-blocking): `\w+` matches exactly one word token between "to" and "per your", so a multi-word target like "Switched to MD5 hash per your update" wouldn't match. Fine for current canary phrasings; widen to `[\w ]+?` if the model starts adding qualifiers.

Test-infra only, no production code. Both scenario files and the runner verified against PR head (ccce6cba). Regression coverage for the deliver-until-acked queue (#2089) stays live. Author can merge.

PR #2092, opened 2026-06-02
fix(gateway): don't track steering/interrupt deliveries (re-deliver-loop fix)

mekenthompson:
## Summary

Follow-up to #2089 (deliver-until-acked). That queue tracked **every** delivered inbound for an `enqueue` ack — including `/steer` and `!`-interrupt messages. Those are delivered mid-turn to **amend the running turn** and never emit a fresh `enqueue`, so they were never acked and the 5s sweep **re-delivered them (duplicate turns / loop)**.

Found by an adversarial stress-test of the v0.14.40 fix — the canary's steer test had passed only by ack-timing luck, so it shipped latent. Live on the fleet since v0.14.40.

## Fix

Gate tracking on `shouldTrackDelivery({ isSteering, isInterrupt })` — track only **non-steering, non-interrupt** inbounds (the fresh-turn messages the #1556 buffer-gate holds-then-delivers, which are exactly the ones that produce an `enqueue` to ack against). Steering/interrupt are the gate's existing carve-outs; they're now equally exempt from the delivery-confirm queue. Also removes the only realistic source of unbounded queue growth (never-acking entries).

## Test plan
- [x] new pure `shouldTrackDelivery` + 4 unit tests (normal→track; steer→no; interrupt→no; both→no)
- [x] full plugin bun surface: **4229 pass / 0 fail**
- [x] tsc, check-bot-api-wrapping clean
- [ ] canary on test-harness (steer + interrupt round-trips, confirm 0 re-deliveries) + staggered rollout — post-merge

## Risk
Low + narrowing: it makes the queue track *fewer* messages (strictly the ones that ack). No change to the fresh-turn no-drop behavior #2089 fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,23 @@
 # Changelog
 
-## v0.14.40 — Reliable inbound delivery: deliver-until-acked
+## unreleased — Delivery-confirm: don't track steering/interrupt (re-deliver-loop fix)
+
+Follow-up hardening to v0.14.40's deliver-until-acked queue. That change
+tracked **every** delivered inbound for an `enqueue` ack — including
+`/steer` and `!`-interrupt messages. But those are delivered mid-turn to
+*amend the running turn* and never emit a fresh `enqueue`, so they were
+never acked and the 5s sweep re-delivered them (duplicate turns / a
+re-delivery loop). An adversarial stress-test of the v0.14.40 fix caught
+this before it bit in anger (the canary's steer test had passed only by
+ack-timing luck).
+
+- **Track only fresh-turn deliveries.** `shouldTrackDelivery` gates
+  tracking to non-steering, non-interrupt inbounds — exactly the messages
+  the #1556 buffer-gate holds-then-delivers, which are the ones that
+  produce an `enqueue` to ack against. Steering/interrupt are the gate's
+  carve-outs and are now equally exempt from the delivery-confirm queue.
+  Pure helper + unit tests pin the invariant. Also removes the only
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- `telegram-plugin/gateway/inbound-delivery-confirm.ts:77-82` — `shouldTrackDelivery` is a pure, total predicate (`!isSteering && !isInterrupt`). Correct: only fresh-turn inbounds emit the `enqueue` session-event that `ackDelivery` keys on (case at gateway.ts:7876 → ackDelivery at 7949); steering/`!`-interrupt amend the running turn and never enqueue, so tracking them left a never-acked entry the 5s sweep re-delivered forever.
- `telegram-plugin/gateway/gateway.ts:10677` — gate added at the track call site. Verified both args are in scope: `isSteering` finalized at 10275 (`priorTurnInFlight && isSteerPrefix`), `interrupt.isInterrupt` from 10628. This is the shared delivery path steering/interrupt messages actually reach (same `markClaudeBusyForInbound` at 10671), so the bug was real and the fix lands where it strands.
- No regression to #2089's no-drop fix: a genuine fresh-turn message has neither flag set, so it is still tracked. The change strictly narrows what's tracked — also closing the only unbounded-queue-growth path (never-acking entries).
- Tests: ran `tests/inbound-delivery-confirm.test.ts` against PR head (a7134e4) — 13 pass / 0 fail, including the 4 new cases (normal→track, steer→no, interrupt→no, both→no).

CHANGELOG entry accurate; behind the existing `DELIVERY_CONFIRM_ENABLED` flag. Author can merge.

Issue #5, opened 2026-04-24
Quota per model


mekenthompson:
Start script quota should show total quote per model/quota tier

mekenthompson:
Closing as **upstream-blocked**.

Anthropic's API rate-limit headers don't expose per-model dimensions today — the `anthropic-ratelimit-unified-5h-utilization` and `-7d-utilization` headers are aggregates across all models on the subscription. There's no API surface that lets us derive "how much of your Sonnet budget is left vs Haiku" without either:

1. Probing each model separately (4-5 extra getMessages requests per check; expensive enough to defeat the purpose)
2. Scraping a different surface (no documented one exists)
3. An upstream change to expose per-model dimensions in the unified header set

If/when (3) happens, we'll have a clean implementation path. Until then, this isn't a switchroom-side problem — it's an upstream feature request.

If anyone wants to track this externally, file an Anthropic API feedback request and link it back here.

mekenthompson:
Closing as upstream-blocked per audit comment from 2026-04-30. Anthropic API doesn't expose per-model rate-limit headers; no implementation path on switchroom's side.

PR #2093, opened 2026-06-02
fix(gateway): harden deliver-until-acked queue (6 audit findings)

mekenthompson:
## Summary
An exhaustive adversarial audit of the inbound→claude→reply path — run *after* the deliver-until-acked queue (#2089) and the steer-fix (#2092) — found that the re-delivery **sweep was bolted on without inheriting the protections the fresh-inbound path earned**. This PR fixes the six directly-reachable-and-harmful defects, all in the queue's own correctness.

JTBD: **always-available** — an operator message must never be silently dropped or trigger a re-delivery loop.

## Why (the six fixes)
1. **Mid-turn re-delivery** — the sweep re-cleared the composer + re-sent even while a turn was genuinely running, re-creating the exact wedge the queue exists to prevent. Now gated on `currentTurn == null` (the enqueue-confirmed idle signal) + no pending permission/ask_user.
   ⚠️ The audit's *suggested* fix (`!turnInFlightForGate()`) was **unsound**: `claudeBusyKeys` is set eagerly at delivery and stays set through a strand, so it would have broken re-delivery for the very stranded case. Corrected here.
2. **Cross-source ACK collision (silent drop)** — `enqueue` fires for *every* turn start (cron / subagent-handback / vault-resume / restart-marker), so a key-only ack let a synthetic turn clear — and drop — a real user message waiting under the same key. Ack now matches the tracked **message id** (preserved end-to-end via the channel XML).
3. **Empty-body re-delivery loop** — `/queue` with no text never enqueues → re-delivered every cycle forever (self-inflicted DoS). `shouldTrackDelivery` now excludes empty text + synthetic `meta.source`.
4. **Env-timeout coercion loop** — `Number(env) || 15000` let a *negative* override through → infinite re-delivery. Clamp to positive+finite.
5. **Orphaned replay of steer/interrupt** — the delivery-failure buffer push was unguarded; a steer/`!`-interrupt that failed delivery offline was persisted + replayed as a fresh orphaned turn after restart. Guarded with the same carve-out.
6. **Re-delivery ack-race / gate incoherence** — a successful re-send neither re-marked busy nor survived an ack racing the dynamic import. Re-mark + re-affirm tracking (idempotent); only `enqueue` clears tracking.

## Test plan
- [x] +7 unit cases in `inbound-delivery-confirm.test.ts` (messageId-matched ack incl. the synthetic-collision scenario; empty/whitespace/synthetic carve-outs)
- [x] Full `telegram-plugin` bun suite green — **3924 pass / 0 fail**
- [x] `tsc --noEmit` clean
- [ ] UAT (no-drop rapid-fire + steer + interrupt round-trips) on canary before fleet rollout

## Risk
Tightly scoped to the queue's own logic (sweep gate, ack match, tracking carve-outs, env clamp). The pure module is fully unit-tested; the gateway wiring reuses the in-scope `isSteering`/`interrupt.isInterrupt`/`effectiveText`/`inboundMsg.messageId` already used at the same callsite. Kill-switch `SWITCHROOM_INBOUND_DELIVERY_CONFIRM=0` unchanged.

## Deferred to a focused follow-up (NOT bundled — separate concerns / lower reachability, avoiding a risky mega-diff)
- turn_end gate-wedge backstop (a throw before purge wedges the #1556 gate — **bounded today by the 300s silence-poke**, so degraded not permanent)
- centralizing drain-path tracking via a shared `deliverAndTrack` helper (a buffer-drained message that re-strands isn't re-delivered)
- reaction-flush gate (reactions bypass the #1556 gate mid-turn)
- bridge-side messageId dedup (re-delivery double-execution of a slow-to-ack side-effectful command)

Related: #2089, #2092.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1343,7 +1343,15 @@ const DELIVERY_CONFIRM_ENABLED = process.env.SWITCHROOM_INBOUND_DELIVERY_CONFIRM
 // a clean delivery, so 15s won't false-positive on a healthy turn. Tunable
 // (env) for tests/forensics; a too-low value re-delivers healthy slow turns
 // (duplicate turn), which is why the default is comfortably above ack latency.
-const DELIVERY_CONFIRM_TIMEOUT_MS = Number(process.env.SWITCHROOM_INBOUND_DELIVERY_TIMEOUT_MS) || 15_000
+const _deliveryTimeoutRaw = process.env.SWITCHROOM_INBOUND_DELIVERY_TIMEOUT_MS
+const _deliveryTimeoutParsed =
+  _deliveryTimeoutRaw != null && _deliveryTimeoutRaw !== '' ? Number(_deliveryTimeoutRaw) : 15_000
+// Clamp to a positive, finite value: a negative / zero / NaN env override would
+// make the sweep treat every tracked entry as stranded and re-deliver every
+// cycle forever (a self-inflicted re-delivery loop). To disable the feature,
+// use SWITCHROOM_INBOUND_DELIVERY_CONFIRM=0, not a degenerate timeout.
+const DELIVERY_CONFIRM_TIMEOUT_MS =
+  Number.isFinite(_deliveryTimeoutParsed) && _deliveryTimeoutParsed > 0 ? _deliveryTimeoutParsed : 15_000
 const DELIVERY_CONFIRM_SWEEP_MS = 5_000
 const deliveryQueue = createDeliveryQueue<InboundMessage>()
 
@@ -4107,7 +4115,16 @@ async function redeliverStrandedInbound(p: PendingDelivery<InboundMessage>): Pro
     if (selfAgent) clearAgentComposer({ agentName: selfAgent })
   } catch { /* best-effort; re-deliver regardless */ }
   const ok = ipcServer.sendToAgent(selfAgent, p.inbound)
-  if (!ok) {
+  if (ok) {
+    // Keep the #1556 gate coherent with the re-sent delivery, and survive an
+    // ack that raced the `await import` above: only `enqueue` clears tracking,
+    // so if a concurrent ack removed the entry, re-affirm it — never drop.
+    // Both ops are idempotent.
+    markClaudeBusyForInbound(p.inbound)
+    if (!deliveryQueue.pending.has(p.key)) {
+      trackDelivery(deliveryQueue, p.key, p.inbound, Date.now(), p.messageId)
+    }
+  } else {
     // Bridge offline between attempts — hand off to the offline buffer
     // (bridgeUp drains it) and stop tracking here; the spool owns it now.
     pendingInboundBuffer.push(selfAgent, p.inbound)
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

All six fixes verified against PR head (0c9ef48); the pure module + gateway wiring are type-coherent and the load-bearing id flow is real.

- Fix #2 (cross-source ack collision) — `inbound-delivery-confirm.ts:173-185`: ack now requires `entry.messageId === enqueueMessageId`. Verified the comparison is string-vs-string, not a silent type mismatch: track records `String(inboundMsg.messageId)` (gateway.ts:10721) and the enqueue `ev.messageId` is a `string | null` parsed from the channel-XML `message_id` attribute (session-tail.ts:92,142), not the `number` on `InboundMessage`. Same id already drives reply-anchoring (`turn.sourceMessageId`, gateway.ts:7833-7834), so the end-to-end flow is the one that already works. Null-fallback to key-only ack is correct for legacy entries.
- Fix #1 (mid-turn re-delivery) — `gateway.ts:10745-10755` (sweep): gated on `currentTurn != null` + `pendingPermissions.size`/`pendingAskUser.size`. All three are module-scoped (1570, 2637, 2678). The inline comment correctly documents why the audit's suggested `!turnInFlightForGate()` was unsound — `claudeBusyKeys` is set eagerly at delivery and persists through a strand, so it would suppress re-delivery for the exact stranded case. `currentTurn` (set only by enqueue, nulled at turn-end) is the right idle signal.
- Fix #3 (empty-body loop) + synthetic — `inbound-delivery-confirm.ts:218-231`: `shouldTrackDelivery` excludes `hasSource` and empty `effectiveText` (trim-checked, and only when supplied so existing callers are unaffected). Track callsite passes the steer/queue-stripped `effectiveText` (10096). Correct.
- Fix #4 (env clamp) — `gateway.ts:1346-1351`: `Number.isFinite && > 0` guard; negative/zero/NaN falls back to 15s. Closes the negative-override infinite-redelivery path.
- Fix #5 (orphaned steer/interrupt replay) — `gateway.ts:10728-10739`: the `!delivered` buffer push is now guarded by the same `shouldTrackDelivery` carve-out, so a steer/interrupt that failed delivery offline isn't persisted+replayed as a fresh orphaned turn.
- Fix #6 (re-delivery ack-race) — `gateway.ts:4118-4126`: successful re-send re-marks busy and re-affirms tracking only if not concurrently acked; both idempotent. Sound.

Tests: ran `inbound-delivery-confirm.test.ts` + `session-tail.test.ts` against head — 53 pass / 0 fail, including the synthetic-collision and empty/whitespace carve-out cases. Behind `SWITCHROOM_INBOUND_DELIVERY_CONFIRM` kill-switch; deferred items explicitly scoped out and reasonable.

Watch (non-blocking): (a) the sweep early-returns on ANY in-flight turn, so a stranded message waits until the current turn ends before re-delivery — correct/bounded in single-tenant, not a drop; (b) if a real user turn's enqueue ever loses its channel XML (`enqueueMessageId` null while tracked id is set), the strict-match branch declines to ack and re-delivers — acceptable given XML is reliably present, but the UAT canary should confirm 0 spurious re-deliveries on the normal path. Author can merge.

Issue #2094, opened 2026-06-02
Delivery-path hardening follow-ups (deferred from #2093 audit)


mekenthompson:
The exhaustive adversarial audit behind #2093 surfaced four further findings that were **deliberately deferred** from that PR to keep it tightly scoped + low-risk (the reachable-and-harmful queue-correctness fixes shipped in #2093). Tracking them here so nothing is silently dropped.

### 1. turn_end gate-wedge backstop  (severity: medium — bounded by silence-poke)
The `turn_end` session-event handler (`telegram-plugin/gateway/gateway.ts` ~8422–9077) runs a synchronous prelude before reaching its first `purgeReactionTracking`/`endCurrentTurnAtomic`. A throw in a pre-purge op (e.g. `redactOutboundText`, `progressDriver?.takeOverCard`) exits before purge → `activeTurnStartedAt` + `claudeBusyKeys` stay populated → the #1556 inbound gate wedges closed → subsequent inbounds buffer.
**Mitigation today:** the 300s silence-poke fallback recovers the wedged turn (degraded, not permanent) — confirm this clears the gate state.
**Fix:** wrap the handler body in a guarded `try { … } finally { if (activeTurnStartedAt.has(key)) purgeReactionTracking(key, turn) }` — a backstop that only acts when a throw skipped the canonical purge. Deserves its own focused PR + test (650-line handler; don't bundle).

### 2. Centralize drain-path tracking via a shared `deliverAndTrack` helper  (severity: medium)
`trackDelivery` is called at only ONE site (the fresh handleInbound path). The buffer-drain / bridgeUp / idle-drain re-delivery callbacks (`gateway.ts` ~1792, ~1887, ~4060, ~5214, and `inbound-delivery-machine-dispatch.ts` drainBuffer) call `markClaudeBusyForInbound` but NOT `trackDelivery`, so a message drained from the offline buffer that then strands is invisible to the sweep. Introduce one `deliverAndTrack(agent, msg)` helper (sendToAgent + markBusy + conditional trackDelivery via `shouldTrackDelivery`) and route all fresh-turn delivery sites through it. This is the "centralize tracking into a shared send helper" the audit recommended.

### 3. Reaction-flush gate  (severity: medium)
[...]

PR #2096, opened 2026-06-02
fix(gateway): approval cards undeliverable to DM operators (topic-id bug)

mekenthompson:
## Summary
Fixes a live, recurring wedge (root-caused from a marko incident): **HIGH-RISK approval/permission cards an agent raises from inside a supergroup topic never reach the operator**, so the request auto-denies after its 10-min TTL and the agent sits at the open prompt — wedged, unable to do anything else (including reporting on its topics).

JTBD: **you hold the leash** — an approval the operator must tap has to actually arrive.

## Root cause
The "PR4b emitter sweep" routed every approval card's `message_thread_id` from `resolveAgentOutboundTopic(originThreadId)` — a forum topic id valid **only in the agent's supergroup**. But the cards fan out to the operator's `allowFrom` chats, which are normally **DMs**. A DM `sendMessage` carrying a `message_thread_id` is rejected by Telegram with `400 Bad Request: message thread not found` → card never arrives → auto-deny at TTL → agent wedges.

Live evidence (marko): a Panorama "Winter Special" **Brevo EDM** approval looping `permission_request send to 8248703757 failed … message thread not found` → `permission TTL expired — auto-deny tool=mcp__brevo__post`, while marko sat at the prompt and stopped reporting on its topics.

Affects **5 emitters**: permission fan-out, drive-approval, ms365-approval, config-approval, and the no-turn re-emit fallback.

## Fix
A pure, unit-tested guard:
```ts
topicForRecipient({ recipientChatId, resolvedTopic, supergroupChatId })
// → resolvedTopic only when recipient IS the owning supergroup chat; else undefined
```
Attach the topic **only** to the recipient that owns it (the agent's `channels.telegram.chat_id`); send thread-less to operator DMs. The turn-initiated re-emit branch (already posts to the originating chat+thread) is unchanged.

## Test plan
- [x] +6 unit cases on `topic-router` (supergroup→topic, DM→none, wrong-supergroup→none, no-topic, no-supergroup-mode, string/number id equivalence) — **39 pass**
- [x] `tsc --noEmit` clean
- [ ] Canary: verify an approval card from a supergroup-topic turn now lands in the operator DM (tappable) before fleet rollout

## Risk
Low + well-scoped. The guard only ever *removes* an invalid `message_thread_id` (the one that was 400-ing); it never adds one where there wasn't. Worst case for a misconfig is a thread-less card in a DM — which is exactly what a DM should get. No behavior change for the already-correct turn-initiated supergroup path.

Root-caused while investigating the marko wedge; related delivery-path work: #2093, #2094.

### DIFF
--- a/src/telegram/topic-router.test.ts
+++ b/src/telegram/topic-router.test.ts
@@ -10,6 +10,7 @@
 import { describe, expect, it } from "vitest";
 import {
   resolveOutboundTopic,
+  topicForRecipient,
   validateCronTopicAliases,
   type OutboundEvent,
   type TopicRouterConfig,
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `src/telegram/topic-router.ts:157` — `topicForRecipient` is a pure guard: returns `resolvedTopic` only when `String(recipientChatId) === String(supergroupChatId)`, else `undefined`. Verified against head (e8c7f3e); all 6 documented cases reproduce (supergroup→topic, DM→none, wrong-supergroup→none, no-topic→none, no-supergroup-mode→none, string/number equivalence). The guard only ever *strips* an invalid `message_thread_id` — never adds one — so worst-case misconfig is a thread-less DM card, which is correct.
- `telegram-plugin/gateway/gateway.ts` — all 5 emitters wired through the guard (lines 2287 permission re-emit, 4681 permission fan-out, 4826 drive, 4904 ms365, 4988 config). Count matches the PR body's claimed 5.
- `gateway.ts:2273-2290` — the permission-resume change is correctly scoped: the `turn != null` (turn-initiated, posts to originating chat+thread) branch is untouched; only the `allowFrom` DM fan-out fallback gets the guard. No regression to the already-correct supergroup path.
- `gateway.ts:11105` — `resolveAgentSupergroupChatId` reads `channels.telegram.chat_id` via the config cascade and swallows errors to `undefined` (→ guard returns `undefined` → thread-less, the safe default). Sound failure mode.

Note: could not run the vitest suite from a shallow clone (deps not installed); verified the guard logic directly via node and confirmed every cited symbol/call site exists at head. PR body claims +6 cases / 39 pass + tsc clean.

Rationale: a pure, strictly-narrowing guard that fixes the exact 400 "message thread not found" → auto-deny → wedge path, applied uniformly across all 5 card emitters, with the correct supergroup path preserved.

Issue #1664, opened 2026-05-22
Gateway: streamed final answers (sendMessageDraft) never finalized — replies vanish from chat & history


mekenthompson:
## Summary

Streamed final answers are delivered to Telegram via the **`sendMessageDraft`** ("answer lane" / draft transport) and are **never finalized into a permanent `sendMessage`**. The draft is visible in the chat transiently, then gets cleared when the next turn's lane opens — so from the user's side, replies *appear and then vanish*. They are also absent from the gateway's SQLite history buffer.

Reported by the user as: "you seem to be deleting messages that were sent."

## Evidence (chat 8248703757, clerk agent, 2026-05-22)

Per-turn pattern in `gateway-supervisor.log`:

- **Ack** (sent via the `reply` MCP tool) → `tg-post method=sendMessage` → `reply: finalized messageIds=[NNNN]` → **persists**, lands in history buffer. ✓
- **Final answer** (ended as plain assistant transcript text) → `tg-post method=sendMessageDraft` + `answer_lane_update ... transport:"draft"` → **no follow-up `sendMessage`** → never finalized. ✗

Concrete log lines — five consecutive turns where the final answer was draft-only:

```
10:08:07 sendMessageDraft bytes=375   (answer_lane_update charCount=375 transport=draft)
10:10:50 sendMessageDraft bytes=861
10:13:47 sendMessageDraft bytes=1232
10:23:41 sendMessageDraft bytes=910
10:28:57 sendMessageDraft bytes=573
```

None of those five has a matching `tg-post method=sendMessage` afterwards.

**Contrast** — the one turn that *did* persist (msg 11660), a short 321-char answer:

```
10:18:28.639 sendMessageDraft bytes=321 hash=d4abe64c5729
10:18:29.244 sendMessage      bytes=321 hash=d4abe64c5729   <- same hash, promoted to real message
```

So short answers occasionally get promoted; the larger streamed ones do not.

Every `turn_end` carries `"suppressClearedCount":1`, and `answer_lane_update transport:"draft"` fires *after* `turn_end` — the answer is written to the draft lane post-turn-end and the draft→committed-message finalize step never runs.

History buffer corroboration: `get_recent_messages` shows the acks (11652, 11655, 11658, 11662, 11664) but **none of the final answers** for those turns — the buffer jumps straight from each ack to the next inbound user message. Message-id gaps (11653, 11656, 11645–11648, …) line up with cleared drafts.

## Likely root cause

[...]

PR #2098, opened 2026-06-02
fix(telegram): route sub-agent handback/progress to the dispatching topic

mekenthompson:
## Summary

Background sub-agent status was reliable in DMs but **mis-routed in supergroups**. When a background worker/researcher finishes, the agent's in-voice "here's what I found" handback reply landed in the chat's *last-seen* topic instead of the topic the work was dispatched from — reading as **silence** in the topic the user was actually watching. First of a series closing the "status must work in channels" gaps (operator-reported).

## Why

`resolveSubagentOriginChat` resolves **both** the origin chat and topic from the parent turn, but:
- the handback (beat 4, `gateway.ts:18592`) and progress (beat 3, `gateway.ts:18754`) callsites took only `.chatId` and **dropped `.threadId`**;
- the inbound builders (`subagent-handback-inbound-builder.ts`, `subagent-progress-inbound-builder.ts`) had **no thread carrier** at all.

So the synthesized handback turn was born thread-blind. Replies are auto-routed (the model is never told the topic — `server.ts:25`), so `executeReply` fell back to `chatThreadMap` (chat last-seen topic) → wrong topic. The worker-feed **running** cue already threaded correctly (`origin?.threadId`, `gateway.ts:18744`), making the asymmetry stark: live activity in the right topic, the result in the wrong one. Background workers run for minutes, so cross-topic activity is the norm in one-agent-owns-supergroup mode — this was a routine mis-route, not an edge case.

## What changed (pure jsonl-tail → render; no model call — inside the subscription-honest boundary)

- **Both builders** now carry the origin topic: top-level `threadId` (→ the enqueued turn's `sessionThreadId`, which routes the handback turn's live activity feed) **and** `meta.message_thread_id` (model-visible channel attribute), mirroring the real-inbound shape at `gateway.ts:10557`. Applied **only** when the origin chat resolved — the owner-DM fallback is topic-less.
- **`executeReply`** prefers THIS turn's own originating topic (turn-pinned, #1664) over the chat's last-seen topic when no explicit `message_thread_id`. Fixes synthetic turns (handback/progress/cron) and is strictly more correct under multi-topic concurrency. **DM behavior unchanged** (both undefined).
- Progress path's pending-ticker clear is keyed on the origin thread so the right lane is yielded in a supergroup.

## Test plan

- [x] `tsc --noEmit` clean
- [x] +24 new assertions across handback/progress **builders + decisions**: thread carried when origin chat wins; omitted on DM fallback / when no origin thread (`subagent-handback-inbound-builder.test.ts`, `subagent-handback-decision.test.ts`, `subagent-progress-inbound-builder.test.ts`)
- [x] Full bun unit suite: 3933 pass / 0 fail
- [x] Full vitest suite green (after `dist` build; remaining failures are unrelated build-artifact/process-spawn env flakes)

## Risk

Low. Builder/decision changes are additive and off the hot send path. The one behavioral change is `executeReply`'s thread precedence — identical result for normal DM/topic turns (turn thread == chat last-seen), only diverges for synthetic turns (the bug) and concurrent multi-topic (more correct). Follow-up PRs in this series: worker-feed DM-fallback, sibling-topic purge, post-ack parent feed, channel UAT coverage.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -5577,7 +5577,21 @@ async function executeReply(args: Record<string, unknown>): Promise<{ content: A
 
   assertAllowedChat(chat_id)
 
-  let threadId = resolveThreadId(chat_id, args.message_thread_id as string | undefined)
+  // Thread resolution precedence: (1) an explicit message_thread_id the
+  // model passed, else (2) THIS turn's own originating topic
+  // (turn-pinned, #1664), else (3) the chat's last-seen topic
+  // (chatThreadMap). Preferring the turn's own thread over the chat
+  // last-seen heuristic fixes synthetic turns (subagent handback/progress,
+  // cron) — whose topic the model is never told and which never write
+  // chatThreadMap — and is strictly more correct under multi-topic
+  // concurrency (a reply lands in the topic the turn came from, not
+  // whichever topic most recently received a message). DM: both are
+  // undefined → unchanged.
+  let threadId = resolveThreadId(
+    chat_id,
+    (args.message_thread_id as string | undefined) ??
+      (turn?.sessionThreadId != null ? turn.sessionThreadId : undefined),
+  )
 
   if (reply_to == null && quoteOptIn && HISTORY_ENABLED) {
     try {
@@ -18581,6 +18595,7 @@ void (async () => {
                   })
                 }
 
+                const handbackOrigin = resolveSubagentOriginChat(agentId)
                 const decision = decideSubagentHandback({
                   handbackEnvValue: process.env.SWITCHROOM_SUBAGENT_HANDBACK,
                   outcome,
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

Verified against PR head 26f4c79.

- `executeReply` thread precedence (gateway.ts:5590-5594): `resolveThreadId(chat_id, args.message_thread_id ?? turn?.sessionThreadId)` correctly implements explicit → turn-own-topic → chatThreadMap last-seen. `turn = currentTurn` is pinned at entry (gateway.ts:5417), so it's in scope. `resolveThreadId` (gateway.ts:2297) returns `Number(explicit)` when non-null else `chatThreadMap.get` — DM path stays undefined→undefined, behavior unchanged. Strictly more correct for synthetic turns and multi-topic concurrency.
- Handback decision (subagent-handback-inbound-builder.ts:202-205): `const threadId = input.fleetChatId && input.originThreadId != null ? input.originThreadId : undefined` — guards the topic-less owner-DM fallback so a stray thread id can't mis-address it. Same guard on progress (subagent-progress-inbound-builder.ts:250-253). Correct.
- Both builders carry top-level `threadId` → enqueued turn's `sessionThreadId` (routes live activity feed) AND `meta.message_thread_id` (model-visible channel attr, mirrors real-inbound shape at gateway.ts:10557), each emitted only when set (handback builder:117,128; progress builder:177,185).
- Callsites resolve origin once and spread `originThreadId` conditionally (gateway.ts:18595-18610 handback, 18768-18775 progress). Pending-ticker clear now keyed on `progressOrigin?.threadId` (gateway.ts:18801-18804) so the right lane is yielded in a supergroup; chat-level for DM agents.
- +24 assertions across the two builders and two decision fns cover origin-chat-won (thread carried), DM fallback (omitted), and no-origin-thread (omitted) — the exact guard boundaries.

Rationale: pure jsonl-tail→render plumbing of an already-resolved origin topic into the synthesized handback/progress turn; no new model call, off the hot send path, DM behavior provably unchanged, and every behavioral branch is test-covered. Couldn't run the vitest/bun suites from the shallow clone (deps absent), but the logic and test coverage verify by inspection. Clean to merge.

mekenthompson:
Addressed the reviewer's one flagged residual in aacc8d6e: the streamed-reply path (`executeStreamReply`) now applies the same turn-thread precedence as `executeReply`, so a streamed beat-4 handback (the common modern-fleet path per #1683) also routes to the originating supergroup topic. Fresh-process reviewer verdict: **APPROVE** (verified routing correctness, DM safety, the origin-chat-won guard, executeReply blast radius, and constraint compliance). tsc clean; targeted suites green (95 tests incl. 49 stream-reply-handler).

PR #2099, opened 2026-06-02
feat(telegram): nest foreground sub-agent tool steps, not just prose

mekenthompson:
## Summary

A foreground sub-agent (Task/Agent, no `run_in_background`) that runs tools **without narrating** (a researcher reading files, a worker running commands) currently surfaces **nothing** in the parent turn's nested ↳ activity block. Its steps were invisible — delegated work reads as silence between "Delegating: …" and the final result. Second in the status-visibility series (after #2098).

## Why

Model A (#2027/#2032) nests a foreground sub-agent's live steps under the parent's activity feed, fed by the watcher's `onProgress`. But `onProgress` fired **only** on `sub_agent_text` (narrative prose) events — never on `sub_agent_tool_use` (`subagent-watcher.ts:774` updated `lastTool`/`toolCount` but didn't fire). And the gateway keyed the nested line solely on `latestSummary` (the prose, `gateway.ts:18729`). So a tool-only sub-agent produced zero ticks → zero nested visibility. This is the foreground half of the visibility-parity goal (JTBD-2).

## What changed (pure jsonl-tail → render; no model call)

- **`subagent-watcher.ts`**: fire `onProgress` from the `sub_agent_tool_use` branch too, carrying a friendly `describeToolUse` label as a new optional `progressLine` arg ("Reading CLAUDE.md", "Running a command") — the **same renderer** the main-turn activity feed uses, so foreground steps read identically. `latestSummary` stays the worker's narrative result, **never** polluted with tool labels (the beat-4 handback payload depends on it).
- **`gateway.ts` foreground branch**: render `progressLine ?? latestSummary` — tool ticks surface their step, text ticks keep showing prose, interleaved chronologically like the main feed.

## Background side-effect (intentional, harmless)

The background worker feed ignores `progressLine` (keeps using `latestSummary`), so its content is unchanged. It may first-paint slightly sooner on a tool tick — throttled (2.5s min edit interval, 8s first-paint floor), harmless, arguably an improvement.

## Test plan

- [x] `tsc --noEmit` clean
- [x] New watcher test: a prose-less `Read` tool tick fires `onProgress` with `progressLine === "Reading CLAUDE.md"`, and `latestSummary` stays `''` (handback payload intact)
- [x] `subagent-watcher.test.ts` (22), `foreground-nesting.test.ts` (9), `tool-activity-summary.test.ts` (21), `subagent-watcher-handback-gaps.test.ts` (10) green
- [x] bun watcher/subagent suites: 77 pass / 0 fail

## Risk

Low–medium. The watcher change is additive (a new onProgress firing path); the gateway change is one line (`progressLine ?? latestSummary`). The gateway accumulate/compose seam itself remains gateway-internal (untested directly — same limitation flagged for the series; the dedicated test-coverage PR will extract a pure helper + add channel UAT). No model call; subscription-honest boundary untouched.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -18736,7 +18736,7 @@ void (async () => {
               // suppresses stale-after-restart delivery (a 4-h-old
               // "still working (5m)" would be a lie). Sweep on handback
               // lives in the `onFinish` block just above.
-              onProgress: ({ agentId, description, latestSummary, elapsedMs, prevBucketIdx, setBucketIdx, lastTool, toolCount }) => {
+              onProgress: ({ agentId, description, latestSummary, elapsedMs, prevBucketIdx, setBucketIdx, lastTool, toolCount, progressLine }) => {
                 let fleetChatId = ''
                 try {
                   const fleets = progressDriver?.peekAllFleets() ?? []
@@ -18776,7 +18776,15 @@ void (async () => {
                     nestingEnabled: foregroundNestingEnabled,
                     replyCalled: turn.replyCalled,
                   })) return
-                  const child = latestSummary.trim().slice(0, 120)
+                  // Prefer the tick's own display line: `progressLine` (a
+                  // friendly tool-step label) on tool ticks, else the
+                  // worker's narrative (`latestSummary`) on text ticks. This
+                  // lets a foreground sub-agent that runs tools without
+                  // emitting prose still nest its steps under the parent
+                  // feed (the foreground blindspot) — mirroring the
+                  // main-turn activity feed, which surfaces both tool labels
+                  // and prose.
+                  const child = (progressLine ?? latestSummary).trim().slice(0, 120)
                   if (child.length === 0) return
                   let narrative = turn.foregroundSubAgents.get(agentId)
                   if (narrative == null) {
--- a/telegram-plugin/subagent-watcher.ts
+++ b/telegram-plugin/subagent-watcher.ts
@@ -42,6 +42,7 @@ import { basename, join } from 'path'
 import { homedir } from 'os'
 import { projectSubagentLine, sanitizeCwdToProjectName, detectErrorInTranscriptLine } from './session-tail.js'
 import { sanitiseToolArg } from './fleet-state.js'
+import { describeToolUse } from './tool-activity-summary.js'
 import { escapeHtml, truncate } from './card-format.js'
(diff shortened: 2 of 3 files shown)


mekenthompson:
APPROVE

Verified against PR head 58b258c.

- `subagent-watcher.ts:792-803`: the `sub_agent_tool_use` branch now fires `onProgress` carrying `progressLine = describeToolUse(...)`, gated on `onProgress != null && entry.state === 'running' && !entry.historical` and a non-empty label. Wrapped in try/catch with a log on callback error — consistent with the existing firing path. Correct fix for the foreground blindspot (tool-only sub-agents previously fired nothing).
- `describeToolUse` (tool-activity-summary.ts:72) is the same renderer the main-turn feed uses; Read→`Reading <basename>` (line 117), so the test's expected `'Reading CLAUDE.md'` is accurate. Foreground steps read identically to the main feed.
- `latestSummary` is left as `entry.lastResultText` (the narrative), never overwritten with the tool label — the beat-4 handback payload that depends on it is unaffected. Test asserts `latestSummary === ''` on the tool tick, confirming no pollution.
- `gateway.ts:18787`: `const child = (progressLine ?? latestSummary).trim().slice(0, 120)` — tool ticks surface their step, text ticks keep prose; empty-after-trim returns early; dedup against the preceding line (18796-18800) absorbs the watcher's per-tick re-emit. One-line, behavior-preserving for text ticks.
- Background feed ignores `progressLine` (still keyed on `latestSummary`), so its content is unchanged; the extra firing path is throttled (2.5s/8s floors) — harmless, as the author notes.
- Tests: new watcher assertion covers the prose-less Read tick (progressLine set, latestSummary empty); existing onProgress wiring in the sync harness is additive (`...(opts.onProgress ? {...} : {})`).

Rationale: pure jsonl-tail→render plumbing of a friendly tool label into an already-existing nesting path; no model call, handback payload provably untouched, the one gateway change is a null-coalesce that preserves prior behavior on text ticks. The gateway accumulate/compose seam stays untested directly (author flagged it for the series' dedicated coverage PR) but the watcher firing and the render fallback both verify by inspection and test. Clean to merge.

mekenthompson:
Fresh-process reviewer verdict: **APPROVE**. Verified: tool-tick onProgress correctly guarded; no latestSummary/handback-payload pollution (the critical check); background blast-radius is provably content-neutral (worker feed + dead-by-default bucket relay both ignore progressLine, bucket-gated); foreground edit path coalesces (no rate-limit risk); describeToolUse is a zero-import leaf (no circular import). 52 targeted tests pass, tsc clean. Residual (non-blocking): a one-line canary spot-check on a real background-worker DM to confirm the marginally-earlier first-paint reads fine.

PR #2100, opened 2026-06-02
test(uat): accept "switched to X as you asked" steer narration

mekenthompson:
## Summary
The `jtbd-rapid-followup` steer assertion's `narratesSteer` regex required "switch(ed) to <algo> **per your** <qualifier>", but on a v0.14.42 canary the model narrated the steer as **"Switched to MD5 as you asked"** — a phrasing #2091's broadening still missed → recurring **false** UAT failure, even though the agent steered correctly (switched algorithms, **zero re-deliveries** — gateway log confirmed #2092's no-loop fix held).

Accept `as (you) asked/requested/instructed/wanted/said` as an alternative qualifier after "switch(ed) to <algo>".

## Why it's safe
- Stays distinct from the QUEUED path, which uses its own `/queued|new task/` matcher (broadening `narratesSteer` doesn't touch it).
- Validated against the real canary strings: the failing "Switched to MD5 as you asked" now matches; back-compat "per your update" still matches; the queued fresh-answer "The MD5 of … is …" correctly does **not** match.

## Context
Found while running the release-critical mtcute UAT against the v0.14.42 fleet. The *product* validated cleanly (no-drop re-run passed + gateway log showed zero drops; steer switched algorithms with no loop; always-on-after-restart ≤2min). The two scenario failures were both **test-matcher artifacts** — this PR fixes the steer one. Test-only change.

Test plan: regex validated against real canary strings; no product code touched.

### DIFF
--- a/telegram-plugin/uat/scenarios/jtbd-rapid-followup-dm.test.ts
+++ b/telegram-plugin/uat/scenarios/jtbd-rapid-followup-dm.test.ts
@@ -56,14 +56,16 @@ describe("uat: rapid follow-ups — steering vs queued classification", () => {
             const txt = m.text;
             const mentionsMd5 = /\bmd5\b/i.test(txt);
             // Steer narration: the agent acknowledges amending the in-flight
-            // task. Accept the phrasings the model actually uses — including
-            // "Switched to MD5 per your update/follow-up" (the 2026-06-02
-            // canary reply that the old regex wrongly rejected). Anchored on
-            // "per your <qualifier>" / continuation language so it stays
-            // distinct from the QUEUED path (a fresh answer with no such
-            // course-correction narration).
+            // task. Accept the phrasings the model actually uses — "Switched
+            // to MD5 per your update/follow-up" (2026-06-02 canary) AND
+            // "Switched to MD5 as you asked" (2026-06-03 canary) — i.e. a
+            // "switch(ed) to <algo>" acknowledgement qualified by EITHER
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

Test-only change, single file (jtbd-rapid-followup-dm.test.ts:24).

- The `narratesSteer` regex (jtbd-rapid-followup-dm.test.ts:24) is broadened by turning the trailing `per your <qualifier>` into an alternation: `(?:per your (?:update|follow-?up|guidance|request|steer)|as (?:you )?(?:asked|requested|instructed|wanted|said))` after `switch(?:ed|ing)? to \w+`. Well-formed, balanced groups, anchored on the same `switch...to <algo>` stem.
- Validated by running the exact regex in node against the canary strings: 'Switched to MD5 as you asked' → matches (the 2026-06-03 failure); 'Switched to MD5 per your update' → still matches (back-compat); 'The MD5 of foo is abc123' (queued fresh answer) → does NOT match. All as the PR body claims.
- Safe scope: the QUEUED path uses its own `/queued|new task/` matcher, untouched here, so broadening the steer matcher can't bleed into queued-vs-steer classification.
- No product code touched — purely loosens a UAT assertion that was producing a false failure while the product (algorithm switch, zero re-deliveries) behaved correctly per the gateway log.

Rationale: a UAT matcher artifact fix that aligns the assertion with phrasing the model actually emits; regex correctness and non-regression verified by direct execution against real canary strings. Clean to merge.

PR #2101, opened 2026-06-02
test(uat): supergroup channel scenarios + driver dialog priming

mekenthompson:
## Summary

Closes the **DM-only UAT coverage gap**: all 44 existing scenarios are `-dm`, so the status/reply path had zero real-Telegram coverage in a supergroup — the operator's "status must work in channels" requirement and the live proof for the status-visibility series (#2098, #2099). Third in that series.

## What's added

- **`driver.primeDialogs()`** — the mtcute driver runs on `MemoryStorage` (empty peer cache every connect). A supergroup referenced by its marked id (no public username) isn't resolvable until the dialog list is fetched (it carries the channel `access_hash`). New helper drains `iterDialogs` to prime the cache; required before posting to/observing a supergroup. Requires the driver account to be a group member.
- **`jtbd-supergroup-reply-channel.test.ts`** — base channel proof: the agent replies **inside** the supergroup (chatId === supergroup, from the bot), not the operator DM.
- **`fuzz-supergroup-channel.test.ts`** — the DM fuzz corpus's channel twin: realistic inbounds → assert meaningful, leak-free replies land in the supergroup.

## Self-skip (CI-safe)

Both scenarios skip **green** when `SWITCHROOM_UAT_CHAT_ID` is unset or the chat isn't a postable forum supergroup (e.g. still a basic group, or the driver account isn't a member). Verified locally: 9 tests pass (skipped) with a clear operator warning. So this can't red CI on unwired hosts.

## Caveats

- mtcute (this repo's version) has **no forum-topic create/enumerate API**, so the live scenarios use the supergroup's **General topic**. Fine-grained "correct topic among many" routing is pinned by the gateway **unit thread-assertions in #2098**; these scenarios assert the live DM-vs-channel boundary mtcute *can* observe (real chat messages, not drafts — per `feedback_mtcute_cannot_observe_drafts`).
- To run live: wire `test-harness` supergroup-owned on a **forum supergroup** (Topics enabled) + set `SWITCHROOM_UAT_CHAT_ID` (`uat/SETUP.md §2`). A basic group will not work (no topics).

## Test plan

- [x] `tsc --noEmit` clean
- [x] Both scenarios skip-green against an unwired host (9 tests pass)
- [ ] Live validation — pending a forum supergroup + a release containing #2098/#2099 pinned to test-harness

## Risk

Very low. Additive: one new driver method (additive) + two new self-skipping scenario files. No existing scenario or harness path touched.

### DIFF
--- a/telegram-plugin/uat/driver.ts
+++ b/telegram-plugin/uat/driver.ts
@@ -178,6 +178,25 @@ export class Driver {
     }
   }
 
+  /**
+   * True if `chatId` is resolvable (its access_hash is known) — i.e. a
+   * peer the account can address. Call after {@link primeDialogs}.
+   * Non-intrusive: sends nothing. A forum supergroup the driver account
+   * is in resolves true; a chat referenced by a wrong/foreign marked id
+   * (e.g. a BASIC group given a supergroup-style `-100…` id, or a chat
+   * the driver isn't a member of) resolves false. Used to skip supergroup
+   * scenarios cleanly when the test forum isn't wired.
+   */
+  async canResolve(chatId: number): Promise<boolean> {
+    const c = this.requireClient();
+    try {
+      await c.resolvePeer(chatId);
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
   async sendText(
     chatId: number,
     text: string,
--- a/telegram-plugin/uat/driver.ts
+++ b/telegram-plugin/uat/driver.ts
@@ -156,6 +156,28 @@ export class Driver {
     this.client = null;
   }
 
+  /**
+   * Populate the local peer cache with the account's dialogs so a
+   * supergroup referenced by its marked id (e.g. `-100…`) becomes
+   * resolvable. The driver runs on `MemoryStorage`, which starts EMPTY
+   * every connect — a bot username resolves on demand (server lookup),
+   * but a supergroup with no public username has no resolution path
+   * until mtcute has seen it via the dialog list (which carries the
+   * channel's `access_hash`). Call this once before sending to /
+   * observing a supergroup. Best-effort: drains up to `limit` dialogs.
+   * Requires the driver account to be a MEMBER of the supergroup — if a
+   * later `sendText` still throws "Peer … not found in local cache",
+   * the account isn't in the group.
+   */
+  async primeDialogs(limit = 200): Promise<void> {
+    const c = this.requireClient();
+    let seen = 0;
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

Test-only, additive. Verified against PR head 07efba5.

- `driver.primeDialogs()` (uat/driver.ts:172-179): wires to the real `requireClient()` (used identically at :186/:202) and the mtcute client's `iterDialogs({ limit })`; the MemoryStorage-starts-empty rationale checks out (:135). The marked-id-needs-access_hash-from-dialogs reasoning is sound. Minor: the `{ limit }` arg already bounds the iterator, so the extra `if (++seen >= limit) break` is redundant — harmless defensive double-bound, not worth changing.
- `jtbd-supergroup-reply-channel.test.ts`: asserts `reply.chatId === SUPERGROUP_ID` and `reply.fromBot === true` — both real fields on `ObservedMessage` (driver.ts:55,64). Uses a unique `sgproof-${Date.now().toString(36)}` nonce so the matcher can't latch a stale group message; `senderFilter:{ notUserId: sc.driverUserId }` (compiled at assertions.ts:191) correctly means "from the bot, not the driver."
- `fuzz-supergroup-channel.test.ts`: the `-dm` fuzz corpus's channel twin — same four JTBD invariants (in-channel, no credential leak via SECRET_PATTERNS, meaningful reply, optional shape match). Same harness/assertion wiring; `beforeAll` probe posts once and flips `postable`, gating every case.
- Self-skip is a plain early `return` when `SWITCHROOM_UAT_CHAT_ID` is unset/non-finite or the probe `sendText` throws (basic group / non-member) — skips GREEN, so unwired CI/dev hosts can't red. No existing scenario or harness path is touched.

Rationale: closes the DM-only UAT coverage gap with two self-skipping real-Telegram supergroup scenarios plus the dialog-priming primitive they need; every API the new code calls (`requireClient`, `iterDialogs`, `spinUp`, `driverUserId`, `expectMessage`, `senderFilter`) and every asserted field (`chatId`, `fromBot`) exists at head. `tsc --noEmit` clean per the PR; live validation is correctly marked pending a wired forum supergroup. Clean to merge.

mekenthompson:
Fresh-process reviewer: **APPROVE** (test-only, CI-isolated — uat/** excluded from gating vitest/bun-test; tsc-clean; skip-green verified). Folded in both non-blocking notes: non-intrusive `canResolve` probe (no junk message left in the group) + feed-message filter in the fuzz matcher.

PR #2104, opened 2026-06-03
test(uat): live supergroup handback channel scenario (#2098)

mekenthompson:
Adds `jtbd-supergroup-handback-channel` — the live validation of #2098. A background worker dispatched from a supergroup finishes and its result is handed back **in the supergroup**, not the DM (the pre-#2098 thread-blind handback bug). **Passed live on v0.14.43** (18.4s) against the wired test supergroup. Self-skips when no supergroup is wired. Test-only; uat/** is CI-isolated.

### DIFF
--- a/telegram-plugin/uat/scenarios/jtbd-supergroup-handback-channel.test.ts
+++ b/telegram-plugin/uat/scenarios/jtbd-supergroup-handback-channel.test.ts
@@ -1,0 +1,77 @@
+/**
+ * JTBD scenario — background-worker HANDBACK lands in the supergroup (#2098).
+ *
+ * This is the live validation of the headline channel fix: when a background
+ * sub-agent (Task/Agent `run_in_background:true`) dispatched from a supergroup
+ * finishes, the agent's in-voice "here's what the worker found" handback (beat
+ * 4) must land IN the supergroup — not the operator DM (the pre-#2098 bug,
+ * where the synthesized handback inbound was thread-blind and the reply fell
+ * back to the chat's last-seen topic / owner DM).
+ *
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

- New file `telegram-plugin/uat/scenarios/jtbd-supergroup-handback-channel.test.ts:1-83` — purely additive live UAT for the #2098 headline fix (bg-worker handback must land in the originating supergroup, not the owner DM). No existing path touched.
- Every API it leans on exists on main: `spinUp` (harness.ts:140), `driverUserId` (harness.ts:71/178), `driver.primeDialogs` (driver.ts:172), `driver.canResolve` (driver.ts:190), `driver.sendText` (driver.ts:200), `expectMessage(driver, chatId, match, opts)` (assertions.ts:147). Asserted fields `text`/`chatId`/`fromBot` are all on `ObservedMessage` (driver.ts:55-64). `senderFilter:{notUserId}` is the shape the harness itself uses.
- Self-skips green via early `return` when `SWITCHROOM_UAT_CHAT_ID` is unset/non-finite (line 42) or the supergroup is not resolvable after `primeDialogs` (line 49), so unwired CI stays green. uat/** is CI-isolated per the PR body.
- Race-safe: `expectMessage` only observes messages emitted after the call starts, but the worker sleeps ~8s before echoing the token, so the post-handback relay arrives well inside the live stream window; the unique `HBK<base36-ts>` token (line 56) prevents latching onto unrelated chatter.

Rationale: test-only, additive, correctly wired to real existing APIs, self-skipping so it cannot red an unwired pipeline; provides the live proof for #2098 that the unit thread-assertions cannot.

PR #2107, opened 2026-06-03
fix(telegram): honest silence-poke when blocked on an approval tap

mekenthompson:
## Summary
The 300s framework-fallback says *"still working… (no update from agent in 5 min)"* even when the turn is parked on an **approval card waiting for the operator's tap**. A prod-log investigation across the fleet (gymbro 14×, clerk 6×) found the recurring "turn-wedges" are *mostly this benign class* — claude is alive and mid-turn, blocked on a Write/Edit/Bash/MCP permission prompt, self-recovering the instant the human answers (no drops). In that state the copy actively **lies**: the agent isn't computing, it's waiting on **you**.

JTBD: **you hold the leash** — the status surface must tell you when the ball is in your court.

## Fix
The reaction controller already tracks the approval park (`setAwaiting()` on the permission-request — the same signal that suspends the stall watchdog). Expose it via `isAwaiting()`, read it at the `onFrameworkFallback` callsite, and have `formatFrameworkFallbackText` emit:

> waiting for your approval — tap Approve or Deny on the card above (N min)

instead of *"still working…"*. Takes precedence over the in-flight-tool framing (a tool *awaiting approval* isn't "running"). **Pure copy/branch — the 300s self-recovery behaviour is unchanged.**

## Test plan
- [x] +3 `formatFrameworkFallbackText` cases (approval copy; precedence over an in-flight tool; back-compat default) + `isAwaiting()` state-transition test
- [x] 73 status-reactions + silence-poke tests pass
- [x] `tsc --noEmit` clean; `check-bot-api-wrapping.sh` clean

## Risk
Minimal — a copy branch gated on an existing state signal. When `isAwaiting()` is false (every non-approval turn) the wording is byte-identical to today. No change to *when* the fallback fires or to the unwedge/self-recovery.

Source: the turn-wedge root-cause investigation (follow-up thread 1). 1 of 3 fix-now follow-ups; #5 (post-ack feed) and #2 (sibling-purge) follow as separate PRs.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -3912,10 +3912,18 @@ silencePoke.startTimer({
       // (CC-4 in `docs/status-ask-cause-classes.md`). Derives "N min" suffix
       // from `ctx.silenceMs` so the wording stays honest if the 300s
       // threshold is tuned.
+      // Honesty: if the turn is parked on an approval card (the dominant
+      // benign "wedge" class — claude is alive, waiting on the operator's
+      // tap), say so instead of "still working…". The reaction controller
+      // already tracks this (setAwaiting on the permission-request park).
+      const blockedOnApproval = activeStatusReactions
+        .get(statusKey(ctx.chatId, ctx.threadId))
+        ?.isAwaiting() ?? false
       text = silencePoke.formatFrameworkFallbackText(
         ctx.fallbackKind,
         ctx.silenceMs,
         ctx.inFlightTools,
+        blockedOnApproval,
       )
     }
     try {
--- a/telegram-plugin/silence-poke.ts
+++ b/telegram-plugin/silence-poke.ts
@@ -264,8 +264,17 @@ export function formatFrameworkFallbackText(
   fallbackKind: 'working' | 'thinking',
   silenceMs: number,
   inFlightTools: ToolSnapshot[] = [],
+  blockedOnApproval = false,
 ): string {
   const minutes = Math.max(1, Math.round(silenceMs / 60_000))
+  // The turn isn't stalled — it's parked on an approval card waiting for YOUR
+  // tap (the dominant live "wedge" class is benign approval-latency, not a
+  // hang). Saying "still working…" here actively lies; name the real blocker so
+  // the operator knows the ball is in their court. Takes precedence over the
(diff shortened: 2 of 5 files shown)


mekenthompson:
APPROVE

- `status-reactions.ts:283-287` (`isAwaiting()`) reads the new `awaitingApproval` flag gated on `!finished`. The flag is set in `scheduleState` (line ~291: `this.awaitingApproval = state === 'awaiting'`), which runs BEFORE every early-return in that method (emoji-null, emoji-coalesce), so the flag is correctly latched even when the 🙏 emoji is suppressed or coalesced. ✓
- Un-park is real: `setThinking()`/`setTool()` route through `scheduleState` with a non-awaiting state → `awaitingApproval = false`. `setAwaiting()` → `scheduleState('awaiting')` → true. `finalize()` → `finishWithState` sets `finished = true` (status-reactions.ts:319), so post-finalize `isAwaiting()` is false. The +`isAwaiting()` state-transition test (status-reactions.test.ts) asserts exactly this sequence. ✓
- Gateway callsite `gateway.ts:3913-3915`: `activeStatusReactions.get(statusKey(ctx.chatId, ctx.threadId))?.isAwaiting() ?? false` — `activeStatusReactions` is the live `Map<string, StatusReactionController>` (gateway.ts:1254), `statusKey(chatId: string, threadId?)` matches (gateway.ts:1703); `?.` + `?? false` safely degrade to today's wording if no controller is registered. ✓
- Precedence is correct: `formatFrameworkFallbackText` returns the approval copy (silence-poke.ts:270-272) BEFORE the in-flight-tool / "still working…" branches, so "a tool awaiting approval isn't running" holds. Default param `blockedOnApproval = false` keeps every existing caller byte-identical. The +3 format tests pin the approval copy, the tool-precedence case, and the back-compat default. ✓
- Behaviour-preserving: pure copy/branch on an existing state signal. No change to WHEN the 300s fallback fires or to the unwedge/self-recovery path. `cancel()` sets `finished` without clearing `awaitingApproval`, but `isAwaiting()` gates on `!finished`, so that's harmless.

Rationale: honest-status copy fix, correctly wired to the pre-existing `setAwaiting` park signal, precedence and back-compat both right, and fully covered by the added unit tests. No product-behaviour change beyond the message string when a turn is genuinely parked on an approval card.

PR #2108, opened 2026-06-03
fix(telegram): keep activity feed alive after an ack-first reply (#5)

mekenthompson:
## Summary
The live activity feed is deleted on the **first** reply (`wasFirstReply`/`replyCalled`). But **ack-first** ("On it" → then delegate/work) is the live fleet pattern (Option-B), so the interim ack deletes the feed and the subsequent sub-agent/tool work renders into nothing — the **"agent went silent after On it"** gap.

JTBD: **you hold the leash** — you should see what the agent is doing between the ack and the real answer.

## Fix
Gate the feed lifecycle on `turn.finalAnswerDelivered` instead of "any reply":
- **feed-clear** (tool_use → reply tool, gateway.ts ~8167): clear only when the FINAL answer landed.
- **feed-stop** (tool_label driver, ~8206): keep narrating until the final answer, so post-ack work has a live surface.

`finalAnswerDelivered` is set by `executeReply` (via `isFinalAnswerReply`) **before** the bridge tool_use event fires — reliably false for interim acks, true for the real answer. The **turn_end `clearActivitySummary` stays unconditional** as the idempotent no-reply / ordering-race safety net.

## Test plan
- [x] tsc + 52 feed/activity/handback/watcher unit tests + bot-api-wrapping clean
- [ ] **Canary UAT (DM + supergroup):** `jtbd-foreground-subagent-activity` + `jtbd-worker-activity-feed` (exercise ack-first → post-ack work). Answer-stream-lane contention is a runtime interaction — the session-event handler isn't unit-testable in isolation, so UAT is the validation tool. **MUST pass before fleet rollout.**

## Risk
Moderate — touches the feed↔answer-stream hand-off. Mitigations: the change is a 2-condition swap on an existing field; turn_end's unconditional clear remains the net; gated behind the same finalAnswerDelivered classifier executeReply already trusts. Source: turn-wedge follow-up thread 3, item #5 (1 of 3 fix-now). Sibling: #2107 (honesty), #2 sibling-purge (next).

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -8154,17 +8154,19 @@ function handleSessionEvent(ev: SessionEvent): void {
       // Phase tracking removed in #553 PR 5 — phases only fed the
       // placeholder-heartbeat label, which has been retired.
       if (isTelegramReplyTool(name)) {
-        const wasFirstReply = !turn.replyCalled
         turn.replyCalled = true
         if (turn.orphanedReplyTimeoutId != null) {
           clearTimeout(turn.orphanedReplyTimeoutId)
           turn.orphanedReplyTimeoutId = null
         }
-        // The model's real reply takes over as the authoritative
-        // surface, so delete the activity feed message — the user
-        // sees the real reply land in the same beat the feed
-        // disappears. turn_end is the no-reply safety net.
-        if (wasFirstReply) {
+        // Delete the activity feed only when the FINAL answer has landed —
+        // NOT on an ack-first interim reply ("On it"). Gating on the first
+        // reply deleted the feed on the ack, so the post-ack work
+        // (sub-agents/tools) rendered into nothing — the "agent went silent
+        // after On it" gap. `finalAnswerDelivered` is set by executeReply
+        // (isFinalAnswerReply) before this tool_use event fires; turn_end
+        // (below) clears unconditionally as the idempotent no-reply / race net.
+        if (turn.finalAnswerDelivered) {
           clearActivitySummary(turn)
         }
       }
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

- `gateway.ts:8169` (feed-clear) and `gateway.ts:8209` (feed-stop) both swap the gate from `turn.replyCalled` → `turn.finalAnswerDelivered`. The field is real (`CurrentTurn.finalAnswerDelivered`, gateway.ts:1489), initialized false per turn (gateway.ts:8001), and set true only under the `isFinalAnswerReply` classifier in `executeReply` (gateway.ts:5831-5836) and `stream_reply` (gateway.ts:6483-6490) — the same #1664 classifier already trusted elsewhere. No fabricated symbols.
- Core fix is statically sound: an interim ack ("On it") is `disable_notification:true` + short, so `isFinalAnswerReply` is false → `finalAnswerDelivered` stays false → the feed survives the ack and keeps narrating post-ack sub-agent/tool work. That is exactly the "silent after On it" gap, closed.
- Safety net intact: `turn_end` still calls `clearActivitySummary` unconditionally (idempotent no-reply / race net), so the feed is always cleaned up even if the final-answer path is missed.
- Residual risk is event-ordering, not logic: the feed-clear at 8169 runs inside the reply tool's `tool_use` session-event handler, and relies on `executeReply` having set `finalAnswerDelivered` before that IPC event is processed. If the order ever inverts on the FINAL answer, the worst case is the feed lingering until `turn_end` (or a brief re-render flicker via 8209) — a benign degradation, never a stuck/missing feed. This is the runtime interaction the PR flags as not unit-testable; the canary UAT (`jtbd-foreground-subagent-activity` + `jtbd-worker-activity-feed`, DM + supergroup) is the right validation and its checkbox is still open.

Rationale: clean 2-condition gate swap on an existing, correctly-set field; fixes the real ack-first feed-deletion bug; the only unverifiable-by-static-analysis piece (tool_use vs executeReply ordering) fails safe via the unconditional turn_end clear. Merge gated on the stated canary UAT passing.

PR #2109, opened 2026-06-03
fix(telegram): don't purge live sibling supergroup topics on silence-poke (#2)

mekenthompson:
## Summary
The silence-poke fallback sweeps sibling turn-state for the firing chat (`purgeStaleTurnsForChat`) as defense-in-depth against dangling keys — matching on **chatId only**, with the now-false comment *"any sibling is by definition stale."* In **one-agent-owns-supergroup**, every forum topic shares the agent's chatId, so a 300s poke on topic A purges a **LIVE sibling topic B's** reaction controller + typing loop — B silently loses its status surface mid-turn (the gymbro/klanker held-mid-turn wedge class, now reachable per-topic).

JTBD: **always-available** — a live topic must keep its status surface when an unrelated topic times out.

## Fix
Gate each sibling on its **own** silence clock:
- Add an `isStale` predicate to `purgeStaleTurnsForChat` (default always-stale → DM / single-topic **back-compat**).
- The gateway passes a predicate that purges a sibling only when it is itself silent ≥ the fallback threshold (its own poke would fire too) **or** has no silence-poke state (dangling).
- New `silencePoke.silenceMsForKey(key, now)` returns `now − (lastOutboundAt ?? turnStartedAt)` — **silence**, not turn-start age, so a long-but-actively-narrating turn isn't mistaken for stale.

The #1556 dangling-key cleanup is preserved; live siblings are spared.

⚠️ **Caveat baked in (per the investigation):** the predicate uses an **absolute per-sibling silence threshold** (matching the fallback), NOT a chatId match — using anything weaker would reopen the #1556 held-forever wedge.

## Test plan
- [x] +3 `turn-state-purge` cases (live sibling spared; all-live purges nothing; default back-compat) + `silenceMsForKey` (silence from outbound/start; null when unknown)
- [x] 54 unit tests pass; tsc + bot-api-wrapping clean
- [ ] Canary: a 2-topic supergroup where topic A times out while topic B is live — confirm B keeps its reaction/typing surface. (mtcute has no multi-topic create API, so this is partly pinned by the unit tests; UAT covers the single-topic no-regression.)

## Risk
Low-moderate — touches the silence-poke unwedge primitive. Mitigation: back-compat default preserves DM/single-topic behaviour exactly; the predicate only ever makes the sweep MORE conservative (skips live siblings), never less; the firing key itself is always purged. Source: turn-wedge follow-up thread 3, item #2 (3 of 3 fix-now). Siblings: #2107 (honesty, merging), #2108 (post-ack feed, merging).

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -4041,16 +4041,29 @@ silencePoke.startTimer({
     // SAME chat (different threads, or a `null` vs `undefined`-thread
     // variant left over from a normal turn-end path that nulled
     // currentTurn without invoking purgeReactionTracking — the
-    // gymbro/klanker held-mid-turn symptom, 2026-05-20). Any sibling
-    // for fbChatId is by definition stale when THIS fallback fires
-    // (the chat has been silent ≥5 min); sweep them via the same
-    // purger. Multi-chat-safe — only touches keys for fbChatId, so
+    // gymbro/klanker held-mid-turn symptom, 2026-05-20); sweep them via
+    // the same purger. Multi-chat-safe — only touches keys for fbChatId, so
     // #1546's intentional cross-chat safety guard is preserved.
+    //
+    // BUT a sibling is NOT "by definition stale": in one-agent-owns-supergroup
+    // every forum topic shares fbChatId, so a chatId-only sweep would purge a
+    // LIVE sibling topic's reaction controller + typing loop when THIS topic's
+    // poke fires. Gate each sibling on its OWN silence clock — purge only those
+    // also silent ≥ the fallback threshold (their own poke would fire too),
+    // sparing topics that are actively mid-turn. Use silence, not turn-start
+    // age, so a long-but-narrating turn isn't mistaken for stale.
     // See turn-state-purge.ts.
+    const fbNow = Date.now()
     const fbExtraPurge = purgeStaleTurnsForChat(
       fbChatId,
       activeTurnStartedAt.keys(),
       purgeReactionTracking,
+      (siblingKey) => {
+        if (siblingKey === fbKey) return true // the firing key is genuinely stale
+        const sib = silencePoke.silenceMsForKey(siblingKey, fbNow)
+        // No silence-poke state → dangling (turn ended, key not purged) → stale.
+        return sib == null || sib >= silencePoke.DEFAULT_THRESHOLDS.fallback
+      },
     )
     // Null `currentTurn` if it's still pointing at the wedged turn —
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- `turn-state-purge.ts:59` adds `isStale: (key) => boolean = () => true` and applies it at line 80 (`if (!isStale(key)) continue`) AFTER the `keyChat !== chatId` guard and BEFORE `purger(key)`. So it only ever filters chatId-matched siblings, and the default preserves DM/single-topic back-compat exactly (every sibling stale). #1546 cross-chat safety is untouched. ✓
- `silence-poke.ts:silenceMsForKey` returns `now - (s.lastOutboundAt ?? s.turnStartedAt)` against the real state shape (`turnStartedAt: number` line 61, `lastOutboundAt: number | null` line 63) — the same clock `tick()` uses, and null when no state exists. Correctly measures SILENCE not turn-age, so a long-but-narrating turn (recent `noteOutbound` → low silence) is spared. ✓
- Gateway callsite (gateway.ts:4060-4067): `fbKey = ctx.key` (3957), `fbChatId = ctx.chatId` (3958), `fbNow = Date.now()`. Predicate: firing key (`siblingKey === fbKey`) → always stale (the timed-out topic is always unwedged); else purge only when `sib == null` (dangling → preserves #1556 cleanup) OR `sib >= DEFAULT_THRESHOLDS.fallback` (300_000, line 87 — its own poke would also fire). A live sibling with low silence is spared. ✓
- The predicate only ever makes the sweep MORE conservative (never purges a sibling the old chatId-only match would have spared), and the firing key is always purged — so the unwedge primitive can't regress while live topics are protected.
- Tests cover the matrix: live-sibling-spared, all-live-purges-nothing, default back-compat, and `silenceMsForKey` (from outbound / from start / null on unknown+ended). ✓

Nit (non-blocking): a sibling topic genuinely mid-turn but emitting no outbound for ≥300s would be classed stale and purged — but that is exactly the threshold at which that topic fires its OWN fallback, so the behaviour is self-consistent with existing silence-poke semantics, not a new gap.

Watch: the 2-topic canary (topic A times out while B is live) is only pinned by unit tests — mtcute has no multi-topic create API, so the live multi-topic path isn't UAT-exercised; single-topic no-regression is.

Rationale: correct, minimal, strictly-more-conservative gate on the sibling purge; every symbol (state fields, DEFAULT_THRESHOLDS.fallback, fbKey/fbChatId) verified present; back-compat default preserves DM behaviour and the #1556 dangling-key cleanup; fully unit-covered.

PR #2111, opened 2026-06-03
fix(telegram): surface operator MCP tools in the live activity feed

mekenthompson:
## Summary
A research turn driven by operator MCP tools (perplexity, webkite, …) showed **no activity feed** — only a typing dot + 👀 — so it read as 'I can't see what it's doing' (operator report; confirmed in the session JSONL: `perplexity_ask`×2, `perplexity_search`, `webkite_read`×3 over 73s, **zero `tool_label` events**).

## Root cause
`tool-label-pretool.mjs` `computeLabel` **allowlisted only switchroom-telegram + hindsight** and returned `null` for every other `mcp__*` tool. The live activity feed is driven off this sidecar, so research tools produced nothing. (The gateway's `describeToolUse` already has a generic MCP fallback — the sidecar never got it.)

## Fix
Generic MCP fallback in `computeLabel` mirroring `describeToolUse`: friendly per-server labels (perplexity → 'Searching the web for …', webkite → 'Reading <host>', gdrive/gmail/notion/calendar), else a model-authored field, else 'Using <tool>'. switchroom-telegram surface tools stay suppressed.

## Test plan
- [x] tsc clean; +6 computeLabel cases (perplexity/webkite/gdrive/unknown MCP labeled; surface tools null; built-ins unchanged); tool-label-sidecar + tool-labels suites green (68 total)
- [ ] Live: re-run an MCP research turn post-release → activity feed shows 'Searching the web for …' / 'Reading …'

## Risk
Low. Pure label-derivation change in a PreToolUse hook (never blocks tools — exit 0). Baked into the agent image → needs a release + rollout to take effect.

### DIFF
--- a/telegram-plugin/hooks/tool-label-pretool.mjs
+++ b/telegram-plugin/hooks/tool-label-pretool.mjs
@@ -133,8 +133,9 @@ export function computeLabel(toolName, input) {
     }
   }
 
-  // MCP allowlist.
+  // MCP tools.
   if (typeof toolName === 'string' && toolName.startsWith('mcp__')) {
+    // Explicit labels / suppressions for the built-in servers.
     switch (toolName) {
       case 'mcp__switchroom-telegram__reply':
       case 'mcp__switchroom-telegram__stream_reply':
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `telegram-plugin/hooks/tool-label-pretool.mjs:135-194` — the old `mcp__*` branch returned `null` for everything not on the switchroom-telegram/hindsight allowlist, so operator-configured MCP servers (perplexity, webkite, gdrive, …) drove the activity feed silently. The new generic fallback fixes that: friendly per-server labels, else a model-authored field (`description`/`query`/`title`), else a humanized tool name. Additive only — the named `switch` cases above are unchanged.
- `tool-label-pretool.mjs:166` — surface/control safety holds: any `mcp__switchroom-telegram__*` tool not in the explicit switch (e.g. `progress_update`) falls into the generic block and is force-suppressed by `if (server === "switchroom-telegram") return null`. Verified: `progress_update` → null.
- Helpers all exist on head (42ca03f9): `clip` (:42), `urlHostPath` (:59), `i = input ?? {}` (:75). No undefined refs in the new code.
- Verified by importing the real `.mjs` and running every test assertion through `computeLabel` directly (deps not installed locally for vitest): all 14 cases pass — perplexity query label, webkite host/path, gdrive/gmail/notion friendly labels, model-authored `description` passthrough, and the `Using <tool>` humanized fallback. The +70-line test file (`tests/tool-label-pretool.test.ts`) covers the regression (no generic MCP tool returns null) plus the built-in allowlist as a guard.
- Nit (non-blocking): a model-authored `description`/`query`/`title` is surfaced verbatim (clipped to 60) into the feed — same trust model as the existing `Bash`/`WebSearch` labels, and the feed is the operators own chat, so no exfil concern. Worth a glance only if untrusted MCP servers are ever added.

Rationale: purely additive label-coverage fix; named cases byte-identical, telegram surface tools still suppressed, all symbols verified on head, and every documented behavior reproduced by direct execution of the shipped function.

mekenthompson:
Fresh-process reviewer: **APPROVE**. Verified parity with describeToolUse, switchroom-telegram surface tools stay suppressed (incl. unlisted ones via the lowercased server guard), no crash path (try/catch → exit 0), built-ins unchanged. 68/68 tests, tsc clean, CI CLEAN. Residual (non-blocking, pre-existing class): the activity-feed live-send path bypasses redact() — widening surfaced MCP fields marginally grows leak blast radius, same class as existing Bash/WebSearch surfacing, bounded by the inbound fail-closed gate; the tracked follow-up is the live-send transport-mask (apply redact() in drainActivitySummary).

Issue #421, opened 2026-04-30
auth: pin banner only in non-default slot state; soft-confirm /auth use when turn is in-flight


mekenthompson:
## Problem A — no always-visible slot indicator after failover

Post-failover, the current scrolling Telegram notification message tells the user the swap happened, but it scrolls out of view as the conversation continues. The user has no always-visible answer to "what slot am I on right now?" without typing `/auth status`. After a long conversation following a failover, this is easy to forget.

## Problem B — `/auth use` silently kills in-flight turns

A manual `/auth use <slot>` swap silently kills any in-flight turn. A fat-finger tap on the wrong slot costs the user whatever work was in progress. There's no confirmation when it matters and no friction when it doesn't.

## Proposed solution A — conditional pinned banner

Pin a slot-state banner **only** when the active slot is NOT the default. Auto-unpin when the user reverts to the default.

- State = default → no pin (avoids pin clutter when nothing is wrong).
- State ≠ default → one pinned message: "Active slot: `<name>` (failover from `<default>`)" or similar.
- In forum chats, pin per-topic so each topic shows its own banner if relevant.
- Update the pinned message in place when the slot changes again; unpin when state returns to default.

This gives the user a persistent answer to "where am I" exactly when they need it, and zero noise when everything is normal.

## Proposed solution B — soft-confirm `/auth use` only when a turn is active

When `/auth use` is triggered and a turn is currently running, reply with a single-line warning instead of silently killing the turn:

> Current turn will abort if you swap. Tap `/confirm` (or send `/auth use <slot> --force`) to proceed.

When `/auth use` is triggered while idle, swap proceeds with no prompt — no friction tax for the common case.

This makes the destructive case explicit without adding ceremony to safe swaps.

## Acceptance criteria

- Banner pin appears only when active slot ≠ default.
- Banner unpins automatically when the active slot returns to default.
- In forum chats, pin is per-topic, not chat-wide.
- `/auth use <slot>` while idle: swaps immediately with no confirm prompt.
- `/auth use <slot>` while a turn is in flight: replies with confirm prompt; swap happens only on `/confirm` (or `--force`).
- Existing slot subcommands `add`/`list`/`rm` are unaffected.

## Test plan

- Integration tests for both scenarios in both states (idle + in-flight × default + non-default).
- Manual test on a real Telegram chat to verify pin/unpin behavior in forum topics specifically (the per-topic pin behavior is easy to get wrong).
[...]

PR #2113, opened 2026-06-03
fix(telegram): topic-safe ops-lane sends + threaded typing (supergroup)

mekenthompson:
## Summary
From the DM-hardcoding → supergroup audit (gaps #3 + #4). Two ops-lane outbound paths attach a supergroup-resolved topic to operator **DM** recipients **without** the per-recipient guard — the same latent `400 message thread not found` class #2096 fixed for approval cards, except here the event **silently fails to deliver** for a supergroup-owned agent (DM `allowFrom`):
- **operator-event broadcast** (boot-crash / restart-watchdog / model-unavailable) — topic attached to every `allowFrom` chat unconditionally.
- **boot card** — topic attached to `envChat` / `allowFrom[0]` unconditionally.

Both now route through `topicForRecipient({recipientChatId, resolvedTopic, supergroupChatId})`: a DM recipient gets a **thread-less** send (no 400); the supergroup owner gets the boot/alerts lane. **No destination redesign** — DM-config agents are byte-identical.

Plus #4: the main `handleInbound` **typing** indicator was un-threaded → showed in General on a supergroup-topic inbound. Now carries the inbound's `message_thread_id`.

## Scope notes (verified during the audit re-check)
- The audit's "marko wedge in approval emitters" and "reactions in wrong topic" findings were **false** (already guarded by #2096; `setMessageReaction` has no thread param). This PR is the *genuinely-unguarded* ops-lane subset.
- The early-ack typing is already DM-only-guarded (not a gap). The fleet auto-fallback broadcast is *deliberately* DM-only/no-thread (no 400) — routing it to the alerts lane is a destination-design choice, deferred with the slot-banner (#421).

## Test plan
- [x] `topicForRecipient` already unit-tested (`topic-router.test.ts`); wiring verified by tsc
- [ ] Canary: a supergroup agent fires a boot/operator-event → lands (no silent 400)

## Risk
Low — DM agents unchanged (guard returns undefined → no thread, exactly today's behavior); only ever *removes* an invalid thread or routes a supergroup recipient correctly. 1 of the supergroup-gap PR series (vault-grant #2 + setup/docs #1 follow).

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -3537,18 +3537,24 @@ function emitGatewayOperatorEvent(event: OperatorEvent): void {
   // Fleet-shared / DM agents see `undefined` → no `message_thread_id`
   // is added to the broadcast opts → behavior unchanged.
   const opEventTopic = resolveAgentOutboundTopic({ kind: 'compact-watchdog' })
+  const opEventSupergroup = resolveAgentSupergroupChatId()
 
   process.stderr.write(
     `telegram gateway: operator-event posting agent=${agent} kind=${kind} to ${access.allowFrom.length} chat(s)` +
       (opEventTopic != null ? ` topic=${opEventTopic}` : '') + '\n',
   )
   for (const chat_id of access.allowFrom) {
+    // The resolved topic is valid ONLY in the agent's supergroup — attaching
+    // it to an operator DM recipient yields 400 "message thread not found" and
+    // the event silently fails to deliver (the marko #2096 class). Guard it:
+    // DM recipients get a thread-less send; the supergroup owner gets the lane.
+    const opEventThread = topicForRecipient({ recipientChatId: chat_id, resolvedTopic: opEventTopic, supergroupChatId: opEventSupergroup })
     // grammy's Other<...> opts type is generated and stricter than our
     // call shape; runtime accepts both. Cast through unknown.
     const opts = {
       parse_mode: 'HTML' as const,
       ...(renderedKeyboard ? { reply_markup: renderedKeyboard } : {}),
-      ...(opEventTopic != null ? { message_thread_id: opEventTopic } : {}),
+      ...(opEventThread != null ? { message_thread_id: opEventThread } : {}),
     }
     // Comment-only context for the reader; the lint marker on the
     // very next line is what unlocks the raw bot.api call.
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

- `gateway.ts:3551` — operator-event send (compact-watchdog etc.) previously attached the resolved supergroup topic to EVERY recipient in `access.allowFrom`, including operator DMs → 400 "message thread not found", event silently dropped (the marko #2096 class). Now `topicForRecipient({recipientChatId: chat_id, resolvedTopic: opEventTopic, supergroupChatId: opEventSupergroup})` is computed per-recipient inside the loop, with the supergroup resolved once outside it. Correct: a DM recipient gets `undefined` → thread-less send; the supergroup owner gets the lane.
- `gateway.ts:10170-10174` — typing indicator in `handleInbound` was an un-threaded `sendChatAction`, so on a supergroup-topic inbound it showed "typing" in General instead of the user's topic. Now passes `message_thread_id: messageThreadId` when set. `messageThreadId` is the inbound's own thread (defined at :9586, `undefined` in a DM → `{}`), matching the existing thread-guard pattern already used at :9801. Correct.
- `gateway.ts:11111,11119` — boot-chat resolution: `supergroupBootTopic` is now per-recipient guarded via `topicForRecipient` on both the env-var and history branches, so a DM owner gets a thread-less boot card instead of a 400. Consistent with the two op-event/typing fixes.
- Verified `topicForRecipient` semantics on head (`src/telegram/topic-router.ts:157`): returns `resolvedTopic` ONLY when `String(recipientChatId) === String(supergroupChatId)`, else `undefined` — exactly the per-recipient guard the comments describe. `resolveAgentSupergroupChatId` (gateway.ts:11170) reads `channels.telegram.chat_id` from the merged agent config, best-effort try/catch → `undefined` on any failure (degrades to thread-less, never throws). This PR extends a pattern already live at 5 other callsites (:2287, :4708, :4853, :4931, :5015).
- Nit (non-blocking): no new unit test in the diff (+21/-4, gateway only). The load-bearing logic is the `topicForRecipient` helper, which is centralized and unit-testable at the topic-router level; the three new callsites are mechanical applications of it. Worth a one-line test that an op-event to a DM recipient drops the thread, but not a blocker.

Rationale: tightly-scoped, correct extension of a proven per-recipient topic guard — eliminates the DM 400 on operator-event + boot sends and fixes topic-local typing; helper semantics verified on head, degrades safely to thread-less on every failure path.

PR #2114, opened 2026-06-03
fix(telegram): route vault grant/save/secret outcomes back to the work topic

mekenthompson:
## Summary

Supergroup mode follow-up (gap #2 of the DM-hardcoding→supergroup audit). When an agent fires `vault_request_access`, `vault_request_save`, or `request_secret` from inside a forum topic, the synthetic wake-up inbound the gateway injects after the operator taps **Approve / Save / Provide** carried **no thread** — so the resumed turn's reply landed in **General** instead of the topic the agent was working in.

This threads the originating topic end-to-end across all three credential flows.

## Why

This is the same class as #2096 (approval-card THREAD_NOT_FOUND) and #2098 (cron synthesized inbounds), but for the *resume* side of the vault/secret request flows. An agent working in topic *X* that needs a credential would, after approval, reply in General — invisible in the topic where the work lives. Supergroup mode (`channels.telegram.chat_id` + `default_topic_id`) makes this user-visible; DM agents were unaffected (no threads).

## What changed

Each pending-request struct captures the working topic at card-post time (`args.message_thread_id`):
- `PendingVaultRequestAccess`, `PendingVaultRequestSave`, `PendingSecretRequest`, `ArmedSecretCapture` each gain an optional `threadId`.

Every outcome inbound now carries it **both ways** (the established two-carrier pattern):
- **top-level `threadId`** → drives the gateway's per-topic busy-key / deliver-until-acked keying (`markClaudeBusyForInbound`).
- **`meta.message_thread_id`** (stringified) → rendered into the `<channel message_thread_id="…">` envelope, which `session-tail`'s `parseChannelMeta` re-extracts to set `currentTurn.sessionThreadId`, which the reply tool defaults to.

Covers the 8 outcome inbounds across the three flows: `vault_grant_approved`, `vault_grant_denied`, `vault_save_completed`, `vault_save_failed`, `vault_save_discarded`, `secret_provided`, `secret_provide_failed`, `secret_declined`.

**Purely additive** — DM / non-topic requests leave both fields absent, so the thread-less send is unchanged (an empty/0 thread is a Telegram 400 "message thread not found").

## Test plan

- [x] `vault-grant-inbound-builders.test.ts`: 10 new cases pin `threadId` set → top-level `threadId` + stringified `meta.message_thread_id` on all 5 builders, and `threadId` absent → both omitted (DM stays thread-less). 23 pass.
- [x] `tsc --noEmit` clean
- [x] adjacent vault / channel-envelope / session-tail suites green

## Risk

Low — optional field, no behavior change on the DM path. Worst case under a bug would be a thread-less send (current behavior), never a crash.

JTBD: *always-available specialist that holds the leash* — credential approvals resume the conversation where the user is looking.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -2913,6 +2913,10 @@ interface PendingVaultRequestSave {
   chat_id: string
   /** Card message id (filled in after we send the card). */
   card_message_id?: number
+  /** Supergroup forum topic the agent was working in when it requested the
+   *  save — carried into the save-outcome inbound so the resumed reply lands
+   *  back in that topic, not General. */
+  threadId?: number
   /** Currently-suggested slug; may be renamed by the user. */
   key: string
   /** Storage shape — 'string' (default) or 'binary'. */
@@ -2954,6 +2958,10 @@ interface PendingVaultRequestAccess {
   chat_id: string
   /** Card message id (filled in after we send the card). */
   card_message_id?: number
+  /** Supergroup forum topic the agent was working in when it requested (the
+   *  card's originating thread). Carried into the grant-outcome inbound so the
+   *  resumed reply lands back in that topic, not General. */
+  threadId?: number
   /** Vault key the agent wants to read. */
   key: string
   /** 'read' (default) or 'write'. */
@@ -7080,6 +7088,8 @@ async function executeVaultRequestSave(args: Record<string, unknown>): Promise<{
   // crashing the tool call.
   const text = renderVaultRequestSaveCard(pending, agentSlug)
   const threadId = args.message_thread_id != null ? Number(args.message_thread_id) : undefined
+  // Remember the agent's working topic so the save-outcome inbound resumes in it.
+  if (threadId != null) pending.threadId = threadId
   const sent = await retryWithThreadFallback<{ message_id: number }>(
     robustApiCall,
     (tid) =>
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `gateway.ts:7090-7091, 7210-7211, 7565-7566` — at card-send time each pending vault request (save / secret / access) now records `pending.threadId = threadId` (derived from `args.message_thread_id`), only when non-null. This is the originating-topic capture; without it the outcome inbound has no topic to resume into. Correct and guarded.
- `vault-grant-inbound-builders.ts` (all five builders: approved/denied/save-completed/failed/discarded) and the three inline secret inbounds (`gateway.ts:7287-7298, 7319-7332, 7405-7418`) carry the topic as a TWO-FOLD carrier: top-level `threadId` + stringified `meta.message_thread_id`, both via conditional spread. Verified both consumers exist on head:
  - top-level `threadId` → `markClaudeBusyForInbound` (gateway.ts:1322-1326) reads `m.threadId` first (falls back to `meta.message_thread_id`) for the per-topic busy / deliver-until-acked key.
  - `meta.message_thread_id` → rendered into the `<channel message_thread_id="…">` XML; the reply tool defaults `args.message_thread_id` from `turn.sessionThreadId` (gateway.ts:6261-6262), which is parsed from that meta. Drop either and the resumed reply mis-routes — the test comment's claim is accurate.
- `InboundMessage.threadId?: number` confirmed on the IPC contract (`ipc-protocol.ts:6`), so the top-level field is type-valid, not a silently-dropped property.
- DM / non-topic safety: every add is `...(threadId != null ? {…} : {})`, so when the request did not originate in a topic BOTH halves are omitted — the DM inbound is byte-identical to before (an empty/0 thread would be a Telegram 400). The +10 parametrized tests assert exactly this for all five builders in both the set and absent cases.
- Pattern parity: this mirrors the existing `subagent-handback-inbound-builder` dual-carrier convention (same file family), so it is not inventing a new routing mechanism — it extends a proven one to the vault outcomes.

Rationale: purely additive topic-plumbing fix; conditional spreads keep the DM path unchanged, both carrier halves are verified against their real consumers on head, and the new behavior is fully covered by parametrized tests across every outcome builder.

PR #2115, opened 2026-06-03
feat(setup): discoverable supergroup mode — wizard step, example, guide

mekenthompson:
## Summary

Gap #1 of the DM-hardcoding→supergroup-mode audit: **discoverability**. Supergroup mode (one agent owns a Telegram forum supergroup, routing its work into per-topic threads) was real and shipping, but an operator had no way to *learn it exists* or turn it on without hand-editing `channels.telegram.chat_id` and reading the RFC. The setup wizard's forum step had been retired to a sentinel, the only user-facing mention was an aside in `scheduling.md`, and there was no example block.

This closes the gap on three surfaces.

## Why

The **docs test** (principles.md): can someone use this without opening `docs/`? Before this PR, no — supergroup mode failed it. The fix makes the feature self-revealing at setup time while keeping the **defaults test** intact: the wizard step is opt-in and **skip-by-default (one keystroke)**, so a fresh `switchroom setup` with zero config still lands DM mode exactly as before.

## What changed

1. **Wizard** — resurrect Step 4 as an optional *"Supergroup mode"* step (`src/cli/setup.ts`). Defaults to **no**. On opt-in: pick the owning agent, paste the supergroup `chat_id`, validate the `-100…` shape, and write `agents.<name>.channels.telegram.chat_id` via a pure YAML editor. The schema smart-defaults `default_topic_id` to General, so the one field is enough to activate.
2. **Example** — commented per-agent supergroup block in `examples/switchroom.yaml` (`chat_id` / `default_topic_id` / `topic_aliases`) with the id-discovery tip inline.
3. **Guide** — new `docs/supergroup-mode.md` (what it is, when to use it, how to find `chat_id` + `thread_id`, the config block, a field reference, gotchas), linked from `docs/README.md` next to `scheduling.md`.

New pure helper `src/cli/supergroup-setup-yaml.ts` — `isValidSupergroupChatId` + `setAgentSupergroupChatId` — mirrors `auth-active-yaml.ts` (parseDocument → mutate → atomic write, comments/siblings preserved; idempotent; coerces a bare `channels:` key).

## Test plan

- [x] `tests/supergroup-setup-yaml.test.ts` — 9 cases: write-under-agent, idempotent, sibling-preserving, bare-channels coercion, malformed-chat_id rejection, missing-agent throw, and a **schema-parse assertion** proving the written config validates and `default_topic_id` smart-defaults to `1`.
- [x] `tsc --noEmit` clean
- [x] merge / setup / config suites (130 tests) green
- [x] `examples/switchroom.yaml` still parses (commented block inert)

## Risk

Low. The wizard step is opt-in and skip-by-default; the helper is new and only invoked on explicit opt-in; the docs/example additions are inert. No change to any runtime routing path.

JTBD: *a standing team that knows you / always-available* — an operator can stand up a supergroup-owned specialist from the wizard instead of reverse-engineering the config.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -34,6 +34,7 @@ operators jump to **Operating**; contributors and reviewers use
 | [status-ask-cause-classes.md](status-ask-cause-classes.md) | Cause-class catalog for driving the status-ask rate to zero. |
 | [session-optimization.md](session-optimization.md) | Managing context/tokens in long-running agents. |
 | [sub-agents.md](sub-agents.md) | Sub-agent delegation ("Opus plans, Sonnet implements"). |
+| [supergroup-mode.md](supergroup-mode.md) | One agent owns a Telegram forum supergroup, routing work into per-topic threads (opt-in). |
 | [google-workspace.md](google-workspace.md) | Agents reading Google Docs/Sheets/Slides/folders. |
 | [webhook-ingest.md](webhook-ingest.md) | Pushing external events into a specific agent's log. |
 | [posthog.md](posthog.md) | Telemetry: product analytics + error tracking via PostHog. |
--- a/docs/supergroup-mode.md
+++ b/docs/supergroup-mode.md
@@ -1,0 +1,125 @@
+# Supergroup mode
+
+By default every switchroom agent talks to you in a **private DM** — one
+bot, one chat. **Supergroup mode** is the alternative topology: a single
+agent **owns a Telegram forum supergroup** and routes all of its traffic —
+your conversations *and* its automated events — into **per-topic threads**
+inside that one group.
+
+Use it when one agent does enough varied work that a flat DM becomes a
+firehose. A marketing agent, say, can keep "Meta Campaigns", "CRM", and
+"Planning" as separate topics, so each thread reads as its own focused
+conversation instead of an interleaved scroll.
+
+DM mode stays the default and needs zero of this. Supergroup mode is
+opt-in and per-agent — you can run a mixed fleet (most agents DM, one owns
+a supergroup).
+
+> Internals and the routing design live in
+> [rfcs/supergroup-mode.md](rfcs/supergroup-mode.md) and
+> [rfcs/supergroup-easy-defaults.md](rfcs/supergroup-easy-defaults.md).
+> This page is the operator how-to.
+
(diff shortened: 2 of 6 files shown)


mekenthompson:
APPROVE

Verified against head e7a4d93. Docs + an opt-in wizard step + a new pure YAML helper with tests — no runtime routing path touched.

- `src/cli/supergroup-setup-yaml.ts:35` (`isValidSupergroupChatId`) uses `/^-\d+$/`, byte-identical to the schema's chat_id regex (`src/config/schema.ts:811`) — the wizard can never write a value the schema would later reject. Good.
- `setAgentSupergroupChatId` (supergroup-setup-yaml.ts:46) mirrors `auth-active-yaml.ts` (parseDocument→mutate→stringify): preserves comments/siblings, idempotent on equal value, and coerces a bare/null `channels:` key to a fresh map before `.set` — the same crash hazard the comment calls out. Correct.
- The load-bearing claim (wizard writes only chat_id; default_topic_id is auto-supplied) holds: `schema.ts:870-871` returns `{...tg, default_topic_id: 1}` when chat_id is set and default_topic_id is null. The test at tests/supergroup-setup-yaml.test.ts:88 asserts this against the real `SwitchroomConfigSchema`. Verified.
- Defaults test intact: `stepSupergroupMode` (setup.ts) defaults the `askYesNo` to `false` and short-circuits non-interactive — a zero-config `switchroom setup` still lands DM mode. All prompt helpers (ask/askYesNo/askChoice) and `atomicWriteFileSync` resolve at head.
- Docs accurate: `docs/supergroup-mode.md` field reference ("Forbidden when dm_only: true", General=topic 1) matches the schema's dm_only mutual-exclusion (schema.ts:2166) and the smart-default. examples/switchroom.yaml block is commented/inert.

Nit (non-blocking): the wizard writes raw YAML and never runs the schema, so default_topic_id isn't materialized until `switchroom apply` — the success message correctly tells the operator to run apply, so no gap, just worth knowing the written file shows only chat_id.

Rationale: purely additive discoverability surface, opt-in and skip-by-default, every cross-file claim (regex parity, smart-default, dm_only exclusion) confirmed against head; 9 tests including a real-schema parse assertion.

PR #2117, opened 2026-06-03
fix(telegram): don't lose inbounds redelivered into a not-ready session on restart

mekenthompson:
## The live bug (clerk, 2026-06-03, hit twice)
A DM that arrived **while the agent was mid-restart** was buffered, redelivered on `bridge registered`, then **lost** — the user saw "your message is queued…" then "still working… (5 min)" then nothing. Both clerk incidents had the identical trace:

```
inbound  bridge_dead → [bufferInbound, persistInbound]      ✓ buffered
bridge registered → drainBuffer → drained=1 redelivered=1   ✓ pulled + socket-write
…claude produced ZERO activity (still booting, Hindsight timed out)…
turnEnd outbound=false; silence-poke … drained_buffered=0/0  ✗ dropped, not retried
```

## Root cause
`bridge registered` ≠ claude session **ready**. The bridgeUp drain (and the silence-poke fallback + flap flushes) ack delivery on a bare `socket.write` returning `true`, and — unlike the live `handleInbound` path — **never call `trackDelivery`**. So the existing 5s deliver-until-acked **sweep** (re-delivers while `currentTurn==null` until `enqueue` acks — it would have rescued this verbatim) never enrolled the message. The spool was also tombstoned on the socket-write, so nothing retried it.

## Phase A — stop losing it
- `redeliverBufferedInbound` gains an `onDelivered(merged)` hook (fired on confirmed delivery).
- `DispatchCtx.onUserInboundDelivered` threads it into the bridgeUp drain; the fallback + both flap flushes pass it too.
- `trackRedeliveredInbound` enrols the redelivered inbound in the deliver-until-acked queue, **keyed `chatKey(chatId, threadId)` (DM and supergroup topics identical)** with `String(messageId)` so the `enqueue` ack matches — mirroring the live tracking. `shouldTrackDelivery` still excludes steer/interrupt/synthetic-source/empty (those never enqueue → would re-deliver forever).
- Net: redelivered inbound re-sent every 5s until claude boots + enqueues, instead of dropped at 300s. **Would have caught both clerk incidents.**

## Phase C — 3h staleness failsafe (operator spec)
`selectResumeBuilder(endedVia, { ageMs, maxAgeMs })` downgrades a stale auto-resume (interrupted turn older than `RESUME_MAX_AGE_MS`, default 3h, env `SWITCHROOM_RESUME_MAX_AGE_MS`) to the passive `report` — the user is told but hours-old work never replays unprompted.

## Deferred (noted) — to avoid a storm
- **B1** (track the boot-resume synthetics) needs the resume builder to emit `meta.message_id` first — else its `enqueue` carries `null`, the deliver-until-acked ack never matches, and it **re-delivers forever**. Verified: `buildResumeInterruptedInbound`'s meta has no `message_id`. Clean follow-up (+ reply-quote side-effect check), ideally coordinated with the `feat/honest-restart-resume` work.
- **B2** (resume a background turn shadowed by a foreground turn) — `turns-schema` scan widening.

## Test plan
- [x] tsc clean; **84 vitest + 27 bun** affected tests pass; delivery-machine/ipc/gateway-bridge/spool regression suites green (68)
- [x] onDelivered fires per confirmed group / not on re-buffer; 3h cap downgrades resume→report past maxAge, keeps within, never upgrades, legacy preserved
- [ ] **MANDATORY canary** (this is the most incident-dense path): pin test-harness, restart **with Hindsight forced down**, send a DM in the boot window → must be answered (was the live failure); + a supergroup-topic variant; tail for `enqueue`/`ackDelivery` follow-through (not `drained_buffered=0/0`). Then **staggered** roll.

## Risk
Pure orchestration (inject_inbound, acked by session `enqueue`; no claude -p/SDK/API). Kill-switch: existing `SWITCHROOM_INBOUND_DELIVERY_CONFIRM`. Highest residual: ack-key correctness — guarded by tracking the merged anchor `messageId` (matches the live path) + the synthetic-source exclusion.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1064,7 +1064,19 @@ try {
   const pending = findLatestTurnIfInterrupted(turnsDb)
   const selfAgent = process.env.SWITCHROOM_AGENT_NAME ?? ''
   if (pending != null && selfAgent) {
-    const kind = selectResumeBuilder(pending.ended_via)
+    // 3h staleness failsafe (operator spec, 2026-06-03): never AUTO-resume
+    // interrupted work older than RESUME_MAX_AGE_MS — selectResumeBuilder
+    // downgrades a stale 'resume' to the passive 'report' so the user is told
+    // ("I was working on X ~Nh ago") but nothing replays unprompted. Env
+    // override SWITCHROOM_RESUME_MAX_AGE_MS (ms); set very high to disable.
+    const RESUME_MAX_AGE_MS = (() => {
+      const v = Number(process.env.SWITCHROOM_RESUME_MAX_AGE_MS)
+      return Number.isFinite(v) && v > 0 ? v : 10_800_000 // 3h
+    })()
+    const kind = selectResumeBuilder(pending.ended_via, {
+      ageMs: Math.max(0, Date.now() - pending.started_at),
+      maxAgeMs: RESUME_MAX_AGE_MS,
+    })
     if (kind === 'resume') {
       bootResumeInbound = { agent: selfAgent, msg: buildResumeInterruptedInbound({ turn: pending }) }
     } else if (kind === 'report') {
@@ -1801,6 +1813,7 @@ function purgeReactionTracking(key: string, endingTurn?: CurrentTurn): void {
           return d
         },
         inboundSpool,
+        trackRedeliveredInbound,
       )
       if (fr.redelivered > 0) {
         process.stderr.write(
@@ -1896,6 +1909,7 @@ function releaseTurnBufferGate(key: string): void {
           return d
         },
         inboundSpool,
+        trackRedeliveredInbound,
       )
       if (fr.redelivered > 0) {
         process.stderr.write(
@@ -4110,6 +4124,7 @@ silencePoke.startTimer({
         return d
       },
       inboundSpool,
+      trackRedeliveredInbound,
     )
     process.stderr.write(
       `telegram gateway: silence-poke framework-fallback ended wedged turn ` +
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

Verified against PR head (pull/2117/head). This closes a real, twice-reproduced lost-inbound boot-race: `bridge registered` ≠ claude session ready, and the redelivery paths acked on a bare socket-write without enrolling in the deliver-until-acked queue, so a DM injected into a still-booting session was silently dropped at the 300s mark.

- gateway.ts:4163-4197 — `trackRedeliveredInbound` mirrors the live tracking site (gateway.ts:10895-10902) exactly: same `shouldTrackDelivery` gating (excludes steer/interrupt/synthetic-source/empty), same `trackDelivery(deliveryQueue, key, msg, now, String(messageId))`. The key `chatKey(merged.chatId, merged.threadId)` matches the `ackDelivery` key at gateway.ts:8151 (`chatKey(ev.chatId, ev.threadId)`), so the `enqueue` ack resolves the right entry for both DMs and forum topics. Correct.
- pending-inbound-buffer.ts:90-124 — `onDelivered` fires only after `spool.ack` on confirmed delivery, and is NOT called on the re-buffer branch. Tests assert both (pending-inbound-buffer.test.ts:224-249). Correct.
- inbound-delivery-machine-dispatch.ts:48-117 — `onUserInboundDelivered` threaded through the bridgeUp drain; wired at every redelivery path (drain at gateway.ts:4434, silence-poke fallback, both flap/reply-gate flushes). Optional, so non-cutover path is unaffected.
- resume-inbound-builder.ts:172-196 — 3h staleness failsafe downgrades a stale auto-resume to passive `report`. Requires BOTH `ageMs` and `maxAgeMs` to be set, so legacy blanket-resume is preserved when omitted; the cap only ever downgrades resume→report, never upgrades report/null. Tests cover all four cases (resume-inbound-builder.test.ts:185-201).
- gateway.ts:1067-1079 — `RESUME_MAX_AGE_MS` parses the env override with `Number.isFinite && > 0` guard, default 3h. Sound.

Deferred work (B1: tracking the resume synthetics) is honestly blocked on `buildResumeInterruptedInbound` emitting `meta.message_id` — without it the ack never matches and it would re-deliver forever; correctly NOT attempted here. Synthetic-source exclusion in `shouldTrackDelivery` is what keeps the resume/cron/vault inbounds out of this queue, so the deferral is safe.

Non-blocking:
- The PR's own test plan marks the MANDATORY canary (restart with Hindsight forced down, DM in boot window, plus supergroup-topic variant) as not-yet-done. This is pure orchestration behind the `SWITCHROOM_INBOUND_DELIVERY_CONFIRM` kill-switch, so it's mergeable, but the canary is the real proof — run it before a wide roll.

Rationale: targeted fix to the exact dropped-ack path, mirrors the proven live tracking semantics, kill-switch guarded, tests assert the new hook fires-on-confirm / not-on-rebuffer and the age cap's directionality. No runtime LLM path touched.

mekenthompson:
Fresh-process reviewer (most-incident-dense-path brief): **APPROVE**. Verified the #1 storm risk is safe — for a real user inbound top-level `messageId` === `meta.message_id` (gateway.ts:10719/10734), the merge preserves the anchor identity+meta together, so tracked-id === enqueue-id → ack always fires; synthetic-source (cron/handback/resume, no meta.message_id) is excluded so it can't be tracked-and-stormed. No double-delivery (sweep gated currentTurn==null; spool acked on drain). Thread-keyed per (chat,thread) → DM + supergroup topics recover independently. B1 correctly deferred (buildResumeInterruptedInbound has no meta.message_id). Phase C correct. 216 tests green, tsc clean. **Verdict on the key question — can any tracked entry re-deliver forever? NO.**

Folded in the reviewer's two non-blocking gaps (88a4d8c0): the kill-switch fallback drain + the idleDrainTick post-flap drain now also enrol via trackRedeliveredInbound, so EVERY drain→deliver path is covered, not just bridgeUp.

PR #2118, opened 2026-06-03
fix(telegram): route permission cards to the work topic + name the operation

mekenthompson:
## Summary

Two operator-reported defects on the high-risk **permission approval card** for supergroup-owned agents (marko). Distinct from #2114 (which fixed the vault-grant *resume* inbound) — this is the Approve/Deny card itself.

1. **Routing** — the card landed in the operator DM, never the topic it was asked from. `onPermissionRequest` iterated `access.allowFrom` unconditionally; for a supergroup agent that's only the operator DM user-ids (the supergroup chat id is never in `allowFrom`), so a card raised from the "CRM (Brevo)" topic could only reach the DM. The post-verdict resume message already routed correctly — the card didn't.
2. **Clarity** — "post (Brevo)" / "put (Brevo)" didn't say *what* was being written. `naturalMcpAction` did `void input` and rendered the bare HTTP verb; the endpoint/payload was forwarded in `inputPreview` but dropped.

## Root cause (code-verified, adversarially)

- Routing: `gateway.ts` `onPermissionRequest` recipient set = `access.allowFrom` (operator DMs). `currentTurn.sessionChatId`/`sessionThreadId` (the originating supergroup+topic) were sitting unused. The resume emitter (`postPermissionResumeMessage`) already targeted the turn's chat+thread.
- Clarity: `permission-title.ts` `naturalMcpAction` discards the input; the REST-wrapper tools (brevo/meta/postiz via `rest-server.mjs`) take `{ path, body?, query? }`.

> Note: the earlier `400 message thread not found` / auto-deny logs were the **pre-#2096** wedge (marko's logs predate its v0.14.42 roll). On v0.14.46 the card reaches the DM thread-less — wrong place, but no 400. This PR makes it reach the *topic*.

## What changed

- **Routing** — both card and resume now route through one shared helper `resolvePermissionCardTargets()` (turn → originating chat+topic; no-turn → operator DMs thread-stripped), so they can't drift. Card send wrapped in `retryWithThreadFallback` → a stale/recreated topic re-sends thread-less into the main chat instead of vanishing → auto-deny.
- **Clarity** — action line names the resource ("POST /smtp/email (Brevo)") + a third card line summarizes the payload ("↳ subject: …, templateId: 12, to"). Every value passes the plugin `redact()` chokepoint (a token in the body is masked — pinned by a test), is length-capped, nested objects show as the bare key (no dump). Falls back to "post (Brevo)" on unknown shapes.

## Test plan
- [x] `permission-title.test.ts` — REST endpoint naming, payload summary, **redaction** (fake token masked), DM/non-REST cards unchanged. 8 new cases.
- [x] `permission-card-routing.test.ts` — structural pin: card routes via the shared helper, no longer iterates `allowFrom`, send wrapped in `retryWithThreadFallback`, resume uses the same helper. (gateway.ts isn't unit-importable; routing *logic* covered by `topic-router.test.ts`.)
- [x] plugin unit suite (`tests/`) **3964 pass, 0 fail**
- [ ] supergroup UAT (post-merge canary): card lands in the topic, tappable, resume in the same topic

## Risk
Low. Routing reuses the proven resume-emitter logic + already-tested `topicForRecipient`; DM agents unchanged. Clarity is redaction-guarded with a safe fallback.

JTBD: *you hold the leash* — approvals appear where the work is, and say enough to decide.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -2241,6 +2241,47 @@ function resumeReactionAfterVerdict(): void {
 }
 
 /**
+ * The recipient set for a permission card (the initial Approve/Deny card
+ * AND the post-verdict resume message — they MUST route identically, so
+ * both go through this one helper).
+ *
+ * Turn-initiated (the normal case — a permission gate fires mid-tool-use
+ * with an active turn): send to the ORIGINATING chat+thread. For a
+ * supergroup-owned agent working in a forum topic that is the supergroup +
+ * the topic, so the card lands IN the topic the operator asked from (e.g.
+ * marko's "CRM (Brevo)" topic) — not the operator's DM. For a DM agent the
+ * originating chat IS the operator's DM (thread-less), unchanged.
+ *
+ * No active turn (cron / background / a swept turn at TTL): fall back to the
+ * configured operator DMs (`allowFrom`), thread-stripped via
+ * `topicForRecipient` so a DM never gets a `message_thread_id` (the 400
+ * "message thread not found" → auto-deny wedge, #2096).
+ *
+ * Before this helper the INITIAL card emitter iterated `allowFrom`
+ * unconditionally, so a supergroup card could only ever reach operator DMs —
+ * the topic chat id is never in `allowFrom`. The resume message already
+ * routed correctly; the card now matches it (marko, 2026-06-03).
+ */
+function resolvePermissionCardTargets(): Array<{ chatId: string; threadId: number | undefined }> {
+  const turn = currentTurn
+  if (turn != null) {
+    return [{ chatId: turn.sessionChatId, threadId: turn.sessionThreadId }]
+  }
+  const sg = resolveAgentSupergroupChatId()
+  const topic = resolveAgentOutboundTopic({
+    kind: 'permission',
+    turnInitiated: false,
+    originThreadId: undefined,
+  })
+  // allowFrom is normally operator DMs — attach the topic only to a
+  // recipient that owns it (the supergroup), never a DM (marko wedge).
+  return loadAccess().allowFrom.map(chatId => ({
+    chatId,
+    threadId: topicForRecipient({ recipientChatId: chatId, resolvedTopic: topic, supergroupChatId: sg }),
+  }))
+}
+
+/**
  * Post the agent-voiced "got your verdict — continuing" message the
  * instant the operator answers a permission card. Travels right beside
  * `resumeReactionAfterVerdict()` at every operator-driven verdict site so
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

Verified against PR head. Two distinct operator-facing fixes to the high-risk permission card, both code-confirmed and well-guarded.

- gateway.ts:2256-2280 (`resolvePermissionCardTargets`) — sound unification. Turn-initiated → `[{turn.sessionChatId, turn.sessionThreadId}]` (the originating supergroup+topic); no-turn → `loadAccess().allowFrom` mapped through `topicForRecipient` so a DM is never handed a `message_thread_id` (the #2096 400/auto-deny wedge). `currentTurn.sessionChatId/sessionThreadId` confirmed to exist (gateway.ts:1467-1468).
- gateway.ts:4709-4740 — the initial card emitter now consumes the shared helper instead of the unconditional `for (const chat_id of access.allowFrom)` fan-out that could only ever reach operator DMs. Send wrapped in `retryWithThreadFallback`; the call shape (robustApiCall, `(tid) => sendMessage`, `{threadId, chat_id, verb}`) matches the signature at retry-api-call.ts:229 exactly, so a stale/recreated topic re-sends thread-less rather than vanishing into a 10-min auto-deny.
- gateway.ts:2313 — `postPermissionResumeMessage` now also calls `resolvePermissionCardTargets()`, so card and resume share one routing source and can't drift. The structural pin (permission-card-routing.test.ts) guards exactly this: shared helper used by both, no raw `allowFrom` loop, retry wrapper present.
- permission-title.ts:204-237 (`restResourcePhrase`) — `naturalMcpAction` drops the old `void input`; external (non-`INTERNAL_MCP_SERVERS`) REST-wrapper tools with a resource key render "POST /smtp/email (Brevo)". Internal servers and unknown shapes fall back to the bare verb phrase (tested).
- permission-title.ts:248-287 (`mcpArgSummary`) — third "↳" line summarizes `body`/`query`: every scalar passes `redact()` (import at :22, module present at secret-detect/redact.ts), nested objects/arrays surface as bare key names (no dump), capped at 4 keys / 40 chars-per-value / 180 line. The redaction is load-bearing and pinned by a fake-token test asserting the token never appears.

Nit (non-blocking): with an active turn, the card now goes ONLY to the originating chat, not every `allowFrom` operator. Identical for the single-operator norm and it correctly matches the resume emitter, but in a multi-operator DM setup only the triggering operator sees the card. Intended per the PR, worth a line in the changelog if multi-operator is supported.

Test plan is solid: 8 new permission-title cases (endpoint naming, payload summary, redaction, DM/non-REST unchanged) + the routing structural pin; PR reports full plugin suite 3964 pass. Couldn't execute vitest here (deps not vendored in a shallow clone) — relied on source verification + the reported run. Supergroup UAT correctly left as a post-merge canary.

Rationale: targeted, reuses the already-proven resume-emitter routing and `topicForRecipient`, redaction-guarded with safe fallbacks, DM agents unchanged.

PR #2121, opened 2026-06-03
feat(telegram): edit boot card in place so routine reboots never notify

mekenthompson:
## Summary

Operator request: agent reboot "back up" cards should produce **zero notification**. They already pass `disable_notification: true` (no sound, no banner — verified fleet-wide, #1990), but a freshly *sent* Telegram message still bumps the **unread badge**, and the Bot API has no flag to suppress that. The only way to surface boot info in the chat without a badge is to stop sending a new message — so a routine reboot now **edits the prior boot card in place** (edits never touch the badge).

## What changed

- **`gateway/boot-card-msgid.ts`** (new) — persists the last boot card's `message_id` per (chat, topic) under the agent's bind-mounted state dir (`.boot-card-msgid.json`), surviving gateway restarts. Mirrors the existing `config-snapshot` / `boot-issue-cache` stores; missing/corrupt → "no prior card".
- **`startBootCard`** — routine reboot (no `ackMessageId`) with a persisted id + a bot that supports `editMessageTextStrict` → **edit in place** (zero notification). Falls back to a fresh silent send on first boot (one badge, ever) or if the prior card was deleted.
- **`editMessageTextStrict`** (new optional `BotApiForBootCard` method) — the shared retry policy *swallows* "message to edit not found" to `undefined`, so the boot path couldn't tell a deleted message from a successful edit. This reports `'gone'` (→ fresh send) vs `'edited'` (incl. identical content). Raw call, lint-annotated.
- **Telegram `/restart`** (`ackMessageId` set) still sends a fresh card replying to the operator's command — they asked and are watching; edit-in-place targets the routine badge-flood (operator update / cli rollout / crash / fresh boot).

## Trade-off (operator-approved)

The chat shows each agent's **current** status (one card, updated in place) rather than a scroll-back of every past reboot.

## Test plan
- [x] `boot-card-msgid.test.ts` — persistence (round-trip, per-chat isolation, corrupt-file, non-positive guard, idempotent), isolated tmpdir
- [x] `boot-card-edit-in-place.test.ts` — first-boot-sends+persists; second-boot-edits-NO-send; deleted→fallback-send; /restart→fresh-send; no-state-path→send; no-strict-method→send
- [x] existing boot-card suite (82) green; full unit suite **3984 pass, 0 fail**
- [ ] canary on test-harness: reboot twice, confirm the 2nd edits in place (no badge) via gateway log `edit_in_place=true notify=none`

## Risk
Low, opt-in by data: no state file (or a bot without the strict method) → always-send, exactly today's behavior. `disable_notification:true` retained on every fresh send. DM + supergroup both covered (keyed by chat+topic).

JTBD: *you hold the leash* — reboots are a silent, in-place status record, not a notification flood.

### DIFF
--- a/telegram-plugin/gateway/boot-card-msgid.ts
+++ b/telegram-plugin/gateway/boot-card-msgid.ts
@@ -1,0 +1,70 @@
+/**
+ * Cross-reboot persistence for the boot card's Telegram message id.
+ *
+ * Why: a freshly *sent* Telegram message always bumps the chat's unread
+ * badge — `disable_notification: true` removes the sound/banner but not the
+ * badge (there is no Bot API flag for that). To make routine reboots produce
+ * ZERO notification (operator request, 2026-06-03), the gateway reuses the
+ * PRIOR boot card's message and EDITS it in place instead of sending a new
+ * one — and edits never touch the badge.
+ *
+ * That requires remembering the last boot card's `message_id` across gateway
+ * restarts, keyed by the chat (+ forum topic) it lives in. This module is the
+ * tiny JSON store for that, mirroring `config-snapshot.ts` /
+ * `boot-issue-cache.ts`: one file under the agent's (bind-mounted, reboot-
+ * surviving) state dir, read once on boot, written once after the id is
+ * established. All failures are non-fatal — a missing/corrupt file just means
+ * "no prior card", so the boot path falls back to a fresh (silent) send.
+ */
+
+import { readFileSync, writeFileSync } from 'node:fs'
+
+/** Stable key for a boot-card target: chat id + optional forum topic. A DM
+ *  agent always boots to the same `<chatId>:` key; a supergroup agent keys by
+ *  `<chatId>:<threadId>` so a topic change starts a fresh card. */
+export function bootCardChatKey(chatId: string, threadId: number | undefined): string {
+  return `${chatId}:${threadId ?? ''}`
+}
+
+type Store = Record<string, number>
+
+function readStore(path: string): Store {
+  try {
+    const parsed = JSON.parse(readFileSync(path, 'utf8')) as unknown
+    if (parsed && typeof parsed === 'object' && !Array.isArray(parsed)) {
+      return parsed as Store
+    }
+  } catch {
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- `gateway/boot-card.ts:687-705` — edit-in-place gate is correct: `reuseId` only computed when `ackMessageId == null && bootCardStatePath != null && editMessageTextStrict != null`. On `'edited'` → reuse; on `'gone'`/throw, `messageId` stays -1 and the code falls through to a fresh silent send. No path leaves `messageId` unset.
- `gateway/gateway.ts:2609-2625` — `editMessageTextStrict` correctly bypasses `robustApiCall` (which swallows "message to edit not found" → undefined) with a deliberate single-attempt raw call, lint-annotated `allow-raw-bot-api`. Classifies "not modified" as `'edited'` (message exists) and every other error as `'gone'` → fresh send. Worst case on a transient error is one badge — acceptable degrade, documented.
- `gateway/boot-card-msgid.ts:38-69` — load/save guard ids with `Number.isFinite && > 0`, merge-on-write preserves other chats, idempotent no-rewrite on unchanged id, all I/O failures non-fatal → "no prior card". Keyed by chat+topic so DM and supergroup topics stay independent.
- `gateway/gateway.ts:4601,18493` — both boot-card call sites wired with `bootCardStatePath` under the bind-mounted agent dir (survives reboot). Consistent with the existing `.config-snapshot.json` pattern.
- Tests: `boot-card-msgid.test.ts` (round-trip, per-chat isolation, corrupt-file, non-positive guard, idempotent) and `boot-card-edit-in-place.test.ts` (first-boot-sends+persists, second-boot-edits-no-send, deleted→fallback, /restart→fresh, no-state-path→send, no-strict-method→send) cover every branch. State written to mkdtemp, not ~/.switchroom. Author reports full suite 3984 pass / 0 fail, tsc + check-bot-api-wrapping clean.

Rationale: opt-in by data with always-send fallback on every failure mode means the blast radius is at most one stray badge; `disable_notification:true` retained on all fresh sends. Only the canary checkbox is open — an operator pre-rollout gate, not a merge blocker.

PR #2122, opened 2026-06-03
fix(telegram): boot-resume turns get origin routing + survive a not-ready restart

mekenthompson:
## Summary

The boot-resume synthetics — the wake-ups that make an agent **pick up work it was doing when it restarted** — carried only `meta.source` (no `chat_id`, `message_thread_id`, or `message_id`). Two real gaps fall out of that, both in the operator's "if it was restarted mid doing work it should pick up that work without prompting" area:

1. **No `currentTurn` for resume turns.** The gateway's enqueue handler gates the entire `currentTurn` construction on `ev.chatId` (parsed from the channel XML, which renders from `meta`). With no `meta.chat_id`, a resumed turn ran with **no progress card and no silence-poke protection**, and its reply fell back to the agent's **default chat** instead of the topic the interrupted work lived in (a supergroup mis-route — same class as the handback / permission-card topic-routing fixes; handback already carries this meta, resume did not).

2. **Not rescued if dropped.** The resume wake is buffered + spool-replayed and drained on bridge-up exactly like a user inbound, but `trackRedeliveredInbound` excludes all `meta.source` synthetics — so a restart that dropped it into a still-booting (slow MCP) session **silently failed to pick the work back up**. This is the resume-side of the #2117 lost-message race.

## What changed

- `buildResumeInterruptedInbound` / `buildResumeWatchdogReportInbound` carry `chat_id` + `message_thread_id` in meta (mirrors the real-inbound + `subagent_handback` shapes) → resumed/report turns now get a `currentTurn` (progress card + silence-poke) and route to the originating chat/topic.
- `buildResumeInterruptedInbound` additionally carries `message_id` (= its fabricated `ts`) so its enqueue id round-trips and the deliver-until-acked queue can ack **this** synthetic by its own enqueue — enrolled via a narrow carve-in (`isTrackableResumeSynthetic`) so it re-delivers until claude consumes it and **cannot re-deliver forever** (the storm the original deferral feared). The watchdog `report` stays untracked (no `message_id`); every other synthetic (cron/vault/handback) stays excluded.

## Why it's storm-safe + reply-safe (the two risks)

- **No storm:** the ack fires (it's gated on `ev.chatId`, now present via `meta.chat_id`) and matches (`ev.messageId` = `meta.message_id` = `String(ts)` = the tracked id). Unit-tested: tracked → acked by own enqueue → sweep empty; a *different* enqueue id does not false-ack.
- **No bad reply-quote:** the fabricated `message_id` is never a Telegram `reply_to`. The model's reply tool quotes the real prior user message via `getLatestInboundMessageId` (`role='user'`; synthetics aren't in history); the activity feed anchors with `allow_sending_without_reply`.

## Test plan

- [x] `isTrackableResumeSynthetic` carve-in (resume-with-id tracked; resume-without-id / report / cron / handback / real all excluded)
- [x] resume tracked → acked by own enqueue → never storms; strands-then-rescues
- [x] builder meta: `chat_id`/`message_thread_id`/`message_id` present on resume; report has chat_id/thread but **no** message_id
- [x] 3981 bun + 3563 vitest green, `tsc --noEmit` clean, bot-api-wrapping + plugin-references clean
- [ ] Canary: real interrupted turn on test-harness (DM) → restart → resume turn shows a progress card + is rescued if it lands in a not-ready session; tail for `boot-resume queued kind=resume` → enqueue → reply
- [ ] Supergroup topic routing rests on the `subagent_handback` precedent (same mechanism) + gateway unit thread-assertions — mtcute can't create non-General topics

## Risk

Low. Rides the existing `SWITCHROOM_INBOUND_DELIVERY_CONFIRM` kill switch. The behavior change is that resume/report turns now become first-class turns (get a `currentTurn`) — which is the intended improvement (visibility + wedge protection). Note: a cap-downgraded stale resume routes its "report" to origin too (correct).

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -287,6 +287,7 @@ import {
   sweep as sweepDeliveryQueue,
   forgetDelivery,
   shouldTrackDelivery,
+  isTrackableResumeSynthetic,
   type PendingDelivery,
 } from './inbound-delivery-confirm.js'
 import { createPendingPermissionBuffer } from './pending-permission-decisions.js'
@@ -4192,15 +4193,17 @@ _deliveryMachineTick.unref?.()
 // re-deliver forever.
 function trackRedeliveredInbound(merged: InboundMessage): void {
   if (!DELIVERY_CONFIRM_ENABLED) return
+  // The boot-resume synthetic ('resume_interrupted') is the ONE synthetic we DO
+  // enrol: a restart can drop it into a not-ready session exactly like a user
+  // inbound, leaving the interrupted work silently un-resumed. Safe iff it
+  // carries the chat_id + message_id that make its enqueue ack-able — see
+  // isTrackableResumeSynthetic. Every other synthetic stays excluded below.
+  const isTrackableResume = isTrackableResumeSynthetic(merged.meta)
   if (
+    !isTrackableResume &&
     !shouldTrackDelivery({
       isSteering: false,
       isInterrupt: false,
-      // Synthetic inbounds (cron / vault / handback / resume) carry a source
-      // and are NOT tracked here — they enqueue under their own semantics, and
-      // (for the resume synthetics) tracking them safely first needs the
-      // resume builder to emit meta.message_id so the deliver-until-acked ack
-      // matches its enqueue. Tracked separately as a follow-up (see PR notes).
       hasSource: merged.meta?.source != null,
       effectiveText: merged.text,
     })
--- a/telegram-plugin/gateway/inbound-delivery-confirm.ts
+++ b/telegram-plugin/gateway/inbound-delivery-confirm.ts
@@ -158,3 +158,29 @@ export function shouldTrackDelivery(input: {
   if (input.effectiveText !== undefined && input.effectiveText.trim().length === 0) return false
   return true
 }
+
+/**
+ * The ONE synthetic-source inbound that IS safe to enrol in the
+ * deliver-until-acked queue: the boot-resume synthetic
+ * (`meta.source === 'resume_interrupted'`). A restart can drop it into a
+ * not-ready (slow MCP boot) session exactly like a user inbound, leaving the
+ * interrupted work silently un-resumed — so it needs the same re-delivery
+ * backstop. It is safe to track ONLY when it carries BOTH:
+ *   - `meta.chat_id` — so the gateway's enqueue handler (gated on `ev.chatId`)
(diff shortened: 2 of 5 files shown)


mekenthompson:
APPROVE

- `gateway/gateway.ts:4201-4212` — carve-in is correct: `!isTrackableResume && !shouldTrackDelivery(...)` short-circuits so a trackable resume bypasses the `hasSource` exclusion and IS enrolled, while every other synthetic still falls through to `shouldTrackDelivery` and stays excluded. No regression for cron/vault/handback.
- `gateway/gateway.ts:4214-4220` — `trackDelivery` keys the tracked id on `String(merged.messageId)`, which for the resume synthetic = `String(ts)` = `meta.message_id`. Confirmed against the ack path: `enqueue` fires `ackDelivery(..., ev.messageId)` gated on `ev.chatId` (gateway.ts:8120, 8174-8178). `meta.chat_id` makes `ev.chatId` present so the ack fires; `meta.message_id` makes `ev.messageId` match the tracked id so it acks THIS synthetic. Storm-safe by construction.
- `gateway/inbound-delivery-confirm.ts:175-183` — `isTrackableResumeSynthetic` correctly requires `source==='resume_interrupted'` AND a non-empty `message_id`; the watchdog `report` (no message_id) and all other synthetics return false. Matches the builder: only `buildResumeInterruptedInbound` emits `message_id`.
- `gateway/resume-inbound-builder.ts:80-105` — `buildResumeInterruptedInbound` carries `chat_id`, optional `message_thread_id`, and `message_id=String(ts)`; top-level `messageId` is the numeric `ts` so it round-trips through the enqueue XML. `buildResumeWatchdogReportInbound` (lines 149-158) carries chat_id+thread but deliberately NO message_id → routing fix without tracking. Consistent with the carve-in gate.
- `message_id` reuse as a fabricated id is safe: it's a 13-digit ms timestamp, never collides with a real per-chat Telegram message_id, and is never used as a Telegram reply_to (the model quotes the real prior user msg). Tests `inbound-delivery-confirm.test.ts:240-266` pin tracked→acked-by-own-enqueue→no-storm, strands-then-rescues, and that a different enqueue id does NOT false-ack.

Rationale: a narrow, well-tested carve-in that fixes both the missing-currentTurn origin mis-route and the resume-side lost-message race (#2117 sibling), rides the existing SWITCHROOM_INBOUND_DELIVERY_CONFIRM kill switch, and is provably storm-safe because the ack matches the synthetic's own enqueue id. Canary checkboxes (real interrupted turn + supergroup topic) are open but are pre-rollout operator gates, not merge blockers.

PR #2125, opened 2026-06-03
test(telegram): permanent restart-resume regression gates (#2117/#2122)

mekenthompson:
## Summary

Promotes the two throwaway canaries from the v0.14.48 (#2117) and v0.14.50 (#2122) rollouts into **permanent UAT regression gates** for the restart/delivery bug class.

These live in `telegram-plugin/uat/scenarios/` — **non-gating** (uat/** is excluded from gating CI) and **self-skip green** without NOPASSWD sudo (or, for the channel one, without a wired forum supergroup). They're runnable via `bun run --cwd telegram-plugin test:uat <name>`.

## Gates added

- **`jtbd-message-during-restart-dm`** / **`-channel`** — a message sent *DURING* the restart boot window is still answered (the #2117 lost-message gate). Distinct from `jtbd-always-on-after-restart-dm`, which sends ~5s *after* the restart returns (live path); these send during boot so the message goes through the restart-redeliver path. `SWITCHROOM_UAT_BOOT_SEND_DELAY_MS=6000` lands it deep enough in the not-ready window to exercise the strand-rescue sweep.
- **`jtbd-interrupted-turn-resumes-dm`** — a turn interrupted by a restart is resumed and the resume turn *completes* (not silently dropped) — the #2122 gate. Asserts the resume **framing** (`/resum|picking .*back|interrupted|cut off/i`), not an end-token: the resume synthetic only carries the first ~160 chars of the original prompt, so tail instructions aren't guaranteed to survive.

## Test plan

- [x] All three discovered by the uat runner (`vitest list --config ../vitest.uat.config.ts`)
- [x] `jtbd-interrupted-turn-resumes-dm` verified green live (56s — real interrupt → resume → completion)
- [x] during-restart DM/channel are byte-identical logic to the canaries already proven during the v0.14.48 rollout (DM answered 13.8s, supergroup 13.6s, strand-rescue fired at 6s delay)
- [x] `tsc --noEmit` clean
- CLAUDE.md release-critical scenarios list updated

## Risk

Minimal — additive test scenarios (non-gating, self-skipping) + a doc list update. No source changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -753,6 +753,16 @@ lifecycle changes):
 - `jtbd-always-on-after-restart-dm` — first message after restart
   ≤60s (5-min wedge regression gate).
 - `jtbd-memory-survives-restart-dm` — memory persistence across restart.
+- `jtbd-message-during-restart-dm` / `-channel` — a message sent
+  *during* the restart boot window is still answered (v0.14.48 / #2117
+  lost-message gate; sends DURING boot, unlike always-on which sends
+  after). Set `SWITCHROOM_UAT_BOOT_SEND_DELAY_MS=6000` to land it deep
+  enough in the not-ready window to exercise the strand-rescue sweep.
+- `jtbd-interrupted-turn-resumes-dm` — a turn interrupted by a restart
+  is resumed (and the resume turn completes, not silently dropped)
+  (v0.14.50 / #2122 gate). Asserts the resume *framing*, not an
+  end-token (the resume synthetic only carries the first ~160 chars of
+  the original prompt).
 - Any feature-specific scenario.
 
 **Supergroup / channel UAT (status-routing work — required).** Every
--- a/telegram-plugin/uat/scenarios/jtbd-interrupted-turn-resumes-dm.test.ts
+++ b/telegram-plugin/uat/scenarios/jtbd-interrupted-turn-resumes-dm.test.ts
@@ -1,0 +1,90 @@
+/**
+ * JTBD — always-on: a turn interrupted by a restart is RESUMED without
+ * re-prompting, and the resume turn runs to completion (it is not silently
+ * dropped into a not-ready session). Regression gate for v0.14.50 / #2122.
+ *
+ * Flow: send a long-running task, let it get in-flight, then restart the agent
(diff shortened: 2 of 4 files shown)


mekenthompson:
APPROVE

Tests-only PR: promotes the #2117 / #2122 rollout canaries into permanent UAT regression gates. Verified against the real diff (290 additions, 4 files, 0 deletions, no source changes).

- `jtbd-interrupted-turn-resumes-dm.test.ts:1-90` — sends a long task, restarts `--force` (interrupt, not drain) at MID_TURN_MS=10s, asserts the resume reply matches RESUME_FRAMING `/resum|picking .*back|interrupted|cut off|just restarted/i` (test.ts:38) and length>40. Correct call to assert framing not an end-token, since the resume synthetic truncates the prompt to ~160 chars — the rationale is documented and accurate.
- `jtbd-message-during-restart-dm.test.ts:1-95` — kicks restart, waits BOOT_SEND_DELAY_MS (default 12000), sends nonce DURING boot window, asserts the redelivered inbound is answered within 180s. Guards exactly the #2117 wedge (bridge-registered != session-ready, silent drop + phantom 300s silence-poke). Sound.
- `jtbd-message-during-restart-channel.test.ts:1-95` — supergroup twin; self-skips green when SWITCHROOM_UAT_CHAT_ID unset or chat unresolvable (channel.test.ts:55,62). mtcute forum-topic caveat (uses General topic) is honestly documented.
- `CLAUDE.md:756-765` — release-critical scenarios list updated to match the three new gates; copy matches the test behavior.

Notes (non-blocking):
- `jtbd-interrupted-turn-resumes-dm.test.ts:2` header says 'JTBD — always-on' but the scenario is resume-specific; cosmetic.
- PR body quotes RESUME_FRAMING as `/resum|picking .*back|interrupted|cut off/i`, dropping the `|just restarted` alternation present in the code (test.ts:38). The code is the authority and is fine; body is slightly stale.
- All three self-skip without NOPASSWD sudo and uat/** is excluded from gating CI, so zero CI-blocking risk.

Rationale: additive, non-gating, self-skipping test coverage for a real incident class (lost-message + interrupted-turn) with accurate rationale and no source changes. Nothing to fix.

PR #2126, opened 2026-06-03
fix(telegram): resumed turn recovers the FULL original prompt, not a 160-char slice

mekenthompson:
## Summary

A turn resumed after a restart only saw a **truncated preview** of the original request. The turns table stores the first ~200 chars (`TURN_PREVIEW_MAX`), and the boot-resume synthetic sliced that *again* to 160 (`promptClause`) — so instructions in the tail of a long/detailed task were silently dropped on resume. (Surfaced by the v0.14.50 interrupt-resume canary: a token placed at the end of a ~340-char prompt never made it into the resumed turn.)

## Why this is the right fix

The full original message is **always recoverable**: `recordInbound` stores it verbatim in the SQLite history buffer (no truncation — `history.ts`), `get_recent_messages` returns it verbatim, and that buffer **survives restarts** (it's the documented post-restart recovery path). The resume synthetic just never told the model to fetch it. So rather than bloat the turns table or the synthetic, the fix points the resumed turn at the existing full-fidelity source.

## What changed (`buildResumeInterruptedInbound`)

- `promptClause` no longer re-slices below the storage cap — the ~200-char stored preview is included verbatim as a hint, framed as *"the start of the request"* (so the model knows it's partial). Fixes both resume + report builders (shared helper).
- The resume body now instructs the model to call `get_recent_messages` for this chat (it has lost in-memory context across the restart) to read the **full** original message + surrounding context **before** continuing — so it resumes the complete task, including tail instructions, not just the preview.

## Test plan

- [x] `tsc --noEmit` clean; plugin-references + PII clean
- [x] Updated unit tests: a 180-char preview is included **verbatim** (no 160-cut); the body instructs `get_recent_messages` recovery and frames the quote as partial
- [x] 3996 bun + 3577 vitest green; no other test asserts on the old text
- [ ] Reviewer
- [ ] Canary: real interrupt with a long, tail-loaded prompt → resume → confirm the model fetches recent messages and completes the *full* task (the v0.14.50 canary's end-token case)

## Risk

Low. Text/prompt-only change to the resume synthetic + dropping a redundant slice; no delivery/routing/storm surface touched. The model is *instructed* to fetch recent messages — worst case it doesn't and behaves as before (still has the ~200-char preview, which is now 200 not 160).

### DIFF
--- a/telegram-plugin/gateway/resume-inbound-builder.ts
+++ b/telegram-plugin/gateway/resume-inbound-builder.ts
@@ -66,9 +66,12 @@ function threadIdNum(turn: Turn): number | undefined {
 function promptClause(turn: Turn): string {
   const p = turn.user_prompt_preview?.trim()
   if (!p) return ''
-  // Quote-trim so a long preview doesn't bloat the channel body.
-  const snippet = p.length > 160 ? p.slice(0, 160) + '…' : p
-  return ` The request was: "${snippet}".`
+  // The stored preview is already capped at the first ~200 chars of the user
+  // message (TURN_PREVIEW_MAX). Include it verbatim as a hint — do NOT
+  // re-truncate (the old 160-char slice dropped instructions that lived in the
+  // tail of a longer request). The FULL original is recovered via
+  // get_recent_messages; the resume body tells the model to do that.
+  return ` The start of the request was: "${p}".`
 }
 
 /**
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

Verified against the real diff (34 additions, 15 deletions, 2 files) and the live tree.

- `resume-inbound-builder.ts:66-74` — `promptClause` no longer re-slices to 160; it now emits the stored preview verbatim framed as "The start of the request was…". Confirmed the stored preview is already capped at 200 by `TURN_PREVIEW_MAX` (gateway.ts:1697, slice at :1702), so this drops a redundant second truncation rather than uncapping anything. Correct.
- `resume-inbound-builder.ts:120-134` — resume body now instructs the model to call `get_recent_messages` for the chat BEFORE continuing, to recover the full original message + surrounding context. `get_recent_messages` is the documented restart-survivable recovery path (SQLite history buffer stores the inbound verbatim), so the full prompt is genuinely recoverable — the synthetic just never told it to fetch. Sound fix that avoids bloating the turns table or the synthetic body.
- Shared helper: `promptClause` is consumed by both `buildResumeInterruptedInbound` (:119) and the report builder (:178), so the de-truncation applies to both paths. Matches the PR's claim.
- `tests/resume-inbound-builder.test.ts:109-124` — old 'truncates a long prompt' test replaced with two: a 180-char preview must appear verbatim (no 160-cut) and `meta.original_prompt` round-trips; and the body must contain `get_recent_messages` + frame the quote as partial. Tests match the new behavior.

Note (non-blocking):
- The storage-side cap at gateway.ts:1702 slices to 200 with NO ellipsis, so a preview that was truncated at exactly 200 now appears with no '…' marker (the old promptClause added one). The 'start of the request' framing covers intent, so the model still knows it's partial — cosmetic only.

Rationale: prompt-text-only change plus removal of a redundant slice that was silently dropping tail instructions on resume; no delivery/routing surface touched, worst case degrades to prior behavior with a 200-char (vs 160) preview. Tests green per PR (3996 bun + 3577 vitest, tsc clean).

PR #2128, opened 2026-06-03
fix(telegram): default the visible answer-stream OFF (no unformatted flash)

mekenthompson:
## Summary

On every reply (DMs **and** supergroups) an unformatted message appears, then a properly-formatted message is sent, then the unformatted one is deleted. This disables the mechanism causing it.

## Root cause (verified live)

The **visible answer-stream** (default-on since 2026-05-25) posts a "preliminary" chat message seeded with the model's **raw markdown** and edits it as tokens stream. When the model calls the **reply tool (≈every turn)**, the reply posts a **separate** canonical formatted message and the stream **`retract()`s — deletes — its preliminary**. Net: an unformatted bubble flashes and vanishes, replaced by the formatted reply. In supergroup topics it also mis-routes (preliminary → General, reply → topic).

Live evidence (clerk DM + marko SG): the preliminary gets only **1–2 tiny edits (5–13 bytes)** before deletion — Telegram rate-limits `editMessageText` to ~1/s so real "watch it type" streaming is impossible, and the model emits almost no interstitial `assistant.text` (thinks → tool → reply). So the visible stream delivered ~no value while imposing the flash + delete every turn.

## Fix

Flip the `SWITCHROOM_VISIBLE_ANSWER_STREAM` default to **OFF** (extracted to `parseVisibleAnswerStreamEnabled`, unit-tested). With it off the answer lane uses the invisible compose-box draft (the original default) and the **reply tool is the single canonical, formatted message** — no preliminary post, no delete. Opt back in per agent with `SWITCHROOM_VISIBLE_ANSWER_STREAM=1`.

## Anti-silence note (the one risk to validate)

The visible stream was originally flipped on to cut a **~19% framework-fallback ("still working…") rate**. That role is now covered by the live **activity feed** (tool-use streaming), the **"…typing"** chat-action loop, and `thinking_effort: low` (fast tool-less turns). **Canary will confirm the framework-fallback rate does not regress before fleet rollout** — revert is a one-flag flip.

## Test plan
- [x] `answer-stream-flag.test.ts` — default OFF; opt-in only on `1/true/on/yes`
- [x] `tsc` + full lint chain clean; plugin unit suite **3999 pass, 0 fail**
- [ ] canary test-harness: tool-using + tool-less turns, DM **and** forum topic → (a) NO `answer-stream: sent`/`retracted` + single formatted reply, (b) framework-fallback rate not regressed

## Risk
Low + reversible. No change to the reply path; the lane reverts to draft transport. `retract()` only deletes a posted message (`answer-stream.ts:621`), so off ⇒ no `deleteMessage`.

### DIFF
--- a/telegram-plugin/answer-stream-flag.ts
+++ b/telegram-plugin/answer-stream-flag.ts
@@ -1,0 +1,18 @@
+/**
+ * Parse the `SWITCHROOM_VISIBLE_ANSWER_STREAM` env flag.
+ *
+ * Default **OFF** (flipped 2026-06-03, operator request) — see the rationale
+ * on the `ANSWER_STREAM_VISIBLE_ENABLED` gate in `gateway/gateway.ts`. When
+ * off, the answer lane streams to the invisible compose-box draft and the
+ * reply tool is the single canonical formatted message — no unformatted
+ * preliminary that flashes and gets deleted. Opt back IN per agent with
+ * `SWITCHROOM_VISIBLE_ANSWER_STREAM=1` (also accepts true/on/yes).
+ *
+ * Extracted as a pure function so the default + parsing are unit-testable
+ * (gateway.ts is not importable in isolation — top-level side effects).
+ */
+export function parseVisibleAnswerStreamEnabled(raw: string | undefined): boolean {
+  if (raw == null) return false
+  const v = raw.trim().toLowerCase()
+  return v === '1' || v === 'true' || v === 'on' || v === 'yes'
+}
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -8437,9 +8437,19 @@ function handleSessionEvent(ev: SessionEvent): void {
             // tool_use stream (case 'tool_use' above) where the real
             // signal lives. assistant.text keeps its visible-message
             // home; the reply tool stays the canonical answer.
+            // Flag OFF (default): use the compose-box draft for DMs, and set
+            // minInitialChars effectively-infinite so the lane NEVER opens a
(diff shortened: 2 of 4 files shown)


mekenthompson:
APPROVE

- `telegram-plugin/answer-stream-flag.ts:14-18` — `parseVisibleAnswerStreamEnabled` cleanly inverts the contract: `raw == null → false` (default OFF), opt-in only on `1/true/on/yes` (case/space-insensitive). Correct and the only sensible parse.
- `telegram-plugin/gateway/gateway.ts:3766-3768` — the old default-ON IIFE (explicit-off list) is replaced by a single call to the extracted helper at the one gate (`ANSWER_STREAM_VISIBLE_ENABLED`). No other gateway path reads the env var, so no stale default-on assumption survives. The retract/delete path is gated entirely behind this flag, so OFF ⇒ no preliminary post and no `deleteMessage` (answer-stream.ts:621).
- `telegram-plugin/tests/answer-stream-flag.test.ts:11-26` — pins the contract: default OFF, falsey/unrecognized OFF, truthy opt-in. Guards against an accidental flip back. Solid coverage.
- Other repo readers of `SWITCHROOM_VISIBLE_ANSWER_STREAM` (UAT `visible-answer-stream-dm.test.ts`, draft-mirror RFC) explicitly set `=1`, so they remain correct under the new default.

Rationale: a small, well-tested, fully reversible behavior flip (one env default) that removes a real UX defect — the unformatted-preliminary flash + delete on every reply, plus topic mis-routing — with the anti-silence role already covered by the activity feed / typing loop / low thinking_effort.

Watch (non-blocking): the canary checkbox (framework-fallback rate not regressing) is still unchecked in the test plan. Confirm that before fleet rollout; revert is a one-flag flip.

PR #2130, opened 2026-06-03
fix(telegram): bind turn-pacing 'answer' to the reply tool (cut orphaned-reply flushes)

mekenthompson:
## Summary

Fleet-wide, agents (esp. marko on long Postiz tool sequences) sometimes end a turn with the answer written as plain transcript text and **no reply-tool call**. Telegram only delivers text sent via the reply tool, so the gateway's 30s `orphaned-reply` backstop flushes the captured text **~30s late** — which reads to the operator as "marko isn't posting updates." Measured at **15–40% of turns across the fleet** (marko ~25%, lawgpt ~40%, gymbro ~39%) — not marko-specific.

## Root cause (investigation, adversarially verified)

A **guidance gap**, not a timing bug. The per-turn `turnPacingDirective` (the live UserPromptSubmit hook the model reads each turn) said "just answer / stop after the answer" but never bound **"answer" to a reply-tool call**. After a long tool sequence the model lets its closing narration stand as the answer and stops — never calling reply — so only the 30s backstop surfaces it.

## Fix — the one verified-safe lever (prompt-only)

Append a CRITICAL clause to `turnPacingDirective` (`src/agents/scaffold.ts`): *"answer" means a call to the reply tool; transcript text is NEVER delivered to Telegram; after a long tool sequence, pass your closing narration to the reply tool.* Takes effect on restart; reverts by reverting the commit; no kill switch needed (a prompt clause can't itself cause a wrong send). The existing NO_REPLY carve-outs (greetings, genuine non-prompts) are intact.

## Deferred (with evidence — do NOT ship)

- **Tighten the 30s timeout — REFUTED unsafe.** `resetOrphanedReplyTimeout` resets only on assistant *text* (single callsite, never on tool_use), so a shorter timer fires mid-tool-sequence → premature partial flush, and the model's later real reply also sends (exact-string dedup misses partial-vs-full) → **double-send**.
- **A turn-end reply nudge — already exists** (`silent-end-interrupt-stop.mjs`). A second one widens the flush-vs-reply race.
- **De-flicker the answer-stream — moot** (visible stream OFF since v0.14.52 / #2128).

## Test plan

- [x] New regression test (`tests/scaffold.test.ts`): the rendered `settings.json` turn-pacing hook command contains the reply-tool binding + the two new sentences (pins the live carrier).
- [x] 188 vitest pass; `tsc --noEmit` clean.
- [x] Fresh-process reviewer APPROVE (live-carrier confirmed, no NO_REPLY regression, deferred-levers validated).
- [ ] UAT: after roll, tail `orphaned-reply timeout` / `turn-flush firing` vs `reply: invoked` on a long-tool-sequence turn — expect the turn-flush ratio to drop.

## Risk

Low — prompt-only, no code-path change, no new emitter (so no new double-send surface), timeout unchanged.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -4139,7 +4139,15 @@ export function buildSettingsHooksBlock(p: HooksBlockParams): Record<string, unk
     '"Sent.", "Hope that helps." as a separate message once you have ' +
     'already replied. Your answer is the last thing the user should ' +
     'see; a follow-up "Done." is dead-air clutter (and the user\'s ' +
-    'device already pinged on the answer). Stop after the answer.</turn-pacing>';
+    'device already pinged on the answer). Stop after the answer.\n\n' +
+    'CRITICAL: "answer" means a call to the reply tool ' +
+    '(mcp__switchroom-telegram__reply, or stream_reply with done=true). ' +
+    'Your terminal/transcript text is NEVER delivered to Telegram — the ' +
+    'user sees only what you send through the reply tool. After a long ' +
+    'tool sequence (scheduling, multi-step research, sub-agent handback), ' +
+    'do not let your closing narration stand as the answer: end the turn ' +
+    'by passing that narration to the reply tool. No reply tool call = the ' +
+    'user got nothing, however much text you wrote.</turn-pacing>';
   const switchroomUserPromptSubmit: Array<Record<string, unknown>> = [
     ...(useHotReloadStable
       ? [
--- a/tests/scaffold.test.ts
+++ b/tests/scaffold.test.ts
@@ -2464,6 +2464,44 @@ describe("scaffoldAgent with global defaults cascade", () => {
     ]);
   });
 
+  it("turn-pacing directive ties the answer to the reply tool", () => {
+    // Regression for the orphaned-reply visibility defect: the model
+    // (esp. marko on long Postiz tasks) ended turns with plain transcript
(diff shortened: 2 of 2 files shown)


mekenthompson:
APPROVE

- `src/agents/scaffold.ts:4142-4150` — appends a CRITICAL clause to `turnPacingDirective` binding "answer" to a reply-tool call (`mcp__switchroom-telegram__reply` / `stream_reply done=true`), stating transcript text is never delivered and that a long tool sequence must end with the reply tool. Prompt-only; no code path changes. Correct lever for the orphaned-reply-flush defect.
- `tests/scaffold.test.ts:2467-2504` — regression test asserts the rendered settings.json UserPromptSubmit hook command contains `mcp__switchroom-telegram__reply`, `terminal/transcript text is NEVER delivered`, and `No reply tool call = the user got nothing`. All three substrings match the new directive text verbatim; the test pins the live carrier so the guidance can't silently regress.
- Existing NO_REPLY carve-outs (greetings, genuine non-prompts) are untouched — the append doesn't contradict them.

Rationale: prompt-only guidance change that closes a documented guidance gap (15-40% of fleet turns ending without a reply-tool call); no new emitter, timeout unchanged, so no new double-send surface. Deferred levers (tighten 30s timeout, second turn-end nudge) are correctly refuted as unsafe in the PR body. Safe to merge once CI green.

PR #2132, opened 2026-06-03
feat(telegram): configurable status-message clear-on-completion (default off)

mekenthompson:
## Summary

Make the live activity/status feed's **delete-on-completion** behaviour a per-agent switchroom.yaml knob, **default off** — the operator prefers the status message to *stay* in the chat as a finalized record beside the reply, instead of being posted then deleted the moment the answer lands.

New field: `channels.telegram.clear_status_on_completion` (zod boolean, optional, per-key cascade).

- **`false` (default)** — `clearActivitySummary` now **finalizes** the message: edits it to an all-done render (every step marked `✓`, no frozen `→ in-progress` arrow) and stops tracking it. No post-then-delete flicker.
- **`true` (opt-in)** — restores the prior #2108 behaviour: the status message is **deleted** when the final answer lands, leaving only the reply.

## Why

#2108 deletes the activity feed on `finalAnswerDelivered`. On trivial turns this produces a visible post-then-delete (~31-byte message appears, then vanishes). The operator confirmed the "status stays" UX reads better and asked to keep it by default, with delete available as an opt-in.

## What changed

- `src/config/schema.ts` — `clear_status_on_completion` on `TelegramChannelSchema`.
- `src/agents/scaffold.ts` — `channelsToEnv` emits `SWITCHROOM_TG_CLEAR_STATUS_ON_COMPLETION` only when explicitly set (gateway's default-off parse owns unset).
- `telegram-plugin/gateway/gateway.ts` — `CLEAR_STATUS_ON_COMPLETION` const; `clearActivitySummary` branches finalize-vs-delete; `composeTurnActivity(turn, final)`.
- `telegram-plugin/tool-activity-summary.ts` — `renderActivityFeed` / `renderActivityFeedWithNested` gain a `final` param so the persisted feed reads done, not stuck mid-step.
- Tests + `docs/configuration.md` cascade-table row.

## Test plan

- [x] `tool-activity-summary.test.ts` — `final=true` → all lines `✓` (flat + nested); single-line; default-false keeps `→`; nested no-children delegates to finalized flat
- [x] `src/config/schema.test.ts` — accepts true/false (round-trip), optional unset → undefined, rejects non-boolean
- [x] `tsc --noEmit` clean; full lint chain clean
- [x] full plugin bun suite (4006 pass) + config/scaffold vitest (226 pass) green

## Risk

Low. Default preserves the no-delete UX the operator confirmed; the delete path is byte-identical to the prior #2108 behaviour, just gated. Per-key cascade so an agent/profile can opt in.

JTBD: *you hold the leash* — awareness without babysitting; the finalized status trail is the "what it did" record without a vanishing message.


mekenthompson:
APPROVE

- `src/config/schema.ts:497-507` — `clear_status_on_completion` added as optional zod boolean on TelegramChannelSchema. Correct: optional means unset stays undefined so the gateway's default-off parse owns the unset case (verified by schema.test.ts:52-62).
- `src/agents/scaffold.ts:1360-1366` — `channelsToEnv` emits `SWITCHROOM_TG_CLEAR_STATUS_ON_COMPLETION` only when `!== undefined`, mapping bool→"1"/"0". Correct: unset emits nothing, so gateway default wins; no spurious env on the common path.
- `telegram-plugin/gateway/gateway.ts:3774-3786` — `CLEAR_STATUS_ON_COMPLETION` const parses 1/true/on/yes, defaults false on null/anything-else. Robust parse; matches the documented default-off intent.
- `telegram-plugin/gateway/gateway.ts:8128-8158` — `clearActivitySummary` refactor is sound: nulls `activityMessageId` before branching, so the dual-call contract (first reply + turn_end safety net) stays idempotent — the second call hits the `== null` early-return. Delete branch is byte-identical to prior #2108 behaviour; finalize branch edits to the terminal render and swallows only 'message is not modified'. No double-send surface.
- `telegram-plugin/tool-activity-summary.ts:203-218, 247-275` — `final` param threads cleanly through `renderActivityFeed`/`renderActivityFeedWithNested`; `final=true` drops the `→ in-progress` arrow on the newest line so the persisted feed doesn't freeze mid-step. Default false preserves the live render exactly. Nested no-children path delegates with `final` forwarded.
- Tests cover both render modes (flat/nested, single-line, default-false keeps →) and the schema (round-trip true/false, optional-unset undefined, non-boolean rejected). docs/configuration.md cascade-table row matches the schema describe().

Rationale: additive per-agent knob, default preserves current UX; the only behavioural change (finalize-instead-of-delete on default) is correctly gated and the delete path is unchanged. Idempotency and env-default ownership are both correct. No code-path risk.

PR #2135, opened 2026-06-03
fix(agents): export gateway-consumed env before the gateway fork in start.sh

mekenthompson:
## Summary

The docker runtime forks the gateway daemon in start.sh's **outer** pass and then re-execs into a tmux **inner** pass for claude. `channels.telegram.*` env knobs (`SWITCHROOM_TG_*`, from `channelsToEnv` → `userEnvQuoted`) were rendered only in the inner pass — *after* the gateway fork — so the gateway daemon, which reads them from `process.env` at startup, never saw them. **Every `channels.telegram` env knob silently defaulted on docker agents.**

Same env-fork landmine class as #1997 (`SWITCHROOM_HANDOFF_SHOW_LINE`).

## How it surfaced

Shipped `channels.telegram.clear_status_on_completion` (#2132 / v0.14.54). Canarying the opt-in on test-harness: setting `clear_status_on_completion: true` did **not** delete the status feed. `/proc/<gateway-pid>/environ` showed **0** `SWITCHROOM_TG_*` vars — the gateway took the default branch because it never received the var.

## Fix

Hoist the `userEnv` export block into the docker outer pass, **before** the gateway fork (`bun "$_gateway_bundle"`). The inner pass keeps its own copy — tmux does not reliably propagate arbitrary env across the re-exec, so claude still needs the inner export.

## Blast radius

**Zero fleet behavior change.** No agent currently sets any `channels.telegram` env knob (verified against the live config — only test-harness's canary `clear_status_on_completion`). This purely makes the knobs functional where set.

## Test plan

- [x] `tests/scaffold.gateway-env-order.test.ts` — `SWITCHROOM_TG_*` export precedes the gateway-fork line (outer pass); same vars re-exported after the fork (inner pass) for claude
- [x] `tsc --noEmit` clean
- [x] full scaffold suite (210) green

## Risk

Low. The `start.sh.hbs` change is additive (extra exports in the outer pass); the inner-pass block is unchanged. Effect takes hold on agent restart (start.sh is re-rendered at reconcile).

### DIFF
--- a/profiles/_base/start.sh.hbs
+++ b/profiles/_base/start.sh.hbs
@@ -33,6 +33,24 @@ if [ "$SWITCHROOM_RUNTIME" = "docker" ] && [ -z "$SWITCHROOM_DOCKER_TMUX_INNER" 
   # same path the rest of start.sh + the MCP sidecar expects.
   export TELEGRAM_STATE_DIR="{{agentDir}}/telegram"
 
+  # Gateway-consumed env MUST be exported HERE, before the gateway fork
+  # below. The gateway daemon reads channels.telegram.* knobs (and any
+  # agent env) from process.env at startup — e.g. SWITCHROOM_TG_STREAM_
+  # THROTTLE_MS, SWITCHROOM_TG_EDIT_BUDGET_THRESHOLD,
+  # SWITCHROOM_TG_CLEAR_STATUS_ON_COMPLETION. The inner tmux pass
+  # re-exports the SAME user-env block below for claude, because tmux
+  # does not reliably propagate arbitrary env across the
+  # re-exec. Without this hoist these vars only ever reach the inner
+  # (claude) pass and the already-forked gateway never sees them — every
+  # channels.telegram env knob silently defaults on docker agents (the
+  # start.sh env-fork landmine, same class as #1997 /
+  # SWITCHROOM_HANDOFF_SHOW_LINE). Pinned by the scaffold-order test
+  # asserting these exports precede the gateway-fork line.
+{{#if userEnvQuoted}}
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `profiles/_base/start.sh.hbs:49-50` — `userEnvQuoted` export block hoisted into the docker outer pass, inside the `[ -z "$SWITCHROOM_DOCKER_TMUX_INNER" ]` guard. Verified at PR head: this renders at line 49, the gateway fork (`bun "$_gateway_bundle" &`) at line 127, so the SWITCHROOM_TG_* exports now precede the fork — the daemon reads them from process.env at startup. Correct fix for the reported defect.
- `profiles/_base/start.sh.hbs:317-322` — inner-pass re-export block is untouched and still present (after the `SWITCHROOM_DOCKER_TMUX_INNER=1` marker at line 166), so claude continues to get the vars across the tmux re-exec. The dual-export rationale holds: tmux doesn't reliably propagate arbitrary env. Uses the same `{{{this}}}` triple-stache (pre-quoted) as the inner block — consistent escaping.
- `tests/scaffold.gateway-env-order.test.ts:55-72` — first occurrence (outer hoist) asserted `< forkIdx`; `tests/...:78-92` asserts a second occurrence `> forkIdx` (inner re-export). Both invariants match the rendered template I inspected. Solid regression guard against re-introducing the landmine.

Rationale: additive template change placing gateway-consumed env before the fork; ordering confirmed against the rendered head template (49 < 127 < 322), inner pass preserved, test pins both halves. Effect on restart only; zero behavior change for agents not setting a channels.telegram knob.

PR #2137, opened 2026-06-03
fix(telegram): deterministic per-topic reply routing for multi-topic supergroups

mekenthompson:
## Summary

In a Telegram forum **supergroup**, asking an agent two questions in two topics back-to-back could land **both answers in one topic**. Operator-reported on marko (Brevo + Meta Campaigns). An agent is **one sequential Claude CLI** with a singleton `currentTurn`, so it can only reply sequentially — but each question must be answered **in its own topic**.

## Root cause (two coupled defects)

1. **Buffer drained on *any* turn-end, not gated on a reply** — when the first topic's turn ended without yet replying, the buffered second-topic message drained immediately, orphaning the first answer into the second turn.
2. **Late replies attribute to `currentTurn` at reply-time** — by the time the first answer landed, `currentTurn` had flipped to the second topic, so the reply routed to the *successor's* thread. (Reply paths were already turn-pinned — but to the *live* flipped pointer.)

`currentTurn→Map` is **not** needed; strict serialization + a stable per-turn id fixes it.

## What changed (5 components, each behind a kill-switch, default ON)

1. **Deliver-before-drain gate** (`SWITCHROOM_SERIALIZE_UNTIL_REPLIED`) — a buffered cross-topic message drains only after the current turn delivered its reply to its own thread.
2. **Bounded no-reply escape hatch** (`SWITCHROOM_SERIALIZE_NOREPLY_DRAIN_MS`=2500) — a turn that legitimately never replies (greeting/handback/NO_REPLY) force-drains the queue after the bound, so it **can't wedge**. (Plus the pre-existing ungated `idleDrainTick` 5s + 300s silence-poke remain independent long-stop backstops.)
3. **Turn-origin routing** (`SWITCHROOM_TURN_ORIGIN_ROUTING`) — a `turnId` stamped on the inbound is carried to the reply; a late reply routes to its **origin** topic, never the successor's.
4. **Per-turn topic framing** (`SWITCHROOM_TOPIC_FRAMING`) — the model is told to answer only the current message's topic.
5. **Queued-status UX** (`SWITCHROOM_QUEUED_STATUS_UX`) — the waiting topic shows one self-updating line: "Queued — replying in #X" → "On it" → **deleted on answer** (+ reaped on any abnormal turn-end and on spool-escalation, so it never dangles).

Two new pure modules (`serialize-drain-gate.ts`, `answer-thread-resolve.ts`) keep the gate predicate + routing precedence unit-testable.

## Test plan

- [x] 45 new/updated tests incl. a **fake-timer liveness proof** the no-reply queue force-drains at the bound (held at MS-1, drained at MS+1) — proves no wedge
- [x] `tsc --noEmit`, `check-bot-api-wrapping`, `check-plugin-references`, `check-no-pii-secrets` all clean
- [x] telegram-plugin: 4045 bun + 5840 vitest pass
- [x] Fresh-process reviewer **APPROVE** — independently verified wedge-proofing (4 drain triggers, 3 ungated), turn-origin routing, queued-status lifecycle, kill-switch reverts, compliance
- [ ] **Supergroup canary required** (the marko forum-topic topology) — DM-only UAT would NOT have caught this. Two questions in two topics → each answered in its own topic; queued-status appears + clears.

## Risk

Medium (core turn machine, incident-dense area) — mitigated by five independent kill-switches, the wedge-proofing (bounded timer + two ungated backstops), and a mandatory supergroup canary before fleet roll.


mekenthompson:
APPROVE

Read the full diff (1584 lines, 9 files) and both pure modules end-to-end. No correctness defect found.

- **Primary fix is sound — component 1 makes component 3 belt-and-suspenders, not load-bearing.** `serialize-drain-gate.ts:968-974` (`mayDrainBufferedInbound`) blocks the cross-topic buffer drain until the ending turn's `finalAnswerDelivered === true`, and that drain fires from inside `executeReply` via `releaseTurnBufferGate` (`gateway.ts:690-696`) AFTER the send resolves its thread. So `currentTurn` cannot flip to the successor before the predecessor's reply has executed — the root flip that caused Brevo→Meta can't occur. This matters because component 3's origin routing on the explicit reply path depends on the model echoing `origin_turn_id`; when omitted it falls back to `liveThreadId = turn?.sessionThreadId` (`gateway.ts:667-675`), which would be the flipped turn if a flip had happened. Component 1 removes the precondition for that flip, so the model-echo dependency is non-load-bearing. Good layering.

- **Liveness is proven, not asserted.** The no-reply turn (`finalAnswerDelivered=false`) deliberately blocks the serialize gate forever; `armNoReplyDrainTimer` (`gateway.ts:423-450`) + `shouldArmNoReplyDrain` (`serialize-drain-gate.ts:997-1001`) force-drain at `SERIALIZE_NOREPLY_DRAIN_MS` (2500ms, clamped positive at `gateway.ts:165-169`). `no-reply-bounded-drain.test.ts:1396-1417` is a real fake-timer liveness proof (held at MS-1, drained at MS+1). The 300s silence-poke + `idleDrainTick` remain ungated independent long-stops (`no-reply-bounded-drain.test.ts:1459-1464`).

- **turnId derivation agrees across both sites.** `deriveTurnId` (`gateway.ts:248-255`) is keyed on chat/thread/messageId — message-id based, not the not-yet-known `startedAt` — so the id stamped into the inbound meta at `gateway.ts:835` matches the id the enqueue handler reconstructs at `gateway.ts:777-779`. Synthetic/cron/handback turns (no messageId) get a `#synthetic-` fallback no reply will echo, correctly falling through to live-turn routing. `recentTurnsById` bounded LRU at 32 (`gateway.ts:227-236`).

- **Drain is idempotent under double-fire.** `performBufferDrain` depth-checks first (`gateway.ts:367`); a delivered turn drains via `releaseTurnBufferGate` then again (no-op, depth 0) via `purgeReactionTracking`→`drainBufferedIfAllowed`. The restart/compaction block correctly stays on the bare `!turnInFlightForGate()` signal (`gateway.ts:494`), not the serialize gate.

- **Component 5 placeholder can't dangle.** Reaped on answer (`executeReply`/`executeStreamReply`), on turn-start promote, on abnormal turn-end (`purgeReactionTracking:465-472`), and on spool escalation (`gateway.ts:632`). Post-await re-check guards the post-race (`gateway.ts:298-305`).

Rationale: the core turn machine is incident-dense, but every component is behind a default-on kill switch, the two new modules are pure and exhaustively unit-tested (45 tests incl. totality proofs), and the primary fix removes the flip precondition rather than racing it. Verdict does not block on the unchecked supergroup-canary box — that is a correctly-flagged deploy-time gate, not a code defect. Ship after the canary.

PR #2139, opened 2026-06-03
fix(telegram): finalize the foreground+ack-first activity feed to done (✓)

mekenthompson:
## Summary

Closes the cosmetic follow-up from v0.14.55. When a foreground sub-agent finishes on an **ack-first** turn ("On it" → `Task`), the persisted status feed froze on its last live `→ in-progress` line instead of reading done (`✓`).

**Root cause:** the handoff-clear path deleted the finished sub-agent from `turn.foregroundSubAgents` *before* calling `clearActivitySummary`. The finalize runs async (`inFlight.then`), so `composeTurnActivity(turn, true)` then saw an empty map **and** empty parent `mirrorLines` (ack-first) → rendered `null` → skipped the finalize edit. The message still **persisted** (the operator's ask), so this was purely cosmetic; common (non-foreground) turns finalized to all-`✓` correctly.

## Fix

In the `handoff-clear` branch, capture `composeTurnActivity(turn, true)` **while the finished sub-agent's narrative is still in the map**, then delete it, then pass the captured render to `clearActivitySummary(turn, finalHtmlOverride)` (new optional param). The two non-foreground callers omit the param → compute as before (byte-identical). The `recompose` branch keeps its delete-then-render order (it deliberately collapses the finished block). Switched the foreground block from delete-up-front to `has()`-then-delete so the capture can precede removal; `remainingForeground` is `size - 1` (equivalent given `has()` confirmed presence).

## Test plan

- [x] `tool-activity-summary.test.ts` — pins the capture-before-delete invariant: ack-first (empty parent) + child present → non-null all-`✓` (no `→`); child removed → `null` (the emptied-feed skip the gateway must avoid)
- [x] existing `foreground-nesting` + activity tests green; full plugin bun suite (4047 pass)
- [x] `tsc`, `check-bot-api-wrapping`, `check-plugin-references` clean
- [ ] live canary: re-run `jtbd-foreground-subagent-activity-dm` + confirm the persisted feed finalizes to `✓` (no `→`)

## Risk

Low/medium — touches the delicate foreground handoff path. The behavior change is confined to the `handoff-clear` case (ack-first + last sub-agent); `recompose`, delete-mode, and the non-foreground callers are unchanged.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -8467,8 +8467,18 @@ async function drainActivitySummary(turn: CurrentTurn): Promise<void> {
  * Called on the first reply (hand-off) and again at turn_end (no-reply safety
  * net); finalize edits are idempotent (a 'message is not modified' on the
  * second call is swallowed).
+ *
+ * `finalHtmlOverride` (finalize path only): a render captured by the caller
+ * BEFORE it tore down turn state the finalize render depends on. The
+ * foreground handoff-clear path passes this — it deletes the just-finished
+ * sub-agent's narrative right after this call, so the async
+ * `composeTurnActivity(turn, true)` below would see an emptied feed (and, on
+ * ack-first turns, empty `mirrorLines`), render null, and skip the finalize —
+ * freezing the last live "→ in-progress" line. The captured render keeps the
+ * persisted record reading done (✓). Omitted → compute it here (the common
+ * reply/turn_end callers, where state is stable).
  */
-function clearActivitySummary(turn: CurrentTurn): void {
+function clearActivitySummary(turn: CurrentTurn, finalHtmlOverride?: string | null): void {
   const chat = turn.sessionChatId
   const thread = turn.sessionThreadId
   const inFlight = turn.activityInFlight ?? Promise.resolve()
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `clearActivitySummary` (gateway.ts:8471→21) gains optional `finalHtmlOverride?: string | null`; the finalize body at gateway.ts:8499-8501 uses `finalHtmlOverride !== undefined ? finalHtmlOverride : composeTurnActivity(turn, true)`. The `!== undefined` guard (not a truthy check) correctly honors an explicit `null` override while letting the two non-foreground callers (8747, 9086) omit the param and compute as before — byte-identical to the old single-arg path. Backward-compat confirmed against head.
- Root-cause fix at gateway.ts:19289-19301: switched from delete-up-front to `has()`-then-delete, capturing `composeTurnActivity(turn, true)` WHILE the finished sub-agent's narrative is still in `foregroundSubAgents`, then deleting, then passing the captured render. This removes the emptied-feed/null-render skip that froze the persisted feed on a stale `→ in-progress` line on ack-first turns. Correct: the capture precedes the mutation the render depends on.
- `remainingForeground: turn.foregroundSubAgents.size - 1` (gateway.ts:19295) is equivalent to the old post-delete `size` because `has()` confirmed presence; `removed: true` literal matches the old post-delete truthy value. No behavior change to `foregroundFinishAction`'s branch selection.
- `recompose` branch (gateway.ts:19297-19300) deliberately keeps delete-then-render order so the finished block collapses — correct and intentionally distinct from handoff-clear.
- Test (tool-activity-summary.test.ts:90-107) pins the invariant both directions: ack-first empty parent + child present → non-null all-`✓` (no `→`); child removed → `null`. This is exactly the gateway's pre-delete-capture contract.

Rationale: cosmetic-only fix (message already persisted) with the root cause removed at the correct point — capture before the dependent teardown — and backward-compat preserved for the two unchanged callers. No correctness defect. The unchecked live canary (re-run jtbd-foreground-subagent-activity-dm) is a deploy gate, not a code blocker.

PR #2141, opened 2026-06-04
fix(telegram): keep the live activity feed alive when an ack-first turn keeps working

mekenthompson:
## Summary

A supergroup agent that **acks first** ("on it, checking Brevo…") then does the actual work showed **no live status** during the work — operator-reported on marko's CRM Brevo topic ("not reporting status on what he is doing").

## Root cause (verified live, not guessed)

The activity feed works fine in topics (proved: a tools-then-answer turn narrates live in a non-General topic). The gap: the feed is gated off in the `tool_label` handler by `if (turn.finalAnswerDelivered) return`, and `finalAnswerDelivered` is set by **any** reply `isFinalAnswerReply` deems final (pings **or** ≥200 chars) — including a short pinging "on it" ack. So an ack-first reply darkens the feed for the work that follows. (Two deep-dive theories — missing `$TELEGRAM_STATE_DIR`, missing `message_thread_id` on edits — were checked against the running system and **refuted**.) Long-standing gap (the "surface foreground activity after an ack" item), not introduced by #2137/#2139.

## Fix

When a `tool_label` arrives after `finalAnswerDelivered` **and the prior reply was a short ACK, not a substantive final answer**, re-open a fresh live feed for the post-ack work (reset `finalAnswerDelivered=false`, new feed message). "Substantive" = `isSubstantiveFinalReply` = `done || ≥200 chars` (final-detect minus the ping-only path); tracked per-turn as `finalAnswerSubstantive`. So:
- short pinging "on it" ack → work → **feed re-opens** (the fix);
- genuine final answer (long/stream-done) → post-answer housekeeping (memory write / TodoWrite / Bash) → **no reset, no spurious silent-end re-prompt, no duplicate answer** (the regression the ack-only refinement closes).

Kill-switch `SWITCHROOM_FEED_REOPEN_AFTER_ACK` (default ON; `=0` → exact legacy gate). Pure decision module `feed-reopen-gate.ts` + `isSubstantiveFinalReply`.

## Why it's safe (all 3 `finalAnswerDelivered` consumers, both scenarios — reviewer-verified)

- **silent-end re-prompt** (`finalAnswerDelivered===false` at turn-end): genuine-final + housekeeping keeps the flag true → no spurious re-prompt / duplicate. Ack-then-work-then-no-answer keeps it false → re-prompts (correct).
- **#2137 deliver-before-drain**: ack-first holds the next topic until the real answer (no cross-topic bleed); the bounded no-reply timer still releases the queue → no wedge.
- **feed spam**: one fresh feed per ack-cycle, then edit-in-place (not one message per tool).

## Test plan

- [x] `feed-reopen-gate` + `isSubstantiveFinalReply` unit tests (reopen iff ack & enabled; no reopen when substantive; kill-switch OFF = legacy)
- [x] no-regression: serialize-drain / multitopic-routing / silent-end / no-reply-bounded-drain green; 4060 non-UAT telegram-plugin tests pass
- [x] `tsc`, bot-api-wrapping, plugin-references, PII clean; gateway-only (compliant)
- [x] Two fresh-process reviewers APPROVE (2nd independently verified the duplicate-answer regression is closed)
- [ ] Supergroup canary: ack-first-then-work turn in a topic → live feed narrates the post-ack work; + a genuine-final-then-memory-write turn → no duplicate answer

## Residual (acknowledged, kill-switchable)

A *short pinging real answer* ("Done!") + post-answer housekeeping still re-opens a small feed — much rarer than the long-answer case, mitigated by the over-ping downgrade, strictly better than going dark.

### DIFF
--- a/telegram-plugin/final-answer-detect.ts
+++ b/telegram-plugin/final-answer-detect.ts
@@ -81,3 +81,37 @@ export function isFinalAnswerReply(input: FinalAnswerReplyInput): boolean {
   if (input.text.length >= FINAL_ANSWER_MIN_CHARS) return true
   return false
 }
+
+/**
+ * Pure predicate: was this reply a *substantive* final answer (as opposed
+ * to a reply that is only "final" because it pinged)? `true` if EITHER:
+ *
+ *   - `done === true` — a `stream_reply` terminal call closing the stream.
+ *   - `text.length >= FINAL_ANSWER_MIN_CHARS` — a substantive-length answer.
+ *
+ * This is `isFinalAnswerReply` MINUS the notification-only path. The
+ * distinction matters for the feed-reopen-after-ack gate
+ * (`feed-reopen-gate.ts`): a *short pinging* reply ("on it, checking
+ * Brevo…") is classified final by `isFinalAnswerReply` (because it pings)
+ * yet is NOT substantive — it is an interim ACK. Only such an ack should
+ * cause the live activity feed to re-open when post-ack tool work arrives.
+ *
+ * A genuine final answer (long, or a stream `done: true`) followed by
+ * routine post-answer housekeeping (a memory write / TodoWrite / Bash —
+ * none of which are surface tools, so they reach the tool_label handler)
+ * must NOT re-open the feed and must NOT reset `finalAnswerDelivered`,
+ * otherwise the silent-end re-prompt would spuriously fire and the agent
+ * would re-deliver a duplicate / garbled answer.
+ *
+ * Residual: a reply that is genuinely the final answer yet is BOTH short
+ * (<200 chars) AND pinging (e.g. "Done!") is indistinguishable here from
+ * an ack, so post-answer housekeeping after it still re-opens the feed.
+ * That is much rarer than the housekeeping-after-long-answer case this
+ * predicate protects, and is kill-switchable via
+ * `SWITCHROOM_FEED_REOPEN_AFTER_ACK=0`.
+ */
+export function isSubstantiveFinalReply(input: FinalAnswerReplyInput): boolean {
+  if (input.done === true) return true
+  if (input.text.length >= FINAL_ANSWER_MIN_CHARS) return true
+  return false
+}
--- a/telegram-plugin/gateway/feed-reopen-gate.ts
+++ b/telegram-plugin/gateway/feed-reopen-gate.ts
@@ -1,0 +1,127 @@
+/**
+ * Feed-reopen-after-ack gate (ack-first live-activity visibility).
+ *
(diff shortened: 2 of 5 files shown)


mekenthompson:
APPROVE

Gateway-only feed-visibility fix. Verified against the real diff and the full gateway.ts.

- **Root cause correctly isolated** (gateway.ts:8838-8868). The legacy `if (turn.finalAnswerDelivered) return` dropped every `tool_label` after an ack-first reply, because `isFinalAnswerReply` classifies any pinging OR ≥200-char reply as final — including a short "on it" ack. The fix re-opens a fresh feed for post-ack work via the pure `decideFeedReopen` gate.
- **Coverage of the critical invariant holds** — every site that sets `finalAnswerDelivered = true` also sets `finalAnswerSubstantive` via `isSubstantiveFinalReply`: gateway.ts:6307/6310, 6646/6650, 6989/6996, 9150/9155 (=true), 9426/9431 (=true), plus the fresh-turn init at 8623 (false). No final-delivered site leaves `finalAnswerSubstantive` defaulted, so a genuine final can never be misread as an ack.
- **The harmful direction is unreachable.** Reopen fires only on `finalAnswerDelivered && !finalAnswerSubstantive && enabled` (feed-reopen-gate.ts:159-163). Because both flags are co-set, the dangerous stale-false case cannot occur; the reopen path leaving `finalAnswerSubstantive` unreset (gateway.ts:8865-8867) is the safe direction (only suppresses a future reopen until the real answer re-sets both).
- **Regression it closes is real and tested.** Post-answer housekeeping (memory write / TodoWrite / Bash — non-surface tools that reach the handler) after a substantive answer keeps `finalAnswerDelivered=true` → no reset → silent-end re-prompt does not spuriously fire → no duplicate answer. feed-reopen-gate.test.ts:501-512 pins this; final-answer-detect.test.ts:588-598 pins the ack vs substantive split (short pinging "on it" → final but not substantive).
- **`activityLastSentRender` reset** (gateway.ts:8867) correctly prevents the drain loop's `pending !== lastSent` guard from skipping the fresh feed render after reclassifying the ack.
- **Kill switch** `SWITCHROOM_FEED_REOPEN_AFTER_ACK=0` (gateway.ts:8848) cleanly reverts to the exact legacy gate; covered by tests.

Rationale: pure decision module, fully unit-tested both directions, all final-delivered sites paired, three `finalAnswerDelivered` consumers reasoned through and consistent. Only open item is the unchecked supergroup canary — a deploy-time verification gate, not a code blocker. The acknowledged residual (short pinging real answer + housekeeping re-opens a small feed) is rare, kill-switchable, and strictly better than going dark.

PR #2143, opened 2026-06-04
feat(telegram): activity-feed heartbeat so long single steps don't freeze

mekenthompson:
## Summary (systems-analysis PR1 of 3 — the "no status" half)

The activity feed is **pull-only**: it re-renders only when a `tool_label` event arrives. A long single step (a 26s Meta-ads pull, a long think) emits no new label, so the feed freezes on "→ doing X" for tens of seconds and the turn reads as ghosting. Confirmed on marko: feed froze ~26s, only the typing indicator moving, `cardMsgId=none`.

Add a **deterministic, framework-owned heartbeat**: a 6s wall-clock timer re-renders the live feed through the *same* single-flight drain path with a climbing `· Ns` elapsed on the in-progress line — but only once the current step has run ≥6s (so a normally-advancing feed is untouched). No new transport, no race; it reuses `activityPendingRender` + `drainActivitySummary`. Skips after the final answer (feed handing off) and after turn-end (`activityMessageId` nulled).

This is the low-risk, independent half of the systems analysis — it addresses the **"no status"** freeze only, not the drop (that's PR2, the obligation ledger).

## Implementation
- `renderActivityFeed` / `renderActivityFeedWithNested` gain an optional `liveSuffix` (default `''` → byte-identical for every existing caller; ignored when `final=true` so a finalized record never ticks).
- `CurrentTurn.lastToolLabelAt` re-anchors on each label → the elapsed restarts per step.
- `feedHeartbeatTick` + a single `setInterval`, gated `!STATIC && FEED_HEARTBEAT_ENABLED`.

## Test plan
- [x] `tool-activity-summary.test.ts`: `liveSuffix` on flat + nested newest line; `final=true` ignores it; empty suffix byte-identical to no-suffix
- [x] `tsc --noEmit`, `check-bot-api-wrapping` clean
- [x] full plugin bun suite — 4066 pass / 0 fail
- [ ] canary: confirm the feed visibly advances during a long MCP turn (mtcute-observable, message transport)

## Risk
Low. Kill switch `SWITCHROOM_FEED_HEARTBEAT=0`. Default on. The only new behavior is an extra in-place edit of an already-existing feed message every ~6s during a stale step; subject to the existing drain throttle/budget.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1431,6 +1431,25 @@ const QUEUED_STATUS_UX_ENABLED =
 const FEED_REOPEN_AFTER_ACK_ENABLED =
   process.env.SWITCHROOM_FEED_REOPEN_AFTER_ACK !== '0'
 
+// Activity-feed heartbeat (PR1). The feed is pull-only — it only re-renders on
+// a tool_label event, so a long single step that emits no new label leaves the
+// feed frozen on "→ doing X" for tens of seconds (the marko ~26s freeze). The
+// heartbeat re-renders the live feed every FEED_HEARTBEAT_TICK_MS with a
+// climbing " · Ns" elapsed on the in-progress line, but only once the current
+// step has run >= FEED_HEARTBEAT_MIN_STALE_MS (so a normally-advancing feed is
+// untouched). Kill switch: SWITCHROOM_FEED_HEARTBEAT=0. Default on.
+const FEED_HEARTBEAT_ENABLED = process.env.SWITCHROOM_FEED_HEARTBEAT !== '0'
+const FEED_HEARTBEAT_TICK_MS = 6_000
+const FEED_HEARTBEAT_MIN_STALE_MS = 6_000
+
+/** Compact mm/ss-ish elapsed for the live feed suffix: "18s", "1m05s". */
+function formatFeedElapsed(ms: number): string {
+  const s = Math.floor(ms / 1000)
+  if (s < 60) return `${s}s`
+  const m = Math.floor(s / 60)
+  return `${m}m${(s % 60).toString().padStart(2, '0')}s`
+}
+
 /**
  * Authoritative "is a turn in flight?" for every gate that previously
  * read `claudeBusyKeys.size`. PR 3b cutover (extends PR 3a's bridgeUp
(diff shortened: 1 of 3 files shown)


mekenthompson:
REQUEST_CHANGES

- **Blocker — the heartbeat is a no-op; the elapsed suffix never reaches the render.** `feedHeartbeatTick` calls `composeTurnActivity(turn, false, ` · ${formatFeedElapsed(elapsed)}`)` (gateway.ts:8541, the only 3-arg call site), but `composeTurnActivity` is unchanged in this PR — it still takes `(turn, final = false)` (gateway.ts:8445) and its body calls `renderActivityFeedWithNested(turn.mirrorLines, childLines, final)` with **no** liveSuffix. The third argument is silently discarded. So every heartbeat tick recomposes the feed *without* the ` · Ns` suffix, producing a string byte-identical to the last render (mirrorLines don't change during a stale step). The drain dedup loop (`while (turn.activityPendingRender !== turn.activityLastSentRender)`, gateway.ts:8472) then sees pending == lastSent and sends nothing. Net effect: the marko ~26s freeze is unchanged. Fix: give `composeTurnActivity` a `liveSuffix = ''` param and forward it — `renderActivityFeedWithNested(turn.mirrorLines, childLines, final, liveSuffix)`.
- **The 'tsc --noEmit clean' claim in the test plan is contradicted by the code.** Calling a 2-parameter function with 3 arguments is a TS2554 error ('Expected 1-2 arguments, but got 3') at gateway.ts:8541. Either the type-check wasn't run against this head (93dd633) or it was run against a version where `composeTurnActivity` had the extra param. Re-run `tsc --noEmit` on the actual head after wiring the param.
- **Tests pass but don't cover the integration path.** tool-activity-summary.test.ts exercises `renderActivityFeed`/`renderActivityFeedWithNested` with `liveSuffix` directly (and that plumbing is correct: tool-activity-summary.ts:161-184, 256-281). None of them go through `composeTurnActivity` or `feedHeartbeatTick`, which is exactly where the wire breaks. Add a test that drives the suffix through `composeTurnActivity` so the dead-end is caught.

Rationale: the `renderActivityFeed` plumbing, the gating (`activityMessageId`/`finalAnswerDelivered`/min-stale guards, gateway.ts:8527-8531), the per-step re-anchor (gateway.ts:8944), and the kill switch are all correct — but `composeTurnActivity` is the one link in the runtime chain that wasn't updated, so the feature ships doing nothing. One-line fix, but it's load-bearing.

PR #2088, opened 2026-06-02
fix(auth-broker): route Microsoft provider in dispatch + refresh (RFC #1873 repair)

mekenthompson:
## Summary

The auth-broker **registered** `MicrosoftProvider` but never routed it. The `get-credentials` / `add-account` / `rm-account` dispatch only branched on `anthropic` and `google`; `microsoft` fell through to `INTERNAL: unhandled provider 'microsoft'` (`server.ts`). So the M365 integration was **non-functional end-to-end** despite provider + storage + CLI + launcher all existing — `m365-mcp-launcher`'s `getCredentials("microsoft")` errored, and `auth microsoft account add` errored at the broker. (Found during an out-of-box-OAuth design review; the docs claimed all 5 PRs shipped but the dispatch wiring was missing.)

This wires the Microsoft branch into all three dispatch sites + the refresh loop, mirroring the existing Google ops.

## Why

Prerequisite for any Microsoft connect/out-of-box work — nothing Microsoft functions until the broker can serve its credentials. Pure repair: no behavior change for Anthropic/Google.

## What changed

- **`opMicrosoftGetCredentials`** — per-agent only; account derived from the agent's `microsoft_workspace.account` selector (path-as-identity, never the wire); ACL gated on `microsoft_accounts.<account>.enabled_for[]`. Mirrors `opGoogleGetCredentials`.
- **`opMicrosoftAddAccount` / `opMicrosoftRmAccount`** — admin-gated; label path-traversal guard; `rm` refuses while any agent is still in `enabled_for[]`.
- **`refreshOneMicrosoftAccount` + refresh-tick loop** — keeps the stored access token fresh for the launcher. Passes `priorCredentials` (Google's path doesn't need this) so canonical identity fields (`tenantId`/`accountType`/`homeAccountId`) survive a refresh response that omits `id_token`; persists the rotated refresh token atomically.
- **`list-microsoft-accounts`** op + client method; un-stubs `auth microsoft account list` (was a placeholder). Surfaces `accountType` so personal MSA vs work is visible. No tokens returned.

## Test plan

[...]

PR #2144, opened 2026-06-04
feat(auth): ship a default Microsoft OAuth app — zero-config connect

mekenthompson:
## Summary

Out-of-box external-account connect, **Microsoft-first**. Until now, every operator had to register their own Entra app before any Microsoft account could connect — the code literally said *"switchroom ships no shared Microsoft OAuth app — register your own."* That fails the Defaults principle (*"batteries included, assembly optional — make users opt **into** complexity"*). This flips it: switchroom.ai ships a default Microsoft **public client**; BYO becomes the documented opt-in.

Builds on #2088 (which repaired the broker so Microsoft works end-to-end at all).

## Why

Operator decision (Ken, 2026-06-02): Microsoft gets the full out-of-box treatment (shared multi-tenant public client, no secret); Google stays BYO/CLI (restricted-scope verification + annual CASA make a shared Google app unviable). Serves the `talk-to-agents-from-anywhere` JTBD (non-technical partner, phone-first) and passes the Defaults test.

## What changed

- **`src/auth/default-oauth-clients.ts` (new)** — `DEFAULT_MICROSOFT_CLIENT_ID` (switchroom.ai's multi-tenant + personal-accounts Entra public client; "Allow public client flows" on, **no secret**). A client_id is a public identifier — safe to ship, same pattern as the Anthropic `DEFAULT_CLIENT_ID`. `resolveMicrosoftClientId()` is the single source of truth for precedence **env → config → default** and reports the source.
- **`auth-microsoft.ts`** — connect no longer hard-errors without a `microsoft_workspace` block; resolves the default + prints a one-line "using shipped default; BYO via config" note. Removed the now-dead/contradicted `microsoftClientSetupGuidance`.
- **`broker/server.ts`** — `MicrosoftProvider` now registers **unconditionally** via the resolver (was gated on `microsoft_client_id`). Safe: no accounts → refresh tick is a no-op, get-credentials returns `ACCOUNT_NOT_FOUND`. This is what makes refresh work out of the box.
- **`doctor-microsoft.ts`** — "no block / empty client_id" flips from **fail → ok** ("using shipped default Microsoft OAuth app"). Absence is now the normal zero-config case.
- **`schema.ts`** — `microsoft_client_id` is now `.optional()` so a block can carry only `org_mode`/`authority` while still using the default.

## Validation

- The shipped client_id was checked **live** against `https://login.microsoftonline.com/common/oauth2/v2.0/devicecode` — it returns a `user_code` for the full `Mail.ReadWrite Calendars.ReadWrite Files.ReadWrite.All` scope set, confirming multi-tenant + public-client-flows are correctly configured.
- MCP servers are BYOT, so this is a pure switchroom-side credential change — the MCP servers don't care whose app minted the token.
- Does **not** touch the Claude-native/subscription compliance boundary (that's strictly the model/CLI inference path; Google/MS OAuth is third-party tooling under the vault/leash).

## Test plan

- [x] `src/auth/default-oauth-clients.test.ts` (8) — precedence env>config>default, empty-config→default, vault-ref passthrough, env-trim/blank, GUID shape.
- [x] `server-microsoft.test.ts` extended — get-credentials **and** refresh-tick now succeed with **no** `microsoft_workspace` block (default registers the provider).
- [x] `doctor-microsoft.test.ts` — flipped no-block / empty-id to the new `ok` semantics.
- [x] Full config + broker + MS CLI/doctor suites green (**335**); `tsc --noEmit` clean.

## Risk

Low–moderate. The behavior change is "MicrosoftProvider always registers" — verified harmless for non-Microsoft installs (no accounts → no-op tick; get-credentials gated per-agent). No secret is distributed (public client). BYO override preserved (env + config). The per-agent grant (connecting an account ≠ an agent getting access) is unchanged — the leash holds.

## Follow-ups (not in this PR)

- Phase 2: Telegram-native `/connect microsoft` device-code flow (the phone self-serve win).
- Phase 4: BYO-app walkthrough in `docs/microsoft-workspace.md`; setup-wizard step.

### DIFF
--- a/src/auth/broker/server-microsoft.test.ts
+++ b/src/auth/broker/server-microsoft.test.ts
@@ -334,6 +334,46 @@ describe("AuthBroker — Microsoft get-credentials op", () => {
 
     broker.stop();
   });
+
+  it("serves credentials with NO microsoft_workspace block — the shipped default registers the provider", async () => {
+    // Out-of-box: even without a microsoft_workspace: block, the broker
+    // registers MicrosoftProvider via the shipped default client_id, so
+    // get-credentials for an enabled agent succeeds end-to-end.
+    const h = makeHarness();
+    const config = makeConfig(h, {
+      agents: { ziggy: { microsoft_workspace: { account: "lisa@outlook.com" } } },
+      microsoftAccounts: { "lisa@outlook.com": { enabled_for: ["ziggy"] } },
+      withProvider: false, // NO microsoft_workspace block at all
+    });
+    seedMicrosoftAccount(h, "lisa@outlook.com", {
+      expiresAt: Date.now() + 3600_000,
+      accessToken: "at-lisa-default",
+    });
+
+    const broker = new AuthBroker(config, {
+      home: h.home,
+      stateDir: h.stateDir,
+      socketRoot: h.socketRoot,
+      disableRefreshLoop: true,
+    });
+    await broker.start();
+
+    const resp = (await rpc(join(h.socketRoot, "ziggy", "sock"), {
+      v: 1,
+      id: "ms-default",
+      op: "get-credentials",
+      provider: "microsoft",
+    })) as {
+      ok: true;
+      data: { account: string; credentials: MicrosoftCredentialsShape };
+    };
+
+    expect(resp.ok).toBe(true);
+    expect(resp.data.account).toBe("lisa@outlook.com");
+    expect(resp.data.credentials.microsoftOauth.accessToken).toBe("at-lisa-default");
+
+    broker.stop();
+  });
 });
 
 describe("AuthBroker — Microsoft add-account / rm-account ops", () => {
@@ -651,7 +691,12 @@ describe("AuthBroker — Microsoft refresh-tick", () => {
     broker.stop();
   });
 
-  it("is a no-op when microsoft_workspace config is absent (provider not registered)", async () => {
+  it("refreshes even with NO microsoft_workspace block — the shipped default registers the provider", async () => {
+    // Out-of-box behavior change: a default Microsoft public client_id
+    // ships, so the broker registers MicrosoftProvider unconditionally
+    // (resolveMicrosoftClientId falls back to the default). A near-expiry
+    // account is therefore refreshed even when the operator never added
+    // a microsoft_workspace: block.
     const h = makeHarness();
     const config = makeConfig(h, { agents: {}, withProvider: false });
     seedMicrosoftAccount(h, "alice@outlook.com", {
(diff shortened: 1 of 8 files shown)


mekenthompson:
APPROVE

Verified against the real diff (head 34e9d13) and the touched files. Sound, well-tested, no blockers.

- **Resolver precedence is correct** (default-oauth-clients.ts:284-295): env (trimmed, blank-ignored) → config (verbatim, vault: refs passed through for the caller to resolve) → shipped default. Single source of truth, called by both the broker ctor and the CLI add path so they can't drift. Tests cover every branch (default-oauth-clients.test.ts:144-208).
- **Unconditional broker registration is safe** (broker/server.ts:106-122). Confirmed three ways: `assertConfigConsistent` (server.ts) only validates consumer names — registration is irrelevant to it; `MicrosoftProvider.refresh` is per-account (microsoft-provider.ts:75) and the refresh tick iterates accounts on disk, so zero accounts = no-op; get-credentials is gated per-agent and returns ACCOUNT_NOT_FOUND with no account. The non-Microsoft-install case is genuinely harmless.
- **CLI no longer hard-errors without a block** (cli/auth-microsoft.ts:312-348): resolves the default, prints a gray 'using shipped default; BYO via config' note only when source==='default', and `mw?.` optional-chaining is applied consistently (lines 327, 337, 358). Dead `microsoftClientSetupGuidance` removed cleanly (no remaining callers in the diff).
- **Doctor flip is right** (cli/doctor-microsoft.ts:451-483): 'no block' and 'empty client_id' collapse into one ok branch ('using shipped default'). Matches the updated tests (doctor-microsoft.test.ts:419-444).
- **Schema** (config/schema.ts:488-500): `microsoft_client_id` is now `.min(1).optional()` — present-but-empty still rejected, absent allowed. Correct.
- **Security:** shipping a public-client client_id (no secret, PKCE/device-code) is the standard pattern — same as Azure CLI and the existing Anthropic DEFAULT_CLIENT_ID. A client_id is a public identifier, not a credential. No secret is distributed.

Nits (non-blocking):
- `.min(1).optional()` means a literal `microsoft_client_id: ""` in yaml fails schema validation at load, so the 'empty-string config → default' branch in resolveMicrosoftClientId/doctor is only reachable via direct object mutation (as the tests do), not a real config load. Harmless defensive handling, just slightly redundant.
- The old 'no-op when microsoft_workspace absent (provider not registered)' test was repurposed to seed an account and assert a refresh (server-microsoft.test.ts:57). The pure zero-account / provider-registered no-op — the exact 'harmless for non-Microsoft installs' claim — no longer has a direct test. Consider keeping one.

Rationale: the behavior change ('MicrosoftProvider always registers via the resolved default') is verified harmless for non-Microsoft installs, the BYO override is preserved through both env and config, the per-agent leash is untouched, and the dead BYO-only guidance is fully removed. Safe to merge.

PR #2145, opened 2026-06-04
feat(telegram): deterministic delivery-obligation ledger (flagged off)

mekenthompson:
## Summary (systems-analysis PR2 of 3 — the delivery *guarantee*)

The framework tracks whether an inbound was **delivered** (read by claude), never **answered** — so a message claude reads but verbally defers (the marko msg-715 silent drop) is lost with no error. This adds the single missing invariant, **model-independent by construction**:

> An inbound is an **obligation** keyed by `origin_turn_id`. OPEN at receipt; CLOSED only by an observable framework event — a **substantive** reply resolving to that origin (never the model's narration/promise). The engine re-presents an OPEN obligation at a clean idle until it closes, **bounded** so a mis-close degrades to one operator-visible nudge, not an infinite re-ask.

The close event is a reply echoing back a framework-stamped id — so the 715 *"I'll handle thread 3 as its own turn"* does **not** close it. Premature `turn_end`, no-reply turns, ack-then-ghost — none close it; all get re-presented.

**Default OFF** (`SWITCHROOM_OBLIGATION_LEDGER`). Every hook is a no-op when off → zero behavior change. Safe to merge before the canary; the flag is flipped only after the gating UAT passes.

## Design — pure core + thin gateway hooks (all flagged)
- **`obligation-ledger.ts`** (pure, 12 unit tests): OPEN/CLOSE/`decideAtIdle`/bounded-represent + `buildObligationRepresentInbound` (synthetic must-answer inbound carrying the original `message_id` + `origin_turn_id` so the reply resolves back) + `obligationEscalationText`.
- **OPEN** — `handleInbound`, same gate as delivery-tracking (real user turns only; `deriveTurnId` null-guards synthetics). Idempotent.
- **CLOSE** — `executeReply` ×3, on `finalAnswerSubstantive`, resolved to the origin the answer *routes* to (`findTurnByOriginId ?? live`) — so 713's reply closes 713 even after `currentTurn` flipped to 715.
- **RE-PRESENT** — a 5s idle sweep, only when no turn in flight **and** the buffer is empty (can't double-present with the existing drain); pushes the synthetic through the same buffer→drain path. Bounded (2) → escalate + close.
- **ADDITIVE** — `armNoReplyDrainTimer` / spool / delivery-queue untouched. PR3 retires the now-redundant pieces only after this proves out.

## Why "substantive" closes (not any reply)
A bare interim ack must not close, or ack-then-ghost re-drops (a real marko pattern). Residual: a genuinely short answer reads non-substantive → bounded re-ask then escalation — **the accepted double-ask tradeoff, to be measured in canary, not assumed.** The 715 headline case has *no* reply at all, so it's always re-presented regardless.

## Test plan
- [x] `obligation-ledger.test.ts` (12): open/close/idempotent, oldest-first, bounded represent→escalate, the 715 scenario, builder origin, escalation text
- [x] `tsc`, `check-plugin-references`, `check-bot-api-wrapping` clean
- [x] full plugin bun suite — 4073 pass / 0 fail
- [ ] **GATING canary (flag ON):** 713/715 interleave on test-harness supergroup — deferred topic answered **once** (not zero = old drop, not twice = double-ask); then marko, then fleet

## Risk
Medium, but **inert until the flag is on**. New failure mode = double-ask (bounded + escalation). Land additively now; activate only after the canary.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1765,9 +1765,48 @@ function closeObligationOnSubstantiveReply(
   liveTurn: CurrentTurn | null | undefined,
 ): void {
   if (!OBLIGATION_LEDGER_ENABLED) return
-  const originTurn = findTurnByOriginId(args.origin_turn_id as string | undefined) ?? liveTurn
-  const oid = originTurn?.turnId
-  if (oid != null) obligationLedger.close(oid)
+  const echoed = findTurnByOriginId(args.origin_turn_id as string | undefined)
+  const target = obligationLedger.resolveCloseTarget(echoed?.turnId, liveTurn?.turnId)
+  if (target != null) obligationLedger.close(target)
+}
+
+/**
+ * PR2 obligation-ledger OPEN. Track a fresh user inbound as an unanswered
+ * obligation the moment it is received — called BEFORE the buffer-until-idle /
+ * deliver split so a mid-turn cross-topic inbound (the 715 case: buffered while
+ * another turn runs) is tracked too. Drain paths re-deliver buffered inbounds
+ * via sendToAgent (NOT through handleInbound), so handleInbound is the ONLY
+ * point that sees every fresh inbound — opening here is mandatory for both
+ * branches. Same gate as delivery-tracking (real user turns only; synthetic /
+ * steering / `!` interrupt / empty excluded — they have no origin id / need no
+ * answer; deriveTurnId null-guards them). Idempotent → opening at buffer-time
+ * and any later delivery is safe. No-op unless the flag is on.
+ */
+function openObligationFromInbound(
+  inboundMsg: InboundMessage,
+  gate: { isSteering: boolean; isInterrupt: boolean; effectiveText: string },
+): void {
+  if (!OBLIGATION_LEDGER_ENABLED) return
+  if (
+    !shouldTrackDelivery({
+      isSteering: gate.isSteering,
+      isInterrupt: gate.isInterrupt,
+      hasSource: inboundMsg.meta?.source != null,
+      effectiveText: gate.effectiveText,
+    })
+  ) {
+    return
+  }
+  const oid = deriveTurnId(inboundMsg.chatId, inboundMsg.threadId, inboundMsg.messageId)
+  if (oid == null) return
+  obligationLedger.openIfAbsent({
+    originTurnId: oid,
+    chatId: inboundMsg.chatId,
+    threadId: inboundMsg.threadId,
+    messageId: inboundMsg.messageId,
+    text: inboundMsg.text ?? '',
+    openedAt: Date.now(),
+  })
 }
 
 /**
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

Verified against the real diff (head b42ed12) and the full gateway.ts. Additive, flagged off, inert until activation — safe to merge.

- **Pure core is correct and well-tested** (obligation-ledger.ts): OPEN is idempotent on originTurnId keeping the earliest (lines 263-267), close is null-safe (270-273), decideAtIdle picks the oldest and escalates at maxRepresents (314-319), markRepresented bumps only an open id (322-327). The 12 unit tests (obligation-ledger.test.ts) cover open/close/idempotent/oldest-first/bounded-represent→escalate and the 713/715 interleave. No mutation in decideAtIdle — caller performs the side effect then markRepresented/close. Clean seam.
- **Zero behavior change when off, by construction.** Every hook early-returns on `!OBLIGATION_LEDGER_ENABLED`: OPEN (gateway.ts:11607), closeObligationOnSubstantiveReply (gateway.ts:1766), obligationSweep (gateway.ts:4756), and the sweep `setInterval` itself is gated `if (!STATIC && OBLIGATION_LEDGER_ENABLED)` (gateway.ts:4787) so it isn't even registered. The ledger object is constructed but never populated. Truly no-op.
- **OPEN gate is the right one** (gateway.ts:11607-11627): reuses `shouldTrackDelivery` (the delivery-tracking gate, imported line 298) plus a `deriveTurnId` null-guard — steering / `!` interrupt / synthetic-source / empty all excluded. The re-presented synthetic carries `meta.source='obligation_represent'` so hasSource=true → it never opens a fresh obligation (obligation-ledger.ts:362), and openIfAbsent idempotency is a second guard.
- **CLOSE resolves to the routed origin, not the live turn** (gateway.ts:1763-1770): `findTurnByOriginId(args.origin_turn_id) ?? liveTurn`. A reply echoing 713's origin closes 713 even after currentTurn flipped to 715; a normal reply (no echoed origin) falls back to the live turn. Gated on `turn.finalAnswerSubstantive` (the #2141 signal, set via isSubstantiveFinalReply at gateway.ts:6423) at all three reply sites (6427, 6780, 7125) — a bare ack does NOT close, so ack-then-ghost is re-presented, not re-dropped.
- **Re-present delivery path is real and tested.** obligationSweep pushes the synthetic into `pendingInboundBuffer` only at a clean idle (no turn in flight AND buffer depth 0, gateway.ts:4757-4760), and the existing idle-drain `setInterval` (gateway.ts:5814, same 5s cadence, same `turnInFlightForGate` guard) drains it to the agent. Bounded at 2 then escalate, and escalate **closes first** (gateway.ts:4773) so the loop ends even if the send throws.

Nit (non-blocking): the escalation send is fire-and-forget after close (gateway.ts:4775-4783) — if `robustApiCall` ultimately fails, the obligation is already closed, so that one escalation is silently lost. Deliberate (close-first prevents an infinite loop) and low-probability given the retry wrapper, but it is the residual edge: a failed escalation = no operator-visible nudge. Worth a stderr-only metric to spot it in the canary.

Rationale: the invariant (an inbound is an obligation closed only by an observable substantive reply, re-presented bounded otherwise) is implemented model-independently, the integration symbols all exist with the claimed semantics, and the whole thing is inert behind SWITCHROOM_OBLIGATION_LEDGER. The gating canary (713/715 interleave, answered exactly once) is correctly the activation gate, not a merge blocker. Safe to land.

PR #2149, opened 2026-06-04
fix(telegram): make obligation ledger deterministic across restart + escalation

mekenthompson:
## Summary

The obligation ledger (`SWITCHROOM_OBLIGATION_LEDGER`, **default off**, on test-harness only) tracks whether an inbound was *answered* and re-presents an unanswered one until it closes — the marko-715 verbal-deferral drop fix. An **end-to-end determinism audit** (reasoning, not live observation, per operator steer) found the guarantee holds for every model behaviour on a single obligation, but leaked in two real places. This PR closes both, behind the existing flag, with a property-test proof.

## Why

Operator: *"you should have understanding end to end to be confident it is deterministic … reliable and as simple as possible … without compromise or over-eng."* The audit's honest verdict:

| Hole | Real? | This PR |
|---|---|---|
| Coalesce double-ask (4/5 audit lenses flagged) | **NO — refuted in code** | n/a (`openObligationFromInbound` is inside `handleInbound`, reached once per coalesced group via `onFlush`, not per raw message) |
| Escalation closes *before* it sends → silent drop | **YES** | fixed (close-after-send + thread-fallback + bounded) |
| In-memory ledger lost on restart | **YES** | fixed (durable snapshot) |
| Coalesced partial-answer (two distinct Qs merged, model answers one) | **YES — hard limit** | documented; unchanged (can't detect without becoming a harness over the model — pillar 3) |

## What changed

**1. Escalation terminal** (`gateway.ts` `obligationSweep`) — was `close()` then a fire-and-forget send whose failure dropped the terminal. Now: send via `retryWithThreadFallback` (dead/renumbered topic → `THREAD_NOT_FOUND` retries thread-less — the #2096 pattern), and append the close **only after the send resolves**. Transient failure → stays OPEN, retried next sweep. Permanent failure → bounded by `OBLIGATION_ESCALATE_MAX` → close best-effort (user unreachable; bounded give-up beats an infinite/poison loop). An in-flight guard prevents concurrent sends for the same id.

**2. Durable snapshot** (`obligation-store.ts`, new) — the in-memory `Map` emptied on restart and the spool's boot-replay bypasses `handleInbound` (where obligations OPEN), so a delivered-but-unanswered message lost its obligation across a restart. The store persists the open set via write-tmp + atomic rename (simplest crash-safe shape for a tiny bounded set — no append-log/compaction). `ObligationLedger` gains `onChange`/`hydrate` hooks + a durable `escalateAttempts` counter (so a poison escalation can't re-loop on every boot); the gateway persists each mutation and hydrates at boot.

Flag-off and STATIC mode touch no disk and exercise no new branch (inert).

## Test plan

- [x] `obligation-determinism.test.ts` — drives the **real** ledger + store through **3000 random `{model-behavior × timing × restart}` schedules**, asserting: every inbound reaches answered-or-escalated; no silent loss; no double-ask of an answered message; bounded termination (no infinite loop). This is the determinism proof that replaces "observe on live traffic."
- [x] `obligation-store.test.ts` — round-trip, atomic tmp+rename, torn/corrupt/wrong-version tolerance, malformed-row filtering, write-failure degrades (never throws).
- [x] `obligation-ledger.test.ts` — onChange fires per mutation; idempotent open / unknown close don't persist; `markEscalateAttempt`; hydrate restores counters without firing onChange.
- [x] `tsc --noEmit` clean; `check-plugin-references.mjs` clean; `check-bot-api-wrapping.sh` clean (escalation send wrapped in `retryWithThreadFallback(robustApiCall, …)`).
- [x] Full vitest: no new failures (the 46 pre-existing failures are all `dist/`-build / CLI-shell-out tests that need a build, unrelated to this change).

## Risk

Low. Entirely behind `SWITCHROOM_OBLIGATION_LEDGER` (off fleet-wide, on test-harness). Additive — does **not** touch the production spool/drain machinery; that subsumption is the deferred **PR3** once this proves out. JTBD: *always-available / you-hold-the-leash* — an inbound is never silently dropped.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -283,6 +283,7 @@ import {
   buildObligationRepresentInbound,
   obligationEscalationText,
 } from './obligation-ledger.js'
+import { loadObligations, persistObligations } from './obligation-store.js'
 import { createInboundSpool } from './inbound-spool.js'
 import { purgeStaleTurnsForChat } from './turn-state-purge.js'
 import { decideInboundDelivery } from './inbound-delivery-gate.js'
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

- Escalation terminal inverted correctly (`gateway.ts:4891-4943`): old code did `close()` then a fire-and-forget `robustApiCall` whose failure silently dropped the terminal. New path sends via `retryWithThreadFallback(robustApiCall, (tid) => bot.api.sendMessage(...), { threadId, chat_id, verb: "obligation.escalate" })` and appends `close()` ONLY in `.then()` — so a transient failure leaves the obligation OPEN for the next sweep, and a permanent one is bounded by `OBLIGATION_ESCALATE_MAX=3` → best-effort close. The call signature matches the proven #2096 pattern already used for the restart banner at gateway.ts:4136 (verified at head 2d02d68).
- In-flight guard is sound (`gateway.ts:4904`, `obligationEscalateInFlight` Set + `.finally()` delete): prevents the 5s sweep from firing a second concurrent send for the same origin while the first awaits. It is in-memory only, so a restart mid-send loses it — but `markEscalateAttempt` persists the bump BEFORE the send, so a reboot re-escalates and re-bumps, costing at most one attempt of the bound. Bounded, not a loss. (Non-blocking — worth a one-line comment noting the guard is intentionally non-durable.)
- Durable snapshot is crash-safe (`obligation-store.ts:337-354`): write-sibling-tmp + atomic `renameSync` means a crash leaves either the prior or the new complete snapshot, never a torn file. `loadObligations` fail-opens to `[]` on missing/unreadable/torn/wrong-version/malformed input (obligation-store.ts:311-329) and `persistObligations` never throws on write failure (degrades to in-memory). Correct for a tiny bounded set — snapshot over append-log needs no compaction.
- Ledger hooks correct (`obligation-ledger.ts`): `persist()` fires `onChange` after open/close/represent/escalate-attempt but NOT on idempotent open (line 181 early-return) or unknown close (line 189) or `hydrate` (line 205) — hydrate IS the restoration of on-disk state, so re-persisting would be redundant; counters (`representCount`, `escalateAttempts`) survive intact so `decideAtIdle` resumes at the right rung after restart.
- Wiring is inert when off (`gateway.ts:4040-4056`): `onChange` is `undefined` and boot-hydrate is skipped under `STATIC || !OBLIGATION_LEDGER_ENABLED`; the sweep `setInterval` is only registered under `!STATIC && OBLIGATION_LEDGER_ENABLED`. No disk touched, no new branch exercised, flag-off fleet-wide.
- Tests are thorough and exercise the REAL ledger+store: `obligation-determinism.test.ts` drives 3000 random {answer-attempt × escalate-fail × restart} schedules asserting termination, no silent loss, correct terminal, and no double-ask of a closed obligation; `obligation-store.test.ts` covers round-trip, atomic tmp→rename ordering, torn/corrupt/wrong-version/malformed tolerance, and write-failure degradation; `obligation-ledger.test.ts` covers onChange firing/non-firing and hydrate counter restoration.

Rationale: closes both real determinism holes (escalation silent-drop, restart loss) the audit flagged, with a property-test proof; entirely behind `SWITCHROOM_OBLIGATION_LEDGER` (off fleet-wide), additive, does not touch the production spool/drain machinery. The coalesced partial-answer residual is honestly documented as the one out-of-scope hard limit, not silently ignored. All cited symbols verified present at head 2d02d68.

PR #2152, opened 2026-06-04
fix(telegram): close the obligation-ledger hang-wedge (bound the escalation send)

mekenthompson:
## Summary
A **total** state-machine proof (every reachable state × every input event — not a sampled property test) found the obligation guarantee (`SWITCHROOM_OBLIGATION_LEDGER`, live on marko) was **not deterministic by construction**: it rested on one unproven liveness assumption a sampling test structurally cannot reach — *the escalation send settles*. This closes it.

## The hole (real, reachable, confirmed in code)
The escalation send is fire-and-forget and clears its in-flight guard (`obligationEscalateInFlight`) only in a `.finally`, which runs **only if the promise settles**. grammy's `bot.api` has no request timeout (`new Bot(TOKEN)`, gateway.ts:674) and `retryApiCall`'s `await fn()` bounds *rejections*, not a *hang* (retry-api-call.ts:91-93). So a stalled send (half-open TCP / unresponsive Telegram) never settles → the in-flight id leaks forever → every later sweep early-returns at the guard (gateway.ts:4918) → the obligation is **stuck OPEN: never re-presented, never escalated, never closed**. A silent loss of the answer guarantee, in the live process. The property test modeled send as synchronously succeed/fail, so it *could not* catch this.

## Fix — correct by construction, not by test
- **`withDeadline`** (new, pure, unit-tested) races the send against a 45s deadline so the chain **always** settles → `.finally` always clears the flag → a hang becomes a bounded reject, handled exactly like any other failed attempt, feeding the already-bounded escalate ladder (`escalateAttempts → OBLIGATION_ESCALATE_MAX`) to a terminal. Loser swallowed (no unhandled rejection); timer cleared + `unref`'d.
- **Decouple escalation from the buffer-depth gate**: escalation is a direct Telegram send, bridge-independent, so a represent stranded behind a dead bridge can no longer block the operator nudge. Only the `represent` path waits on an empty buffer. (The proof's second, lesser hole — a delay, not a drop, but it needlessly blocked escalation too.)
- Records the **closed-form determinism argument with the code**: the ledger is a finite FSM with a strictly-decreasing measure μ = (REPRESENT_MAX−representCount) + (ESCALATE_MAX−escalateAttempts); the sweep's termination now rests only on bounded facts (5s tick; gate opens via turn-end / `disconnect-flush.ts` / 300s watchdog; the send is bounded) plus the one irreducible assumption shared by *all* inbound delivery (bridge reconnects / process restarts, self-healing via durable spool + boot-replay).
- The property test is demoted in its own header to a **regression guard** — evidence, never the proof.

## What the proof refuted (don't "fix" non-bugs)
- *"`claudeBusyKeys` is never deleted → gate wedges forever"* — **wrong**: deletes at `disconnect-flush.ts:114/134/160`, `gateway.ts:2237/2359`, plus the 300s watchdog. (This is why the sweep fired live at all.)
- *buffer-drain stall* — backstopped by durable spool + boot-replay; a delivery delay, not a drop.

## Test plan
- [x] `with-deadline.test.ts` (6) — incl. the **hang case** (`never-settles` → bounded reject) and "the `.finally` always runs even when the inner promise hangs".
- [x] `obligation-{ledger,store,determinism}.test.ts` (32) green; determinism test re-headed as a regression guard.
- [x] `tsc --noEmit`, `check-plugin-references.mjs`, `check-bot-api-wrapping.sh` all clean (the escalation send stays wrapped in `retryWithThreadFallback(robustApiCall, …)`).

## Risk
Low; flag-gated (live on marko, off elsewhere). Strictly improves the escalation terminal — turns the one unbounded wait into a bounded one. Follow-up to #2149.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -289,6 +289,7 @@ import {
   obligationEscalationText,
 } from './obligation-ledger.js'
 import { loadObligations, persistObligations } from './obligation-store.js'
+import { withDeadline } from './with-deadline.js'
 import { createInboundSpool } from './inbound-spool.js'
 import { purgeStaleTurnsForChat } from './turn-state-purge.js'
 import { decideInboundDelivery } from './inbound-delivery-gate.js'
@@ -1411,6 +1412,17 @@ const OBLIGATION_SWEEP_MS = 5_000
 // attempts the gateway closes best-effort (loud log): the user is genuinely
 // unreachable, so a bounded give-up beats an infinite/poison loop.
 const OBLIGATION_ESCALATE_MAX = 3
+// Deadline for a single escalation send. grammy/fetch impose NO request timeout,
+// and the in-flight guard (obligationEscalateInFlight) is cleared only in the
+// send's `.finally` — which never runs if the send hangs. Without a bound, one
+// stalled send leaks the in-flight flag forever and the obligation is stuck OPEN
+// (never re-escalated, never closed) — the sole liveness hole a total proof
+// found. Racing the send against this deadline makes the wait bounded BY
+// CONSTRUCTION (see with-deadline.ts): the chain always settles, `.finally`
+// always clears the flag, and a hang becomes a bounded reject that feeds the
+// bounded escalate ladder to a terminal. 45s comfortably exceeds robustApiCall's
+// 3-attempt network backoff so a legitimate slow send isn't cut short.
+const OBLIGATION_ESCALATE_SEND_DEADLINE_MS = 45_000
 // Durable snapshot of the open obligation set on the persistent per-agent
 // volume (STATE_DIR = /state/agent/telegram in prod). Closes the restart hole:
 // the in-memory ledger alone empties on restart and the spool's boot-replay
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- Core fix is correct by construction (gateway.ts:4960-4980 + with-deadline.ts:37-50). The escalation chain `.then(close).catch(ladder).finally(delete inFlight)` previously hung off `retryWithThreadFallback(...)`, and grammy imposes no request timeout, so a stalled send never settled, `.finally` never ran, and `obligationEscalateInFlight` leaked forever — wedging the obligation OPEN. Wrapping only the inner promise in `withDeadline(..., 45_000, ...)` races it against a 45s reject so the chain always settles in <=45s; the unchanged `.finally` (gateway.ts:4980-4982) therefore always clears the flag. Verified against head cf2caa78: the then/catch/finally structure is preserved, only its source promise changed.
- withDeadline is sound (with-deadline.ts:37-50). The loser-swallow `p.catch(() => {})` at line 40 — applied before the race — prevents an unhandled rejection from a hung-then-late-rejecting send; the race still propagates p's real rejection independently. Timer is `unref`'d (line 45) so it never keeps the loop alive, and `clearTimeout`'d in the trailing `.finally` (line 147-149) so it does not dangle on a fast settle. Pure, no gateway coupling.
- Buffer-gate decoupling is right (gateway.ts:4902, 4908-4916). The early `pendingInboundBuffer.depth(agent) > 0` return is moved out of the sweep prologue into the represent branch only, so a represent stranded behind a dead bridge can no longer block the escalation send — escalation is a direct, bridge-independent Telegram send, so this is the correct scope. No regression to the double-present guard, which only ever mattered for the represent path.
- Tests cover the load-bearing cases — with-deadline.test.ts:230-249 exercises the never-settles hang (bounded reject) and asserts the chained `.finally` runs even when the inner promise hangs (the exact gateway invariant), plus the no-unhandled-rejection case (line 251-266). The determinism test is correctly demoted to a regression guard.
- Nit (non-blocking): the "45s comfortably exceeds robustApiCall's 3-attempt backoff" claim (gateway.ts:1417-1427) is asserted, not proven in this diff; if that backoff ceiling ever grows past ~45s a legitimate slow send could be cut to a spurious escalate-attempt. Bounded and self-correcting (feeds the bounded ladder), so harmless — worth a glance if retry backoff is ever retuned.

Rationale: closes the sole unbounded-wait liveness hole in the obligation ledger by making the escalation send settle by construction; flag-gated (live on marko only), strictly improves the terminal, integration verified against gateway head cf2caa78.

PR #2154, opened 2026-06-04
test(telegram): make the obligation escalation hang-path executable

mekenthompson:
## Summary
Makes the obligation-ledger escalation **hang-path** executable in a test. The hang-wedge fix (#2152) was verified by a total proof + a `withDeadline` unit test in isolation + review of the wiring — but the *integrated* path (the `withDeadline` + close/bounded-retry/clear-flag chain in `obligationSweep`) had **never actually run with a hung send**: mtcute can't make Telegram hang, and the synchronous property test can't model a non-settling promise. So the exact behavior the fix changes rested on reasoning, not execution. This closes that gap — **no behavior change**.

## What changed
- Extract the escalation step from `obligationSweep` into **`driveEscalation`** (`escalation-drive.ts`) — pure, dependency-injected (`ledger` / `inFlight` set / `send` thunk / `maxAttempts` / `deadlineMs` / `withDeadlineFn`). The sweep now calls it; the logic is identical (verified by the unchanged gates).
- **`escalation-drive.test.ts`** drives the REAL `driveEscalation` with the REAL `ObligationLedger` and REAL `withDeadline`:
  - **a send that NEVER settles** → the in-flight flag still clears (the wedge is gone), settling at the deadline (proven by elapsed time);
  - **repeated hung sends** → bounded terminal in exactly `maxAttempts` sends, never an infinite loop;
  - transient-failure-stays-open, success-closes, and the concurrency guard (a second concurrent call is a no-op).

## Why (the honest gap this closes)
mtcute is a no-regression check on the *normal* path (it ran green on v0.14.62), but it structurally cannot exercise a hung send. The certainty on the changed behavior was a proof; this turns it into an executed, deterministic regression guard — the behavior-matched test the proof deserves.

## Test plan
- [x] `escalation-drive.test.ts` (5) + `with-deadline` (6) + `obligation-{ledger,store,determinism}` (32) = 43 green.
- [x] `tsc --noEmit`, `check-plugin-references.mjs`, `check-bot-api-wrapping.sh` clean (escalation send still wrapped in `retryWithThreadFallback(robustApiCall, …)`).

## Risk
Minimal — pure extraction, behavior-identical, flag-gated. Follow-up to #2152.

### DIFF
--- a/telegram-plugin/gateway/escalation-drive.ts
+++ b/telegram-plugin/gateway/escalation-drive.ts
@@ -1,0 +1,79 @@
+/**
+ * escalation-drive.ts — the obligation-ledger escalation step, extracted from
+ * obligationSweep so the hang → bounded → terminal behaviour is EXECUTABLE in a
+ * test with a fake hanging send. That path is unreachable by the two harnesses
+ * we have: mtcute can't make Telegram's API hang, and the synchronous property
+ * test can't model a promise that never settles — yet it is exactly the path the
+ * total proof flagged as the determinism hole. This file makes the real drive
+ * logic the sweep uses testable in isolation (escalation-drive.test.ts), so the
+ * fix is verified by execution, not only by reasoning + review.
+ *
+ * Invariant it upholds: a single escalation attempt
+ *  - is guarded so a sweep tick cannot fire a second concurrent send for the
+ *    same obligation while one is awaiting;
+ *  - bounds the send with `withDeadline`, so the chain ALWAYS settles within
+ *    `deadlineMs` ⇒ the in-flight flag ALWAYS clears (a hung send becomes a
+ *    bounded reject, handled like any other failed attempt);
+ *  - closes the obligation ONLY after a successful send; a transient failure
+ *    leaves it OPEN (retried next sweep); a permanent failure
+ *    (attempt ≥ maxAttempts) closes best-effort — so repeated hung/failed sends
+ *    reach a terminal in a bounded number of sweeps, never an infinite loop.
+ */
+import { withDeadline } from './with-deadline.js'
+
+/** The slice of the ledger the escalation step needs. */
+export interface EscalationLedger {
+  markEscalateAttempt(originTurnId: string): number
+  close(originTurnId: string | null | undefined): boolean
+}
+
+export interface DriveEscalationArgs {
+  escId: string
+  /** Set of origin ids with an escalation send in flight (concurrency guard). */
+  inFlight: Set<string>
+  ledger: EscalationLedger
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- Pure extraction, behavior-identical. The escalation step removed from `obligationSweep` (gateway.ts:4950-5003 old) is reproduced verbatim in `driveEscalation` (escalation-drive.ts:57-85): same in-flight guard → `markEscalateAttempt` → `withDeadline(send(), deadlineMs, ...)` → `.then(close)` → `.catch(attempt>=max ? close : retry)` → `.finally(clear flag)`. No control-flow change.
- Wiring is faithful (gateway.ts:4959-4974). `inFlight: obligationEscalateInFlight`, `ledger: obligationLedger`, `maxAttempts: OBLIGATION_ESCALATE_MAX`, `deadlineMs: OBLIGATION_ESCALATE_SEND_DEADLINE_MS`, and `send` is the identical `retryWithThreadFallback(robustApiCall, sendMessage, {threadId, chat_id, verb:'obligation.escalate'})` thunk. `withDeadline` import moved into escalation-drive.ts (gateway.ts:294 swap), still wrapped in `retryWithThreadFallback` so `check-bot-api-wrapping.sh` stays green.
- Ledger API confirmed at head: `markEscalateAttempt` (obligation-ledger.ts:217, returns the bumped 1-based count), `close` (:108), `isOpen` (:132), `openIfAbsent` (:100), `list` (:145), `escalateAttempts` (:46). Every symbol the drive and test touch exists.
- Hang-path test is sound (escalation-drive.test.ts:235-258). Real `driveEscalation` + real `ObligationLedger` + real `withDeadline` against a `new Promise(()=>{})` send: asserts the in-flight flag clears and `elapsed < DEADLINE+500` — proves the wedge settles at the deadline, not "never". This is exactly the path mtcute can't reach and the synchronous property test can't model.
- Bounded-terminal test is correct (escalation-drive.test.ts:260-287). With `maxAttempts=3`, attempts 1–2 reject→stay OPEN, attempt 3 hits `attempt>=maxAttempts`→close best-effort, so `sends===MAX===3` and `drives<=MAX+1`. Matches the `.catch` gate in escalation-drive.ts:71. Concurrency guard test (:289-305) correctly asserts the second concurrent call returns `undefined` and fires no second send.

Rationale: behavior-preserving refactor that makes the #2152 hang→bounded→terminal path executable as a deterministic regression guard. Logic verified identical to head, ledger API confirmed present, all 5 tests assert the real invariants. Flag-gated, follow-up to #2152. Only nit: the extracted escalate log line drops the old "(exhausted N re-presents)" detail (escalation-drive.ts:64 vs the old gateway line) — log text only, no behavior impact.

PR #2155, opened 2026-06-04
fix(telegram): stop '[object Object]' in activity feed for object-valued MCP inputs

mekenthompson:
## Summary
Marko's Meta Campaigns thread showed `✓ [object Object]` status beats. Root cause: the PreToolUse tool-label hook (`telegram-plugin/hooks/tool-label-pretool.mjs`) coerced tool-input fields with `String(i.query ?? '')` before clipping. An operator MCP server (Brevo CRM) passed an **object** in `query`/`description`/`title` (a filter/payload), so `String({...})` → `"[object Object]"`, which is truthy → became the label.

## Why the existing guard missed it
The renderer's `clip()` in `tool-activity-summary.ts` already rejects non-strings (`typeof s !== "string" → null`). But this hook runs *earlier* and writes the precomputed label into the sidecar JSONL via the `tool_label` event — its `String()` coercion happened before the value ever reached the renderer's guard.

## Fix
Add `asText(v)` — string pass-through, number/boolean clean-stringify, **object/array → `''`** — and swap every `String(i.<field>)` callsite to it. Object-valued fields now fall through to the next sibling field or the humanized tool name (`Using get contacts`).

Before → after (verified):
- `{query:{listIds:[3]}}` → ~~`[object Object]`~~ → `Using get contacts`
- `{description:"Day 0 pricelist report"}` → `Day 0 pricelist report` (string still wins)
- `{query:42}` → `42` (numbers still coerce)

## Test plan
- [x] `tests/tool-label-pretool.test.ts` — new case: object query/description/title + array field never yield `[object Object]` (16 pass)
- [x] `tests/tool-activity-summary.test.ts` renderer suite green (34)
- [x] `check-bun-test-imports` clean

## Risk
Low — pure label-string change on an exit-0 fail-silent hook; the only behavior change is that object/array fields stop rendering as `[object Object]`. Ships in the agent image (baked hook), so takes effect on fleet roll.

### DIFF
--- a/telegram-plugin/hooks/tool-label-pretool.mjs
+++ b/telegram-plugin/hooks/tool-label-pretool.mjs
@@ -36,6 +36,26 @@ function readStdin() {
 }
 
 /**
+ * Coerce a tool-input field to display text WITHOUT the `[object Object]`
+ * trap. Only primitives carry a meaningful label: strings pass through,
+ * numbers/booleans stringify cleanly. Objects and arrays return '' so the
+ * caller falls through to its next fallback (a sibling field, or the
+ * humanized tool name) instead of surfacing literal "[object Object]".
+ *
+ * This guards the MCP-tool path in particular: an operator-configured
+ * server (e.g. Brevo CRM) may pass a filter/query OBJECT in `query` /
+ * `description` / `title`, and the old `String(i.query ?? '')` coercion
+ * rendered that as "[object Object]" on the live activity feed. The
+ * renderer's own `clip()` already rejects non-strings; this mirrors that
+ * contract at the hook so the bad value never reaches the sidecar JSONL.
+ */
+function asText(v) {
+  if (typeof v === 'string') return v
+  if (typeof v === 'number' || typeof v === 'boolean') return String(v)
+  return ''
+}
+
+/**
  * One-line, length-bounded escape of a value for inclusion in a label.
  * Newlines collapsed, very long strings truncated with an ellipsis.
  */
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- Root-cause fix is correct and complete. `asText(v)` (tool-label-pretool.mjs:23-27) returns strings as-is, clean-stringifies number/boolean, and maps object/array → `''` so the caller falls through to a sibling field or the humanized tool name instead of surfacing `[object Object]`. Confirmed all 11 `String(i.*)` coercion callsites in the head file are swapped (lines 85/88/106/107/111/115/131/144/180/189-191 at head → asText in the diff); grep for `String(i.` against the branch shows none left behind.
- Primitive behavior is preserved. `asText('')` → `''` (falsy, falls through, same as old `String('' ?? '')`); `asText(42)` → `'42'`; `asText(true)` → `'true'`. The only behavior change is object/array fields stop rendering as `[object Object]`/comma-joined — exactly the intended scope.
- Perplexity callsite change is equivalent (tool-label-pretool.mjs:188). Old `String(i.query ?? i.description ?? '')` → new `asText(i.query) || asText(i.description)`: for a string query both yield the query; for an object query the old code produced `[object Object]` (truthy, wrong), the new code yields `''` and falls to description, then to `'Searching the web'`. Strictly better, no regression on the primitive paths.
- Test covers the real failure (tests/tool-label-pretool.test.ts:114-143): object `query`/`description`/`title` and an array field all assert the label never contains `[object Object]`, sibling string still wins (`Day 0 pricelist report`), and object-only falls through to `Using get contacts`. Matches the Brevo CRM repro in the PR body.
- Layering is right. The renderer's `clip()` already rejects non-strings, but this hook precomputes the `tool_label` event into the sidecar JSONL *before* the renderer's guard runs, so the coercion had to be fixed here too. Fail-silent exit-0 hook, baked into the agent image — takes effect on fleet roll.

Rationale: minimal, well-scoped label-string fix; every coercion callsite converted, primitive behavior unchanged, object/array trap closed, with a test that reproduces the exact `[object Object]` bug. No blockers, no nits. Safe to merge.

PR #2156, opened 2026-06-04
fix(telegram): obligation escalate-grace window — stop false nudges on slow turns

mekenthompson:
## Summary
Fuzz testing on v0.14.62 (obligation ledger live on marko) surfaced a real false-positive: a slow / background-worker / multi-segment turn **ends** (the in-flight gate clears) before its trailing answer's reply lands, so the 5s sweep re-presents then escalates in that gap — firing a user-facing **"⚠️ I may have missed this"** on a message the agent is *actively answering*.

## Evidence
- Live log on test-harness: obligation `#4334` escalated at `13:58:24`, then `reply: invoked "Spinning up a worker…"` at `13:58:25`, answer at `13:58:36`.
- Real frequency: **~14%** of marko's no-reply turn-ends had the answer in flight; on test-harness (fast synthetic traffic) ~37%, breaking **10 fuzz assertions**. marko has fired **0** false cards live (humans pause), so it's latent there — but would fire on a research/worker turn.

## Fix (in the pure, durable ledger — unit-testable + survives restart)
- `Obligation.lastTurnEndedAt`; `noteTurnEnded` stamps it at `turn_end` when the turn ended **without** a final answer (obligation stays open).
- `decideAtIdle({now, graceMs})` skips an obligation whose handling turn ended `< graceMs` ago, picking the **oldest eligible** — a genuinely-stale message still acts while a freshly-ended one waits for its trailing close.
- Default 45s; kill switch `SWITCHROOM_OBLIGATION_ESCALATE_GRACE_MS=0` → byte-identical pre-grace behaviour.

## Correctness
- **Determinism preserved:** `decideAtIdle()` with no args is the exact pre-grace path → the 3000-schedule determinism proof passes unchanged.
- **Ladder still terminates:** grace only ever *delays*; each turn re-stamps `lastTurnEndedAt` at most once and `representCount` is capped → bounded. Pinned by a "grace still terminates the ladder" test + eligibility/ordering tests (7 new).
- **715 unaffected:** a thread that never gets an answer has no trailing reply, grace expires, it re-presents/escalates as before.

## Test plan
- [x] `obligation-ledger.test.ts` 29 (7 new grace) + store/determinism/escalation-drive/with-deadline = 50 green.
- [x] `tsc`, `check-plugin-references`, `check-bot-api-wrapping` clean; full `telegram-plugin/` suite 3484/0.

## Risk
Low; flag-gated (live on marko, off elsewhere), instant rollback via the env knob. Prerequisite for the mid-turn auto-classifier work.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1423,6 +1423,22 @@ const OBLIGATION_ESCALATE_MAX = 3
 // bounded escalate ladder to a terminal. 45s comfortably exceeds robustApiCall's
 // 3-attempt network backoff so a legitimate slow send isn't cut short.
 const OBLIGATION_ESCALATE_SEND_DEADLINE_MS = 45_000
+// Escalate-grace window. A slow / background-worker / multi-segment turn ends
+// (the in-flight gate clears) BEFORE its trailing answer's reply lands, and the
+// 5s sweep would re-present/escalate in that gap — a false "⚠️ I may have missed
+// this" on a message that's actively being answered (fuzz-confirmed on v0.14.62:
+// ~14% of marko's no-reply turn-ends had the answer in flight). An obligation
+// whose handling turn ended < this ago is skipped by decideAtIdle, giving the
+// trailing answer's close a beat to fire. Bounded: each re-present is itself a
+// turn that re-stamps once, representCount is capped → the ladder still
+// terminates. 45s > the observed "answer lands within ~60s, usually <40s" gap.
+// Kill switch: SWITCHROOM_OBLIGATION_ESCALATE_GRACE_MS=0 → pre-grace behaviour.
+const OBLIGATION_ESCALATE_GRACE_MS = (() => {
+  const raw = process.env.SWITCHROOM_OBLIGATION_ESCALATE_GRACE_MS
+  if (raw == null || raw === '') return 45_000
+  const n = Number(raw)
+  return Number.isFinite(n) && n >= 0 ? n : 45_000
+})()
 // Durable snapshot of the open obligation set on the persistent per-agent
 // volume (STATE_DIR = /state/agent/telegram in prod). Closes the restart hole:
 // the in-memory ledger alone empties on restart and the spool's boot-replay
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- Grace logic is correct (obligation-ledger.ts:118-125). `oldestEligible(now, graceMs)` skips any obligation whose `lastTurnEndedAt` is within grace and picks the oldest of the rest by `openedAt` — matching the existing `oldest()` ordering (obligation-ledger.ts:150-156, orders by `openedAt`). A still-queued obligation with no `lastTurnEndedAt` is always eligible (correct — nothing in flight to wait for).
- Determinism is preserved (obligation-ledger.ts:107-113). `decideAtIdle()` with no opts (or `graceMs<=0`) takes the exact `this.oldest()` path — byte-identical pre-grace behaviour, so the 3000-schedule proof holds. The gateway passes `undefined` when the env knob is 0 (gateway.ts:55-59), so the kill switch is a true no-op, not a different code path.
- Durability holds (claim verified). `noteTurnEnded` calls `this.persist()` (obligation-ledger.ts:129-134) and obligation-store.ts persists the whole object (`JSON.stringify({v:1, obligations:[...snapshot]})`, store:96-99); `loadObligations` filters rows with a duck-type check that does NOT strip extra fields (store:84), so `lastTurnEndedAt` round-trips across restart — grace survives a boot as documented.
- Ladder still terminates (gateway.ts:2442-2452 + test:210-223). Grace only ever delays. Each turn-end re-stamps `lastTurnEndedAt` at most once; `representCount` is still capped and incremented by `markRepresented`, so represent→represent→escalate reaches a terminal — pinned by the "grace still terminates the ladder" test. Worst-case added latency is bounded (~maxRepresents × graceMs) before escalate.
- No double-fire race. The represent turn injected after `markRepresented` is in flight, so `turnInFlightForGate()` early-returns the sweep (gateway.ts:4950) until that turn ends and re-stamps — the freshly-ended obligation then re-enters grace. The stamp only happens in the `else` branch when `finalAnswerDelivered===false` (gateway.ts:2444-2452); a delivered answer still closes, and `noteTurnEnded` is a no-op for synthetic/closed turns.
- Env parsing is safe (gateway.ts:19-24). Empty/unset → 45_000; non-finite or negative → 45_000; `0` → off. No NaN leak into the comparison.

Rationale: flag-gated, behavior-preserving-when-off fix for a fuzz-confirmed false-escalation race; grace only delays, the ladder stays bounded, determinism and durability are both intact, and 7 new tests cover eligibility, oldest-eligible ordering, all-in-grace, persistence, and ladder termination. No blockers, no nits. Safe to merge.

PR #2157, opened 2026-06-04
fix(telegram): cancel the interrupted turn's obligation on ! interrupt

mekenthompson:
## Summary
An `!` interrupt SIGINT-kills the in-flight turn, but the killed turn doesn't reliably emit `turn_end`, so `endCurrentTurnAtomic` never closes its obligation. The obligation survives → the idle sweep later re-presents/escalates **"⚠️ I may have missed an earlier message"** for a question the user **explicitly cancelled by interrupting**. (Surfaced by the mode × obligation-ledger matrix review.)

## Fix
`cancelInterruptedObligation()` closes the in-flight turn's obligation the moment the SIGINT fires, on **both** the synchronous (`gateway.ts` interrupt-marker path) and **deferred** (`fireDeferredInterrupt`) paths. An interrupt is a deliberate redirect → closing is the correct terminal; the user can re-ask. **Only the interrupted turn's own obligation** is closed (`currentTurn.turnId`); queued siblings are untouched. No-op when the flag is off / no turn in flight / turn not tracked.

## Test plan
- [x] New ledger test pins the interrupt-cancel semantics (close in-flight → sweep won't return it; sibling survives; re-close is a safe no-op).
- [x] `tsc`, `check-plugin-references`, `check-bot-api-wrapping` clean; full `telegram-plugin/` suite **3478/0**.

## Risk
Low; flag-gated (live on marko, off elsewhere). Independent of the grace-window PR (#2156). Pure terminal-on-explicit-cancel.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1999,6 +1999,29 @@ let pendingDeferredInterrupt: PendingDeferredInterrupt | null = null
  * Idempotent: nulls the slot and clears the timer before doing any work so a
  * boundary event and the timeout can't double-fire.
  */
+/**
+ * An `!` interrupt SIGINT-kills the in-flight turn. That turn was handling a
+ * user message with an open obligation, and the killed turn does NOT reliably
+ * emit turn_end (so endCurrentTurnAtomic never closes it) — so without this the
+ * obligation survives and the idle sweep later re-presents/escalates "you have
+ * an earlier message you never answered" for a question the user EXPLICITLY
+ * cancelled. An interrupt is a deliberate redirect, so closing that obligation
+ * is the correct terminal (the user chose to interrupt; they can re-ask). Only
+ * the interrupted turn's OWN obligation is closed — queued siblings (other open
+ * obligations) are untouched. No-op when the flag is off, no turn is in flight,
+ * or the turn isn't a tracked obligation (synthetic / already closed).
+ */
+function cancelInterruptedObligation(): void {
+  if (!OBLIGATION_LEDGER_ENABLED) return
+  const turn = currentTurn
+  if (turn == null) return
+  if (obligationLedger.close(turn.turnId)) {
+    process.stderr.write(
+      `telegram gateway: obligation cancelled by interrupt origin=${turn.turnId}\n`,
+    )
+  }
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- Fix is correct and well-scoped (gateway.ts:2021-2030). `cancelInterruptedObligation()` reads the module global `currentTurn` (declared `let currentTurn: CurrentTurn | null` at gateway.ts:1795) and closes only `turn.turnId` — the interrupted turn's own obligation. Queued siblings (other open obligations) are untouched, exactly as the PR claims; verified by the new sibling-survives test.
- Both interrupt paths are covered. Synchronous interrupt-marker path calls it after the SIGINT (gateway.ts:10814) and the deferred path calls it after the SIGINT in `fireDeferredInterrupt` (gateway.ts:2053). On both, the call lands AFTER the kill but BEFORE the replacement body is delivered as a fresh turn, so `currentTurn` still points at the interrupted turn when `close()` runs — the close targets the right obligation.
- Premise holds and the no-op cases are safe. The killed turn not reliably emitting `turn_end` means `endCurrentTurnAtomic` (which nulls `currentTurn` at gateway.ts:2432) hasn't run, so `currentTurn` is still set when this fires. If turn_end DID fire, the obligation is already closed and `close()` returns false → harmless. Guarded for flag-off (`OBLIGATION_LEDGER_ENABLED`), `currentTurn==null`, and untracked/synthetic turns (`close()` no-ops on an unknown id).
- `close()` returns boolean and gates the log line (gateway.ts:2025), so the "cancelled by interrupt" stderr only fires on a real close — no false log on a no-op.
- Test is in scope and asserts the real semantics (obligation-ledger.test.ts:228-241). It lives in the `describe` at line 169, which has its own local `input()` helper (line 170), so it compiles; it pins close-in-flight → removed from `decideAtIdle`, sibling still actionable, and re-close/unknown → safe `false`. Matches `cancelInterruptedObligation`'s behaviour.

Rationale: flag-gated, behaviour-preserving-when-off terminal-on-explicit-cancel; only the interrupted turn's obligation closes, both SIGINT paths are wired before replacement delivery, every no-op path is safe, and the test pins the semantics. No blockers, no nits. Safe to merge.

PR #2158, opened 2026-06-04
feat(telegram): mid-turn auto-classify (steer-vs-queue) in shadow mode

mekenthompson:
## Summary
First step toward letting users steer/queue by **intent** instead of explicit `/steer` `/q` prefixes: a deterministic, **model-free** classifier using **topic-vs-active-turn + reply-recency** as proxies. Strong signal is structural — a message in a *different* forum topic than the in-flight turn is a separate conversation (queue); same-topic + recent is a steer candidate.

**Ships SHADOW-ONLY** behind `SWITCHROOM_AUTOCLASSIFY_MIDTURN_SHADOW` (default OFF): logs what it *would* decide; **behaviour unchanged** (no-prefix still queues). Gathers the real-world distribution to tune the auto-steer window + measure how often same-topic continuations happen — *before* any behaviour flips on. Honours "you hold the leash" (auto-steer is an irreversible injection → must earn its way in with data) and the April-2026 lesson (unconditional implicit-steer was too blunt).

## What's in it
- **`auto-classify-mid-turn.ts`** — pure decision fn: prefix wins → cross-topic→queue (strong) → same-topic recent→steer / stale→queue → DM timing-only & OFF by default. Canonical thread compare (null/0/undefined collapse). **11 unit tests.**
- **Recency clock** `lastAgentOutputAt` per `(chat,thread)` — the agent's last *output* (not turn age, so a long narrating worker turn doesn't read "stale"), stamped beside `noteOutbound`, LRU-bounded, **only when the flag is on**.
- **Shadow hook** at the classify site logs `would=/reason=/same_topic=/ms_since_out=/actual=`; `isSteering` (behaviour) untouched.

## Test plan
- [x] `auto-classify-mid-turn.test.ts` 11 (prefix override, cross/same-topic, recency, DM-off, canonical-collapse).
- [x] `tsc`, `check-plugin-references`, `check-bot-api-wrapping` clean; full `telegram-plugin/` suite **3497/0**.
- [x] Flag off → byte-identical + zero overhead (map only maintained when on).

## Risk
Minimal — no behaviour change; shadow logging only, default off. Next PR flips auto-steer behind its own flag after the shadow data + a product call (DM stays off; supergroup window tuned from the data).

### DIFF
--- a/telegram-plugin/gateway/auto-classify-mid-turn.ts
+++ b/telegram-plugin/gateway/auto-classify-mid-turn.ts
@@ -1,0 +1,119 @@
+/**
+ * auto-classify-mid-turn.ts — deterministic, model-free classification of a
+ * mid-turn inbound into STEER (amend the in-flight turn) vs QUEUE (new task),
+ * using TOPIC-vs-active-turn + reply-RECENCY as proxies for intent. No model
+ * inference: the gateway must decide at inbound time while the single CLI is
+ * busy.
+ *
+ * Today a no-prefix mid-turn message always QUEUES (the default flipped
+ * 2026-04-17 away from the blunt "everything steers" — see
+ * reference/steer-or-queue-mid-flight.md). This module is the basis for a
+ * smarter default. It ships first in SHADOW mode (the gateway logs what it WOULD
+ * decide but still queues), to gather real-world data — how often mid-turn
+ * messages are same-topic continuations vs cross-topic new tasks, and the
+ * recency distribution — before any behaviour flips on.
+ *
+ * Signal strength (be honest):
+ *  - TOPIC (supergroup): STRONG + structural. A message in a DIFFERENT forum
+ *    topic than the active turn is, by the supergroup-mode invariant, a separate
+ *    conversation → queue. This needs no timing guess.
+ *  - RECENCY: weaker. Within the same topic it cannot tell "also do X" (steer)
+ *    from "new question, same topic" (queue) — only a tight window + the
+ *    visible/correctable UX (the JTBD doc) makes auto-steer acceptable, and
+ *    that is gated separately. The recency clock is the agent's LAST OUTPUT
+ *    (msSinceLastAgentOutput), NOT turn age: a long actively-narrating worker
+ *    turn must not read "stale".
+ *  - DM: no topic at all → timing-only (the pre-2026-04-17 regime that
+ *    over-steered). DM auto-steer is OFF by default (window 0).
+ *
+ * Pure (no gateway imports) ⇒ unit-testable.
+ */
+
+export type MidTurnClass = 'steer' | 'queue'
+
+export type MidTurnReason =
+  | 'steer_prefix'
+  | 'queue_prefix'
+  | 'not_mid_turn'
+  | 'cross_topic'
+  | 'same_topic_recent'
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- Pure classifier is correct and matches its tests (auto-classify-mid-turn.ts:95-125). Precedence is sound: explicit prefix → not-mid-turn → DM(timing, off by default) → supergroup(topic-disabled / cross-topic→queue / same-topic recent→steer / stale→queue). `recent` requires `msSinceLastAgentOutput != null && >= 0` then `<= window`, so a null/cold clock correctly reads not-recent (auto-classify-mid-turn.ts:102-103). Cross-topic always queues before any timing check (auto-classify-mid-turn.ts:120) — the strong structural signal. Canonical `sameThread` collapses null/undefined/0 (auto-classify-mid-turn.ts:84-87), avoiding the raw-=== key-mismatch class. 11 tests cover every branch.
- Shadow wiring is behaviour-neutral (gateway.ts:11486-11510). `isSteering` is computed above and untouched; the shadow block only runs when `AUTOCLASSIFY_MIDTURN_SHADOW && priorTurnInFlight`, computes, and writes one stderr line. No path mutates steer/queue behaviour — confirmed default-off (`=== '1'`, gateway.ts:11491 / flag at 1452).
- Recency-clock keys align — verified, since this is the whole point of the shadow data. `executeReply` stamps `noteAgentOutputAt(statusKey(chat_id, threadId), …)` (gateway.ts:6566) and the shadow reads `lastAgentOutputAt.get(key)` where `key = statusKey(chat_id, messageThreadId)`; `statusKey` is just `chatKey` (gateway.ts:2146-2148), so write and read keys match including thread canonicalization. The gathered `ms_since_out` will be real, not always-null.
- Zero overhead when off, bounded when on. The `lastAgentOutputAt` map is only written under the flag (gateway.ts:6566), and `noteAgentOutputAt` is LRU-bounded at 512 keys with correct delete-then-set MRU reinsertion (gateway.ts:1459-1466). All referenced symbols exist: `isQueuedPrefix` (gateway.ts:11239), `isDmChatId` (2117), `currentTurn.sessionThreadId` (CurrentTurn iface, 1658).
- Nit (non-blocking): the recency clock is stamped only in `executeReply` (gateway.ts:6566), so "last agent output" is really "last reply-tool send" — reactions, edits, and streamed replies don't stamp it. That's a reasonable proxy for substantive output, but it's narrower than the "last visible output" wording in the doc comment (auto-classify-mid-turn.ts:64) and could skew the recency distribution slightly toward "stale". Worth noting since tuning the auto-steer window depends on this distribution.
- Nit (non-blocking): the shadow site hard-codes `topicSteerWindowMs: 8_000` (gateway.ts:11503). On the first mid-turn message after enabling the flag, before any reply has stamped the clock, `msSinceOut` is null → logs `same_topic_stale`; expected cold-start, self-corrects after the first output, but worth being aware of when reading the early shadow logs.

Rationale: shadow-only, default-off, behaviour byte-identical when the flag is off; the pure classifier is well-factored and fully tested, the recency-clock keys genuinely match so the data it gathers is valid, and the map is flag-gated and LRU-bounded. Honours "you hold the leash" by gathering data before any irreversible auto-steer flips on. No blockers. Safe to merge.

PR #2162, opened 2026-06-05
feat(telegram): instrument the live status-surface lane for triage

mekenthompson:
## Summary
PR **1 of 2** of the marko "status went dark" fix. Pure observability — **no behavior change**. Makes the live status-surface lane (progress card / activity feed / worker feed / typing) explainable from the gateway log, which is the gap that turned two one-line bugs into a multi-agent investigation.

The diagnosis (read-only, adversarially verified) found two latent, version-independent defects darkening the lane, plus a third silent-drop class — none of which left a greppable breadcrumb:
- **A (persistent):** the 300s silence-poke fallback nulls `currentTurn` (which drives the card/feed/typing) on any tool-work stretch >5min without a model reply.
- **B (acute, resume-only):** the resume-interrupted synthetic fabricates a `message_id` from `Date.now()` (~1.78e13) → used as the activity-feed reply anchor → **every** feed send 400s `message_id must be a valid Number` → feed dark the whole first post-restart turn.
- **C:** operator DM notifications carrying a supergroup topic id → `message thread not found`.

The **fixes** land in PR 2; this PR makes those fixes verifiable on the marko canary.

## What's added (additive logging only)
- **New pure module** `gateway/status-surface-log.ts` — `formatTurnLifecycle()` + `detectStatusSurfaceDegraded()`, unit-tested without a live gateway.
- **`currentTurn` lifecycle line** at every set (`enqueue`) and clear (`endCurrentTurnAtomic`): turnId, topic key, tool count, feed state, age, reason. (The other two clear paths — silence-poke fallback and disconnect-flush — already log their own reason.)
- **`CurrentTurn.activityEverOpened`** (sticky) + **`activityDrainFailures`** (counter); the `activity-summary drain failed` log now carries the failing **reply anchor + topic** — would have made bug B self-evident.
- **`status-surface DEGRADED`** line at turn-end when tools ran but the feed never opened because its sends failed (bug B's signature).
- **worker-activity-feed**: log `paint`/`edit`/`finish` on **success** (it previously logged only failures, so a healthy worker feed was invisible — exactly why the worker/sub-agent status looked dark in triage).

## Why (JTBD)
Serves outcome **2 (you hold the leash)** — awareness/control of what an agent is doing. When the live status lane fails, the operator must be able to tell *why* from the log, not reverse-engineer it.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `vitest` `status-surface-log` (8) + `worker-activity-feed` log-sink cases — 43 green
- [x] gateway-area vitest sweep — 197 green
- [x] lint gates clean: plugin-refs, bun-test-imports, bot-api-wrapping, no-pii-secrets, vault-hermeticity, no-broadcast

## Risk
Low — additive logging + two new per-turn fields (initialized at enqueue, read only by the new log sites). No transport/routing change.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -65,6 +65,7 @@ import {
 import { StatusReactionController } from '../status-reactions.js'
 import { DeferredDoneReactions } from '../reaction-defer.js'
 import { createWorkerActivityFeed, isWorkerActivityFeedEnabled } from '../worker-activity-feed.js'
+import { formatTurnLifecycle, detectStatusSurfaceDegraded } from './status-surface-log.js'
 import { isTelegramReplyTool, isTelegramSurfaceTool } from '../tool-names.js'
 import { appendActivityLabel, renderActivityFeedWithNested } from '../tool-activity-summary.js'
 import { toolLabel } from '../tool-labels.js'
@@ -1798,6 +1799,14 @@ type CurrentTurn = {
   activityInFlight: Promise<void> | null
   activityPendingRender: string | null
   activityLastSentRender: string | null
+  // Status-surface observability. `activityEverOpened` is sticky-true once the
+  // feed posts its first message — unlike `activityMessageId`, it is NOT nulled
+  // by `clearActivitySummary`, so the turn-end DEGRADED check can tell "feed
+  // never opened" (the resume-400 signature) from "feed finalized + cleared".
+  // `activityDrainFailures` counts real activity-feed send/edit failures this
+  // turn (429s + "message is not modified" excluded). Both reset per turn.
+  activityEverOpened: boolean
+  activityDrainFailures: number
   // Wall-clock anchor for the newest in-progress feed step — set each time a
   // tool_label re-renders the feed. The heartbeat (`feedHeartbeatTick`) reads
   // it to show a climbing " · Ns" elapsed on the live line so a long single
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

Pure additive observability for the live status-surface lane, no behavior change. Verified against the real diff and gateway.ts at head (a4d9c4d).

- `status-surface-log.ts:1-102` — new pure module (`formatTurnLifecycle` + `detectStatusSurfaceDegraded`), no import back into gateway.ts; `StatusSurfaceTurnView` is satisfied structurally. Confirmed every field it reads (turnId, sessionChatId, sessionThreadId, startedAt, toolCallCount, activityMessageId, activityEverOpened, activityDrainFailures, replyCalled, finalAnswerDelivered) exists on the `CurrentTurn` type (gateway.ts:1690-1815).
- `gateway.ts:9079-9080` — the two new per-turn fields (`activityEverOpened`, `activityDrainFailures`) are initialized at the ONLY `CurrentTurn` construction site (the sole `activityLastSentRender:` init confirms there is just one), so the new log sites never read `undefined`.
- `status-surface-log.ts:88-101` — DEGRADED gate is correctly conservative: returns null on no tool work, on feed-opened (sticky `activityEverOpened` survives `clearActivitySummary` nulling `activityMessageId`), and on zero drain failures. Matches the resume-400 signature without false-positiving the healthy finalize path. The 8 unit tests cover exactly these branches.
- `gateway.ts:8861-8862` — `sourceMessageId` is the real reply anchor (guarded `/^\d+$/` synthesis at :9057), so logging it at :8897 surfaces the fabricated-13-digit-id case that the PR targets.

Nit (non-blocking), `gateway.ts:8887` + `status-surface-log.ts` doc comment: both say 429s are excluded from `activityDrainFailures`, but the increment guard only excludes `message is not modified`. A send that exhausts `robustApiCall`'s 429 retries and throws WILL increment the counter. Harmless for the DEGRADED signal (a 429-blanked feed is genuinely degraded), but the comment overstates the filter — tighten the wording or add the 429 check to match it.

Rationale: additive logging + two enqueue-initialized fields, read only by new log sites; no transport/routing change, fully unit-tested, low risk.

PR #2163, opened 2026-06-05
fix(telegram): stop two always-failing status sends (resume-feed 400 + compact DM thread-leak)

mekenthompson:
## Summary
PR **2 of the marko "status went dark" fix** (companion to #2162 observability). Two **strictly-corrective** fixes — each targets a send that **always 400s today** (Telegram rejects it regardless of our code), so neither changes a working path and neither needs a kill-switch flag.

### Fix B — resume-turn dark feed (acute, every first turn after a restart)
Synthetic boot-resume inbounds fabricate a `message_id` from `Date.now()` (~1.78e13) so the deliver-until-acked queue can ack the synthetic by its own enqueue id. But the enqueue handler also turned that into `turn.sourceMessageId`, which `drainActivitySummary` sends as the activity-feed `reply_parameters.message_id` — **829× over Telegram's 2³¹ max**, so *every* feed send on the first post-restart turn 400'd `message_id must be a valid Number` and the live feed was dark the whole turn (proven in the marko log: 10 failed sends / 0 edits vs 3/11 on a normal turn).

New `gateway/source-message-id.ts` `parseSourceMessageId()` accepts the id as a reply anchor **only when it's a plausible Telegram message id**; an out-of-range value → `null` → the feed posts **unanchored** (still correct; the anchor is a nicety). The synthetic's ack-tracking is untouched — it keys on the enqueue event's own id, not this anchor. `resume-inbound-builder.ts` is unchanged.

### Fix C — proactiveCompact DM thread-leak (silent operator-notice drops)
The compact-watchdog notice attached the agent's **supergroup topic id** to **operator-DM** sends → `400 message thread not found` → the notice silently vanished (29× in the marko log; same class as the #2096 approval-card wedge). `proactiveCompact.start` was the *one* operator-send still missing the `topicForRecipient` guard the sibling sites (operator-event, drive, ms365, config) already use — now applied.

## Why (JTBD)
Outcome **4 (always available / predictable)** + **2 (you hold the leash)**: the live status surface and operator lifecycle notices must actually reach the operator.

## Test plan
- [x] `tsc --noEmit` clean
- [x] new `source-message-id.test.ts` (4) pins the 13-digit-timestamp rejection + in-range / 2³¹-boundary cases
- [x] `topicForRecipient` DM-strip already pinned (`topic-router.test.ts:313` "drops the topic for an operator DM recipient — the wedge fix")
- [x] gateway + topic-router vitest sweep — 206 green; `resume-inbound-builder` bun — 31 green
- [x] lint gates clean (plugin-refs, bun-test-imports, bot-api-wrapping, no-pii, vault-hermeticity, no-broadcast)

## Risk
Low — both fixes correct always-failing sends; no flag, no working-path change.

## Follow-up
**Fix A** (the persistent 300s silence-poke teardown that nulls `currentTurn` mid-work) lands as its **own PR** — it's the delicate, incident-prone one and wants a marko canary against #2162's now-merged telemetry.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -66,6 +66,7 @@ import { StatusReactionController } from '../status-reactions.js'
 import { DeferredDoneReactions } from '../reaction-defer.js'
 import { createWorkerActivityFeed, isWorkerActivityFeedEnabled } from '../worker-activity-feed.js'
 import { formatTurnLifecycle, detectStatusSurfaceDegraded } from './status-surface-log.js'
+import { parseSourceMessageId } from './source-message-id.js'
 import { isTelegramReplyTool, isTelegramSurfaceTool } from '../tool-names.js'
 import { appendActivityLabel, renderActivityFeedWithNested } from '../tool-activity-summary.js'
 import { toolLabel } from '../tool-labels.js'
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

Two strictly-corrective fixes, each targeting a send that always 400s today. Verified against the real diff and gateway.ts at head.

- `source-message-id.ts:34-41` (Fix B) — `parseSourceMessageId` correctly gates the activity-feed reply anchor: rejects null/undefined/empty/non-digit/non-positive/non-integer and anything >= 2^31, so the fabricated 13-digit `Date.now()` boot-resume id (1.78e13) yields null and the feed posts unanchored instead of 400ing every send. Tests at source-message-id.test.ts:1-32 pin the exact failing timestamp, the 2^31 boundary, and in-range acceptance. Replaces the weaker inline `/^\d+$/` guard at the enqueue site (gateway.ts:9067) which had no range check.
- `gateway.ts:2690-2696` (Fix C) — `topicForRecipient({recipientChatId, resolvedTopic, supergroupChatId})` now wraps the compact-watchdog topic, matching the exact signature of the six sibling operator-send sites (gateway.ts:2909, 4243, 5703, 5781, 5865, 12333). An operator-DM recipient gets a thread-less send (no more `message thread not found`), the supergroup owner keeps the lane. Both helpers (`topicForRecipient`, `resolveAgentSupergroupChatId`) resolve at head.
- Synthetic ack-tracking is untouched — it keys on the enqueue event's own id, not `sourceMessageId`; `resume-inbound-builder.ts` is unchanged. No flag needed since neither fix alters a working path.

Rationale: both correct sends that fail 100% today; additive guard + reuse of an established helper, fully unit-tested, no working-path or transport change. Low risk.

PR #2164, opened 2026-06-05
fix(telegram): keep the live feed alive during long in-flight tool work (silence-poke defer)

mekenthompson:
## Summary
PR **3 of the marko "status went dark" fix** — the **persistent, dominant** cause (companions: #2162 observability, #2163 resume/compact fixes).

The 300s silence-poke "framework fallback" nulls `currentTurn` (which drives the progress card / activity feed / typing) when there's been **no model reply** for 5 min. Activity-feed edits don't reset that clock, so a long quiet tool stretch — a foreground sub-agent, a big research pass, *exactly marko's background work* — tripped the fallback and **darkened the live feed mid-work while the agent kept going**. Proven 6× over 3 days in the marko log (e.g. card froze 22:57:51, dark from 23:02:10, answer landed 23:05:55).

The module header's *"the model could be ripping through 20 tool calls and still be silent to the user"* rationale **predates the live activity feed** (#2162): with the feed rendering tool progress, an actively-tool-churning turn is **not** silent to the user, so nulling `currentTurn` (killing that very feed) is counterproductive.

## The fix (opt-in, default OFF — canary first)
When the fallback threshold is crossed **with a parent tool genuinely in flight**, **defer** the unwedge (keep the turn + its live feed alive) instead of firing — **bounded by a hard ceiling** so a hung-mid-tool turn still unwedges and can't pin the conversation.

Safety preserved:
- A turn with **no in-flight tool** fires at the base threshold exactly as before → the #1556 dangling-turn protection is intact.
- A **crashed** agent is independently recovered by the bridge-disconnect sweep (`onDanglingTurnsSwept`) → deferring here doesn't reintroduce the crash-wedge.
- The defer `continue`s WITHOUT setting `fallbackFired`, so once the tool ends and the turn stays silent past threshold (or the ceiling is crossed) it fires promptly.

## Knobs (gateway env → silence-poke deps)
| env | default | effect |
|---|---|---|
| `SWITCHROOM_SILENCE_DEFER_INFLIGHT_TOOLS` | **off** | enable the defer |
| `SWITCHROOM_SILENCE_FALLBACK_MS` | 300000 | base threshold |
| `SWITCHROOM_SILENCE_FALLBACK_HARD_MS` | 900000 | defer hard ceiling |

Default-OFF so it ships dormant; enable on **marko** and verify against #2162's `turn-lifecycle` / `status-surface DEGRADED` telemetry before fleet.

## Why (JTBD)
Outcome **2 (you hold the leash)** — the live status of long background work must stay visible while it runs.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `silence-poke.test.ts` +6 Fix A cases: legacy-off unchanged at 300s; defer holds at 300s; fires once tool ends + silent; fires at hard ceiling (hung-mid-tool); no-tool turn unaffected; no-ceiling defers indefinitely — **49 green**
- [x] gateway + silence-poke vitest sweep — **216 green**
- [x] lint gates clean (plugin-refs, bun-test-imports, bot-api-wrapping, no-pii, vault-hermeticity, no-broadcast)

## Risk
**Medium** — touches the wedge-recovery path. Mitigated by: default-OFF flag, the bounded hard ceiling, the preserved no-in-flight-tool path, and independent bridge-disconnect crash recovery. **Canary on marko before fleet rollout.**

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -4579,7 +4579,27 @@ function ensureIssuesCard(chatId: string, threadId: number | undefined): void {
 // incident fix. In-memory only; a gateway recreate naturally resets it.
 let inFlightUpdate: { requestId: string; startedAt: number } | null = null
 
+// Fix A — silence-fallback tuning (status-surface darkening, 2026-06-05). A long
+// quiet tool stretch (foreground sub-agent / big research) crossed the 300s
+// fallback and nulled currentTurn, darkening the live activity feed mid-work.
+//   SWITCHROOM_SILENCE_FALLBACK_MS         — base threshold (default 300000)
+//   SWITCHROOM_SILENCE_FALLBACK_HARD_MS    — hard ceiling for the in-flight-tool
+//                                            defer (default 900000 = 15min)
+//   SWITCHROOM_SILENCE_DEFER_INFLIGHT_TOOLS=1 — enable the defer (default OFF;
+//                                            canary on marko against #2162 telemetry)
+function parsePositiveMsEnv(name: string, fallbackMs: number): number {
+  const raw = process.env[name]
+  if (raw == null || raw === '') return fallbackMs
+  const n = Number(raw)
+  return Number.isFinite(n) && n > 0 ? Math.floor(n) : fallbackMs
+}
+const SILENCE_FALLBACK_MS = parsePositiveMsEnv('SWITCHROOM_SILENCE_FALLBACK_MS', 300_000)
+const SILENCE_FALLBACK_HARD_MS = parsePositiveMsEnv('SWITCHROOM_SILENCE_FALLBACK_HARD_MS', 900_000)
+const SILENCE_DEFER_INFLIGHT_TOOLS = process.env.SWITCHROOM_SILENCE_DEFER_INFLIGHT_TOOLS === '1'
+
 silencePoke.startTimer({
+  thresholdsMs: { fallback: SILENCE_FALLBACK_MS, fallbackHardCeiling: SILENCE_FALLBACK_HARD_MS },
+  deferFallbackWhileToolInFlight: SILENCE_DEFER_INFLIGHT_TOOLS,
   emitMetric: (event) => {
     // Re-emit through the unified runtime-metrics fan-out (PostHog + JSONL).
     emitRuntimeMetric(event)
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

Opt-in, default-OFF defer of the 300s silence fallback while a parent tool is genuinely in flight. Verified against the real diff and both files at head.

- `silence-poke.ts:408-415` — defer gate is correct: requires `deferFallbackWhileToolInFlight === true` AND `s.inFlightTools.size > 0` AND `silence < (fallbackHardCeiling ?? Infinity)`, then `continue`s WITHOUT setting `fallbackFired` so the next tick re-checks. `inFlightTools` is a real Map (state def silence-poke.ts:85) populated by `noteToolStart` (:233) and drained by the idempotent `noteToolEnd` (:250), so `.size` reflects live tool state and drops to 0 the moment the tool ends — the defer then releases at the base threshold. No way to leave `fallbackFired` wrongly set.
- `silence-poke.ts:401` — the no-in-flight-tool path is untouched: a genuinely silent/wedged turn still fires at `thresholds.fallback` exactly as before, so the #1556 dangling-turn protection is intact; crash recovery stays with the independent bridge-disconnect sweep. Both safety claims check out in the code, not just the PR body.
- `gateway.ts:4602-4604` — wiring always supplies `fallbackHardCeiling` (default 900000 via `parsePositiveMsEnv`), so the unbounded `?? Infinity` defer-forever branch is only reachable by passing the dep without a ceiling (tests only). `parsePositiveMsEnv` correctly rejects non-finite/<=0 and floors. Flag gate is `=== '1'` (ships dormant).
- Tests (silence-poke.test.ts:+6) cover every branch: legacy-off fires at 300s, defer holds at 300s, fires once tool ends + still silent, fires at the hard ceiling on a hung-mid-tool turn, no-tool turn unaffected, no-ceiling defers indefinitely.

Nit (non-blocking): when defer is on and a tool's `noteToolEnd` never arrives (lost tool result), the turn pins until the 15min ceiling — intended and bounded, but worth watching on the marko canary that the ceiling actually fires in practice.

Rationale: bounded, default-OFF, preserves the no-in-flight and crash-recovery paths, fully unit-tested; the premise (post-#2162 the live feed means a tool-churning turn isn't silent to the user) is sound. Medium risk is correctly self-assessed and gated behind a canary flag.

PR #2166, opened 2026-06-05
fix(telegram): route late replies to their origin topic + reply-route telemetry

mekenthompson:
## Summary
Fixes the **2026-06-05 marko triage**: a reply that fires *after* the orphaned-reply backstop ended its turn lost its topic and landed in **General** — so in a forum supergroup the answer vanished from the topic the user was reading (the "out of order / missing message" symptom). Two late replies from the *same* dead turn routed to *different* topics, because resolution depended on the model echoing `origin_turn_id`, which it does inconsistently.

### Fix B2 — deterministic late-reply topic recovery (the robust fix)
New precedence tier (4) in `resolveAnswerThreadId`: when there's **no explicit thread, no echoed origin, AND no live turn** (`currentTurn` already cleared), recover the origin topic from the **most-recently-ended turn for that chat** (`findLatestEndedTurnForChat` over `recentTurnsById`). Deterministic — no model echo required. It is **not** the `chatThreadMap` last-seen heuristic (that caused the original wrong-topic bug); it's the chat's own most-recent turn, which a late reply almost always belongs to. Wired into **both** answer paths (`executeReply` + `executeStreamReply`). Kill switch: `SWITCHROOM_LATE_REPLY_TOPIC_RECOVERY=0`.

### reply-route telemetry — the blind spot this triage exposed
One line per answer reply: which precedence tier won (`via=explicit|origin|live|recovered|none`), the resolved thread, the origin turn + its thread, and whether the reply was **late** (`currentTurn` null). `via=recovered` marks a late reply this fix saved from General; **`UNROUTED`** flags a supergroup reply that still resolved to no topic (the residual gap to watch). Turns "did this answer land in the right topic?" into one `grep`.

### Fix B1 — best-effort directive reinforcement
The turn-pacing directive already strongly binds "answer"=reply tool; reinforced with the specific failure mode (call reply as the FIRST action; don't write the answer as transcript text then call reply after — the backstop flushes unsent text late and out of order). B1 is fundamentally model behaviour; **B2 makes its effect benign regardless** of whether the backstop still occasionally fires.

## Why (JTBD)
Outcome **2 (you hold the leash)** + **4 (predictable)** — the agent's answer must land in the topic the user asked in, in order.

## Test plan
- [x] `tsc --noEmit` clean
- [x] new `answer-thread-resolve.test.ts` (9): full precedence incl. tier-4 recovery, DM origin, recovery-doesn't-override-live, no-candidate legacy result
- [x] `scaffold.test.ts` (188) green — directive assertion is a substring match
- [x] gateway vitest sweep 176 green; lint gates clean

## Risk
Low/medium — tier 4 only adds a fallback where the old result was `undefined` (General); behind a kill switch; live-turn + origin precedence unchanged. Telemetry is additive. Will be exercised in the MTCute DM + channel UAT for this release.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -4153,7 +4153,11 @@ export function buildSettingsHooksBlock(p: HooksBlockParams): Record<string, unk
     'tool sequence (scheduling, multi-step research, sub-agent handback), ' +
     'do not let your closing narration stand as the answer: end the turn ' +
     'by passing that narration to the reply tool. No reply tool call = the ' +
-    'user got nothing, however much text you wrote.</turn-pacing>';
+    'user got nothing, however much text you wrote. Call the reply tool as ' +
+    'your FIRST action when you have the answer — do not write it out as ' +
+    'transcript text first and call reply afterward: a framework backstop ' +
+    'flushes unsent text after a delay and then your real reply lands late ' +
+    'and out of order.</turn-pacing>';
   const switchroomUserPromptSubmit: Array<Record<string, unknown>> = [
     ...(useHotReloadStable
       ? [
--- a/telegram-plugin/gateway/answer-thread-resolve.test.ts
+++ b/telegram-plugin/gateway/answer-thread-resolve.test.ts
@@ -1,0 +1,85 @@
+import { describe, it, expect } from 'vitest'
+import { resolveAnswerThreadId } from './answer-thread-resolve.js'
+
+describe('resolveAnswerThreadId — precedence', () => {
+  it('(1) explicit model thread wins over everything', () => {
+    expect(
+      resolveAnswerThreadId({
+        explicitThreadId: 7,
+        originResolved: true,
+        originThreadId: 3,
+        liveThreadId: 4,
+        lastEndedResolvedForChat: true,
+        lastEndedThreadIdForChat: 9,
+      }),
+    ).toBe(7)
+  })
+
+  it('(2) origin turn thread wins over the live turn (the Brevo→Meta fix)', () => {
+    expect(
+      resolveAnswerThreadId({ originResolved: true, originThreadId: 3, liveThreadId: 4 }),
+    ).toBe(3)
+  })
+
+  it('(2) a DM origin (resolved, thread undefined) pins to undefined, not the live thread', () => {
+    expect(
+      resolveAnswerThreadId({ originResolved: true, originThreadId: undefined, liveThreadId: 4 }),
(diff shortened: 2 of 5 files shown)


mekenthompson:
APPROVE

- `answer-thread-resolve.ts:87-103` — tier ordering is correct: explicit (1) → origin (2) → live (3) → tier-4 recovery → undefined. Tier 4 only adds a fallback where the old result was `undefined` (General); tiers 1-3 are byte-for-byte unchanged, so no existing routing path moves. Pure + fully covered by the 9-case `answer-thread-resolve.test.ts`.
- `gateway.ts:191-197` (`findLatestEndedTurnForChat`) — `recentTurnsById` is populated at enqueue (gateway.ts:9130) and never deleted on turn end (only LRU-evicted at 32). So for an *in-flight* General reply (live turn threadless, `sessionThreadId` undefined in a supergroup) the iteration returns the live turn itself → recovered thread undefined → resolves to General, unchanged. No diversion of legitimate General replies. The genuine late case (currentTurn cleared) recovers the ended turn's topic. Behavior is correct in both.
- `gateway.ts:218-221` — recovery gate `liveTurn?.sessionThreadId == null` is consistent with the pure resolver's tier-3 guard (`liveThreadId != null`), so the two never disagree on which tier fires.
- Kill switch `SWITCHROOM_LATE_REPLY_TOPIC_RECOVERY` (gateway.ts:178) defaults ON, restores legacy on `=0`. Wired into both `executeReply` (gateway.ts:6602) and `executeStreamReply` (gateway.ts:7259). Telemetry is additive (stderr only).
- Nit (non-blocking) — `gateway.ts:191` name `findLatestEndedTurnForChat` is a misnomer: it returns the latest turn for the chat *including the still-live one* (registry isn't pruned on turn end). Functionally fine; rename to `findLatestTurnForChat` to match behavior.
- Nit (non-blocking) — `gateway.ts:240-246`: a legitimate General-topic reply in a supergroup during a live General turn logs `via=recovered` + `UNROUTED(supergroup→no-topic)` even though General genuinely has no topic and routing is correct. UNROUTED will fire on every General reply, not just true gaps, diluting it as a grep signal; `late=false` while `via=recovered` is likewise slightly off from the PR's framing. Consider suppressing UNROUTED when the recovered owner is the live turn.

Rationale: deterministic late-reply topic recovery, bounded strictly to the prior-General fallback, behind a kill switch, with tiers 1-3 untouched and the live-turn-in-registry property making in-flight General replies safe — low risk. Nits are log-clarity only.

PR #2168, opened 2026-06-05
test(uat): real-work coverage (mcp/tool/sub-agent/research) + reply-route DM telemetry fix

mekenthompson:
## Summary
The existing fuzz scenarios send **conversational** prompts ("hey how's it going", emoji, markdown edge-cases) → trivial fast replies, so they never exercise the status-surface / reply-ordering bugs — those only appear under **genuine work** (MCP/tools/sub-agents/research that runs long enough to cross the silence-poke / orphaned-reply thresholds). This adds that coverage and fixes a defect the new telemetry caught in itself.

### New coverage (`telegram-plugin/uat/`)
- **`real-work-prompts.ts`** — human-style prompts that provoke multi-tool work, web research, sub-agent delegation, and background workers; a `collectTurn` that captures the **whole** bot-message sequence; and `analyzeTurn` bug detectors:
  - **hard:** `no-answer`, `orphaned-fragment` (the backstop signature: a stub lands ≥8s before the real answer), `wrong-surface` (leaked out of the channel).
  - **soft:** `surface-missing` — whether the agent dispatches a worker/sub-agent is its judgment, so a missing feed is *reported* (for the bug hunt), not failed.
- **`fuzz-real-work-dm.test.ts` / `fuzz-real-work-channel.test.ts`** — the set in a DM and a forum supergroup (self-skips green without `SWITCHROOM_UAT_CHAT_ID`).

### Live-validated on test-harness @ v0.14.66
- `step-by-step` → **real `df`/`du` output, 8 activity-feed edits over 51s**, feed stayed alive, one coherent answer (no orphaned fragment). Coverage + status surface confirmed working.
- The new `reply-route` telemetry surfaced a defect **in its own marker** (fixed below).
- The gateway log surfaced a **residual status-dark case** (see Findings).

### reply-route telemetry fix (`gateway.ts`)
The late-reply recovery gated on `liveTurn?.sessionThreadId == null` — also true for a **threadless DM with a live turn** — so every DM reply logged `via=recovered`/`RECOVERED`, drowning the real supergroup recoveries the marker exists to surface. Routing result was already correct (DM → undefined); now gates on `liveTurn == null` (the precise late-reply condition).

## Findings (reported, not all fixed here)
1. **Residual status-dark (follow-up):** a long *pure-composition* turn (the agent drafts a big answer, no tracked tools) still trips the 300s silence-poke and nulls `currentTurn`. Fix A (#2164) defers only on `inFlightTools`, not answer-stream/draft liveness. The "background worker" prompt hit this (agent composed inline for 5 min, no tools, no worker dispatched).
2. **Telemetry gap:** the silence-poke teardown isn't emitted in the `turn-lifecycle` format (it logs its own line), so a fallback-nulled turn has no `clear` line — harder to correlate.
3. **Methodology:** mtcute can't observe drafts (a draft-heavy turn reads as `botMsgs=0`); concurrent UAT runs against one agent cross-contaminate (sequential only).

## Test plan
- [x] `tsc --noEmit` clean
- [x] live run on test-harness (step-by-step pass; bgworker/research drove the residual finding)
- `uat/**` is excluded from gating CI — never reds a CI run.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1930,11 +1930,17 @@ function resolveAnswerThreadWithLog(
   liveTurn: CurrentTurn | null,
   surface: 'reply' | 'stream_reply',
 ): number | undefined {
+  // Recover ONLY for a genuinely LATE reply — no live turn at all. Gating on
+  // `liveTurn?.sessionThreadId == null` (the original) also fired for a
+  // threadless DM that still had a live turn, marking every DM reply
+  // `via=recovered`/RECOVERED in the telemetry (routing result unchanged —
+  // DM → undefined — but it drowned the real supergroup recoveries the marker
+  // exists to surface). `liveTurn == null` is the precise late-reply condition.
   const recovered =
     LATE_REPLY_TOPIC_RECOVERY_ENABLED &&
     explicitThreadId == null &&
     originTurn == null &&
-    liveTurn?.sessionThreadId == null
+    liveTurn == null
       ? findLatestEndedTurnForChat(chatId)
       : null
   const threadId = resolveAnswerThreadId({
--- a/telegram-plugin/uat/real-work-prompts.ts
+++ b/telegram-plugin/uat/real-work-prompts.ts
@@ -1,0 +1,332 @@
+/**
+ * Real-work UAT coverage — human-style prompts that trigger actual work
+ * (multi-tool, web research, sub-agents, background workers) plus a turn
+ * collector + bug detectors for the failure classes the conversational fuzz
+ * never exercised.
+ *
+ * Why this exists: the existing fuzz scenarios send conversational prompts
+ * ("hey how's it going", emoji, markdown edge-cases) → trivial fast replies.
+ * The status-surface and reply-ordering bugs (live feed going dark mid-work,
+ * the orphaned-reply backstop flushing a fragment then the real answer landing
+ * late and out of order, late replies misrouting) only manifest when the agent
+ * does REAL work — uses tools/MCPs, spawns sub-agents, researches long enough to
+ * cross the silence-poke / orphaned-reply thresholds. These prompts provoke that
+ * work in a human voice; `collectTurn` captures the whole bot-message sequence;
+ * `analyzeTurn` flags the known bug signatures.
+ *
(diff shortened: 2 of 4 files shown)


mekenthompson:
APPROVE

- `gateway.ts:9-20` (`resolveAnswerThreadWithLog`) — recovery gate tightened from `liveTurn?.sessionThreadId == null` to `liveTurn == null`. Routing-neutral: the two gates only diverge when a live turn EXISTS but is threadless (a DM, or General in a supergroup). In that case the old gate computed `recovered` = the live turn itself (it's in `recentTurnsById`, never deleted), `resolveAnswerThreadId` tier-4 returned its undefined thread → DM/General, same as the new gate's tier-3 `liveThreadId=undefined` → undefined. Result identical; only `via`/RECOVERED labelling changes. This is exactly the false-RECOVERED telemetry nit I flagged on #2166 — correctly fixed.
- The genuine late-reply case (`liveTurn == null`) is unchanged by this gate — both old and new gates are true there, so supergroup recovery still fires.
- `real-work-prompts.ts:208-270` (`collectTurn`) — observer iterator is created before `sendText`, with a settle window after the first substantive answer to catch the orphaned-fragment/late-misroute signature. Sound; the documented mtcute race (observeMessages backfills nothing) is acknowledged and acceptable for a UAT harness.
- `real-work-prompts.ts:301-352` (`analyzeTurn`) — hard violations (no-answer, orphaned-fragment ≥8s, wrong-surface) vs soft warnings (surface-missing) split is correct; a generic agent's choice to dispatch a worker is judgment, so reporting-not-failing it is right.
- CI safety verified: `vitest.config.ts:131` excludes `**/telegram-plugin/uat/**` from the default `vitest run` path; the three new files all live under `telegram-plugin/uat/`, so they never red a gating run (only `bun run test:uat` invokes them). PR-body claim confirmed.
- Findings #1-3 in the PR body (residual pure-composition status-dark, silence-poke teardown not in turn-lifecycle format, mtcute can't observe drafts) are correctly reported as follow-ups, not fixed here.

Rationale: one routing-neutral telemetry-gate fix (resolving a #2166 nit) plus net-new UAT coverage that is excluded from gating CI — low risk, no production behavior change beyond cleaner reply-route logging.

PR #2169, opened 2026-06-05
fix(telegram): count feed/draft activity as liveness so long work doesn't dark out

mekenthompson:
## Summary
Closes the residual status-dark case the 2026-06-05 UAT surfaced. A long turn that is **visibly producing** — the activity feed editing in place, or the answer-stream draft updating — still tripped the **300s silence-poke fallback** and nulled `currentTurn`, killing the very feed/draft the user was watching. Fix A (#2164) only deferred while a tracked **tool** was in flight; it didn't count feed/draft renders.

Proven on test-harness: a 5-min compose-only turn (`tools=0`, draft updating) → `silence-poke framework-fallback ended wedged turn silence_ms=303389 currentTurn_nulled=true`.

**Root:** the silence clock resets only on a real reply (`noteOutbound`). The header's "the model ripping through 20 tool calls is still silent to the user" rationale predates the live activity feed (#2162) + compose draft — those **are** user-visible now.

## Fix
- New `silence-poke.noteProduction(key, now)` (resets the clock like a reply), called on every successful **activity-feed render** (`drainActivitySummary`) and every **answer-stream draft update** (`onMetric`). The fallback now fires only on **genuine silence** — no reply, no feed, no draft, no tool events for the window — a real wedge. A wedged agent produces nothing observable, so its clock is never reset and it still recovers. Default ON; `SWITCHROOM_SILENCE_LIVENESS_PRODUCTION=0` restores legacy.
- **Telemetry gap fix:** emit `turn-lifecycle clear reason=silence_fallback` at the teardown, so a fallback-nulled turn has a lifecycle line like every other clear path (was only the differently-shaped framework-fallback line).

## Test plan
- [x] `tsc --noEmit` clean
- [x] `silence-poke.test.ts` +4 `noteProduction` cases (resets clock; 10-min steady render stays alive; production STOPS → fires 300s after last render; unknown-key no-op) — 53 green
- [x] gateway + status-surface tests green; lint gates clean
- [ ] end-to-end: long-tool-turn feed-cadence UAT after deploy to test-harness (this rollout)

## Risk
Medium (wedge-recovery path). Mitigated: only NEW renders reset (feed/draft dedup → a stuck agent re-rendering identical content doesn't reset); a genuine wedge (no output) still fires; default-ON kill switch; the bridge-disconnect sweep still recovers crashes independently.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -9012,13 +9012,6 @@ async function drainActivitySummary(turn: CurrentTurn): Promise<void> {
           )
         }
         turn.activityLastSentRender = target
-        // Production-liveness: the live activity feed just rendered (send/edit) —
-        // the agent is visibly working, so reset the silence-poke clock. Without
-        // this, a long tool turn whose feed keeps updating still tripped the 300s
-        // fallback and nulled currentTurn, killing the very feed the user watches.
-        if (SILENCE_LIVENESS_PRODUCTION) {
-          silencePoke.noteProduction(statusKey(chat, thread), Date.now())
-        }
       } catch (err) {
         const msg = err instanceof Error ? err.message : String(err)
         if (!msg.includes('message is not modified')) {
@@ -9480,6 +9473,16 @@ function handleSessionEvent(ev: SessionEvent): void {
         // the " · Ns" elapsed restarts from this step (and the feed itself just
         // advanced, so it isn't stale).
         turn.lastToolLabelAt = Date.now()
+        // Production-liveness: a NEW model-driven activity label is genuine
+        // liveness (the model emitted a new step), so reset the silence-poke
+        // clock — this is the safe site, NOT drainActivitySummary, because the
+        // framework feedHeartbeatTick also drains (climbing-elapsed re-renders)
+        // and would falsely reset the clock forever on a hung-mid-tool turn,
+        // reintroducing the #1556 dangling-turn wedge. Only the model emitting a
+        // fresh label reaches here.
+        if (SILENCE_LIVENESS_PRODUCTION && currentTurn === turn) {
+          silencePoke.noteProduction(statusKey(turn.sessionChatId, turn.sessionThreadId), Date.now())
+        }
         // Recompose so any active foreground sub-agent's nested block (Model A)
         // is preserved when the parent appends its own step. composeTurnActivity
         // == the flat render when no foreground sub-agent is active.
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- `silence-poke.ts:192-197` (`noteProduction`) — byte-for-byte equivalent to `noteOutbound`: sets `lastOutboundAt = now`, `fallbackFired = false`, no-ops on unknown key. Since `tick()` measures `silence = now - (lastOutboundAt ?? turnStartedAt)` (silence-poke.ts:397-401), production resets the clock exactly like a reply — so a turn producing observable output every <300s stays alive indefinitely, identical to the reply path. Intentional and test-pinned.
- Wedge recovery preserved (the core safety property): the activity-feed reset is gated by the `while (pendingRender !== lastSentRender)` dedup loop (gateway.ts:8966) AND sits only in the try-SUCCESS path — the catch ("message is not modified" / 400) does NOT call `noteProduction` (gateway.ts:9012 vs the catch at ~9019). So a stuck agent re-emitting identical content never enters the loop, never resets, and the 300s fallback still fires. A genuinely wedged claude emits no new tool_uses → no new render → recovers. Confirmed by the "production STOPS → fires 300s after last render" test.
- `gateway.ts:4901` telemetry fix — added `wedgedTurn != null` guard before emitting `turn-lifecycle clear reason=silence_fallback` and nulling. No regression: the only newly-skipped case is wedgedTurn==null while currentTurn==null, where the old `currentTurn = null` was a harmless no-op. Net win: the dark-out now greps in the same lifecycle shape as every other clear path.
- Nit (non-blocking) — `gateway.ts:9640` draft path: `noteProduction` is called from `onMetric`, which fires on stream metric events, NOT on a content-deduped visible draft change like the feed path. It mirrors the pre-existing `signalTracker.noteSignal` at the same site (so no new looseness), but it means draft-path liveness is really "the stream is producing tokens" rather than strictly "the user saw a new draft" — slightly weaker than the PR's "feed/draft dedup" framing. Harmless (token production still means claude is alive), but worth knowing the two paths aren't symmetric.
- Default ON, kill switch `SWITCHROOM_SILENCE_LIVENESS_PRODUCTION=0` (gateway.ts:4684); the Fix-A in-flight-tools defer branch and its hard ceiling (silence-poke.ts:412) are untouched; bridge-disconnect crash sweep still recovers independently. 4 new tests, 53 green.

Rationale: closes the residual compose-only/long-tool status-dark by counting genuinely-visible production as liveness, with the wedge-recovery guarantee intact via the success-path + dedup gating, behind a kill switch. Medium risk correctly self-assessed; the one asymmetry (draft path) is a non-blocking observation.

Issue #1017, opened 2026-05-11
broker: doesn't see new agents added after broker container start (no socket bind)


mekenthompson:
## Symptom

After `switchroom agent add <new-agent>` on a Docker host:

1. `switchroom apply` writes the new per-agent broker-socket volume into the compose file.
2. The new agent container is created and mounts the volume.
3. The agent's gateway tries to talk to the broker → broker isn't listening on `/run/switchroom/broker/<new-agent>/sock`.
4. Inside the agent: `VAULT-BROKER-DENIED: broker not running and stdin is not a TTY`.

Reason: the broker enumerates per-agent socket dirs **once at boot** (`vault-broker: per-agent socket listening agent=<name> sock=...` log lines emit at startup). New agents added later show up in the compose file and have their volume created, but the broker doesn't bind a socket for them until it's restarted.

## Repro

```bash
# Take an existing healthy fleet, then:
switchroom agent add new-agent --profile default --topology dm \
  --bot-username some_bot --allow-from $UID
# Wait for agent to come up.

docker exec switchroom-vault-broker ls /run/switchroom/broker/new-agent/
# (nothing — no sock file)
docker exec -u <new-uid> switchroom-new-agent switchroom vault list
# VAULT-BROKER-DENIED ...
```

Fix:

```bash
docker compose -p switchroom -f ~/.switchroom/compose/docker-compose.yml \
  up -d --force-recreate --no-deps vault-broker
```

## Proposed fix

Either:

1. **Auto-recreate broker** as part of `switchroom apply` when a new per-agent volume mount lands on the broker service. Cheapest; `apply` already runs `docker compose up -d` and could pass `--force-recreate vault-broker` conditionally.
2. **Broker watches its mount-root**, picks up new per-agent subdirs at runtime and binds sockets dynamically. Cleaner long-term; needs a watcher and a `chown` plumbing path.
3. **`agent add` itself** force-recreates the broker as a final step. Most local fix.

(3) is the smallest; ~5 LOC in `src/agents/add-orchestrator.ts` after the final `apply` call.

## Encountered while

Setting up the `test-harness` UAT agent for vault-UX scenarios (#866 / #1012). The agent appeared healthy but every vault op failed silently from inside the container; diagnosis required reading the broker's boot log and noticing only the pre-existing agents had per-agent socket lines.

## Related
- #1016 (just merged) — `admin: true` env propagation. Same shape of bug: yaml change without an obvious restart-the-singleton step.

PR #2170, opened 2026-06-05
fix(rollout): recreate stale singletons on pin bump + doctor drift check

mekenthompson:
## Summary

Fixes a silent operational bug: **the staggered fleet-rollout recipe never recreates the singleton containers**, so a `release.pin` bump can leave `vault-broker` / `approval-kernel` / `switchroom-auth-broker` frozen on an old image while every agent rolls forward.

## The incident (root cause)

The documented recipe is: edit `release.pin` → `switchroom apply` (regenerates compose with new image tags on **every** service) → per-agent `switchroom agent restart <name> --wait --force`. But `agent restart`'s only docker command is `docker compose up -d --force-recreate --no-deps agent-<name>`. The `--no-deps` flag — correct, so a routine bounce doesn't bounce the brokers — means **nothing in the recipe ever recreates the singletons** after a pin bump. The compose file says v0.14.66; the running broker stays on whatever it last recreated at.

On 2026-06-05 this surfaced live: `switchroom-auth-broker` had been on **v0.14.24 for ~2 weeks** while the fleet reached v0.14.66, so `/connect microsoft` failed with *"provider 'microsoft' is not registered"* — the broker code predated the Microsoft provider by 42 versions. (Resolved live by manually `docker compose up -d`-recreating all three singletons.)

## The fix

- **`src/agents/singleton-reconcile.ts` (new)** — `detectSingletonDrift` compares each singleton's running image (`docker inspect`) to the compose-pinned image (parsed from the generated compose); `reconcileSingletons` recreates **only the drifted ones** via plain `up -d --no-deps <svc>` (no `--force-recreate` → a true no-op when in sync). Best-effort: a recreate failure is logged + collected, never thrown. All IO injectable → unit-tested without docker.
- **`agent.ts` `restart` handler** — calls `reconcileSingletonImages()` **once before the per-agent loop**, so the staggered recipe self-heals stale singletons (brokers first, then agents). Drift-aware ⇒ near-zero overhead on a routine single-agent restart.
- **`docker-fleet.ts` `agent add`** — added the **omitted `switchroom-auth-broker`** to the singleton recreate loop (it has the same per-agent socket-dir enumeration as vault-broker/kernel — the same #1017 class of bug the vault-broker entry already guards, just never extended to auth-broker).
- **`doctor-docker.ts` `checkSingletonImageDrift` (new)** — the check that would have caught this before a feature broke. Warns + names the stale singleton & tags; degrades to `ok` when docker is unreachable.

## Why this is the right cut

`switchroom update` (whole-project `up -d`) already recreates singletons — but the canary recipe deliberately avoids `update` to skip the bulk bounce, and in doing so dropped singleton recreation entirely. The fix lives in the path the recipe actually uses (`agent restart`), is drift-aware so it's cheap, and adds a doctor check as the safety net.

## Test plan

- [x] `src/agents/singleton-reconcile.test.ts` (12) — drift detection (in-sync / stale / absent / build-local), reconcile (recreates only drifted, no-op when synced, best-effort on recreate failure, **regression guard that `switchroom-auth-broker` is covered**).
- [x] `tests/docker/doctor-checks.test.ts` (+6) — drift check ok / warn-with-tags / docker-unreachable / wired-into-`runDockerChecks`.
- [x] `tsc --noEmit` + full lint chain green.
- [x] Verified live: manual recreate of all three singletons to v0.14.66 fixed `/connect microsoft` (the incident); vault-broker auto-unlocked, agents kept cached creds.

## Risk

Low. `reconcileSingletons` is drift-aware and best-effort — when images match the pin (the common case) it's 3 `docker inspect` reads + 1 compose parse and zero recreations. Only a genuine pin-vs-running mismatch triggers a recreate, which is exactly the desired self-heal. `--no-deps` keeps it from cascading into agent bounces.

> ⚠️ Note: 2 pre-existing `update.test.ts` failures in the dev sandbox (hostd re-exec shells `docker exec` → status 127) are **environmental, unrelated to this change**, and reproduce on a pristine base with this branch's changes stashed — they pass in CI.

### DIFF
--- a/src/agents/docker-fleet.test.ts
+++ b/src/agents/docker-fleet.test.ts
@@ -84,7 +84,7 @@ describe("bringUpAgentService", () => {
     });
   });
 
-  it("recreates broker + kernel then brings up agent — three compose calls in fixed order (#1017)", () => {
+  it("recreates broker + kernel + auth-broker then brings up agent — four compose calls in fixed order (#1017)", () => {
     const home = mkdtempSync(join(tmpdir(), "docker-fleet-"));
     (execFileSync as any).mockReturnValue(Buffer.from(""));
 
@@ -96,11 +96,12 @@ describe("bringUpAgentService", () => {
       stdio: "ignore",
     });
 
-    // Order matters: the two singletons must be recreated BEFORE the
-    // new agent comes online so their per-agent socket-dir mounts
-    // include the new agent's subdir. See #1017 for the bug this
-    // ordering fixes.
-    expect(execFileSync).toHaveBeenCalledTimes(3);
+    // Order matters: all THREE singletons (vault-broker, approval-kernel,
+    // switchroom-auth-broker) must be recreated BEFORE the new agent comes
+    // online so their per-agent socket-dir mounts include the new agent's
+    // subdir. See #1017 for the bug this ordering fixes. auth-broker was
+    // added to close the same gap for OAuth-credential sockets.
+    expect(execFileSync).toHaveBeenCalledTimes(4);
     const composePath = resolve(home, "compose", "docker-compose.yml");
     const callArgs = (execFileSync as any).mock.calls.map(
       (c: [string, string[], unknown]) => c[1],
@@ -126,6 +127,16 @@ describe("bringUpAgentService", () => {
       "approval-kernel",
     ]);
     expect(callArgs[2]).toEqual([
+      "compose",
+      "-f",
+      composePath,
+      "up",
+      "-d",
+      "--no-deps",
+      "--force-recreate",
+      "switchroom-auth-broker",
+    ]);
+    expect(callArgs[3]).toEqual([
       "compose",
       "-f",
       composePath,
--- a/src/agents/docker-fleet.ts
+++ b/src/agents/docker-fleet.ts
@@ -139,8 +139,14 @@ export function bringUpAgentService(
   // but the broker never binds the matching socket, so every vault
   // operation from the new agent fails with `VAULT-BROKER-DENIED:
   // broker not running` (#1017). `--no-deps` keeps us from recursively
-  // bouncing other agents (only the broker + kernel are touched).
-  for (const svc of ["vault-broker", "approval-kernel"] as const) {
+  // bouncing other agents (only the singletons are touched).
+  // switchroom-auth-broker is included alongside vault-broker +
+  // approval-kernel: it has the same per-agent socket-dir enumeration
+  // (#1017), so a newly-added agent's /run/switchroom/auth-broker/<agent>
+  // socket isn't bound until the auth-broker is recreated — omitting it
+  // strands the new agent's OAuth-credential reads the same way the
+  // vault-broker omission once stranded vault reads.
+  for (const svc of ["vault-broker", "approval-kernel", "switchroom-auth-broker"] as const) {
     execFileSync(
       dockerBin,
       [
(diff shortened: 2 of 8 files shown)


mekenthompson:
APPROVE

- Core correctness confirmed: `SINGLETON_SERVICES` (singleton-reconcile.ts:305-309) — `vault-broker`, `approval-kernel`, `switchroom-auth-broker` — match the generated compose service names exactly (compose.ts:785 `vault-broker:`, :1016 `approval-kernel:`, :1108 `switchroom-auth-broker:`). So `readPinnedSingletonImages` resolves a pin for all three and the fix is NOT a silent no-op (the failure mode it's curing). Container names line up too under the default prefix: `singletonContainerName` yields `switchroom-vault-broker`/`-approval-kernel`/`-auth-broker`, matching compose's `${containerNamePrefix}-…` (compose.ts:787/1018/1110) with the default `switchroom` prefix.
- `singleton-reconcile.ts:410` `needsRecreate = pin !== null && running !== pin` — correct: build-local (no `image:`) → pin null → never flagged; absent container → running null ≠ pin → flagged for bring-up; in-sync → equal → skipped. `detectSingletonDrift` is a pure read; `reconcileSingletons` recreates only flagged services and is genuinely best-effort (singleton-reconcile.ts:463-470 collects, never throws), so it can't fail the restart it piggybacks on.
- `singleton-reconcile.ts:424-428` recreate uses `up -d --no-deps <svc>` (no `--force-recreate`) — recreates only because the `image:` line changed, and `--no-deps` prevents cascading agent bounces. Correct, and a true no-op when in sync since reconcile skips non-drifted services before ever shelling out.
- `docker-fleet.ts:18` — adds the omitted `switchroom-auth-broker` to the `agent add` singleton recreate loop. Genuine sibling fix: auth-broker has the same per-agent socket-dir enumeration (#1017) as vault-broker/kernel, so without it a newly-added agent's `/run/switchroom/auth-broker/<name>` socket goes unbound until the next recreate.
- `doctor-docker.ts:563` `checkSingletonImageDrift` — filters `d.running !== null` so an absent container reads as 'down' (owned by health checks), not 'drift'; degrades to `ok` when docker is unreachable. Wired into `runDockerChecks` only when composeYaml is present (doctor-docker.ts:293). Correct safety-net placement.
- Nit (non-blocking) — `singletonContainerName` (singleton-reconcile.ts:314) hard-codes the `switchroom-` container prefix. Under a custom `containerNamePrefix` (test-isolation, compose.ts:288) the `docker inspect` target is wrong → running=null → always flags drift → a redundant `up -d --no-deps <svc>` that compose resolves to a no-op (it finds the real container in sync). Harmless and prod uses the default prefix, but the detection is prefix-blind; consider threading the project/prefix if isolation ever needs accurate drift reporting.
- Nit (non-blocking) — drift is an exact string compare of `.Config.Image` vs the compose `image:` tag. Right for switchroom's tag-bump release model (the incident), but it won't detect a same-tag digest change (re-pushed tag). Out of scope here; worth knowing the check is tag-granular, not digest-granular.

Rationale: closes a real silent rollout gap (singletons frozen on an old image after a pin bump) in the exact path the staggered recipe uses, drift-aware so it's cheap, best-effort so it can't break a restart, with a doctor check as the net and the parallel auth-broker `agent add` omission fixed. 18 unit tests incl. an auth-broker regression guard. Service-name match verified against compose.ts — low risk.

PR #2172, opened 2026-06-05
feat(web): Connections dashboard tab — Google + Microsoft + Notion visibility

mekenthompson:
## Summary

Adds a unified **Connections** tab to the web dashboard (renamed from the old read-only "Google" tab) showing all three external integrations — Google, Microsoft 365, Notion — and, for each account/database, **which agents have access** (the `enabled_for` ACL). This is the visibility foundation for managing connections from the UI.

## Why

The dashboard could show Google accounts but had **no view of Microsoft or Notion**, and Microsoft is the freshly-shipped integration (the `/connect microsoft` flow) with zero UI. Operators asked to *see and manage* external connections + per-agent access from the management UI. This PR is the read half; the write half (connect / choose-agents) is the follow-up.

## What changed

- **`api.ts`** — `handleGetMicrosoftAccounts` mirrors `handleGetGoogleAccounts`: unions the `microsoft_accounts` config ACL with the broker's `listMicrosoftAccounts` inventory, adds `accountType` (personal/work), degrades to config-only when the broker is unreachable. `handleGetNotionWorkspace`: Notion has no per-account model, so it surfaces the declared `databases` + which agents' `notion_workspace.databases` grant each — the token itself is **never returned**, only its vault-key reference.
- **`server.ts`** — `GET /api/microsoft-accounts` + `GET /api/notion-workspace` via the existing `parseRoute`→switch dispatch.
- **`ui/index.html`** — the Google tab becomes **Connections**, fetching all three endpoints in parallel and rendering Google / Microsoft 365 / Notion sections. OAuth account cards show enabled-for agents, expiry, scope count, client; Microsoft cards add a personal/work pill; Notion shows each database + granted agents. Empty states point at the connect path.

## Read-only

Pure visibility — no write/mutation. Managing access (connect a service, choose which agents) is the follow-up PR (needs the config-write-path decision: direct yaml editor vs hostd).

## Validation

Smoke-tested **live against the running v0.14.67 broker**: the view surfaces the real connected Microsoft account (`ken.thompson@outlook.com.au`, personal, minted by the shipped default app `9dff88fa…`) with `enabledFor: []` — exactly the "connected but not yet wired to an agent" state the view is meant to make visible.

## Test plan

- [x] `tests/web.test.ts` (+6): Microsoft merge / broker-degrade / config∪broker union (mirrors the Google cases) + Notion not-configured / databases-with-per-agent-grants / token-never-leaked.
- [x] `tsc --noEmit` + full lint chain green; UI inline JS parses (`new Function` over each `<script>`).
- [x] Live handler smoke against the real broker.

## Risk

Low. Two new **read-only** GET endpoints mirroring an existing one; the broker calls (`listMicrosoftAccounts`) already shipped (#2088) and degrade gracefully when unreachable. The UI change renames one tab and adds two sections; no other tab touched. Dashboard auth/CSRF unchanged (same bearer/tailnet gate as the existing GETs).

### DIFF
--- a/src/web/api.ts
+++ b/src/web/api.ts
@@ -34,6 +34,10 @@ import type { ApprovalDecisionMeta } from "../vault/broker/protocol.js";
 import { captureEvent, captureException } from "../analytics/posthog.js";
 import { resolveAgentsDir } from "../config/loader.js";
 import { resolveAgentConfig } from "../config/merge.js";
+import {
+  agentCanAccessNotionDB,
+  shouldEmitNotionMcp,
+} from "../config/notion-workspace-acl.js";
 import { getAccountInfos, type AccountInfo } from "../auth/account-store.js";
 import {
   AuthBrokerError,
@@ -772,16 +776,27 @@ export interface NotionWorkspaceDashboard {
   vaultKey: string | null;
   /** Declared friendly-name → id databases with their agent grants. */
   databases: NotionDatabaseInfo[];
+  /**
+   * Agents with UNRESTRICTED Notion access (a `notion_workspace:` block
+   * with no `databases` filter → every DB the integration can see). They
+   * also appear in each database's `enabledFor`, but are surfaced here
+   * so the UI can render "full access" once instead of in every row.
+   */
+  fullAccessAgents: string[];
 }
 
 /**
  * Notion workspace view for the dashboard. Unlike Google/Microsoft,
  * Notion has no per-account concept (one operator-owned internal
  * integration token); access is per-DATABASE via each agent's
- * `notion_workspace.databases` friendly-name list. This surfaces the
- * declared databases and, for each, which agents are granted it — the
- * "which agents can reach Notion (and which DBs)" answer. The token
- * itself never leaves the vault and is never returned here.
+ * `notion_workspace.databases` friendly-name list — OR unrestricted when
+ * an agent has a `notion_workspace:` block with no `databases` filter.
+ * This surfaces the declared databases and, for each, which agents are
+ * granted it — the "which agents can reach Notion (and which DBs)"
+ * answer. Per-DB grants are computed via the canonical
+ * `agentCanAccessNotionDB` so the dashboard exactly matches runtime
+ * authorization (including the empty-filter = full-access case). The
+ * token itself never leaves the vault and is never returned here.
  */
 export function handleGetNotionWorkspace(
   config: SwitchroomConfig,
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- Token-leak check (the security-sensitive bit): `api.ts:109-139` `handleGetNotionWorkspace` returns only `vaultKey: nw.vault_key` — the vault KEY REFERENCE, never the token (which lives in the vault, not in config). No token field is read or emitted. Test asserts the serialized payload contains no token value (web.test.ts:490). Clean.
- Auth gate inherited correctly: both new routes dispatch through the same `parseRoute`→`checkAuth`→switch path, and `checkAuth(req, token, server)` runs at server.ts:501 BEFORE the `switch (route.handler)` at :504 — so `getMicrosoftAccounts`/`getNotionWorkspace` get the identical bearer/tailnet gate as every existing GET. No per-handler bypass. Confirmed.
- `api.ts:30-80` `handleGetMicrosoftAccounts` faithfully mirrors the Google handler: unions config ACL ∪ broker inventory, adds `accountType`, catches `AuthBrokerUnreachableError` and rethrows everything else (api.ts:55) so a real bug still surfaces. Degraded path renders ACL with live fields null. Correct.
- UI: every dynamic value is `escapeHtml`-wrapped in `renderOAuthAccountCard` and the Notion card (account, accountType, enabledFor, scope, clientId, db.name, db.id, vaultKey at index.html:364). `Promise.all` degrades each endpoint independently (`.then(r => r.ok ? r.json() : [])`), so one failing fetch doesn't blank the tab. Sound.
- Nit (non-blocking, PRE-EXISTING — shared with Google): the ACL lookup builds `keys` from `Object.keys(cfgAccounts).map(toLowerCase)` then reads `cfgAccounts[key]` with the lowercased key (api.ts:66-68). If a config email key is mixed-case (e.g. `Lisa@Outlook.com`), the lookup misses and `enabledFor` silently renders `[]`. Identical pattern already ships in `handleGetGoogleAccounts` (api.ts:665-672), so this PR doesn't regress anything — but worth normalizing the config map to lowercase once, in both handlers, in the follow-up.

Rationale: two read-only GET endpoints mirroring an existing, already-reviewed one; the two genuinely security-sensitive surfaces (Notion token, auth gate) verified clean; UI escapes all dynamic content and degrades per-endpoint. 6 new tests incl. a token-never-leaked assertion. Low risk.

PR #2173, opened 2026-06-05
feat(telegram): retire the invisible compose-box draft → testable visible answer lane

mekenthompson:
## Summary
The live answer-stream rendered to the **compose-box draft** (`sendMessageDraft`), which the **mtcute UAT harness cannot observe** — so the live answer surface was untestable (it's why the real-work UAT kept reporting `botMsgs=0`). Per the operator's request, retire it and switch live status to a surface we can **test reliably**.

**Design** (workflow-investigated, hybrid): retire the draft **and** flip the answer lane to a **visible edit-in-place message** (`sendMessage` + `editMessageText`) — fully mtcute-observable. Default RETIRED; kill switch `SWITCHROOM_DRAFT_ANSWER_LANE=0` restores the legacy draft.

### Why not just disable the lane
The #2169 `onMetric → noteProduction` silence-liveness reset is the **only** reset for a long pure-compose DM turn (no tools, no reply yet). Disabling the lane — or dropping the draft without flipping `minInitialChars` to 1 — would lose that reset → a 300s false "still working" nudge over a composing agent (a **#2169 regression**). Switching to visible **preserves** the reset and **broadens** it from DM-only to DM+supergroup (supergroups had zero answer-lane reset before).

### Changes (single chokepoint + paired flips, all behind the kill switch)
- `answer-stream-flag.ts`: `parseDraftLaneRetiredEnabled` (default true).
- `gateway.ts`: `DRAFT_ANSWER_LANE_RETIRED` hoisted above the draft probe (no TDZ); `sendMessageDraftFn` gated off when retired (drops the draft from **both** the answer-stream config and the `:7437` `stream_reply` consumer — which already falls back to message transport when the draft API is absent, as in channels); answer-stream config → `{minInitialChars:1}` (visible) when retired; the **materialize-as-answer guard broadened in lockstep** (else a text-only no-reply turn would `retract()` + delete the user's only answer copy); boot log clarified.

### Flash
Bounded by model behavior: on the common think→tool→reply turn the model emits ~zero interstitial text past the 150ms preamble window, so the visible lane **never opens** (no flash); reply-tool turns retract the preview and the single formatted reply is canonical; the only fresh-send+delete is the one-shot beat-5 ping on text-only no-reply turns (the same path that already ran in visible mode).

## Test plan
- [x] `tsc` clean; `parseDraftLaneRetiredEnabled` unit tests (6)
- [x] **wiring guards (5)** pin the PRIMARY (must flip `minInitialChars:1`), SECONDARY (materialize lockstep), TDZ-order, and onMetric-preserved risks the design flagged
- [x] answer-stream + silence-poke + silence-liveness-wiring regression (91) green; lint clean
- [ ] post-deploy canary (**DM + supergroup**): `visible-answer-stream-dm` passes with **no env override** (visible is now default); `jtbd-fast-trivial-dm` shows no streamed-then-deleted preliminary (no per-turn flash); a long pure-prose turn doesn't trip the 300s fallback (onMetric reset survived); kill-switch verified.

## Risk
Medium-low, fully kill-switched (env var + restart, no image rebuild for the flag).

### DIFF
--- a/telegram-plugin/answer-stream-flag.ts
+++ b/telegram-plugin/answer-stream-flag.ts
@@ -16,3 +16,22 @@ export function parseVisibleAnswerStreamEnabled(raw: string | undefined): boolea
   const v = raw.trim().toLowerCase()
   return v === '1' || v === 'true' || v === 'on' || v === 'yes'
 }
+
+/**
+ * Draft-answer-lane retirement (2026-06-05). The compose-box draft transport
+ * (`sendMessageDraft`) is invisible to the mtcute UAT harness, so the live
+ * answer-stream surface couldn't be tested. Retired by DEFAULT: the answer lane
+ * now opens a real, observable edit-in-place message instead of the compose-box
+ * draft (and the onMetric silence-liveness reset from #2169 now fires on visible
+ * sends in BOTH DMs and supergroups, not just DM drafts). Kill switch
+ * `SWITCHROOM_DRAFT_ANSWER_LANE=0` (also false/off/no) restores the legacy
+ * invisible draft.
+ *
+ * Returns true when the draft lane is RETIRED (the default — env unset or any
+ * truthy value); false only for an explicit disable of the retirement.
+ */
+export function parseDraftLaneRetiredEnabled(raw: string | undefined): boolean {
+  if (raw == null) return true
+  const v = raw.trim().toLowerCase()
+  return !(v === '0' || v === 'false' || v === 'off' || v === 'no')
+}
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -98,7 +98,7 @@ import * as pendingProgress from '../pending-work-progress.js'
 import { writeSilentEndState, clearSilentEndState, recordUndeliveredTurnEnd } from '../silent-end.js'
 import { isFinalAnswerReply, isSubstantiveFinalReply } from '../final-answer-detect.js'
 import { createAnswerStream, type AnswerStreamHandle } from '../answer-stream.js'
-import { parseVisibleAnswerStreamEnabled } from '../answer-stream-flag.js'
+import { parseVisibleAnswerStreamEnabled, parseDraftLaneRetiredEnabled } from '../answer-stream-flag.js'
 import { type SessionEvent } from '../session-tail.js'
 import {
   shouldSuppressToolActivity,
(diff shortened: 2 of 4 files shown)


mekenthompson:
APPROVE

- `answer-stream-flag.ts:23` — `parseDraftLaneRetiredEnabled` defaults RETIRED (true) on null/unrecognized, false only on explicit 0/false/off/no; mirrors the sibling parser's normalization. Unit-tested (answer-stream-flag.test.ts:126). Correct.
- `gateway.ts:51` — `DRAFT_ANSWER_LANE_RETIRED` hoisted above the sendMessageDraft boot-probe block; the TDZ risk is real (the const is read at gateway.ts:65 inside the probe) and the ordering is pinned by draft-retirement-wiring.test.ts:176. Correct.
- `gateway.ts:65` — single chokepoint: `!DRAFT_ANSWER_LANE_RETIRED && typeof _rawSendMessageDraft === 'function'` forces `sendMessageDraftFn = undefined` when retired. Verified BOTH consumers null-guard it: the answer-stream config at gateway.ts:9550 only reads it in the legacy (non-retired) MAX-gate branch, and the `stream_reply` consumer at gateway.ts:7437 spreads `sendMessageDraftFn != null ? {sendMessageDraft} : {}` — so undefined drops the draft and both fall back to message transport. No null-deref.
- `gateway.ts:80` — PRIMARY flip: `ANSWER_STREAM_VISIBLE_ENABLED || DRAFT_ANSWER_LANE_RETIRED ? {minInitialChars:1} : {sendMessageDraft, MAX_SAFE_INTEGER}`. Without this the retired lane (no draftFn) would sit behind the MAX gate and never open — a total answer-lane + #2169 onMetric-reset no-op. The paired flip is what keeps the silence-liveness reset alive (and broadens it DM→DM+supergroup). Guard-tested at draft-retirement-wiring.test.ts:183.
- `gateway.ts:95` — SECONDARY lockstep: materialize-as-answer guard broadened to the same predicate. Confirmed the else-branch is retract(); without the broadening a text-only no-reply turn (replyCalled=false) that streamed a visible preview would retract()+delete the user's only answer copy. Materialize path correctly gates the cleanup delete on a numeric returned id (gateway.ts:~9895), preserving the preview when materialize dedup-suppresses. Guard-tested at draft-retirement-wiring.test.ts:189.
- `gateway.ts:104` — boot log now reports `lane=visible(draft-retired)|visible|draft` + `draftFn=available|off`. Accurate.

Rationale: clean, fully kill-switched (env + restart, no rebuild) retirement of the untestable compose-box draft in favor of an mtcute-observable edit-in-place message. The two design-flagged silent-break modes (forget minInitialChars flip; forget materialize broadening) are both flipped in lockstep and pinned by source-level wiring guards, and the #2169 onMetric reset is preserved and broadened. Flash is bounded by model behavior on the common think→tool→reply path. No production regression.

PR #2175, opened 2026-06-05
test(uat): multi-topic + cross-surface ordering stress + UNROUTED telemetry false-alarm fix

mekenthompson:
## Summary
Stress coverage for **message ordering / status placement / multiple channels**, using the test supergroup's General topic + a real topic (`thread=5`). The driver sends into a topic via `messageThreadId`→`replyTo` and reads each observed message's `threadId`, so *which* topic each answer landed in is provable.

### New scenarios (non-gating, self-skip green without a resolvable supergroup)
- **`fuzz-multitopic-routing-channel`** — two questions in two forum topics back-to-back (the #2137 bleed shape): each answer must land in **its own topic**, no swap/drop. Plus a **slow-topic-5 + fast-General** variant where the slow answer lands **late** (after `currentTurn` flipped) and must still route to topic-5 (#2166 late-reply recovery).
- **`fuzz-cross-surface-ordering-channel`** — a DM question + a channel question back-to-back: neither answer may leak across surfaces (singleton-`currentTurn` cross-contamination).

**All pass live on test-harness @ v0.14.68** — routing is correct in the simple, late-after-flip, and cross-surface cases.

### Telemetry fix found *by* this stress
The `reply-route` `UNROUTED(supergroup→no-topic)` flag fired on **every legitimate General-topic reply** (a General turn correctly resolves to no thread — `thread=-` is right for General). It now gates on **`ownerTurn == null`** — fires only when a supergroup reply resolved to no topic with **no owner turn** to attribute it to (genuinely lost). General replies no longer false-alarm, so the flag is useful for catching real bleeds again.

> Note: the stress *initially* flagged a "bleed", which turned out to be my test's `/42/` regex matching `~42Gi` in the slow turn's `free -h` tool output — the real General answer routed to General correctly. Fixed with an unambiguous word token (`ZUCCHINI`). Good reminder that answer-tokens must not collide with tool output.

## Test plan
- [x] `tsc` clean; `answer-thread-resolve` + `multitopic-routing-wiring` (26) green; lint clean
- [x] 3 stress scenarios pass live on test-harness @ v0.14.68

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -1932,8 +1932,10 @@ function findLatestEndedTurnForChat(chatId: string): CurrentTurn | null {
  * timestamps. This wrapper logs, per reply: which precedence tier won (`via`),
  * the resolved thread, the origin turn + its thread, and whether the reply was
  * late (turn already ended). `via=recovered` marks a late reply this fix saved
- * from General; `UNROUTED` flags a supergroup reply that still resolved to no
- * topic (the residual gap to watch).
+ * from General; `UNROUTED` flags a supergroup reply that resolved to no topic
+ * with NO owner turn to attribute it to (genuinely lost — the residual gap to
+ * watch). A General-topic turn legitimately has no thread, so its replies are
+ * NOT flagged.
  */
 function resolveAnswerThreadWithLog(
   chatId: string,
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `gateway.ts:28` — `unrouted` now gates on `ownerTurn == null` in addition to `isSupergroup && threadId == null`. Correct: `ownerTurn` (line 20) = `originTurn ?? recovered ?? liveTurn`, so the flag fires only when a supergroup reply resolves to no topic AND no turn can claim it (genuinely lost). A legitimate General-topic reply has a live/origin turn, so it's no longer false-flagged — the residual-gap signal stays meaningful. No behavior change to routing itself, only to the diagnostic flag.
- `fuzz-multitopic-routing-channel.test.ts:236-237,253-254` — distinct non-overlapping tokens (14 for topic-5, 42 for General) with anchored regex `/(^|\D)NN(\D|$)/`; threadId assertions prove which topic each answer landed in (#2137 swap/drop + #2166 late-recovery). Self-skips green without a resolvable `SWITCHROOM_UAT_CHAT_ID` supergroup (line 195, 207-216).
- `fuzz-multitopic-routing-channel.test.ts:301-302,316-321` — late-after-flip variant uses WORD tokens (CORESDONE/ZUCCHINI) to avoid the numeric tool-output collision the PR body documents (`/42/` matching `42Gi` in `free -h`); honestly logs + warns when the slow answer did NOT actually arrive late, so a non-exercised run isn't silently reported as a pass of the stronger claim.
- `fuzz-cross-surface-ordering-channel.test.ts:108-116,132` — stray detection is symmetric (42-in-DM or 14-in-channel = leak) and asserts `strays` empty plus per-surface chatId; deadline-bounded dual-pump with `iter.return?.()` cleanup. Self-skips green without supergroup.

Rationale: a defensive one-line telemetry gate that removes a per-General-reply false alarm (verified against the resolver's `ownerTurn` definition), plus two non-gating UAT scenarios with unambiguous tokens and honest skip/late reporting. No production routing path changes; `uat/**` is non-gating. Author states tsc/lint clean and 3 scenarios pass live @ v0.14.68.

PR #2176, opened 2026-06-05
fix(telegram): deterministic answer-routing — prove the core, recover origin without the model echo

mekenthompson:
## Summary

Forum-supergroup answer/topic routing was **conditionally deterministic** — correct only when the model echoes `origin_turn_id` (precedence tier 2). A finite-FSM proof (total enumeration of the pure resolver + adversarial trace of the routing FSM that feeds it) found the load-bearing hole and this PR proves the deterministic core, recovers origin without the model where the framework owns the signal, and makes the irreducible residual observable.

## Why

The origin signal that survives a `currentTurn` flip travels through the **model** (it echoes `origin_turn_id`). When the model omits it **and** `currentTurn` has flipped to a successor topic — reachable by construction via the 2.5s no-reply escape-hatch drain or the 300s silence-poke — the reply falls to the live successor's thread = **wrong topic**, and silently (not even `UNROUTED`-flagged). This is the original Brevo→Meta multi-topic bleed, reopened on every model omission while a successor is live. A live fuzz never saw it because the model echoed every time, so it couldn't distinguish "routing is deterministic" from "the model happened to echo."

JTBD: *you hold the leash* (`reference/` outcome 2) — an answer must land where the user is reading, not get stolen by another topic.

## What

1. **Prove the deterministic core.** `answer-thread-resolve.test.ts` is promoted from 9 sampled cases to a **total enumeration** over the whole 64-input space: totality, determinism (idempotence), no-fabrication, the full precedence decision table, and the by-construction invariants — notably **INV-2 origin flip-immunity** (once origin is resolved, the result is independent of `liveThreadId` for *every* combination, so a `currentTurn` flip cannot steal a resolved origin).

2. **Recover origin without the model echo, deterministically.** When the model quotes a specific message (`reply_to`), that message id is a *framework-owned* anchor. A new source-message reverse index (evicted in lock-step with `recentTurnsById`, chat-scoped to avoid cross-chat id collisions) maps it back to the owning turn. `executeReply`/`executeStreamReply` resolve origin as `echo ?? quoted` — echo stays authoritative; the quote is a strictly-additive fallback that resolves the **actual** origin turn (never the live successor), so it cannot mask a misroute. Kill switch `SWITCHROOM_FRAMEWORK_ORIGIN_ROUTING` (default on, independent of `SWITCHROOM_TURN_ORIGIN_ROUTING`).

3. **Alarm the irreducible residual.** A bare no-echo, no-quote late reply is genuinely model-dependent — only the model knows which ended turn it answers. Rather than route it to the live successor silently, emit `MISROUTE_RISK(no-echo→live-successor)` telemetry when a different topic recently had a turn. Observability only; routing unchanged. Mirrors the obligation ledger's `resolveCloseTarget` stance: when ambiguous, surface it, never wrong-attribute.

## Determinism verdict

NOT fully deterministic-by-construction even after this PR — a bare non-quoted late reply with no echo is the irreducible model-dependent tail. The honest ceiling: the deterministic **core** (the echo/quote paths) is now *proven*, the framework recovers everything it owns, and the one case it cannot is **alarmed, not silent**.

## Test plan

- [x] `answer-thread-resolve.test.ts` total-enumeration proof — 18 tests
- [x] `multitopic-routing-wiring.test.ts` — reverse-index population + eviction parity + chat-scope guard + resolve order + residual alarm pinned (22 tests)
- [x] `tsc --noEmit` clean
- [x] 185 gateway vitest + 52 routing/obligation bun tests green
- [ ] Supergroup UAT (`fuzz-multitopic-routing-channel`, `fuzz-cross-surface-ordering-channel`) on the General-topic harness
- [ ] Live marko canary: tail `reply-route` telemetry for `via=` distribution + any `MISROUTE_RISK` / `QUOTED` before fleet roll

## Risk

MEDIUM, kill-switched. The pure-resolver proof is zero-risk (additive, pure fn). The framework-origin recovery only fires on the absent-echo branch (today's buggy path), echo-first so an obedient model is unaffected; chat-scoped so it can never leak a cross-chat thread (worst case it declines to recover → today's behaviour). The alarm is telemetry-only.

### DIFF
--- a/telegram-plugin/gateway/answer-thread-resolve.test.ts
+++ b/telegram-plugin/gateway/answer-thread-resolve.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest'
-import { resolveAnswerThreadId } from './answer-thread-resolve.js'
+import { resolveAnswerThreadId, type AnswerThreadInput } from './answer-thread-resolve.js'
 
 describe('resolveAnswerThreadId — precedence', () => {
   it('(1) explicit model thread wins over everything', () => {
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `gateway.ts:217-227` (findTurnByQuotedMessageId) — verified correct and safe. Resolves origin from the framework-stamped `reply_to` message id via the reverse index, kill-switched (`FRAMEWORK_ORIGIN_ROUTING_ENABLED`), non-numeric/absent guarded, and chat-scoped (`turn.sessionChatId !== chatId` ⇒ null). `CurrentTurn.sessionChatId` is `string` (base 1702) and both call sites pass a string (`chat_id` / `String(args.chat_id)`), so the scope comparison is type-sound. Worst case it declines to recover and falls through to today's live-turn behaviour — it can never emit a wrong-chat thread.
- `gateway.ts:183-200` (rememberRecentTurn eviction) — correct. `const evicted = recentTurnsById.get(oldest)` is captured BEFORE the delete, and the reverse entry is dropped only when it still points at the evicted turn (`=== oldest`), so a newer turn that reused the same message id keeps its mapping. Index can't outgrow `RECENT_TURNS_MAX`. `CurrentTurn.sourceMessageId` confirmed `number|null` (base 1708).
- `gateway.ts:328-330,346-349` — origin precedence is `echo ?? quoted`: `quotedTurn` is only computed when `echoedTurn == null`, so an obedient (echoing) model path is byte-unchanged. Strictly additive fallback on the absent-echo branch (today's buggy path).
- `gateway.ts:258,267,335,354` — the new `originVia` param ('echo'|'quoted'|null) is threaded through `resolveAnswerThreadWithLog` and BOTH call sites pass it (`echoedTurn != null ? 'echo' : 'quoted'`); it only relabels `via` (origin→quoted), never alters `threadId`. Routing is unchanged by the label.
- `gateway.ts:281-284,306-315` (MISROUTE_RISK / hasDifferentThreadedRecentTurn) — observability only; fires on `via==='live'` when a different-threaded recent turn exists for the chat, and does not touch the returned thread. Note: it only catches the bleed when the live successor has a non-null thread (via='live'); a fall-to-General successor reads as 'recovered'/'none', so that sub-case isn't alarmed. Acceptable — it's telemetry, not routing.
- `answer-thread-resolve.test.ts:35-149` — total enumeration over all 64 inputs with an independent `specExpected` reference (separate from the impl, so a shared regression surfaces as divergence). INV-2 flip-immunity (origin resolved ⇒ output independent of liveThreadId) is the real Brevo→Meta guarantee and is asserted for every live/lastEnded combo.

Rationale: additive, kill-switched (`SWITCHROOM_FRAMEWORK_ORIGIN_ROUTING`, independent of `SWITCHROOM_TURN_ORIGIN_ROUTING`) origin recovery from a framework-owned anchor, with the echo path unchanged, a verified cross-chat collision guard, parity-evicted index, and the irreducible no-echo/no-quote tail alarmed rather than silently misrouted. All referenced `CurrentTurn` fields and `findTurnByOriginId` confirmed present in base `main`; signature change applied consistently at both call sites. Author reports tsc clean + 185 gateway + 52 routing/obligation tests green.

Watch (non-blocking): the two supergroup UAT boxes and the live marko canary (`via=` distribution + any MISROUTE_RISK/QUOTED) are unchecked in the test plan — run those before the fleet roll, but the code is correct and safely kill-switched for merge.

PR #2178, opened 2026-06-05
fix(telegram): stop the answer-stream flash (decouple visible-preview from draft retirement)

mekenthompson:
## Summary

Fixes the recurring **"unformatted message → formatted message → unformatted one deleted"** flash users see on most turns. It's a fleet-wide regression where one fix silently undid another, plus proper behavioural regression coverage.

## Why

The flash is the **visible answer-stream**: a raw preview the agent posts as it streams, retracted (deleted) when the formatted `reply` tool fires. v0.14.52 (#2128) turned that surface **OFF by default** *specifically to remove this flash*. v0.14.68 (#2173) retired the invisible compose-box draft transport but wired the preview-open condition as `(ANSWER_STREAM_VISIBLE_ENABLED || DRAFT_ANSWER_LANE_RETIRED)` — and the draft is retired **by default**, so a visible preview opened on every streaming turn and got retracted. **v0.14.68 re-opened the flash v0.14.52 removed**, fleet-wide, with the visible flag still off.

Live evidence (current logs): `clerk` 190, `marko` 128, `gymbro` 92 `answer-stream: retracted preliminary` events.

## What

**The fix** — decouple the two flags. The visible preview gates on `ANSWER_STREAM_VISIBLE_ENABLED` alone; draft retirement controls only the **transport**. With defaults the lane is **dormant** (no preview, no draft) → the `reply` tool is the single formatted message → no flash. No-reply text-only answers still deliver via the turn-flush backstop (the pre-v0.14.68 path).

**The decision is now a pure function** — `resolveAnswerLaneConfig` (in `answer-stream-flag.ts`); the gateway delegates to it at all three use-sites so they can't drift back into the conflation.

## Regression coverage

- **Total-enumeration** of `resolveAnswerLaneConfig` — all 4 flag combos + the load-bearing invariant `opensVisiblePreview === visibleEnabled` (the exact property v0.14.68 broke).
- **End-to-end behavioural test** — the resolved dormant config wired into `createAnswerStream` → a 2000-char answer sends **nothing** (no preview → no flash).
- **Wiring guards** — the gateway delegates to the function; a global negative assertion that `VISIBLE || RETIRED` is gone. (The previous wiring test had *asserted the bug* — it pinned the conflation as a "guard"; rewritten.)

## Verification

Adversarial 5-lens verification (lost-answer, flash-gone, silence-poke, opt-in-visible, kill-switch-draft) → **SHIP**, `lostAnswerSafe: true`, no blockers. The lost-answer path (the severe failure mode) is proven: a no-reply answer is delivered by turn-flush, never retracted.

## Test plan

- [x] `tsc --noEmit` clean
- [x] 240 gateway/answer-stream vitest + 55 bun tests green (both runners)
- [x] total-enumeration + end-to-end dormant-config behavioural tests
- [ ] canary: confirm `answer-stream: retracted preliminary` drops to **zero** on test-harness/marko

## Risk

LOW. Behaviour-equivalent on the opt-in-visible and kill-switch-draft paths; the default path goes from "flash" to "single clean reply". Kill switches unchanged (`SWITCHROOM_VISIBLE_ANSWER_STREAM`, `SWITCHROOM_DRAFT_ANSWER_LANE`).

### DIFF
--- a/telegram-plugin/answer-stream-flag.ts
+++ b/telegram-plugin/answer-stream-flag.ts
@@ -35,3 +35,72 @@ export function parseDraftLaneRetiredEnabled(raw: string | undefined): boolean {
   const v = raw.trim().toLowerCase()
   return !(v === '0' || v === 'false' || v === 'off' || v === 'no')
 }
+
+/**
+ * `minInitialChars` sentinel meaning "never open a visible chat-timeline
+ * preview" — mirrors the `Number.MAX_SAFE_INTEGER` gate the createAnswerStream
+ * call site uses so the lane stays silent.
+ */
+export const ANSWER_LANE_NEVER_OPENS = Number.MAX_SAFE_INTEGER
+
+export type AnswerLaneState = 'visible' | 'draft' | 'dormant'
+
+export interface AnswerLaneConfig {
+  /** `minInitialChars` for createAnswerStream: `1` opens a visible preview on
+   *  the first text chunk; `ANSWER_LANE_NEVER_OPENS` suppresses it. */
+  minInitialChars: number
+  /** Whether the lane streams to the invisible compose-box draft transport. */
+  usesDraftTransport: boolean
+  /** Whether a USER-VISIBLE chat-timeline preview opens — i.e. the surface that
+   *  flashed (raw preview → formatted reply → preview deleted). This is THE
+   *  regression invariant: it must equal `visibleEnabled`, never depend on the
+   *  draft flag. */
+  opensVisiblePreview: boolean
+  /** Label for the boot log. */
+  state: AnswerLaneState
+}
+
+/**
+ * Resolve the answer-lane config from the two INDEPENDENT inputs.
+ *
+ * The visible PREVIEW (the flash surface) is gated on `visibleEnabled` ALONE;
+ * draft retirement controls only the TRANSPORT (whether `sendMessageDraft` is
+ * available). Conflating them was the v0.14.68 regression: retiring the draft
+ * (the default) forced a visible preview that flashed on every streaming turn,
+ * re-opening the flash v0.14.52 had removed. The load-bearing invariant —
+ * `opensVisiblePreview === visibleEnabled` for EVERY `draftFnAvailable` — is
+ * what this function exists to make total-enumerable (the gateway IIFE is not).
+ *
+ *   visibleEnabled                       → 'visible'  (preview opens, minChars 1)
+ *   !visible, draft transport available  → 'draft'    (no preview; draft renders)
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

Verified against the real diff and base main. This is a correct decoupling fix — the v0.14.68 regression (`ANSWER_STREAM_VISIBLE_ENABLED || DRAFT_ANSWER_LANE_RETIRED`) really did force a visible preview on every streaming turn with the visible flag off, since the draft is retired by default.

- `answer-stream-flag.ts:51-77` — `resolveAnswerLaneConfig` is a pure total function over `{visibleEnabled, draftFnAvailable}` (4 states). The invariant `opensVisiblePreview === visibleEnabled` for every `draftFnAvailable` is exactly the property v0.14.68 broke, and it is now total-enumerated in the test. Correct.
- `gateway.ts:4589` (new `ANSWER_LANE`) reads `sendMessageDraftFn != null`; that fn is declared at base `gateway.ts:703`, well above this use-site — no TDZ. Confirmed.
- `gateway.ts:~9691` createAnswerStream config: the three resolved branches match the old behavior table exactly — visible→{minInitialChars:1,no draft}, dormant→{MAX,no draft}, draft→{sendMessageDraft,MAX}. No behavior change on the opt-in-visible or kill-switch-draft paths.
- Lost-answer path holds (the severe failure mode): the materialize-as-answer guard (base `gateway.ts:9971-9982`) is re-gated on `ANSWER_LANE.opensVisiblePreview` alone. In the dormant default it is false, so the branch is unreachable, `streamFinalizedAsAnswer` stays false (base `gateway.ts:9966/9984`), and `flushDecision` falls through to `decideTurnFlush` (base `gateway.ts:10089-10091`) — the no-reply answer is delivered by the turn-flush backstop, never retracted. In dormant `streamedMsgId` is also null, so both guards converge on the safe path.
- Test rewrite is honest: the old `draft-retirement-wiring.test.ts` asserted the conflation as a "guard" (i.e. pinned the bug); the new global negative assertion `not.toMatch(/ANSWER_STREAM_VISIBLE_ENABLED \|\| DRAFT_ANSWER_LANE_RETIRED/)` ensures no path re-introduces it. Plus an end-to-end behavioural test wiring the actual dormant resolver output into createAnswerStream and asserting zero sends.

Kill switches (`SWITCHROOM_VISIBLE_ANSWER_STREAM`, `SWITCHROOM_DRAFT_ANSWER_LANE`) unchanged. Boot log now reports the true 3-state lane instead of mislabeling dormant as visible(draft-retired).

Watch (non-blocking): the canary checkbox — confirm `answer-stream: retracted preliminary` actually drops to zero on test-harness/marko post-merge. Thats the live proof the flash is gone.

PR #2180, opened 2026-06-05
feat(agents): fleet-wide fact-grounding + anti-AI-tell voice

mekenthompson:
## Summary
Two behaviors the operator wants across **all** agents:

1. **Fact-grounding** — agents assert volatile facts (CRM numbers, status, prices) from stale memory/training instead of validating live. Adds a `GROUND BEFORE YOU ASSERT` block to the per-turn `turnPacingDirective` (reaches every agent fresh each turn, incl. custom-CLAUDE.md ones). Ordering-phrased ("pull it THIS turn, then state it"), not a weak "verify" nudge, per the post-rationalization research.

2. **Anti-AI-tell voice** — two layers:
   - *Prompt:* a `VOICE` line in the same directive (no sycophancy openers, no AI-tell filler, lead with the answer, disagree plainly).
   - *Deterministic:* extend the existing gateway `scrubVoice` (already strips em/en dashes on every outbound reply) to also strip a **leading** sycophancy opener and recapitalize. Conservative — message start only, the pure-affirmation set only, only when real content follows (standalone ack survives). Mid-sentence hype stays prompt-side (the prior semantic-loss reasoning is preserved in the header).

## Design notes
- Both prompt behaviors ride the **per-turn** directive (higher salience than the static fleet-invariants wall, and universal). This is the survivor of a 3-angle adversarial review that killed a heavier reply-gate-hook design (turn-boundary races, false-positive regex, user-strand-on-block).
- `scrubVoice.replaced` now totals dash + opener changes so the gateway's `replaced > 0` apply-gate and the metrics fan-out keep working unchanged; new `openersStripped` field is the breakdown. Kill switch unchanged (`SWITCHROOM_DISABLE_VOICE_SCRUB`).
- Honest ceiling: the grounding line is prompt-tier (real but limited). The structural follow-up, if needed, is provenance-wrapping the data MCP tools — out of scope here.

## Test plan
- [x] `text-voice-scrub` +9: opener strip + recap, curly apostrophe / "you are" form, trailing em-dash consumed, **standalone ack preserved**, **bare "you're right" preserved**, mid-message affirmation preserved, code-region safe, kill switch
- [x] `scaffold` +1: turn-pacing directive carries grounding + voice contract
- [x] `tsc --noEmit` + `check-bun-test-imports` clean
- [x] Fuzzy MTCute UAT `fuzz-voice-scrub-dm` — drives real inbounds engineered to bait both tells, asserts neither reaches the user (runs at canary)

## Risk
Low. Prompt additions are additive text. The scrub extension is conservative, deterministic, code-region-aware, kill-switched, and the opener set deliberately excludes load-bearing phrases. No gateway.ts edit (call sites work unchanged via the `replaced` total). Ships in the agent image + gateway; takes effect on fleet roll.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -4146,6 +4146,20 @@ export function buildSettingsHooksBlock(p: HooksBlockParams): Record<string, unk
     'already replied. Your answer is the last thing the user should ' +
     'see; a follow-up "Done." is dead-air clutter (and the user\'s ' +
     'device already pinged on the answer). Stop after the answer.\n\n' +
+    'GROUND BEFORE YOU ASSERT. Any fact in your reply that can change ' +
+    '(a number, a status, a price, a date, who-uses-what, anything ' +
+    '"current" or "latest") must come from a source you actually checked ' +
+    'THIS turn: your data tool, a file, the web. Memory and what you ' +
+    '"already know" are leads to verify, not sources. If you have not ' +
+    'checked it this turn, do not state it as fact: go get it now, or tell ' +
+    'the user you will confirm and then do it. A confident wrong number is ' +
+    'worse than "let me check".\n\n' +
+    'VOICE: write like a sharp colleague, not a chatbot. Do not open with ' +
+    'affirmation ("You\'re absolutely right", "Great question", "Great ' +
+    'catch", "Exactly!"); just answer. Skip AI-tell filler ("smoking ' +
+    'gun", "delve", "it\'s worth noting", "a testament to", "in today\'s ' +
+    'fast-paced..."). Lead with the answer, plain words, kept short. When ' +
+    'the user is wrong, say so directly; flattery is not help.\n\n' +
     'CRITICAL: "answer" means a call to the reply tool ' +
     '(mcp__switchroom-telegram__reply, or stream_reply with done=true). ' +
     'Your terminal/transcript text is NEVER delivered to Telegram — the ' +
--- a/telegram-plugin/runtime-metrics.ts
+++ b/telegram-plugin/runtime-metrics.ts
@@ -124,7 +124,7 @@ export type RuntimeMetricEvent =
    * losing ground; a per-agent spike is prompt drift on that agent.
    *
    *   chatKey   → `<chatId>:<threadIdOrEmpty>` (statusKey shape)
-   *   replaced  → count of dashes rewritten in this single message
+   *   replaced  → total voice changes in this message (dash rewrites + leading-affirmation strips)
    *   site      → which reply path saw the scrub (executeReply / edit / answer-stream)
    */
   | {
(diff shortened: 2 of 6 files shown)


mekenthompson:
REQUEST_CHANGES

- `telegram-plugin/text-voice-scrub.ts:95` (`LEADING_AFFIRMATION_RE`) + `:103` (`stripLeadingAffirmation`): the strip mangles an opener that flows into a continuation clause into a broken sentence fragment. Empirically (ran the actual regex against head):
  - `"You're absolutely right that the config drifted."` → `"That the config drifted."`
  - `"You're absolutely right to worry about races."` → `"To worry about races."`
  - `"Great catch that you noticed the leak."` → `"That you noticed the leak."`
  - `"Spot on about the timing issue."` → `"About the timing issue."`
  - `"Exactly right because the token expired."` → `"Because the token expired."`
  This is the SAME load-bearing hazard the header (`:99-102`) already calls out for bare `"you're right"` — but `"...absolutely right that X"`, `"...to X"`, `"...about X"`, `"...because X"` for the whole affirmation set is not guarded. The standalone-ack guard at `:108` only catches an empty remainder, not a fragment-producing continuation. Result: by default the gate ships grammatically broken wire text to the user on a common AI phrasing — arguably worse than the tell it removes.
  Fix: only strip when the affirmation is a complete clause, i.e. require it to be terminated by punctuation rather than a bare space-into-continuation. Tighten the trailing class so the strip fires on `[,.!?;:—–]` separators but NOT when a word follows directly (or add a negative lookahead for a continuation word: `(?![\s]*(that|to|about|because|on|regarding|when|since|so)\b)`). Add the `"...right that X"` / `"...catch that X"` cases to `text-voice-scrub.test.ts` as preserved.

Everything else is sound and ships as-is once the above is guarded:
- `src/agents/scaffold.ts:4146` grounding + VOICE additions are additive prompt text on the per-turn directive; verified present and asserted in `tests/scaffold.test.ts:2505`.
- `text-voice-scrub.ts:239` `scrubVoice` correctly folds the opener count into the `replaced` total so the gateway `replaced > 0` apply-gate and metrics fan-out keep working on an opener-only strip (`:50-57` test confirms `replaced === 1`).
- Ordering is right: park code/URLs → strip opener → dashes → restore, so the opener pass is code-region safe and consumes a trailing em-dash before the dash pass (`:60-63` test confirms).
- Kill switch `SWITCHROOM_DISABLE_VOICE_SCRUB` preserved across both transforms (`:94-100` test).
- Standalone ack and bare `"you're right that X"` correctly preserved (`:70-80` tests).

Rationale: the prompt half is fine, but the deterministic scrub is the half that actually mutates user-visible text, and on the common `"You're absolutely right that ..."` pattern it emits a broken fragment. Guard the continuation case + add coverage, then this is an APPROVE.

PR #2182, opened 2026-06-06
feat(voice): prompt-first tone, stop context-free word deletion by default

mekenthompson:
## Summary
Operator steer: agent tone should be carried by the agent *understanding* it (prompt, context-aware), not by a hook blindly deleting words. Context-free removal is bad UX (it can strip a sincere "good catch, that was my bug" along with hollow praise).

- **Sycophancy-opener strip is now opt-in, OFF by default** (`SWITCHROOM_VOICE_STRIP_OPENERS=1` re-enables the backstop). By default the deterministic layer removes no words.
- **Em-dash normalization stays deterministic** — it substitutes punctuation, removes no content, and prompt-only was measured to fail at it (em-dashes in 73% of replies despite the SOUL rule). The prompt now also asks for plain punctuation so both layers align.
- **The turn-pacing VOICE directive becomes context-aware**: skip hollow openers and AI-tell filler, but genuine acknowledgement that adds something is fine. The model has context the hook never will, so tone lives there.

## Why this split
The AI-tell that is pure punctuation (em-dash) is safe to enforce deterministically. The AI-tell that is about judgment and context (sycophancy/tone) belongs in the prompt, where the model can keep what is genuine and drop what is hollow. This is the "deterministic for what is provably safe, prompt for what needs context" line the whole codebase already follows.

## Test plan
- [x] `text-voice-scrub` 34: opener tests opt-in via the flag; new default-off block proves openers preserved while em-dashes still fixed
- [x] `scaffold` 189: VOICE directive assertion updated (context-aware)
- [x] `tsc` + `check-bun-test-imports` clean
- [x] Fuzz UAT keeps the deterministic em-dash assertion (hard), downgrades the opener check to a soft warn (now probabilistic)

## Risk
Low and reversible. The opener-strip code is unchanged, just gated behind a default-off flag. Em-dash behavior is unchanged. The prompt change is additive text. Kill switch `SWITCHROOM_DISABLE_VOICE_SCRUB` still disables the whole scrub.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -4154,12 +4154,16 @@ export function buildSettingsHooksBlock(p: HooksBlockParams): Record<string, unk
     'checked it this turn, do not state it as fact: go get it now, or tell ' +
     'the user you will confirm and then do it. A confident wrong number is ' +
     'worse than "let me check".\n\n' +
-    'VOICE: write like a sharp colleague, not a chatbot. Do not open with ' +
-    'affirmation ("You\'re absolutely right", "Great question", "Great ' +
-    'catch", "Exactly!"); just answer. Skip AI-tell filler ("smoking ' +
-    'gun", "delve", "it\'s worth noting", "a testament to", "in today\'s ' +
-    'fast-paced..."). Lead with the answer, plain words, kept short. When ' +
-    'the user is wrong, say so directly; flattery is not help.\n\n' +
+    'VOICE: write like a sharp colleague, not a chatbot. Lead with the ' +
+    'answer, plain words, plain punctuation (commas and periods, not ' +
+    'em-dashes). Skip the hollow openers ("You\'re absolutely right", ' +
+    '"Great question", "Great catch", "Exactly!") and AI-tell filler ' +
+    '("smoking gun", "delve", "it\'s worth noting", "a testament to", "in ' +
+    'today\'s fast-paced..."). Genuine acknowledgement is fine when it is ' +
+    'real and adds something ("good catch, that was my bug"); what to ' +
+    'avoid is the reflexive praise that opens every reply and means ' +
+    'nothing. When the user is wrong, say so directly; flattery is not ' +
+    'help.\n\n' +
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- scaffold.ts:15-24 — VOICE directive rewritten context-aware: still names the hollow openers and AI-tell filler to skip, but now permits genuine acknowledgement ("good catch, that was my bug") and explicitly asks for plain punctuation (no em-dashes). Matches the prompt-carries-tone intent; no contradiction with the deterministic layer.
- text-voice-scrub.ts:158-161 — new openerStripEnabled() gates the word-deleting opener strip behind SWITCHROOM_VOICE_STRIP_OPENERS=1; default off. Correct: the opener-strip code (stripLeadingAffirmation) is untouched, just no longer invoked by default.
- text-voice-scrub.ts:175-177 — scrubVoice now does `openerStripEnabled() ? stripLeadingAffirmation(parked) : {out: parked, count: 0}`. replaceDashes still runs unconditionally, so em/en-dash normalization remains a hard guarantee while no words are removed by default. Logic is sound.
- text-voice-scrub.test.ts:36-94 — existing opener tests correctly opt in via the flag in beforeEach; new default-off block proves openers preserved (openersStripped===0) while em-dashes still fold to commas (replaced===1). Coverage matches the behavior change on both sides.
- fuzz-voice-scrub-dm.test.ts:222-235 — opener invariant downgraded from throw to console.warn, em-dash check stays a hard assert. Right call: a prompt-tier lever is probabilistic and would flake as a gate; the deterministic transform stays enforced.
- scaffold.test.ts:249-253 — assertions updated to the new prompt text (Skip the hollow openers / Genuine acknowledgement is fine / plain punctuation). Consistent with scaffold.ts.

Rationale: clean deterministic-vs-prompt split — punctuation (provably content-safe) stays enforced in the hook, tone (needs judgment) moves to the prompt. Behavior change is gated behind a default-off flag with the opener code intact, em-dash behavior unchanged, and the global SWITCHROOM_DISABLE_VOICE_SCRUB kill switch still present. Tests cover both flag states. Low-risk and fully reversible.

PR #2185, opened 2026-06-06
feat(web): manage agent↔account access from the dashboard (approval-gated)

mekenthompson:
## Summary

UI Phase B: the **Connections** tab gets per-agent access toggles for Google and Microsoft accounts — *"give marko access to lisa@outlook.com"* from the dashboard. Builds on the read-only Connections tab (#2172).

## Security — the load-bearing design

The dashboard does **NOT** write config. An agent on host networking can **forge** the dashboard's loopback auth (it trusts `Tailscale-User-Login` on presence + loopback source, and agents run `network_mode:host`), so a direct-write grant endpoint would be a **self-elevation hole**. Instead a toggle **proposes a unified diff to hostd** over the operator socket; hostd validates it, raises an **Allow/Deny card in the operator's Telegram** (out-of-band — an agent can't forge the tap), and is the **sole writer** (`git apply` on approval). Nothing changes without the human tap. *This is the leash* (vision pillar 2).

> Caught during this work by the operator's question: *"what would stop an eager agent from doing this without approval?"* — the answer was "nothing" for a direct write, which is why this routes through the approval card.

## What changed

- **`config/agent-workspace-account.ts`** — comment-preserving editor for the per-agent `<provider>_workspace.account` selector (set/clear/get). The broker needs **both** the `enabled_for` ACL and this selector, so a grant writes both.
- **`web/config-edit-plan.ts`** — read → apply the ACL+selector transforms → schema-validate the result (same gate the loader runs), **no write**. Rejects a broken edit before raising a card.
- **`web/config-diff.ts`** — generate a git-compatible unified diff with `git diff --no-index` (the same tool family hostd's `git apply` consumes), normalizing path headers to `a/<name>`/`b/<name>` so hostd's `-p1` strip lands on its scratch basename. **Round-trip-tested against hostd's exact `git apply --recount --unidiff-zero` flags.**
- **`web/hostd-config-propose.ts`** — submit `config_propose_edit` to the hostd **operator** socket (resolved from the web container's `/host-home` mount or `~/.switchroom` — no infra change), 11-min timeout (> hostd's 10-min approval window), map completed/denied/error.
- **`api.ts` `handleSetConnectionAccess`** — orchestrates plan → diff → **background** hostd propose (the POST returns immediately; hostd blocks on the human tap) + a pending-status map; `handleGetConnectionAccessStatus` polls it (reaped at 30 min). Fresh-config reload so reads reflect an applied edit.
- **`ui/index.html`** — per-agent enable/disable pills on each account card; on click → *"check Telegram → approve"* → polls status → on applied, offers to restart the agent (it scaffolds the MCP at boot).

## Feasibility (verified live, no infra change)

The `switchroom-web` container (uid 1000) reaches the hostd operator socket (uid 1000, mode 0600, `SWITCHROOM_HOSTD_OPERATOR_UID=1000`) under its `~/.switchroom` mount; `host_control` + `hostd.config_edit_enabled` are on.

## Test plan

- [x] **90 tests** across the touched suites: pin editor (set/clear/get/idempotent/empty-block-drop); **config-diff round-trips through hostd's exact git-apply flags** + empty-on-no-op + ConfigDiffError; plan (schema-valid / no-op / rejects-invalid-and-unparseable, no write); propose (completed/denied/error mapping, never-throws, socket resolution); handler (bad-input rejection, no-op→no-card, **propose+poll→applied with config-unchanged-on-disk asserted**, denial via poll, hostd-unreachable).
- [x] `tsc --noEmit` + full lint chain green; UI inline JS parses.

## Risk

Moderate (touches the web server + config plane) but mitigated by design: the web server **cannot write config** — hostd is the sole writer and every grant requires the operator's Telegram tap. The plan is schema-validated before a card is raised; the diff round-trips through hostd's exact apply flags. No new dependency (uses the system `git`).

## Deferred (operator's call: "just my feature for now")

The existing dashboard write endpoints (`/api/auth/use`, agent restart) share the forgeable-header weakness — flagged for a separate hardening pass.

### DIFF
--- a/src/config/agent-workspace-account.test.ts
+++ b/src/config/agent-workspace-account.test.ts
@@ -1,0 +1,62 @@
+import { describe, expect, it } from "vitest";
+import {
+  setAgentWorkspaceAccount,
+  clearAgentWorkspaceAccount,
+  getAgentWorkspaceAccount,
+} from "./agent-workspace-account.js";
+
+const BASE = `agents:
+  marko:
+    extends: default
+  ziggy:
+    extends: default
+`;
+
+describe("setAgentWorkspaceAccount", () => {
+  it("creates the workspace block + account selector under a known agent", () => {
+    const after = setAgentWorkspaceAccount(BASE, "microsoft", "marko", "lisa@outlook.com");
+    expect(getAgentWorkspaceAccount(after, "microsoft", "marko")).toBe("lisa@outlook.com");
+    // Other agents untouched.
+    expect(getAgentWorkspaceAccount(after, "microsoft", "ziggy")).toBeNull();
+    // Comment-preserving editor keeps the rest of the file.
+    expect(after).toContain("ziggy:");
+  });
+
+  it("is idempotent (byte-equal when already set to the same value)", () => {
+    const once = setAgentWorkspaceAccount(BASE, "google", "marko", "a@b.com");
+    const twice = setAgentWorkspaceAccount(once, "google", "marko", "a@b.com");
+    expect(twice).toBe(once);
+  });
+
+  it("throws for an undeclared agent", () => {
+    expect(() => setAgentWorkspaceAccount(BASE, "microsoft", "nope", "x@y.com")).toThrow(
+      /not declared/,
+    );
+  });
+});
+
+describe("clearAgentWorkspaceAccount", () => {
+  it("removes the selector and drops the now-empty workspace block", () => {
+    const set = setAgentWorkspaceAccount(BASE, "microsoft", "marko", "lisa@outlook.com");
+    const cleared = clearAgentWorkspaceAccount(set, "microsoft", "marko");
+    expect(getAgentWorkspaceAccount(cleared, "microsoft", "marko")).toBeNull();
+    expect(cleared).not.toContain("microsoft_workspace");
+  });
+
+  it("keeps the workspace block when it carries other keys", () => {
+    const withOrg = `agents:
+  marko:
+    microsoft_workspace:
+      account: lisa@outlook.com
+      org_mode: true
+`;
+    const cleared = clearAgentWorkspaceAccount(withOrg, "microsoft", "marko");
+    expect(getAgentWorkspaceAccount(cleared, "microsoft", "marko")).toBeNull();
+    expect(cleared).toContain("org_mode: true");
+  });
+
+  it("is a no-op for an unset selector / unknown agent", () => {
+    expect(clearAgentWorkspaceAccount(BASE, "microsoft", "marko")).toBe(BASE);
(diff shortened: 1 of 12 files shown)


mekenthompson:
APPROVE

Security design is the right one and it holds up under reading:
- The dashboard never writes config. `handleSetConnectionAccess` (`src/web/api.ts:275-399`) computes + schema-validates the edit (`planConfigEdit`, `src/web/config-edit-plan.ts:704`), builds a unified diff, and routes it to hostd over the **operator** socket; hostd is the sole writer and raises the Telegram Allow/Deny card. The forgeable loopback-auth self-elevation hole is correctly avoided — an agent on host networking can't forge the out-of-band operator tap. The body's reasoning matches the code.
- Diff generation round-trips through hostd's exact apply flags: `generateUnifiedDiff` (`src/web/config-diff.ts:516`) uses `git diff --no-index` and normalizes headers to `a/<name>`/`b/<name>` so hostd's `-p1` strip lands on its scratch basename; `config-diff.test.ts:434` asserts the patch applies under `git apply --whitespace=nowarn --recount --unidiff-zero -p1`. Correct.
- Both halves of the grant are written: `enabled_for` ACL + the per-agent `<provider>_workspace.account` selector (`agent-workspace-account.ts`), composed in `api.ts:308-323`. Disable only clears the per-agent pin if it points at THIS account (`api.ts:316-321`) — won't disturb a pin to a different account. Good.
- Input validation is present: provider allowlist, action allowlist, agent-must-exist, and an email regex that excludes `@\s:` (`api.ts:286-300`). The status map is bounded by a 30-min reaper for settled entries (`api.ts:239`); pending entries always settle (11-min client timeout > hostd's 10-min window, plus a `.catch`).
- `captureEvent`/`captureException` are already imported (`api.ts:34`); no missing-import break. CI: build-dist, lint-run, bun-test, vitest-shard(4) green at review time; remaining vitest shards pending — let them finish before merge.

Nits (non-blocking):
- `src/web/ui/index.html:1190-1193` — the `onclick='toggleAccess(, , , ...)'` embeds JSON.stringify'd account/agent values into a single-quoted attribute without HTML-escaping. Account emails and agent names are operator-controlled config (not attacker input) and the write path validates them, so the risk is theoretical, but a value containing a single quote or `<` would break the attribute. Consider escapeHtml on the JSON for defense-in-depth.
- `hostd-config-propose.ts:943` hardcodes `target_path: /state/config/switchroom.yaml` while the diff is generated from the web container's `configPath`. If those ever diverge in content, `git apply` fails closed (E_PATCH_APPLY_FAILED → state error), so it's safe, but it's an implicit coupling worth a comment.

Rationale: correct, well-tested (90 tests across the touched suites), fail-closed approval-gated design with the human tap as the sole authority; the only findings are a theoretical UI-escaping nit and a documented path coupling, neither blocking.

PR #2190, opened 2026-06-06
fix(hindsight): set --shm-size so PostgreSQL doesn't die on 64MB default

mekenthompson:
## Summary
Hindsight launched with no `--shm-size`, so it got Docker's 64MB default `/dev/shm`. Its embedded PostgreSQL needs far more (observed: a single ~533MB shared-segment request), so every such allocation failed with `could not resize shared memory segment ... No space left on device` and **all memory writes/queries died fleet-wide** (2026-06-06 outage; the session-long HTTP 502s).

## Why
The live host was hot-fixed by recreating the container with `--shm-size=2g`, but `src/setup/hindsight.ts` had no shm flag in **either** launch path, so the next `switchroom apply`/`update`/`setup` would recreate hindsight at the broken 64MB default and re-break everything. This makes the fix durable.

## Changes
- Add `HINDSIGHT_DEFAULT_SHM_SIZE = "2g"` (clears the observed peak with headroom; stays under the 4g mem cap so shm can't starve the app).
- Apply it to the standalone `docker run` args **and** the compose snippet (kept in sync).

## Test plan
- [x] `vitest run tests/setup/hindsight.test.ts` — 19 pass, incl. 2 new (run-arg `--shm-size`, compose `shm_size`, both assert default >=1g so it can't silently regress)
- [x] `tsc --noEmit` clean

## Risk
Low — additive flag on container launch; no behavior change beyond giving PostgreSQL the shared memory it needs. Serves the **always-available** outcome (memory backend stays up across recreations).

JTBD: agents that remember (Hindsight is the memory backend the fleet depends on).

### DIFF
--- a/src/setup/hindsight.ts
+++ b/src/setup/hindsight.ts
@@ -180,6 +180,21 @@ export const HINDSIGHT_DEFAULT_MEM_RESERVATION = "2g";
 export const HINDSIGHT_DEFAULT_PIDS_LIMIT = 1000;
 
 /**
+ * `/dev/shm` size for the hindsight container.
+ *
+ * Hindsight's embedded PostgreSQL allocates large shared-memory segments
+ * for parallel query workers / large sorts (observed: a single segment
+ * request of ~533MB). Docker's DEFAULT shm is only 64MB, so without this
+ * flag every such allocation fails with `could not resize shared memory
+ * segment ... No space left on device` and ALL memory writes/queries die
+ * (2026-06-06 fleet outage). 2g gives comfortable headroom over the
+ * observed peak while staying well under `HINDSIGHT_DEFAULT_MEM_LIMIT`
+ * (4g) so shm can't starve the app. Applies to BOTH the standalone
+ * `docker run` path and the compose snippet below — keep them in sync.
+ */
+export const HINDSIGHT_DEFAULT_SHM_SIZE = "2g";
+
+/**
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `src/setup/hindsight.ts:345` (run path) and `:443` (compose snippet): both launch paths now emit `--shm-size=2g` / `shm_size: 2g` from the single `HINDSIGHT_DEFAULT_SHM_SIZE` constant (`hindsight.ts:179-194`). These are the only two container-launch sites (`execFileSync("docker", args, ...)` at the run path and the compose generator), so the 64MB-default regression is closed in both. Correct root-cause fix: it makes the live hot-fix durable across `apply`/`update`/`setup`.
- Sizing is sound: 2g clears the observed ~533MB peak with headroom and stays under `HINDSIGHT_DEFAULT_MEM_LIMIT` (4g) so shm can't starve the container.
- Regression guards are real: `tests/setup/hindsight.test.ts:95` asserts the run-arg `--shm-size` AND that the default is >=1g (blocks a silent revert to 64MB); `:226` asserts the compose snippet emits `shm_size`, keeping the two paths in sync.
- Build + bun-test green; vitest shards pending but the change is additive and the new assertions are deterministic.

Additive container-launch flag fixing a confirmed fleet-wide memory outage; both launch paths covered and guarded by tests. Safe to merge once vitest lands.

PR #2191, opened 2026-06-06
feat(memory): memory.file gate to retire curated MEMORY.md (hindsight-native)

mekenthompson:
## Summary
Adds a per-agent `memory.file` boolean (default `true` = no behavior change). When `false`, the scaffold stops seeding the curated workspace `MEMORY.md`, so once its content is migrated into Hindsight and the file deleted, the delete **sticks** across reconcile (today MEMORY.md is seed-once → a bare delete gets re-created as an empty template).

## Why
switchroom runs **two** parallel memory systems: Hindsight (auto-recall + auto-retain hooks + agent `sync_retain`) and a curated `MEMORY.md` that's seeded once and auto-loaded every turn. The architecture goal is to converge on Hindsight as the single backend ("hindsight-native"). This is the mechanism that lets us retire the file per-agent after migrating its content into Hindsight (directives for standing rules + auto-synthesized mental models + retained facts).

The per-turn loader already skips absent files gracefully (`loadNamedFile` → `missing: true`), so only the **seed** path needed gating.

## Changes
- `memory.file` field on `AgentMemorySchema` (override cascade, documented).
- `seedWorkspaceBootstrapFiles` skips `MEMORY.md` (both `.hbs` and plain) when `memory.file: false`.
- Caller threads `agentConfig.memory?.file !== false`.

## Test plan
- [x] `vitest run tests/scaffold.test.ts` — 193 pass, incl. 2 new (gate skips MEMORY.md + siblings intact + default still seeds; gate survives reconcile)
- [x] config schema tests — 44 pass
- [x] `tsc --noEmit` clean

## Risk
Low — additive, opt-in, default `true`. No existing agent changes behavior. Serves the **REMEMBER = one backend** box of the agent-context architecture (single source of truth, no file/bank divergence).

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -1520,6 +1520,13 @@ function seedWorkspaceBootstrapFiles(params: {
   created: string[];
   skipped: string[];
   rewrittenWithBackup: string[];
+  /**
+   * When false (agent `memory.file: false`), do NOT seed the curated
+   * workspace MEMORY.md. Hindsight-native agents carry memory in the bank
+   * (directives + mental models + retained facts), so seeding the file
+   * here would re-create it after a migration delete. Defaults to seeding.
+   */
+  seedMemoryFile?: boolean;
 }): void {
   const profileWorkspaceDir = join(params.profilePath, "workspace");
   if (!existsSync(profileWorkspaceDir)) {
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `src/config/schema.ts:139-145`: `memory.file` boolean added to `AgentMemorySchema` with `.default(true)`, so existing agents are unchanged. Documented cascade (override, per-agent wins). Correct.
- `src/agents/scaffold.ts:1549-1556`: the gate `params.seedMemoryFile === false && relPath.replace(/\.hbs$/, "") === "MEMORY.md"` is precise. It matches both `MEMORY.md` and `MEMORY.md.hbs` at the workspace root only; daily notes (`memory/YYYY-MM-DD.md`) contain a slash so they're never caught. Skip is recorded in `skipped[]`, no silent drop.
- `src/agents/scaffold.ts:3148`: caller threads `agentConfig.memory?.file !== false`, so the gate is engaged only on explicit `false` (undefined/true both seed). Matches the schema default.
- Seed-path-only is the right scope: the per-turn loader already handles an absent file (`loadNamedFile` → `missing: true`), so a migrated-and-deleted MEMORY.md now stays deleted across reconcile instead of being re-seeded as an empty template.
- Tests cover the three cases that matter: gate skips MEMORY.md while siblings (USER/IDENTITY/SOUL) survive, default still seeds, and the gate holds across a re-scaffold (`tests/scaffold.test.ts:1153,1181`). bun-test + changes green; lint/build pending but additive.

Additive, opt-in, default-true gate that closes the seed-once re-creation gap for hindsight-native agents. Safe to merge once remaining checks land.

PR #2192, opened 2026-06-06
feat(doctor): memory-backend health check (catch silent hindsight outages)

mekenthompson:
## Summary
Adds `switchroom doctor` checks for the two **silent** hindsight failure modes that an MCP-reachability probe can't see — both of which took down fleet memory for an entire session on 2026-06-06 before anyone noticed.

## Why
The existing check confirms hindsight is reachable + speaking MCP. But the container can be up and serving MCP while every write silently fails:
1. **shm exhaustion** — PostgreSQL needs ~533MB+ shared memory; Docker's 64MB default → `No space left on device` on every write.
2. **OAuth quota 429** — fact-extraction `claude` calls hit the weekly limit → retains accepted but 0 facts extracted → nothing recallable.

Neither shows over the MCP port; both show in the container's shm config + recent logs.

## Changes
- `src/cli/doctor-memory.ts`: pure classifiers `classifyShmSize` (fail < 1g) and `classifyExtractionLogs` (shm exhaustion / LLM-provider error with 0 successes / sustained 0-fact runs), plus a best-effort docker wrapper that **silently skips** when hindsight isn't a local container (remote URL / no docker / inside an agent container).
- Wired into `checkHindsight` under the existing "Memory (Hindsight)" doctor section.

## Test plan
- [x] `vitest run tests/cli.doctor.memory-health.test.ts` — 11 pass, classifiers tested against this session's **real** log + shm samples (64MB shm, the `claude_code_llm` error + `Extract facts: 0 facts`, healthy `5 facts`)
- [x] `tsc --noEmit` clean

## Risk
Low — read-only diagnostics, best-effort, no-ops on remote/dockerless setups. Serves "always available" (vision #4): memory outages get caught by `doctor` instead of by users noticing their agents went amnesiac.

### DIFF
--- a/src/cli/doctor-memory.ts
+++ b/src/cli/doctor-memory.ts
@@ -1,0 +1,154 @@
+/**
+ * Memory-backend health checks for `switchroom doctor`.
+ *
+ * The existing hindsight checks (`checkHindsight`) confirm the backend is
+ * *reachable and speaking MCP* — but the 2026-06-06 outage proved that's not
+ * enough: the container was up and serving MCP while every memory write
+ * silently failed. Two distinct failure modes hid behind a healthy-looking
+ * port:
+ *
+ *   1. **shm exhaustion** — PostgreSQL needs ~533MB+ of shared memory for
+ *      large query/sort segments; Docker's 64MB default made every write fail
+ *      with `could not resize shared memory segment ... No space left on
+ *      device`. (Now prevented at launch by --shm-size, #2190 — but a
+ *      pre-#2190 container or a hand-rolled one can still hit it.)
+ *   2. **OAuth quota exhaustion** — hindsight's fact-extraction `claude` calls
+ *      hit the account's weekly limit (429), so retains were accepted but
+ *      extracted 0 facts → nothing became recallable.
+ *
+ * Neither is visible over the MCP port. Both ARE visible in the container's
+ * shm config and recent logs. These checks surface them so the next outage is
+ * caught by `doctor`, not by a user noticing their agents went amnesiac.
+ *
+ * The pure classifiers below are unit-tested against real log/shm samples; the
+ * docker wrapper is best-effort and silently skips when hindsight isn't a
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `src/cli/doctor-memory.ts:46-64` `classifyShmSize`: clean threshold at `MIN_HINDSIGHT_SHM_BYTES` (1 GiB). Fails below, ok at/above (tested at the exact floor, `test:215-218`). Correctly catches the 64MB default that caused the 2026-06-06 outage.
- `src/cli/doctor-memory.ts:70-123` `classifyExtractionLogs`: the `okFacts === 0` guard on both the `llmError` (fail) and `zeroFacts >= 3` (warn) branches is the right call, it prevents a single transient extraction error amid healthy runs from flipping the check red (asserted at `test:250-260`). shm-exhaustion takes precedence and points back at the shm check. Sound.
- `src/cli/doctor-memory.ts:130-160` `checkHindsightContainerHealth`: best-effort docker wrapper fails closed to `[]` on any inspect error (remote URL / no docker / inside an agent container), and skips just the extraction check if `logs` fails while keeping the shm result. stderr ignored, 8s timeout. No way for the probe itself to fail the doctor run.
- `src/cli/doctor.ts:1013`: wired into `checkHindsight` via spread under the existing Memory section, after the reachability + consumer checks. Correct placement, additive.
- Nit (non-blocking): `doctor-memory.ts:153` uses a fixed `--since 10m` log window. Auto-retain extraction is bursty (~every 10 turns), so on a quiet fleet the window can hold zero `Extract facts:` lines and the check reports "no recent extraction activity" (ok) even while a 429 is live. It catches a sustained outage but can miss an intermittent one during idle periods. Consider widening the window or persisting last-success time. Not a merge blocker.

Read-only diagnostics, pure classifiers unit-tested against the real outage's log + shm samples, best-effort wrapper no-ops on remote/dockerless setups. build-dist + bun-test green, vitest/lint pending. Safe to merge.

PR #2194, opened 2026-06-06
feat(auth-broker): fail consumers over to fallback_order on quota exhaustion

mekenthompson:
## Summary
Gives **consumers (hindsight) the same quota-failover agents already have.** When a consumer's pinned `consumers[].account` is within a mark-exhausted window, `callerAccount` now fails over to the first healthy account in `fallback_order` (reverting automatically when the window passes).

## Why
Agents survive a quota-exhausted account because they ride the swappable `auth.active` (gateway detects 429 → `mark-exhausted` → broker rolls active). Consumers were pinned **unconditionally** — so when hindsight's account quota-exhausted, it went dead (retains accepted, 0 facts extracted) until a manual repoint. That **was** the 2026-06-06 outage.

This is the answer to "can hindsight leverage the same failover mechanism": **yes** — it just needed the broker to honor exhaustion for consumers. Everything else already exists:
- **Exhaustion state** (`this.quota`) — set by `mark-exhausted`, which an agent *sharing* the account already raises.
- **fallback_order** — reused as-is.
- **Propagation** — hindsight's existing background cred-refresh loop re-fetches `get-credentials`, so the swap reaches it with no restart.

## Changes
- `callerAccount` consumer branch → `consumerAccountWithFailover` (+ `isAccountExhausted` helper). Only the consumer path changes; operator/agent paths untouched.

## Test plan
- [x] `vitest run src/auth/broker/server.test.ts` — 49 pass, incl. 3 new (healthy→pinned, exhausted→fallback, no-healthy→stays pinned)
- [x] `tsc --noEmit` clean

## Risk
Low + scoped to consumers. Same trust boundary (all `fallback_order` accounts are the operator's). Availability > isolation only while the pinned account is down. **Follow-up:** a *dedicated* consumer account no agent shares needs `opProbeQuota` on a proactive timer to raise exhaustion; the shared-account path works today.

### DIFF
--- a/src/auth/broker/server.test.ts
+++ b/src/auth/broker/server.test.ts
@@ -373,6 +373,47 @@ describe("AuthBroker — consumer failover on quota exhaustion", () => {
     expect(resp.data.account).toBe("secondary");
     broker.stop();
   });
+
+  it("a consumer's mark-exhausted attributes to its PINNED account, never the failover view", async () => {
+    // Invariant (schema.ts): mark-exhausted from a consumer only affects ITS
+    // account. Failover must NOT leak into attribution — else a consumer being
+    // SERVED a failover account could mark that healthy account exhausted and
+    // cascade the fleet off it. This test fails if get-credentials' failover
+    // view bleeds into mark-exhausted.
+    const h = makeHarness();
+    const config = makeConfig(h, {
+      active: "default",
+      fallback_order: ["default", "secondary"],
+      consumers: [{ name: "hindsight", account: "secondary" }],
+      agents: { marker: { auth: { override: "secondary" } } },
+    });
+    seedAccount(h, "default");
+    seedAccount(h, "secondary");
+    mkdirSync(join(h.agentsDir, "marker"), { recursive: true });
+    const broker = new AuthBroker(config, {
+      home: h.home, stateDir: h.stateDir, socketRoot: h.socketRoot, disableRefreshLoop: true,
+    });
+    await broker.start();
+    // Exhaust secondary (via an agent on it) → consumer failover is now active.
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `consumerAccountWithFailover` (server.ts:130-137) correctly mirrors the existing agent-failover path (server.ts:1984-1985): same `q.exhausted_until > this.now()` exhaustion test, same `readAccountCredentials(cand, this.home)` credential gate, same `fallback_order` iteration. No new state, no new trust boundary.
- `isAccountExhausted` (server.ts:119-122) matches the inline checks already used in `opListState` (server.ts:952-953) — consistent semantics, no off-by-one on the window comparison.
- Scope is precise: only the `identity.kind === "consumer"` branch changes (server.ts:98-110); operator and agent paths untouched (server.ts:97, 112-115). Skips the pinned account as a candidate and skips other exhausted candidates, so it can't fail over to a dead account.
- Degrades safely: returns the pinned account when no healthy fallback has stored creds (server.ts:136) — retry beats serving nothing. Tests cover all three cases: healthy→pinned, exhausted→fallback, no-healthy→stays pinned (server.test.ts:10-86).
- Reversion is automatic: once `exhausted_until` passes, `isAccountExhausted` returns false and the consumer reverts to its pinned account with no manual repoint.

Rationale: minimal, well-scoped reuse of the proven agent-failover mechanism that closes the consumer-pinning gap behind today's hindsight outage; build-dist/bun-test green, vitest/e2e pending. Noted in the PR body: a dedicated consumer account no agent shares still needs a proactive quota probe to raise exhaustion — the shared-account path works today.

PR #2195, opened 2026-06-06
feat(memory): accept memory.file at the defaults tier (fleet-default hindsight-native)

mekenthompson:
## Summary
Adds `file` to the **defaults/profile** memory schema (it was added to the per-agent schema in #2191 but not the parallel defaults one). This lets a single `defaults.memory.file: false` make the whole fleet — and future agents — hindsight-native by default, with per-agent `memory.file: true` as the opt-in.

## Why
After the hindsight-native migration (#2191 + the per-agent rollout), the only file-based memory left is the **empty-template MEMORY.md on the 5 stub agents** + whatever a *new* agent would seed. A fleet default cleanly retires both. The defaults memory schema already has `collection` optional, so `defaults.memory.file: false` needs no collection — just the `file` field.

## Changes
- `file: z.boolean().optional()` on the defaults/profile memory schema (schema.ts ~1471).

## Test plan
- [x] `vitest run src/config/merge.test.ts` — 15 pass, incl. 2 new (agent inherits `defaults.memory.file:false`; per-agent `file:true` opts back in)
- [x] `tsc --noEmit` clean

## Risk
Minimal — additive optional field; no behavior change until an operator sets `defaults.memory.file`. Completes the hindsight-native architecture (single source of truth, file opt-in).

### DIFF
--- a/src/config/merge.test.ts
+++ b/src/config/merge.test.ts
@@ -49,6 +49,21 @@ describe("mergeAgentConfig — release cascade", () => {
   });
 });
 
+describe("mergeAgentConfig — memory.file cascade (fleet-default hindsight-native)", () => {
+  it("an agent inherits defaults.memory.file:false when it sets no memory.file", () => {
+    const defaults = { memory: { file: false } } as AgentDefaults;
+    const result = mergeAgentConfig(defaults, baseAgent());
+    expect(result.memory?.file).toBe(false);
+  });
+
+  it("a per-agent memory.file:true opts back in over the fleet default", () => {
+    const defaults = { memory: { file: false } } as AgentDefaults;
+    const agent = baseAgent({ memory: { collection: "x", file: true } } as Partial<AgentConfig>);
+    const result = mergeAgentConfig(defaults, agent);
+    expect(result.memory?.file).toBe(true);
+  });
+});
+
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- Adds `file: z.boolean().optional()` to the defaults/profile memory schema (schema.ts:1435-1438), mirroring the per-agent `AgentMemorySchema.file` already added in #2191 (schema.ts:139). The two schemas are intentionally parallel since AgentSchema does not spread profileFields, so this fills the documented gap.
- Cascade is correct: `mergeAgentConfig` does a per-field shallow merge of `memory` (merge.ts:388-401) — it seeds from `defaults.memory` then overrides only with agent fields that are defined (skips `undefined`). So `defaults.memory.file:false` inherits when the agent sets nothing, and a per-agent `file:true` overrides it. No special-casing needed beyond what `recall` already has.
- Tests cover both directions: inherit-default-false and per-agent-opt-back-in-true (merge.test.ts:9-22).
- Purely additive optional field, no behavior change until an operator sets `defaults.memory.file`.

Rationale: minimal, correct completion of the hindsight-native migration (#2191) — the schema field lands at the tier the cascade already supports, with tests proving inheritance both ways. build-dist/bun-test green, vitest/lint pending.

PR #2197, opened 2026-06-06
fix(vault): correct secret-guard block message + doctor check for IDs-stored-as-secrets

mekenthompson:
## Summary

Closes the rabbit hole behind the Google Ads / marko incident: a **public identifier stored as a vault secret** that the secret-guard PreToolUse hook then blocked from ever appearing in a tool call — while telling the agent to *"reference it as `vault:<key>` instead"*, a mechanism that **does not exist** for tool arguments. The agent could neither pass the literal (hook blocked it) nor a `vault:` ref (it reached Google verbatim → `INVALID_CUSTOMER_ID`), and confabulated auth/token causes for hours.

Three changes, all addressing *how the system responds* so it self-corrects instead of misleading:

1. **Hook message (the highest-leverage fix).** `secret-guard-pretool.mjs` no longer suggests a phantom `vault:` resolver. The block reason now states the two real resolutions: a genuine secret should never be a tool argument (the launcher env-injects it); a value you must pass literally (an account/customer ID) is a *public identifier* and shouldn't be vaulted at all.

2. **`vault doctor` check (proactive).** New `identifier-stored-as-secret` warn flags vault values that look like public identifiers (≥8 digits, optionally dash/space-grouped) — the exact footgun. The shape is classified in the CLI layer so **raw secret values never enter the pure analysis layer**.

3. **Fleet guidance (`VAULT_GUIDANCE`).** Tells every agent: never pass a secret value or a `vault:` placeholder as a tool argument; if a call is blocked for containing a vaulted secret, report the verbatim message and stop — don't theorise about auth/tokens.

## Why

`vault:` *does* resolve for config values (e.g. `bot_token`), but **not** for MCP tool arguments — so the hook's old advice was a dead end. The miscategorisation (public ID in the vault) is the root cause; #1 + #2 make it cheap to detect and impossible to chase the wrong way. See the marko Google Ads incident.

## Test plan

- [x] `src/vault/doctor.test.ts` (new, 5 tests) — heuristic true/false cases + the check fires only on identifier-shaped keys
- [x] `secret-guard-pretool.test.ts` — asserts the new message + **absence** of the old `reference it as vault:` phrase; still blocks real secrets, still fails open
- [x] scaffold tests (198) green — fleet-guidance rendering unaffected
- [x] `tsc --noEmit` clean · `check-plugin-references.mjs` clean

## Risk

Low. Hook behaviour is unchanged (still blocks the same inputs) — only the human-readable `reason` string changed. The doctor check is purely additive (new warn, defaults off when no value is identifier-shaped). Guidance is prompt-only.

## JTBD

Serves *"you hold the leash"* / predictability: the secret guard still protects tokens, but stops sending agents (and the operator) down a phantom path when it fires.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -286,7 +286,18 @@ name from that list.
 Only call the \`vault_request_access\` MCP tool — which pings the
 operator for a Telegram approval — for a key you've **confirmed via
 \`vault list\` you don't already have.** Requesting access to a guessed
-or already-held key wastes the operator's tap and fails.`;
+or already-held key wastes the operator's tap and fails.
+
+**Never put a secret's value — or a \`vault:<key>\` placeholder — in a
+tool call.** Tool arguments are sent to the underlying API verbatim;
+nothing resolves a \`vault:\` reference for them. A real secret never
+needs to appear in a tool call — the launcher injects it into the
+tool's environment for you. A value you genuinely must pass literally
+(an account or customer ID) is a public identifier, not a secret, and
+belongs in plain config, not the vault. If a tool call is **blocked for
+containing a vaulted secret**, report that exact block message to the
+operator and stop — don't theorise about auth or tokens; it just means
+a stored value reached a tool argument.`;
 
 /**
  * Canonical rendering of the fleet invariants file written to
--- a/src/cli/vault-doctor.ts
+++ b/src/cli/vault-doctor.ts
@@ -19,7 +19,7 @@
 import type { Command } from "commander";
 import chalk from "chalk";
 import { existsSync } from "node:fs";
-import { analyseVaultHealth, type Diagnostic, type DiagnosticLevel } from "../vault/doctor.js";
+import { analyseVaultHealth, looksLikeIdentifierValue, type Diagnostic, type DiagnosticLevel } from "../vault/doctor.js";
 import { openVault, VaultError } from "../vault/vault.js";
 import { loadConfig, resolvePath } from "../config/loader.js";
 import { statusViaBroker, resolveBrokerSocketPath } from "../vault/broker/client.js";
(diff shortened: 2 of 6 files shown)


mekenthompson:
APPROVE

- Hook message fix (`secret-guard-pretool.mjs:194-233`): drops the phantom `reference it as vault:${hit.key}` advice (no `vault:` resolver exists for tool arguments) and replaces it with the two real resolutions — launcher env-injection for genuine secrets, plain config for public identifiers. Behavior unchanged; only the model-read `reason` string changes. Correct.
- Heuristic `looksLikeIdentifierValue` (`doctor.ts:157-164`): length 8–24, digit-bounded `^[0-9][0-9 -]*[0-9]$`, `digitCount >= 8`. The `>= 8` gate matches the hook's `MIN_VALUE_LENGTH_TO_GUARD = 8` (confirmed in `secret-guard-pretool.mjs`), so it only flags values the guard would actually block. Letters/symbols/slashes correctly excluded.
- Clean layering (`vault-doctor.ts:107-118`): value shape-classified in the CLI via `looksLikeIdentifier`, only the boolean reaches `analyseVaultHealth` — raw secret values never enter the pure analysis layer. Good.
- New warn `identifier-stored-as-secret` (`doctor.ts:187-207`) is purely additive: emits only when a key is identifier-shaped, names the offending keys, and the `fix` is actionable. Does not fire when no value matches.
- Fleet guidance (`scaffold.ts:286-297`) and the test asserting absence of the old phrase (`secret-guard-pretool.test.ts:241-253`) are consistent with the hook change.

Rationale: targeted fix for the IDs-stored-as-secret footgun — corrects misleading hook copy, adds a proactive doctor check with safe length bounds, all additive and well-tested (5 new doctor tests + updated guard test). build-dist + changes green, rest pending; no failures.

PR #2198, opened 2026-06-06
feat(scaffold): enable tool-search to defer the ~31k-token MCP schema surface

mekenthompson:
## Summary

The context audit found the per-turn floor is ~63k tok (~32% of a 200k window), and the biggest single slice is the **MCP tool-schema surface (~31k tok across 10–11 servers)** — bigger than all of switchroom's authored prompt text combined, and mostly unused on any given turn.

This enables the unmodified `claude` CLI's **`ENABLE_TOOL_SEARCH`** (confirmed present in the pinned 2.1.159 binary): MCP tool schemas are **deferred** — surfaced by name, full JSON fetched only when the model actually calls the built-in ToolSearch — reclaiming ~25–29k tok of window. It's a flag *on* the native CLI, so it stays inside the Claude-native compliance pillar.

## What

- `buildToolSearchEnvVars()` → `ENABLE_TOOL_SEARCH` default **`auto`** (self-gating: only defers once the tool surface exceeds ~10% of context — every agent's does; small-surface agents untouched). Override `SWITCHROOM_TOOL_SEARCH_MODE`; kill switch `SWITCHROOM_DISABLE_TOOL_SEARCH=1`. Spread into **both** `userEnvQuoted` sites (buildScaffold + reconcile), before `agentConfig.env` so per-agent config wins.
- **`alwaysLoad: true`** pinned on the load-bearing framework servers — **switchroom-telegram** (reply/stream_reply/get_recent_messages/react), **hindsight** (recall/retain), **agent-config**. Without this, `reply` defers and every turn pays a ToolSearch round-trip before it can answer (the orphaned-reply failure class). The heavy integration servers (webkite/perplexity/marketing) deliberately omit it → they defer = the win.
- `McpServerConfig` gains optional `alwaysLoad?: boolean`.

## Why this design

Per the audit's prompt-caching finding, the fixed stance is mostly cache-hot so this is **window-relief** (fewer auto-compactions, more working room) more than per-turn dollars — but the dollar win is real on cold-miss/restart turns. The tool surface was the right lever because it's the largest slice and lazy-loading it changes no prompt text.

## Test plan

- [x] tsc clean; 6 tests (default/override/kill-switch + the alwaysLoad pins + env-order lockstep)
- [ ] **test-harness canary** (the gate): confirm tool-search activates AND `reply` / `mcp__hindsight__recall` / `get_recent_messages` still resolve — always-on, fast-trivial, real-work UAT + a live DM round-trip.

## Risk

MEDIUM, kill-switched. Tool search is a CLI beta; the `alwaysLoad` pins protect the framework tools. Worst case: `SWITCHROOM_DISABLE_TOOL_SEARCH=1` + restart reverts to today's all-tools-loaded behavior. Per-turn ToolSearch round-trips add latency only on integration-heavy turns.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -1413,6 +1413,31 @@ function buildHumanizerEnvVars(
 }
 
 /**
+ * Claude Code tool-search env. Defers (lazy-loads) MCP tool schemas so the
+ * heavy per-agent integration servers (webkite/perplexity/marketing — ~31k
+ * tokens of schema) stop occupying the context window every turn; the model
+ * fetches a schema only when it actually calls that tool. The load-bearing
+ * framework servers (switchroom-telegram, hindsight, agent-config) carry
+ * `alwaysLoad: true` in .mcp.json so reply/get_recent_messages/recall never
+ * defer (no per-turn ToolSearch round-trip).
+ *
+ * Default value `auto` (self-gating: claude only defers once the tool surface
+ * exceeds ~10% of the window, which every agent's does — a small-surface agent
+ * is left untouched). Override per fleet via SWITCHROOM_TOOL_SEARCH_MODE
+ * (e.g. `true` to force always-defer, `auto:N` for an N% threshold). Kill
+ * switch SWITCHROOM_DISABLE_TOOL_SEARCH=1 omits the var entirely → claude
+ * loads all tools (today's behaviour). A per-agent `env: {ENABLE_TOOL_SEARCH}`
+ * in switchroom.yaml still wins (spread after this in the env builder).
+ *
+ * Consumed by claude (inner start.sh pass), not the gateway, so the
+ * env-before-gateway-fork landmine does not constrain placement.
+ */
+export function buildToolSearchEnvVars(): Record<string, string> {
+  if (process.env.SWITCHROOM_DISABLE_TOOL_SEARCH === "1") return {};
+  return { ENABLE_TOOL_SEARCH: process.env.SWITCHROOM_TOOL_SEARCH_MODE ?? "auto" };
+}
+
+/**
  * Top-level settings.json keys that switchroom's scaffold/reconcile
  * pipeline owns and rebuilds on every run. When the settings_raw
  * escape hatch injects additional top-level keys (e.g. `effort`,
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- `buildToolSearchEnvVars()` (`scaffold.ts:1429-1432`): kill switch `SWITCHROOM_DISABLE_TOOL_SEARCH=1` returns `{}` (omits the var → legacy all-tools-loaded), else `ENABLE_TOOL_SEARCH` defaults to `auto` with `SWITCHROOM_TOOL_SEARCH_MODE` override. Correct and well-bounded.
- Env order verified at both sites (`scaffold.ts:2135` buildScaffold, `scaffold.ts:4647` reconcile): `...buildToolSearchEnvVars()` is spread BEFORE `...(agentConfig.env ?? {})`, so a per-agent `env.ENABLE_TOOL_SEARCH` still wins. Test guards both occurrences stay in lockstep (`scaffold.tool-search.test.ts:191-200`).
- `alwaysLoad: true` pinned on the load-bearing framework servers: switchroom-telegram (`scaffold.ts:2996-3001`/`5105-5110`), agent-config (`scaffold.ts:3013-3017`/`5122-5126`), hindsight (`scaffold-integration.ts:130`). Heavy integration servers correctly omit it so they defer — that is the win.
- Survives serialization: `filterMcpServers` (`scaffold.ts:1480`) strips only `secrets:` via rest-spread, so `alwaysLoad` reaches `.mcp.json`. `McpServerConfig.alwaysLoad?: boolean` added (`hindsight.ts:114`).
- Watch (not a blocker): the entire mechanism depends on the pinned `claude` binary actually honoring `config.alwaysLoad` and `ENABLE_TOOL_SEARCH=auto`. That cannot be verified from this repo; the PR asserts it for 2.1.159 and the test-harness canary in the test plan is still UNCHECKED. If `alwaysLoad` is not honored, `reply`/`recall` defer and every turn pays a ToolSearch round-trip (the orphaned-reply class). Run the canary before merge.
- Nit: the source-guard tests (`scaffold.tool-search.test.ts:202-217`) regex-match scaffold.ts text rather than asserting emitted config — brittle to reformatting, but acceptable given the canary covers behavior.

Rationale: clean, kill-switched window-relief change; reclaims the ~31k-token MCP schema slice by deferring integration servers while pinning the framework tools loaded, no prompt text changes. Internally consistent and tested. build-dist + bun-test green, rest pending, no failures. The CLI-honors-alwaysLoad assumption is the only real risk and it is reversible via the kill switch.

PR #2199, opened 2026-06-06
fix(context): memory-prose contradiction, delete-tool bug, true debug-turn budget

mekenthompson:
## Summary

Three of the four context-optimization items from the audit (the fourth, tool-search, shipped in #2198). All low-risk: a correctness fix, a dead-prose removal, and instrumentation.

### #3 — remove the file-memory contradiction (correctness)
The workspace operating protocol (`profiles/default/workspace/CLAUDE.md.hbs`, what AGENTS.md/AGENT.md symlink to, in the system prompt every turn) told agents to "maintain MEMORY.md + memory/YYYY-MM-DD daily files" — **directly contradicting** the cwd CLAUDE.md + fleet invariants ("Hindsight is your single backend, file-memory disabled"). Agents got opposite memory instructions every turn. Memory guidance now defers to the authoritative carriers and points at hindsight. AGENTS.md/AGENT.md stay as symlinks (loader reads them by name; switchroom is claude-CLI-only so they're inert). *Seed-once file → fixes new agents + source of truth; existing fleet needs a one-time re-seed (rollout).*

### #2 — agents call a non-existent delete tool (correctness)
The guidance told agents to call `mcp__hindsight__delete_memory` for forget/correct — but the server has **no such tool** (the real one is `delete_document`; verified live against the 29 server tools). Every forget/correction silently failed with "Unknown tool". Renamed (3 sites). The broader tool-surface *trim* was evaluated and **deferred** (documented): the global allowlist would break host-side mental-model/vault-sweep ops, and the token win is already captured by tool-search (#2198).

### #4 — true per-turn-floor accounting in `debug turn`
The Totals block understated the floor ~3x: double-counted SOUL.md (already in the stable prefix), omitted the --add-dir fleet invariants, omitted the MCP tool surface, used a /4 divisor. Now: drop the SOUL double-count, add fleet invariants, add an honest MCP-surface line (+ "agent-private" note when 0600-blocked), `/3.7` token estimate, and a "Per-turn FLOOR" total. On marko: authored text ~20,180 tok, floor ~63k tok — matches the audit.

## Test plan
- [x] tsc clean; 14 tests (debug floor accounting incl. SOUL-double-count guard; rendered workspace doc is hindsight-only; guidance names delete_document not delete_memory)
- [ ] Rollout: re-seed existing agents' workspace/CLAUDE.md (the #3 seed-once file) — bundled with the tool-search canary.

## Risk
LOW. #3/#2 are prompt-text/correctness; #4 is read-only instrumentation. No behavioral runtime change; no kill switch needed.


mekenthompson:
APPROVE

- Delete-tool fix (`scaffold.ts:58,65,69`): renames `mcp__hindsight__delete_memory` → `delete_document` across the correct/forget/inspect guidance. Verified against the live hindsight toolset — `delete_document` exists, `delete_memory` does not, so the old guidance produced silent "Unknown tool" failures on every forget/correction. Correct.
- Memory-prose contradiction (`CLAUDE.md.hbs:9-34,42-45`): the workspace protocol no longer tells agents to maintain MEMORY.md + memory/ daily files, now defers to Hindsight + the cwd CLAUDE.md. Removes the every-turn contradiction with the fleet invariants. `memory_search`/`memory_get` reference swapped for `mcp__hindsight__recall`. Consistent.
- SOUL double-count fix (`debug.ts:142-149`): `soulMdBytes` dropped from `totalBytes` (it already lives inside `stableResult.concatenated`), now shown as an informational sub-line (`debug.ts:153`). `fleetBytes` (--add-dir invariants) added (`debug.ts:131-149`). Both guarded by source-grep tests (`debug.turn-floor.test.ts:290-297`).
- MCP-surface accounting (`debug.ts:100-114,183-200`): `readMcpServerNames` reads `.mcp.json` without spawning servers, returns `[]` when absent and `null` when present-but-unparseable (0600 agent-private) — correctly distinguished from a misleading 0. Imports (`existsSync`/`readFileSync`/`join`) all present (`debug.ts:2-3`). `/3.7` divisor replaces `/4`, honestly labelled an estimate. Read-only.
- Deferred-trim note (`setup/hindsight.ts:211-230`): documents why the 29-tool allowlist is NOT applied (global filter would break host-side mental-model/vault-sweep callers) and that tool-search (#2198) already captures the token win. Good rationale, no code change.

Nit: the Totals/floor guard tests regex-match debug.ts source rather than asserting rendered output — brittle to refactors, but the behavioral estimateTokens/readMcpServerNames tests cover the logic.

Rationale: a correctness fix (real delete tool), a dead-prose removal (memory contradiction), and read-only instrumentation. No runtime behavior change. Well-tested (14 tests). build-dist + bun-test-run + evals green, rest pending, no failures. Note the unchecked rollout item: existing agents need a one-time workspace/CLAUDE.md re-seed for #3 to reach the live fleet (seed-once file).

PR #2215, opened 2026-06-07
feat(quota-watch): fleet-wide all-exhausted operator alert

mekenthompson:
## Summary

A fleet-wide **all-accounts-exhausted** alert — the one piece of operator visibility the failover trilogy (#2206/#2209/#2212) didn't cover.

The existing trigger-based all-blocked card only fires when an **agent** 429s into the wall. It misses two now-real cases:
- a **quiet period** — the fleet goes all-exhausted but no agent happens to hit it, so you're never told;
- the **consumer (hindsight)** and **cron** paths added in v0.14.81/82, which degrade *silently* (hindsight serves an exhausted account; crons defer to the JSONL audit).

The broker is a separate container with no Telegram path, so it can't alert itself. `quota-watch` already polls broker `listState`, classifies health, and posts with edge-trigger dedup — but had no fleet-level tier. This adds one.

## Design
- **`evaluateFleetAllExhausted`** (pure, edge-triggered) — fires **once** on entering all-exhausted (with earliest-reset line + self-healing reassurance) and **once** on recovery. Keyed off the broker's **authoritative** per-account `exhausted` flag (set by mark-exhausted via failover + the consumer sensor), **not** probe-derived health — so there is **no probe-failure false alarm**.
- State stored under a reserved key in the existing `quota-watch.json` — backward-compatible, survives the load validator, and the per-account loop (which iterates account snapshots, not state-map keys) never sees it.
- **`runQuotaWatch`** — the fleet check runs **before** the per-account early-return (all-exhausted means the per-account loop is all skips → it would return early), sends via the existing `swallowingApiCall` path, and persists immediately so the edge-trigger doesn't re-fire.

## Re-scope note (important)
This **supersedes** the originally-proposed "split all-blocked vs probe-failed in logs." Post-v0.14.80 the all-blocked decision is **broker-authoritative** (`rolledTo===null`), so that log ambiguity no longer exists — the genuine gap was the missing quiet-period/consumer/cron alert, which this fills. I verified that before building rather than ship the now-redundant log split.

## What it is NOT
Observability only — **no behavior/recovery change.** Failover, defer/retry, and self-healing are all unchanged. Purely additive; reuses the existing poll + dedup + send infra.

## Test plan
- [x] pure: entered / still (no re-spam) / recovered / not-all / empty-fleet / reset-unknown — 6 cases
- [x] runs under **both** vitest AND bun (30 tests in file, 0 fail)
- [x] `tsc --noEmit` clean; `check-bot-api-wrapping.sh` clean (new send wrapped); `check-plugin-references.mjs` clean

## Risk
Very low. New edge-triggered alert on authoritative state; dedup prevents spam; off the hot path; no decision/behavior change. Worst case is a missed alert (fails safe), never a wrong action.

## Rollout
Gateway change → fleet agent restart (sidecar), staggered + version-asserted. Or fold into the next release.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -412,6 +412,8 @@ import {
 } from '../credits-watch.js'
 import {
   evaluateQuotaWatchAccount,
+  evaluateFleetAllExhausted,
+  FLEET_ALL_EXHAUSTED_KEY,
   loadQuotaWatchState,
   saveQuotaWatchState,
   patchQuotaWatchState,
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

Verified the wiring end-to-end against a fresh clone.

- `gateway.ts:14807-14848` — fleet block is correctly placed AFTER the snapshot build / state load but BEFORE the per-account "First pass" and its `if (notifications.length === 0) return` early-return (`gateway.ts:~14891`). The placement claim holds: in the all-exhausted case every snapshot classifies `unknown` → `evaluateQuotaWatchAccount` skips all → `pendingTransitions` empty → early return, so a fleet alert placed after would never fire. Confirmed.
- `quota-watch.ts:98-125` `evaluateFleetAllExhausted` — pure, edge-triggered off the broker's authoritative `exhausted` flag (not probe-derived health), so no probe-failure false alarm. `allExhausted` requires `accounts.length > 0`, so empty fleet never alerts (`quota-watch.ts:104`).
- Account shape confirmed: `AccountState.exhausted` / `exhausted_until` are real broker fields (`auth-command.ts:683`, `786-792`), so `listStateData.accounts` feeds `evaluateFleetAllExhausted` with the right data.
- Reserved-key safety holds: `FLEET_ALL_EXHAUSTED_KEY = "__fleet_all_exhausted__"` (`quota-watch.ts:78`) is encoded as a `QuotaWatchAccountState`, so it survives the load validator (`quota-watch.ts:243-245`), and the per-account loop iterates `snapshots` from `buildSnapshotsFromCachedState`, never state-map keys, so it never collides.
- Persist-on-notify (`gateway.ts:14841-14848`) is correct: `watchState` is reassigned via `patchQuotaWatchState` and saved immediately, so the early-return path can't drop the edge-trigger flag and re-fire next poll. `send` goes through the existing `swallowingApiCall` wrapper with verb `quota-watch.fleet-all-exhausted` (`gateway.ts:14837-14840`).
- Helpers all present: `escapeHtml` (`quota-watch.ts:215`), `emptyAccountState` (`:72`), `patchQuotaWatchState` (`:270`), `formatRelative` (imported). Recovery message escapes the account label; the entered message only shows a count, no labels — no injection surface.
- Tests cover the 6 transitions (entered / still / recovered / not-all / empty / reset-unknown) at `tests/quota-watch.test.ts:180-249`. CI: build-dist + changes green; lint/bun-test/vitest pending, none failing.

Nit (non-blocking): on recovery, both this fleet-recovered alert and a per-account "recovered" notification can fire in the same poll (one account freed). Slightly redundant but harmless — different granularity, and the fleet path persists/sends independently.

Observability-only, additive, off the hot path, fails safe (worst case = a missed alert, never a wrong action). Ship it.

PR #2218, opened 2026-06-07
fix(wedge): detect /rate-limit-options weekly-quota menu → trigger failover

mekenthompson:
## Summary

Durable fix for the rate-limit TUI wedge that left **finn silently dead for hours** (2026-06-07). The claude CLI's **weekly** quota wall surfaces as an interactive `/rate-limit-options` menu, *not* a stderr 429 — so the gateway never sees it, `markExhausted` is never called, and **none** of the v0.14.80-83 failover/alert machinery engages. A headless agent then sits forever on a menu no human will answer.

The generic wedge-watchdog doesn't catch it either: the menu footer is "Enter to confirm · Esc to cancel" with no "to select/navigate/↑↓", so `WEDGE_FOOTER_SIGNATURE` fails to match. finn was **wedged-forever-undetected**.

## The fix (small — reuses the entire existing chain)
1. **Detect** in `wedge-watchdog.ts`: a dedicated branch (checked before the generic modal branch, same stability gate + cooldown). Two-anchor `RATE_LIMIT_MENU_SIGNATURE` (`/rate-limit-options` **and** `Switch to usage credits|Upgrade your plan`) — verified against the live pane, zero false-positive risk.
2. **Parse** the weekly reset (`parseWeeklyReset`, tz-aware via `Intl`).
3. **Signal** the gateway (`quota_wall_detected` IPC verb → `onQuotaWallDetected`) which triggers the **existing** `fireFleetAutoFallback`, threading the parsed reset as `markExhausted`'s `until` (with a **+7d fallback** when unparsed — passing undefined would let the ~5h default un-exhaust a weekly wall and re-wedge). The existing chain rolls to a fallback subscription account, or fires #2215's all-exhausted operator alert.
4. **Park** with Esc.

## Compliance (vision.md pillar 3) — central
Recovery is *only* subscription-account failover or graceful Esc-park. The park is **Esc-only** (cancels without selecting) — a unit test asserts the branch can **never** emit `Down`/`Up`/`Enter`/`2`/`3`, so it can never pick "Switch to usage credits" (off-subscription) or "Upgrade your plan". Esc-dismiss + the exact option wording were **verified against finn's live wedged pane** (the menu literally prints "Esc to cancel").

## Reuses (no parallel machinery)
`markExhausted` · `fireFleetAutoFallback`/`runFleetAutoFallback` (incl. #2206's idempotency guard + single-authoritative-chooser) · the operator-announcement broadcast · #2215 `quota-watch` all-exhausted alert · the wedge-watchdog stability/cooldown/Esc seams · the agent-scheduler IPC trust model.

## Design notes
- **Gateway-signal, not sidecar-direct** `markExhausted` — so it goes through the existing idempotency guard + single-chooser (#2206) instead of recreating the divergent path #2206 deleted.
- **Separate from this code fix:** an agent needs a usable fallback account to actually *roll* vs just park+alert. finn rides shared `auth.active=default` with no fallback — a **config** gap handled in recovery, not this PR.

## Test plan
- [x] tsc clean; 27 vitest (detect / parse incl. AEST→UTC / **exhaustive compliance keystroke** / kill-switch / signal client) + 12 bun (IPC verb validation)
- [ ] Live: recover finn (config fallback + restart) and confirm the detector signals failover on the real menu

## Risk
LOW, kill-switched (`SWITCHROOM_RATE_LIMIT_DETECT=0` → exact v0.14.83 behaviour). Esc-only park means a false positive is harmless (just declines a modal). The signal is fire-and-forget soft-fail. No new model-calling path; the claude CLI stays unmodified on OAuth.

### DIFF
--- a/src/agents/rate-limit-signal.ts
+++ b/src/agents/rate-limit-signal.ts
@@ -1,0 +1,108 @@
+// One-shot gateway signal for the rate-limit (weekly-quota) wedge.
+//
+// When the wedge-watchdog (autoaccept-poll sidecar) detects claude's
+// `/rate-limit-options` menu, it must trigger the gateway's EXISTING account-
+// failover chain (markExhausted → roll → quota-watch alert). The watchdog runs
+// in a separate process from the gateway, so it signals over the gateway's
+// per-agent UDS — the same socket + trust model the agent-scheduler uses
+// (anything inside this container's network namespace may connect; the agent
+// name is validated server-side, never trusted from the wire alone).
+//
+// Fire-and-forget, one connection per signal (the signal fires at most once per
+// watchdog cooldown, ~60s). Soft-fail throughout — must NEVER throw, to honour
+// the sidecar's never-throw contract. A dropped signal just means failover
+// isn't triggered this tick; the next stable detection retries.
+
+import { createConnection, type Socket } from "node:net";
+import { join } from "node:path";
+
+/** The NDJSON envelope written to the gateway socket. Mirrors the
+ *  ipc-protocol `QuotaWallDetectedMessage`. */
+export interface QuotaWallSignal {
+  type: "quota_wall_detected";
+  agentName: string;
+  /** Parsed weekly-reset epoch-ms, or omitted when unparseable (the gateway
+   *  substitutes a weekly-scale default — never the ~5h markExhausted default,
+   *  which would un-exhaust a weekly wall and re-wedge). */
+  resetAt?: number;
+}
+
+export interface SignalOptions {
+  socketPath?: string;
+  connectTimeoutMs?: number;
+  /** Test seam: replace createConnection. */
+  _connect?: (path: string) => Socket;
+  /** Test seam: log sink. */
+  _log?: (msg: string) => void;
+}
+
+/** Resolve the gateway socket the same way the agent-scheduler does. */
+export function resolveGatewaySocketPath(): string {
+  const stateDir = process.env.TELEGRAM_STATE_DIR ?? "/state/agent/telegram";
+  return process.env.SWITCHROOM_GATEWAY_SOCKET ?? join(stateDir, "gateway.sock");
+}
+
+/**
+ * Send a `quota_wall_detected` signal to the gateway. Fire-and-forget,
+ * never throws. Returns a Promise that resolves true if the bytes were written
+ * to the socket (the strongest delivery signal a fire-and-forget client gets),
+ * false on any failure. The watchdog does not await it (so a slow/absent socket
+ * can never stall the poll loop), but it is awaitable for tests.
+ */
+export function signalQuotaWall(
+  agentName: string,
+  resetAt: number | null,
+  opts: SignalOptions = {},
+): Promise<boolean> {
+  const socketPath = opts.socketPath ?? resolveGatewaySocketPath();
+  const connect = opts._connect ?? ((p: string) => createConnection(p));
+  const log = opts._log ?? ((m: string) => console.error(m));
+  const connectTimeoutMs = opts.connectTimeoutMs ?? 5_000;
+
+  const msg: QuotaWallSignal = { type: "quota_wall_detected", agentName };
+  if (resetAt != null) msg.resetAt = resetAt;
+  const line = JSON.stringify(msg) + "\n";
+
+  return new Promise<boolean>((resolve) => {
+    let settled = false;
+    const done = (ok: boolean) => {
+      if (settled) return;
+      settled = true;
+      resolve(ok);
+    };
+    let sock: Socket;
+    try {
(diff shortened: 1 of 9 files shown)


mekenthompson:
APPROVE

Durable, well-scoped fix for the weekly-quota TUI wedge that left finn silently dead. Reuses the entire existing failover chain rather than building parallel machinery; verified end-to-end against the PR HEAD.

- **Detection** (`src/agents/wedge-watchdog.ts:144`) — `RATE_LIMIT_MENU_SIGNATURE` requires BOTH anchors (`/rate-limit-options` AND `Switch to usage credits|Upgrade your plan`) via independent lookaheads; no nested quantifiers, so no backtracking risk. Repo grep confirms zero pre-existing occurrences of either anchor → false-positive risk is effectively nil. Checked before the generic modal branch (`:287-290`) so the menu is signalled before any Esc, with the same stability + cooldown gate.

- **Stability hygiene** (`wedge-watchdog.ts:361-366`) — the trailing `else` resets `stableCount`/`lastKey` when neither branch matches, and both firing branches reset after firing, so no stale streak leaks between the rate-limit and generic paths despite sharing state.

- **Parse** (`wedge-watchdog.ts:160` + `wallClockToEpoch:192`) — tz-aware via `Intl.DateTimeFormat` (DST-correct, not `new Date(localString)`), rolls year forward on a past M/D, returns null on any failure. Tested for AEST→UTC, pm/minutes, no-tz, year-roll, and garbage (`tests/rate-limit-menu-detect.test.ts:686`).

- **Weekly-reset fallback** (`gateway.ts:6271-6284`) — when `resetAt` is absent/invalid, the gateway substitutes `now + 7d` rather than letting `markExhausted`'s ~5h default un-exhaust a weekly-walled account and re-wedge it. Correct and the central correctness subtlety of the PR.

- **Reuse, not reinvention** (`gateway.ts:14633`, `:14683`) — `fireFleetAutoFallback(triggerAgent, untilMs?)` threads `untilMs` into `client.markExhausted(until?)` (`auth-command.ts:224`, `auth-broker-client.ts:30` both accept optional `until`). The existing 429-path caller (`gateway.ts:4403`) passes no `untilMs` → unchanged ~5h behavior. Goes through #2206's idempotency guard + single-chooser instead of the divergent direct path #2206 deleted.

- **Compliance — Esc-only park** (`wedge-watchdog.ts:330`) — the park sends only `["Escape"]`; the exhaustive keystroke test (`tests/rate-limit-menu-detect.test.ts:775`) asserts the branch can NEVER emit Down/Up/Enter/2/3/Tab, so it can never select "Switch to usage credits" (off-subscription) or "Upgrade your plan". This is the load-bearing safety property and it is correctly pinned.

- **IPC trust** (`ipc-server.ts:246`) — `quota_wall_detected` validated server-side: `agentName` must pass `AGENT_NAME_RE`, `resetAt` (optional) must be finite. Rejects path-traversal/uppercase/non-finite payloads (`tests/ipc-server-validate-quota-wall.test.ts`). Signal client (`rate-limit-signal.ts`) is fire-and-forget, never-throws, soft-fails on connect/write/timeout.

- **Kill switch** — `SWITCHROOM_RATE_LIMIT_DETECT=0` reverts to exact v0.14.83 behavior (`autoaccept-poll.ts:373`), tested (`rate-limit-menu-detect.test.ts:760`).

Watch (non-blocking, no change required): for an agent with NO fallback account (finn rides shared `auth.active=default`), Esc parks but the next inference re-hits the wall, so the watchdog re-detects and re-signals every ~60s cooldown indefinitely until the config gap is fixed. The #2215 all-exhausted alert is persist-on-notify (fires once) and #2206's gate should suppress repeat fleet announcements — confirm on the live canary that the 60s re-signal doesn't produce repeat operator alerts when no fallback exists. The author already calls the fallback-account config gap out as separate recovery work.

CI is green on every code gate (vitest ×4 shards, bun-test, lint, build-base, build-dist); build-dependents are docker-image builds, not correctness gates. The one unchecked test-plan item is the live finn canary — a deploy-time validation, not a code-correctness gap. Safe to merge.

mekenthompson:
Reviewer note (non-blocking, for the record): the boot-phase `runAutoaccept` `enter-to-confirm` rule (`/Enter.{1,30}confirm/` → Enter) would match this menu's footer 'Enter to confirm · Esc to cancel'. Not reachable in practice — `runAutoaccept` exits after 30s idle and the weekly wall is a mid-session event days later, by which point only the Esc-only watchdog runs. Even a hypothetical boot-window collision selects the *highlighted* option, which the live pane shows is option 1 'Stop and wait' (on-subscription) — never credits/upgrade. Flagging for completeness; candidate for a defensive follow-up (exclude the rate-limit signature from the boot rule) but out of scope here.

PR #2220, opened 2026-06-07
fix(credits-watch): stop false 'out of pre-paid credits' alarms (subscription-only)

mekenthompson:
## Summary

Agents were posting `⚠️ out of pre-paid credits — cron tasks and inbound replies will fail until resolved · check your subscription or pre-paid usage`. **It's a false alarm.**

The card comes from `credits-watch` reading `cachedExtraUsageDisabledReason=out_of_credits` from `.claude.json`. By its name, that flag is *always* "why **extra usage** (the optional pay-as-you-go layer beyond the Pro/Max plan) is disabled." switchroom is **subscription-only by design** (compliance pillar 3 — the plan is the ceiling, no API/pay-as-you-go), so extra usage being **off is the expected, desired state**. Every reason value (`out_of_credits` / `extra_usage_disabled` / `credits_exhausted` / `org_level_disabled` = "org admin disabled *extra usage*") describes that benign state — **none indicates a real switchroom failure.** The card also advised buying credits, which *contradicts* the subscription-honest model.

**Evidence it was benign:** the flag was cleared fleet-wide, agents kept answering (clerk completed a 6-tool reply right after the card), all accounts healthy.

The genuine failure modes are already covered:
- an account's subscription window exhausts → stderr → **fleet auto-fallback** rolls to a healthy account (#2206);
- the **whole** fleet exhausts → the **all-exhausted alert** (#2215).

So credits-watch's `out_of_credits` alarm is redundant noise on top.

## Change
- `DEFAULT_CREDIT_FATAL_REASONS` is now **empty** (subscription-only: extra-usage-off never alarms).
- `resolveCreditWatchFatalReasons(env)` — **opt back in** via `SWITCHROOM_CREDITS_WATCH_FATAL_REASONS` (comma list, or `*`) for an operator who actually runs on pay-as-you-go.
- `evaluateCreditState` takes an optional `fatalReasons` set; gateway passes the env-resolved one.
- The transition **machinery** (entered/changed/exited) is unchanged — only the default *policy* of which reasons are fatal changed (now none). Fully reversible via env.

## Re-scope note
Diagnosis revealed the flag is *categorically* about extra-usage (even `org_level_disabled`), so the right fix was to default the whole alarm off for subscription-only — not just drop one reason. Surfaced + verified before building.

## Test plan
- [x] machinery tests pass an explicit fatal set (policy-agnostic)
- [x] new: default (empty) set → each extra-usage reason **skips** (the fix); explicit opt-in restores the alarm; `resolveCreditWatchFatalReasons` default-empty / comma-parse / `*` / blank
- [x] **28 tests pass under both vitest AND bun**; `tsc` + `check-plugin-references` + `check-bot-api-wrapping` clean

## Risk
Low. Observability-only; reversible via env. Worst case = a real pay-as-you-go operator misses the alarm until they set the env var (switchroom isn't that — subscription-only is the whole product). Genuine exhaustion still surfaces via failover + all-exhausted alert.

## Rollout
Gateway change → staggered fleet restart, or fold into the next release.

### DIFF
--- a/telegram-plugin/credits-watch.ts
+++ b/telegram-plugin/credits-watch.ts
@@ -64,7 +64,8 @@ const KNOWN_CREDIT_REASONS = [
 /**
  * Resolve the active fatal-reason set. Empty by default (subscription-only);
  * if `SWITCHROOM_CREDITS_WATCH_FATAL_REASONS` is set, parse it as a
- * comma-separated list (unknown tokens ignored). `*` opts in all known reasons.
+ * comma-separated list (tokens kept verbatim — an unknown token is harmless,
+ * it simply never matches a real reason value). `*` opts in all known reasons.
  */
 export function resolveCreditWatchFatalReasons(env: NodeJS.ProcessEnv): Set<string> {
   const raw = env.SWITCHROOM_CREDITS_WATCH_FATAL_REASONS;
--- a/telegram-plugin/credits-watch.ts
+++ b/telegram-plugin/credits-watch.ts
@@ -27,20 +27,52 @@ import { join } from "path";
 const STATE_FILE = "credits-watch.json";
 
 /**
- * Possible values of `cachedExtraUsageDisabledReason` in `.claude.json`
- * that warrant a user-facing notification. Other values (null,
- * undefined, transient unknowns) are treated as "no notification
- * needed".
+ * Which `cachedExtraUsageDisabledReason` values warrant a user-facing alarm.
  *
- * Conservative list: only fatal-billing reasons. We don't want to fire
- * on every transient API blip the cache happens to write.
+ * IMPORTANT — empty by default (subscription-only model). The flag name says it
+ * all: `cachedExtraUsageDisabledReason` is *always* "why **extra usage** (the
+ * optional pay-as-you-go layer beyond the Pro/Max plan) is disabled." switchroom
+ * is subscription-only by design (compliance pillar 3 — "the plan is the
+ * ceiling, no API/pay-as-you-go"), so extra usage being OFF is the **expected,
+ * desired** state. Every value (`out_of_credits`, `extra_usage_disabled`,
+ * `credits_exhausted`, even `org_level_disabled` = "org admin disabled extra
+ * usage") describes that benign state — none indicates a real switchroom
+ * failure. Firing a "cron + replies will fail, buy credits" card on it is a
+ * false alarm AND advises something that contradicts the subscription-honest
+ * model.
+ *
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `credits-watch.ts:40` — `DEFAULT_CREDIT_FATAL_REASONS` is now an empty Set (was the 4-reason `FATAL_REASONS`). Correct for the subscription-only model: `cachedExtraUsageDisabledReason` is categorically about the pay-as-you-go layer being off, which is the expected state, so none of its values should alarm. Verified no stray references to the old `FATAL_REASONS` name remain in telegram-plugin.
- `gateway.ts:14746` — the sole caller of `evaluateCreditState` passes `resolveCreditWatchFatalReasons(process.env)`; the env-resolved set threads cleanly. `runQuotaWatch` (gateway.ts:14808, the "mirrors runCreditWatch" function) is the separate quota-watcher that owns the all-exhausted fleet alert — it does not call `evaluateCreditState` and is correctly left unchanged, so the genuine-exhaustion path the PR body cites is intact.
- `credits-watch.ts:84-85` — with an empty default set both `currentIsFatal` and `prevIsFatal` are always false, so credits-watch is fully silent by default and the recovery card never fires. Benign: there is no alarm to recover from. (One-time effect: an account that was alarmed under a prior version won't get a 'recovered' card after upgrade, since prev.lastNotifiedReason is no longer in the fatal set. Transient and harmless — not a blocker.)
- `resolveCreditWatchFatalReasons` (credits-watch.ts:56) — env parsing is correct: blank/whitespace → empty, `*` → all known reasons, comma list → trimmed tokens. Nit: the doc comment at line 53 says "unknown tokens ignored" but the code does NOT filter the comma list against `KNOWN_CREDIT_REASONS` (line 60 keeps every non-empty token). Harmless (an unknown token never matches a real reason value) but the comment is inaccurate — either filter against KNOWN or fix the comment to 'tokens are kept verbatim'.
- Tests: machinery suite now passes an explicit fatal set (policy-agnostic); new suites cover default-empty-skips-each-reason, explicit opt-in restores, and the four `resolveCreditWatchFatalReasons` cases. Good coverage of the behavioral change.

Rationale: correct, well-tested, fully reversible via env, observability-only with low blast radius; genuine exhaustion still surfaces via #2206 failover and #2215 all-exhausted alert. CI: build-dist + changes green, vitest/bun/lint still pending at review time — author should confirm they land green (body claims 28 pass under both runners) before merge. The one doc nit is non-blocking.

PR #2222, opened 2026-06-07
fix(failover): durably hold weekly-quota auto-fallback (clerk wedge)

mekenthompson:
## Summary

clerk went silently dead on claude's `/rate-limit-options` **weekly-quota TUI menu** — the same wedge #2218 was meant to catch. Deep inspection found the auto-failover chain broken in **three complementary places**; all three are required for a weekly wall to fail over *and stay failed over*.

| RC | Where | Bug | Fix |
|----|-------|-----|-----|
| **#1** | `src/agents/wedge-watchdog.ts` | Detector anchored on the `/rate-limit-options` line, which is printed **above** the menu and scrolls **off the top** of the `tmux capture-pane -p` viewport (no scrollback). clerk wedged exactly that way → detector logged **zero fires**. | Re-anchor `RATE_LIMIT_MENU_SIGNATURE` on the option-1 menu-body row (`Stop and wait for…`, pinned to the bottom where the cursor is) + `usage credits` / `Upgrade your plan` / `/rate-limit-options`. Esc-only park unchanged. |
| **#2** | `telegram-plugin/gateway/gateway.ts` (429 path) | 429 path called `fireFleetAutoFallback(agent)` with **no until** → markExhausted's ~5h default. A weekly wall surfacing as a 429 carries calendar wording (`resets Jun 9, 5am`) that `parseResetTime` **cannot read** → `resetAt` undefined → un-exhausts after 5h. | Shared, unit-tested `resolveExhaustUntil()` (`exhaust-until.ts`) with a **+7d floor that never returns undefined**, called from **both** the menu handler and the 429 path. |
| **#3** | `src/auth/broker/server.ts` | Even with a correct weekly `until`, `mark-exhausted` marks quota + mirrors the fallback **once** but never rewrites `auth.active`. The refresh-tick fanout resolved the effective account as `override ?? auth.active` with **no exhaustion check** → when the walled account's OAuth token refreshed (~1h) it **silently re-mirrored walled creds onto the whole fleet** (rollback / flap). | Route agent serving + fanout through a shared `accountWithFailover()` (generalised from the proven `consumerAccountWithFailover`): serve/mirror the first healthy fallback while walled, fall back to pinned/active when none healthy (retry beats dark), auto-revert when the window passes. Operator serving stays attribution-true. |

## Why

This is the durable resolution of the clerk incident (2026-06-07) and closes the latent flap that made auto-failover unreliable since #2206. RC#1 makes the weekly menu *fire* failover; RC#2 makes the exhaustion *window* long enough to cover a weekly wall; RC#3 makes the broker *honour* that window instead of silently re-mirroring the walled account on the next refresh tick.

**Compliance (vision.md pillar 3):** the wedge park remains **Esc-only** — never selects "Switch to usage credits" / "Add funds…" / "Upgrade your plan". The +7d floor errs toward keeping a walled account out of service (never serving it), which is the compliance-safe direction.

## Test plan

- [x] `tests/rate-limit-menu-detect.test.ts` — CLERK scroll-off fixture (no `/rate-limit-options` line + "Add funds…" wording) **matches**; finn fixture still matches; anchor-A-alone & anchor-B-alone stay false; full-watchdog e2e on the scroll-off pane signals + Esc; compliance (Esc-only, exhaustive banned-key) preserved.
- [x] `telegram-plugin/tests/exhaust-until.test.ts` — `resolveExhaustUntil` floors undefined/past/NaN → +7d, passes finite future through; 429-path composition proves rolling (finite, sub-floor) AND unparseable-weekly (+7d, never undefined) thread a non-undefined until.
- [x] `src/auth/broker/server.test.ts` — agent on a walled `auth.active` served its healthy fallback; refresh-tick fanout during exhaustion never re-mirrors the walled account; a weekly until holds past the ~5h default then auto-reverts; all-walled fleet keeps retrying the active account.
- [x] `tsc --noEmit` clean · `check-bot-api-wrapping.sh` clean · `check-plugin-references.mjs` clean · full build green · 569 vitest + bun IPC tests green.

## Risk

RC#3 touches the broker's hot agent-serving/fanout reads. Mitigated: `accountWithFailover` is a 1:1 generalisation of the production-proven consumer path, diverges from `auth.active` **only** when it's within a mark-exhausted window, and **always** falls back to the pinned/active account when no healthy alternative exists (an all-walled fleet retries, never goes dark). Healthy-case behaviour is identical (all 57 existing broker tests unchanged). RC#3 ships broker-side → the auth-broker **singleton must be recreated explicitly** on rollout (it's skipped by the staggered per-agent restart); RC#1/#2 are gateway-side and ride the per-agent roll.

Follow-up to #2218.

### DIFF
--- a/src/agents/wedge-watchdog.ts
+++ b/src/agents/wedge-watchdog.ts
@@ -59,15 +59,28 @@ export const WEDGE_FOOTER_SIGNATURE =
  * Hence a dedicated detector.
  *
  * Two INDEPENDENT anchors, both required, so it can never false-positive on
- * normal model output (verified against the live wedged pane, 2026-06-07):
- *   (a) the literal slash-command claude prints for this menu, which cannot
- *       appear in ordinary output: `/rate-limit-options`;
- *   (b) an option string ONLY this menu shows: "Switch to usage credits" or
- *       "Upgrade your plan".
- * (Grep of the repo confirmed zero pre-existing occurrences of any of these.)
+ * normal model output (verified against the live wedged panes, 2026-06-07):
+ *   (a) the interactive selector's option-1 row, which sits at the BOTTOM of
+ *       the pane (where the cursor is) and is therefore always inside the
+ *       captured viewport: "Stop and wait for …";
+ *   (b) an option string ONLY this menu shows: "usage credits" (covers both
+ *       "Switch to usage credits" and the newer "Add funds to continue with
+ *       usage credits" wording), "Upgrade your plan", or the literal
+ *       `/rate-limit-options` slash-command claude prints for the menu.
+ *
+ * v0.14.84 (#2218) anchored (a) on `/rate-limit-options` itself — but that
+ * line is printed ABOVE the menu options, so on a pane with enough preceding
+ * output it scrolls OFF the top of the `tmux capture-pane -p` viewport (which
+ * grabs only the visible ~50 lines, no scrollback). clerk wedged exactly this
+ * way on 2026-06-07 and the detector stayed silent (zero log fires). The fix
+ * re-anchors (a) on the option-1 row, which is pinned to the bottom of the
+ * pane and cannot scroll off; `/rate-limit-options` survives only as one of
+ * the (b) alternatives. "Stop and wait for" (not the full "…limit to reset")
+ * absorbs minor CLI wording drift (e.g. "…for your limit to reset").
+ * (Grep of the repo confirmed zero pre-existing occurrences of these.)
  */
 export const RATE_LIMIT_MENU_SIGNATURE =
-  /(?=[\s\S]*\/rate-limit-options)(?=[\s\S]*(?:Switch to usage credits|Upgrade your plan))/;
+  /(?=[\s\S]*Stop and wait for)(?=[\s\S]*(?:usage credits|Upgrade your plan|\/rate-limit-options))/;
 
 const MONTHS: Record<string, number> = {
   jan: 0, feb: 1, mar: 2, apr: 3, may: 4, jun: 5,
--- a/src/auth/broker/server.test.ts
+++ b/src/auth/broker/server.test.ts
@@ -2090,20 +2090,22 @@ describe("AuthBroker — historical-bug regressions (2026-05-14 fanout incident)
 });
 
 describe("AuthBroker — agent serving-failover (Layer 2, #2218 weekly-wall durability)", () => {
-  // Fleet: active=ken, fallback ken→pixsoul, one agent (clerk). The auto-
-  // fallback path (mark-exhausted) marks ken's quota + mirrors the fallback
-  // ONCE but never rewrites auth.active, so the exhaustion-BLIND refresh-tick
-  // fanout used to silently re-mirror ken back onto the fleet within ~1h.
-  // These pin that the serving + fanout reads now honour the exhaustion window.
+  // Fleet: active=default, fallback default→secondary, one agent (clerk). The
+  // auto-fallback path (mark-exhausted) marks the active account's quota +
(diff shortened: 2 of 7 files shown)


mekenthompson:
APPROVE

Verified the three RCs end-to-end against the touched files and the broker code on `main`; the mechanism, the tests, and the call graph all line up.

- RC#1 `src/agents/wedge-watchdog.ts:35-37` — re-anchoring `RATE_LIMIT_MENU_SIGNATURE` from `/rate-limit-options` (printed above the menu, scrolls off the `capture-pane -p` viewport) to the bottom-pinned option-1 row (`Stop and wait for`) + the `usage credits|Upgrade your plan|/rate-limit-options` alternation is the correct fix for the clerk scroll-off blind spot. Both lookaheads required; `tests/rate-limit-menu-detect.test.ts:490-494` (CLERK fixture, no slash-cmd line) and the anchor-A-alone / anchor-B-alone negatives (`:503-505`) prove no single-anchor false-positive. Esc-only park unchanged.
- RC#2 `telegram-plugin/gateway/exhaust-until.ts:307-315` + `gateway.ts:4402-4403` — `resolveExhaustUntil` never returns undefined, and threading it into the 429 path (`fireFleetAutoFallback(agent, untilMs)`) closes the real gap: that function's `untilMs` reaches `client.markExhausted(untilMs)` (gateway.ts:14688), and the broker defaults a missing `until` to `MARK_EXHAUSTED_DEFAULT_MS` = 5h (server.ts:119,1264). A weekly 429 (`parseResetTime` can't read `Jun 9, 5am` → undefined) previously got that 5h default; now it floors to +7d. The menu path (`onQuotaWallDetected`, gateway.ts:6285) is refactored to the same helper with identical behavior. Tests at `telegram-plugin/tests/exhaust-until.test.ts` cover undefined/past/NaN/Infinity → +7d and the real `detectModelUnavailable`→`resolveExhaustUntil` composition for both rolling and unparseable-weekly wordings.
- RC#3 `src/auth/broker/server.ts:973-219` — generalizing `consumerAccountWithFailover`→`accountWithFailover` and routing agent serving (`servingAccount`, now everything except `operator`) AND both fanout reads (`fanoutToAffectedAgents:248`, `fanoutForAgent:261`) through it is the durable half. Matching fanout on the failover-resolved account means a refresh tick on the walled account resolves to its fallback, matches no agent, and cannot re-mirror walled creds; a refresh of the live fallback still re-mirrors correctly. The dedup `existing >= exhaustedUntil` (server.ts:1212) lets the weekly window persist past the 5h default. The four new broker tests (`server.test.ts:81,98,114,136`) pin walled-active serving, the rollback guard, weekly-holds-then-auto-reverts, and all-walled-retries-not-dark. With nothing exhausted the resolution is identical to the old `override ?? auth.active`, so the healthy path is unchanged.

Watch (non-blocking): RC#3 now also fails over agents that are explicitly `auth.override`-pinned to a walled account (`fanoutForAgent:261` / `servingAccount` resolve `override` through `accountWithFailover`). That's the compliance-safe direction (never serve walled creds) and matches the stated intent, but it is a behavior change for operator-pinned agents — worth a note in the rollout so it isn't mistaken for the pin being ignored.

Nit: the RC#1 anchor `Stop and wait for` is generic prose; safe here only because the second lookahead is mandatory. If a future CLI build drops/renames the option-1 row, both anchors must be re-checked together (the comment at wedge-watchdog.ts:31 already flags the wording-drift assumption).

Rationale: each RC has a clear, verified mechanism and a regression test for the exact failure it fixes; healthy-case behavior is unchanged; the +7d floor and all-walled-retry both err toward never serving exhausted creds. Heed the RC#3 rollout note in the PR body — the auth-broker singleton must be recreated explicitly, not just per-agent rolled.

PR #2224, opened 2026-06-07
feat(agents): file-delivery guidance — deliver files, never hand back a container path

mekenthompson:
## Summary
Adds a `DELIVERY_GUIDANCE` fleet-invariant prompt block so agents **deliver** files instead of replying with a local container path the user can't reach. This is **layer 1** of the file-delivery outcome (operator-requested) and uses capabilities that already exist.

## Why
An agent runs headless in a container. When it writes a file and replies `/state/agent/workspace/report.pdf` (or `/tmp/...`), the user — who is not on the box — can't open it. The work didn't land. There was no guidance telling agents how to get a file *out*.

## The rule it encodes
- **Telegram (default):** the `reply` tool *already* attaches files via `files: ["/abs/path"]` (photos for images, documents otherwise, 50MB). The guidance points agents at it — no new tool needed.
- **Drive/OneDrive (keep/edit, or >50MB):** upload via the existing `gdrive` / `ms-365` MCP into a **`Switchroom/<agent-name>/`** folder + reply with the link.
- **Never:** hand back a local container path.

## Scope / follow-ups
This is layer 1 (guidance) of the full system the operator approved. Layers still to come as separate PRs:
- **Layer 3:** a deterministic `Switchroom/<agent>/` find-or-create helper (so the folder is reliable, not model-guessed).
- **Layer 4:** a standing pre-approval for writes to the agent's *own* delivery folder, so routine delivery doesn't fire a per-write Telegram approval (while writes elsewhere in the user's Drive still gate).

(Layer 2 — Telegram file send — turned out to already exist via `reply.files`, so no new tool was added.)

## Test plan
- [x] `vitest run tests/scaffold.memory-prompt.test.ts` — 7 pass incl. new (delivery block present, names channels, forbids local-path dumps, ordered sandbox→delivery→memory)
- [x] `tsc --noEmit` clean

## Risk
Minimal — prompt-only, additive fleet invariant. Serves the "always available" outcome: a file the user can't open is work that didn't land.

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -99,6 +99,45 @@ Example response shapes:
   sandbox. Cloning into \`$HOME/workspace\` instead."`;
 
 /**
+ * File delivery. One of the fleet invariant blocks (sibling to
+ * SANDBOX_GUIDANCE — that one covers where files live INSIDE the
+ * container; this one covers how to get a file OUT to the user).
+ *
+ * The recurring failure this fixes: an agent writes a file and hands
+ * back its local container path (\`/state/agent/...\`, \`/tmp/...\`).
+ * The user is not on the box and cannot open it. See SANDBOX_GUIDANCE
+ * above for the constant-vs-\`.hbs\` rationale.
+ */
+const DELIVERY_GUIDANCE = `## Delivering a file to the user
+
+You run in a container. A local path like \`/state/agent/...\` or
+\`/tmp/...\` means **nothing** to the user — they are not on your box and
+cannot open it. When you produce a file the user should have, **deliver
+it**, don't print its path.
+
+### Send it in Telegram (default for most files)
+The \`reply\` tool attaches files: pass \`files: ["/abs/path"]\` (images
+send as photos, everything else as documents; 50 MB max each). This is
+the right channel for almost anything — a report, a CSV, a chart, a
+generated doc — the user gets it straight in the chat. Prefer this.
+
+### Put it in their Drive (for files they'll keep or edit, or > 50 MB)
+If you have Google Drive / OneDrive tools wired (the \`gdrive\` /
+\`ms-365\` MCP), upload the file there and reply with the share link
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `src/agents/scaffold.ts:19-46` — new `DELIVERY_GUIDANCE` constant is accurate: it points agents at the real `reply` attachment mechanism (`files: ["/abs/path"]`, photos vs documents, 50MB) which matches the actual telegram-plugin reply tool, and at the existing `gdrive`/`ms-365` MCPs for keep/edit or >50MB. No invented capabilities — all referenced tools exist. Consistent constant-not-`.hbs` choice with sibling SANDBOX_GUIDANCE.
- `src/agents/scaffold.ts:55-56` — insert places `DELIVERY_GUIDANCE` between SANDBOX_GUIDANCE and TELEGRAM_GUIDANCE; on main the order is sandbox→telegram→memory→webfetch→vault (scaffold.ts:328-334), so the new order is sandbox→delivery→telegram→memory. Matches the test's ordering assertions exactly.
- `tests/scaffold.memory-prompt.test.ts:68-82` — new test asserts presence of `files: ["/abs/path"]`, the `Switchroom/<your-agent-name>` folder convention, the not-on-your-box rule, and sandboxIdx < deliveryIdx < memoryIdx. Confirmed both referenced headers exist verbatim in main (scaffold.ts:59 `## Sandbox: you're running in a switchroom container`, scaffold.ts:200 `## Memory — proactive, conversational`), so the index lookups resolve and the ordering holds.

Watch: layer-1 guidance only. The `Switchroom/<agent>` folder is model-guessed (create-if-missing) until the layer-3 deterministic helper lands, and routine delivery writes still fire per-write Drive approval until layer-4 pre-approval. Both are acknowledged as separate follow-up PRs in the description, not defects here.

Prompt-only, additive fleet invariant. No runtime behavior change beyond the injected system-prompt block; every claim in the guidance maps to a real, already-shipped capability. Safe to merge.

PR #2225, opened 2026-06-07
feat(delivery): switchroom deliver-file — upload to Switchroom/<agent> + link (OneDrive)

mekenthompson:
## Summary
`switchroom deliver-file <path>` uploads a file the agent produced into the user's **`Switchroom/<agent>/`** folder on their connected drive and prints a **share link**. The agent runs it (via Bash — `switchroom` is on PATH) and replies with the link, never a local container path the user can't reach. This is **layers 3+4** of the file-delivery outcome, combined into one tool (the operator chose the `deliver_file` approach).

## Why a CLI verb, not an MCP tool
The agent already has Bash + `switchroom` on PATH, and a switchroom-owned write to the agent's **own** `Switchroom/<agent>/` folder is pre-authorized by design — so it **bypasses the per-write MCP approval gate** (which exists to stop agents writing over the user's *existing* docs). That collapses **layer 3** (deterministic folder — switchroom owns find-or-create, no model-guessed names) and **layer 4** (friction carve-out — no per-file approval tap) into one.

## Scope
- Ships the **OneDrive (Microsoft)** provider — its broker creds already carry an access token.
- **Google Drive** is the identical-shape follow-up (refresh-token exchange + Drive v3 calls).
- DELIVERY_GUIDANCE (#2224) updated to point at `switchroom deliver-file`.

## Changes
- `src/delivery/onedrive.ts` — find-or-create `Switchroom/<agent>` (path lookup → create, 409-race safe), inline (≤4MB) + resumable upload, share link (anonymous → org → item webUrl fallback). All HTTP via **injectable fetch** (the `src/drive/docs-get.ts` pattern).
- `src/cli/deliver-file.ts` — testable core (injectable broker + provider + fs) + thin CLI wrapper; clean errors that point at the `reply` 50MB fallback when no drive is connected.

## Test plan
- [x] `vitest run src/delivery/ src/cli/deliver-file.test.ts` — 14 pass (folder find/create/race, inline+session upload, link fallback, full orchestration; core success/missing/empty/no-drive/upload-failure)
- [x] guidance test green; `tsc --noEmit` + plugin-refs clean
- [ ] **Live canary** on clerk (OneDrive connected): `switchroom deliver-file /tmp/test.pdf` → confirm it lands in `Switchroom/clerk/` and the link opens. Cloud-API code can't be live-exercised in CI without touching the user's drive, so request shapes are asserted with mocked fetch and validated by canary.

## Risk
Moderate (writes to the user's OneDrive), bounded: writes only into the agent's own `Switchroom/<agent>/` folder, never elsewhere; no-drive degrades to a clear message pointing at the `reply` fallback. Behind a brand-new verb — nothing else changes.


mekenthompson:
APPROVE

- `src/delivery/onedrive.ts:148-152, 481-549` — Graph integration is correct end-to-end. Path-addressed folder lookup (`root:/Switchroom`, `root:/Switchroom/clerk`), root-vs-nested create URL split, and the parent-id+child-path upload form (`items/<id>:/<name>:/content`) all match the documented Graph API. Verified the broker seam: `getCredentials` returns `credentials` verbatim (protocol.ts:364) and `server-microsoft.test.ts:249` asserts `credentials.microsoftOauth.accessToken`, so the cast at deliver-file.ts:146 resolves a real live token.
- `src/delivery/onedrive.ts:603-610` — resumable upload PUT correctly omits the Authorization header (the createUploadSession `uploadUrl` is pre-authorized; sending a bearer would break it). Chunk size 5 MiB is a valid 320-KiB multiple per Graph rules. 200/201 captures the final item, 202 continues, anything else throws.
- `src/delivery/onedrive.ts:536-539` — 409 create-race is handled (re-fetch on conflict), covered by onedrive.test.ts:306.
- `src/delivery/onedrive.ts:626-647` — share-link fallback chain anonymous→organization→item.webUrl is sound; throws only if all fail AND no webUrl. Right call (a sign-in-required item URL still beats a container path).
- `src/cli/deliver-file.ts:178-197` — error handling is clean and actionable: missing file, empty file, and no-connected-drive each return a distinct message; the no-drive path points the agent at the `reply` 50MB fallback (deliver-file.ts:185-187), matching the #2224 guidance. Injectable broker/provider/fs seams make the core testable; 14 tests cover the paths.
- `src/cli/index.ts:228,236` + `src/agents/scaffold.ts:15-22` — verb registered alongside peers; DELIVERY_GUIDANCE updated to point at `switchroom deliver-file`, consistent with the just-merged #2224.

Watch (non-blocking):
1. `src/cli/deliver-file.ts:164,192` — `readFileSync` buffers the entire file into memory before upload. The Drive path exists precisely for >50MB files, so a several-hundred-MB delivery is fully resident in RAM. Fine for typical reports; revisit with a streaming read if large media becomes common.
2. `src/delivery/onedrive.ts:600-616` — a failed chunk throws with no resume/retry despite the session being resumable. Acceptable for v1; a transient mid-upload network blip restarts from scratch on the next invocation.
3. Token freshness depends entirely on the broker handing a live (refreshed) accessToken — no 401-refresh-retry in the provider. The broker appears to refresh server-side, so this is a watch, not a defect.

Nit: `src/cli/deliver-file.ts:191-197` — the try wraps both `read()` and `deliver()`, so a file vanishing between stat and read surfaces as 'upload failed' rather than 'file not found'. Cosmetic, extremely unlikely.

Writes are correctly bounded to the agent's own `Switchroom/<agent>/` folder via the pre-authorized CLI path (never the user's existing docs), with a clean degrade when no drive is connected. Cloud-API behavior can't be exercised in CI; request shapes are asserted with mocked fetch and the live canary checkbox is appropriately left unchecked. Safe to merge once the canary passes.

PR #2226, opened 2026-06-07
feat(auth): `auth schedule` — per-account 5h vs weekly quota view

mekenthompson:
## Summary

Adds `switchroom auth schedule`: a per-account quota view that separates the **5-hour rolling** window from the **7-day weekly** window, surfacing the staggered weekly reset schedule that `auth list`/`auth show` hide (they collapse quota into a single "soonest reset" column).

Per account it shows pool role (active / fallback rank / excluded), **5h % + reset**, **weekly % + reset DAY**, and a STATE distinguishing `healthy` / `5h-walled` / `weekly-walled`.

```
  ACCOUNT                        POOL        5H WINDOW       WEEKLY WINDOW              STATE
● pixsoul@gmail.com              active #1   9% · 3h 35m     12% · Sun 11:00 (6d 16h)   healthy
! ken.thompson@outlook.com.au    #2          0% · now        100% · Tue 05:00 (1d 10h)  weekly-walled · 1d 10h
✓ me@kenthompson.com.au          excluded    31% · 3h 55m    20% · Sun 20:00 (1h 25m)   healthy
```

## Why

The weekly anchor is a fixed day-of-week per subscription, **different per account** — so the fleet can keep coverage by preferring whichever account has weekly headroom. But "exhausted" in `auth list` is ambiguous: a 5h blip recovers in hours, a weekly wall in days. This view makes the distinction (and the staggered schedule) visible at a glance. Real example above: ken.outlook's 5h window resets in minutes, but it's **weekly-walled till Tuesday** — the view says so instead of mislabelling it "5h-walled".

## Design notes

- **Live probe by default.** Uses the existing broker `probe-quota` op (the same one the web dashboard uses; operator-initiated, one `max_tokens:1` call per account). The cached `last_quota` is empty after a broker restart, so a cached default would render an all-`—` table. `--cached` skips the live probe; `--json` dumps raw state. No new API callsite — it's the broker's existing op.
- **Weekly-maxed precedence.** A weekly window ≥99.5% classifies as `weekly-walled` with the weekly reset as the horizon, even when the exhaustion ledger points at a sooner 5h reset — so a weekly-capped account isn't shown as recovering in minutes.
- **Pure, tested core.** `buildScheduleRows` / `classifyState` / formatters live in `src/cli/auth-schedule.ts` behind `_testing`; tests call them directly (broker IO not mocked — the `auth-google.test.ts` pattern). House-style hand-rolled `chalk`+`padEnd` table, color applied last to dodge the ANSI/pad-width trap.

## Test plan
- [x] 20 unit tests: the 5h-vs-weekly classifier (precise-match + duration heuristic + weekly-maxed precedence), window resolution (live-preferred, cached ISO→Date fallback, none), row building (pool ranks/sort), and deterministic tz-pinned rendering (no ANSI in `--color=false`).
- [x] `tsc --noEmit` clean · full `bun lint` chain clean (PII/bun-test-imports/etc.).
- [x] Ran live against the production broker: default (live probe), `--cached`, `--json` all correct.
- [x] Pre-existing `update.test.ts` env-sensitive failures verified unrelated (fail identically with this change stashed).

## Risk
Additive, read-only CLI verb. Default does a live probe (operator-initiated, trivial cost); `--cached` for a pure read. No change to existing verbs or broker/serving code.

### DIFF
--- a/src/cli/auth-schedule.test.ts
+++ b/src/cli/auth-schedule.test.ts
@@ -46,7 +46,7 @@ describe("formatResetDay (deterministic across host tz)", () => {
     expect(formatResetDay(WEEKLY_RESET, "UTC")).toBe("Sun 01:00");
     // Melbourne is UTC+10 → 01:00 UTC = 11:00 local, still Sunday.
     expect(formatResetDay(WEEKLY_RESET, "Australia/Melbourne")).toBe("Sun 11:00");
-    // ken.outlook's anchor (Mon 19:00 UTC) lands Tuesday morning in Melbourne.
+    // A Mon 19:00 UTC weekly anchor lands Tuesday morning in Melbourne (UTC+10).
     expect(formatResetDay(new Date("2026-06-08T19:00:00Z"), "Australia/Melbourne")).toBe("Tue 05:00");
   });
 });
@@ -111,7 +111,7 @@ describe("classifyState — the 5h-vs-weekly distinction", () => {
   });
 
   it("maxed weekly (>=99.5%) → weekly-walled even when exhausted_until is the sooner 5h reset", () => {
-    // The live ken.outlook case: 5h resets in ~27m, but weekly is 100% till Tue.
+    // The real-world case: the 5h window resets in ~27m, but weekly is 100% till Tue.
     const fiveSoon = new Date("2026-06-07T06:27:00Z");
     const weeklyTue = new Date("2026-06-09T19:00:00Z");
     expect(
@@ -288,7 +288,7 @@ describe("formatScheduleRows (no-color)", () => {
   it("active row shows the ● glyph, pool, and weekly reset day", () => {
     expect(lines[1]).toContain("●");
     expect(lines[1]).toContain("active #1");
-    expect(lines[1]).toContain("Sun 11:00"); // pixsoul-style Sunday anchor, Melbourne
+    expect(lines[1]).toContain("Sun 11:00"); // a Sunday weekly anchor, Melbourne
     expect(lines[1]).toContain("healthy");
   });
 
--- a/src/cli/auth-schedule.test.ts
+++ b/src/cli/auth-schedule.test.ts
@@ -1,0 +1,300 @@
+// Unit tests for the pure helpers behind `switchroom auth schedule`. Broker IO
+// is NOT mocked — we test the model/classify/format helpers directly with the
+// broker response shapes as plain typed args (mirrors auth-google.test.ts).
+// The load-bearing assertion is classifyState: the 5h-vs-weekly distinction.
+
+import { describe, it, expect } from "vitest";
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `src/cli/auth-schedule.ts:386-405` — `resolveWindow` field names verified correct against the real types: live path reads `entry.result.data.{fiveHourUtilizationPct,sevenDayUtilizationPct,fiveHourResetAt,sevenDayResetAt}` (client.ts:214-217, Date-revived at client.ts:451-452) and the cached path reads the same names off `last_quota` (client.ts:147-153, ISO strings correctly wrapped in `new Date()`). No name drift — the table will populate, not render all-em-dash.
- `src/cli/auth-schedule.ts:437-444` — weekly-wall precedence is correct: a ≥99.5% weekly with a future reset classifies `weekly-walled` ahead of the exhaustion-ledger 5h timestamp, which is exactly the ken.outlook case the PR describes. Tested at auth-schedule.test.ts:119.
- `src/cli/auth-schedule.ts:475` — `exhausted = a.exhausted && (exhaustedUntil?.getTime() ?? 0) > now` correctly drops stale-walled rows whose until has passed (precedence is fine: `>` binds before `&&`). The `until===undefined → walled` branch in classifyState is therefore unreachable via buildScheduleRows, but kept as a defensive direct-call path — harmless.
- `src/cli/auth-schedule.ts:553-559` — `formatStateCell` horizon shows the classified window's own reset (weekly reset for a weekly wall) rather than the ledger timestamp, so a weekly-walled account isn't shown recovering in minutes. Tested at auth-schedule.test.ts:247.
- `src/cli/auth-schedule.ts:576-590, 619-643` — alignment is sound: raw cells are `padEnd`'d at fixed widths and color is applied only to the single-glyph slot and the unpadded final state column, dodging the ANSI/pad-width trap. `--color=false` emits no escape codes (asserted at auth-schedule.test.ts:291).
- `src/cli/auth-schedule.ts:676-687` — live probe via the existing `probe-quota` broker op with `.catch(() => undefined)` falling back to the cached snapshot; `--cached` skips it, `--json` dumps raw. No new broker callsite. `brokerCall`/`withConfigError` confirmed to exist (broker-call.ts:63, helpers.ts:9).
- `src/cli/auth.ts:475` — subcommand registered alongside the google/microsoft auth subcommands; help text updated (auth.ts:16).

Watch (non-blocking): the 6h–20h remaining band with no matching reset timestamp falls through to a generic `walled` label — only reachable when both probe and cache are absent, so cosmetic. And the weekly column (W.weekly=26) truncates with an ellipsis if a horizon ever exceeds ~25 chars (e.g. a 3-digit-day reset) — alignment is preserved either way.

Additive, read-only verb; pure tested core (20 tests covering classifier, window resolution, row sort, tz-pinned rendering); no change to existing verbs or broker/serving code. Safe to merge.

PR #2228, opened 2026-06-07
feat(delivery): Google Drive provider for deliver-file + fix one-shot hang

mekenthompson:
## Summary
Completes the file-delivery outcome across **both clouds** and fixes a process hang the OneDrive canary exposed.

- **Google Drive provider** (`src/delivery/gdrive.ts`): find-or-create `Switchroom/<agent>` → multipart upload → shareable link. Same injectable-fetch shape as OneDrive.
- **Provider selection**: `runDeliverFile` resolves the first connected drive (OneDrive, else Google Drive) and reports which.
- **Hang fix**: the broker client keeps a connection open; the one-shot verb now closes it (`finally`) — without this the process hangs after printing its result.

## The token simplification
Google needed no token exchange / client_secret / vault-broker: the auth-broker keeps a **fresh** Google access token on disk (its background refresh-tick, same as Microsoft), so the CLI passes `googleOauth.accessToken` straight in — the Google path is identical to OneDrive. (I'd drafted a refresh-token exchange; the broker made it unnecessary, so it's removed — no dead code.)

## The hang (caught by the live canary)
The OneDrive canary on `clerk` delivered the file and printed a real `1drv.ms` link — but `EXIT=124`: the process didn't exit. Root cause: `AuthBrokerClient` keeps one connection open (built for long-running consumers); a one-shot CLI verb must `close()` it or the open socket keeps the event loop alive and the agent's Bash call hangs after success. Fixed.

## Test plan
- [x] `vitest run src/delivery/ src/cli/deliver-file.test.ts` — 31 pass (gdrive provider: folder find/create, multipart upload, link-fetch fallback, full orchestration; deliver-file: provider selection OneDrive/Google, scope resolvers for both clouds, slug-guard, error paths)
- [x] `tsc --noEmit` clean
- [x] **OneDrive path canary-proven live** on clerk (real upload + share link)
- [ ] **Google canary** once an agent has Google Drive connected (mocked-fetch tested; live validates the real Drive upload)

## Risk
Moderate (writes to the user's drive), bounded: only ever `Switchroom/<agent>/`; agentName slug-guarded; share scope configurable (`SWITCHROOM_DELIVER_LINK_SCOPE`). The hang fix is a pure correctness fix on the merged OneDrive path.

### DIFF
--- a/src/cli/deliver-file.test.ts
+++ b/src/cli/deliver-file.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect } from "vitest";
-import { runDeliverFile, resolveLinkScopes, safeAgentName } from "./deliver-file.js";
+import { runDeliverFile, resolveLinkScopes, resolveGoogleLinkScopes, safeAgentName } from "./deliver-file.js";
 
-describe("resolveLinkScopes", () => {
+describe("resolveLinkScopes (OneDrive)", () => {
   it("defaults to anonymous-first", () => {
     expect(resolveLinkScopes({} as NodeJS.ProcessEnv)).toEqual(["anonymous", "organization"]);
   });
@@ -10,6 +10,15 @@ describe("resolveLinkScopes", () => {
   });
   it("honors anonymous-only", () => {
     expect(resolveLinkScopes({ SWITCHROOM_DELIVER_LINK_SCOPE: "anonymous" } as NodeJS.ProcessEnv)).toEqual(["anonymous"]);
+  });
+});
+
+describe("resolveGoogleLinkScopes", () => {
+  it("defaults to anyone-with-link", () => {
+    expect(resolveGoogleLinkScopes({} as NodeJS.ProcessEnv)).toEqual(["anyone"]);
+  });
+  it("maps organization → domain", () => {
+    expect(resolveGoogleLinkScopes({ SWITCHROOM_DELIVER_LINK_SCOPE: "organization" } as NodeJS.ProcessEnv)).toEqual(["domain"]);
   });
 });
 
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- **Hang fix verified correct** (`deliver-file.ts:173-193`): `brokerToken` wraps `getCredentials` in try/catch and `client.close()` in `finally`. Confirmed `AuthBrokerClient.close()` exists and `sock.destroy()`s the open socket (`client.ts:371-382`) — that is exactly the open-socket-keeps-event-loop-alive cause of the canary's EXIT=124. Pure correctness fix on the merged OneDrive path.
- **Google token cast resolves a live token** (`deliver-file.ts:182-187`): `creds.googleOauth.accessToken` matches the real broker shape `GoogleAddAccountCredentials.googleOauth.accessToken: string` (`client.ts:302-303`); `getCredentials` returns `credentials` verbatim (`client.ts:388-396`), and `ProviderNameSchema` accepts `"google"` (`protocol.ts:60`), so the wire call is valid rather than always falling into the null path.
- **Provider selection sound** (`deliver-file.ts:198-214`): OneDrive-first, Google fallback, null when neither connected → actionable error pointing at the reply 50MB fallback (`deliver-file.ts:149-157`). Token simplification (no exchange/client_secret, broker keeps Google fresh) matches the OneDrive design — no dead code.
- **Drive v3 endpoints correct** (`gdrive.ts`): folder find-or-create with query escaping + `trashed=false` (`:492-505`), multipart `uploadType=multipart` with hand-built `multipart/related` body and correct boundary framing (`:536-556`), permissions POST then `webViewLink` fallback-fetch chain (`:579-597`). Upload omits no required auth.
- 31 tests cover both providers, scope resolvers, slug-guard, and error paths.

Watch (non-blocking, all acknowledged in the PR): multipart buffers the whole file in RAM with no resumable-session path (docstring flags >5MB as follow-up, `gdrive.ts:522-527`); the `anyone` default is anyone-with-link over-share (documented SECURITY NOTE + `SWITCHROOM_DELIVER_LINK_SCOPE` knob, `gdrive.ts:566-571`); Google live canary still unchecked (mocked-fetch only). None block merge.

Rationale: completes file-delivery across both clouds and fixes a real process hang; every integration seam checks out against the merged broker code.

PR #2233, opened 2026-06-07
fix(routing): framework owns the reply topic, not the model's message_thread_id

mekenthompson:
## Summary

A General-topic question to **marko** got answered in the **CRM (Brevo)** topic — so the user, reading General, saw silence. Proven in the gateway logs: the inbound came from the General topic (which carries **no** `message_thread_id`), and the model answered with an explicit `message_thread_id`, which `resolveAnswerThreadId` let win **outright** over the framework's record of the origin. The model could redirect *any* reply to *any* topic, and a thread-less General origin had nothing to pull it back.

This can't be fixed with a model prompt (that's the same model-dependency that caused it). The fix inverts the precedence so the **framework's turn anchor owns the reply topic**:

```
origin turn  →  live in-flight turn  →  (model explicit)  →  late-ended recovery
```

The model's `message_thread_id` is demoted to a **last resort** — consulted only when there is neither a resolved origin nor a live turn (a genuinely orphaned/proactive send). The live tier keys off the turn's **presence** (`liveTurnPresent`), not its thread value, so a General live turn (thread `undefined`) still anchors the reply to General.

Net effect: topic-targeting for proactive posts becomes **config-driven** (a cron's `topic:` sets the turn thread) instead of model-driven — the deterministic behaviour we want.

## Determinism (proven, not sampled)

Per the operator standard (`feedback_prove_finite_fsm_not_sample`), the resolver's **total-enumeration proof** is rewritten to cover **all 54 reachable inputs × both modes**, including:
- **`INV-ANCHOR`** (the load-bearing guarantee): under framework authority, the output is **independent of `explicitThreadId`** whenever `originResolved || liveTurnPresent` — i.e. the model's explicit thread can *never* redirect an anchored reply.
- An explicit proof of the **marko case**: General origin (resolved, thread `undefined`) + model explicit ⇒ General, not the model's topic.
- TOTAL / DETERMINISTIC / NO-FABRICATION / PRECEDENCE for both the new default and the legacy (kill-switch) path.

## Controls & observability

- Kill switch `SWITCHROOM_REPLY_TOPIC_AUTHORITY=0` restores the legacy explicit-first precedence.
- New telemetry tag `EXPLICIT_OVERRIDDEN(model→X,routed→Y)` on the `reply-route` log line surfaces every time the framework corrects a model topic-grab — no silent override.
- `via=` now reflects the active precedence so telemetry matches routing.

## Test plan
- [x] `telegram-plugin/gateway/answer-thread-resolve.test.ts` — total-enumeration proof (54×2) + INV-ANCHOR + marko-case + legacy mode (20 tests).
- [x] Updated the sibling `tests/answer-thread-resolve.test.ts` assertions that pinned the old explicit-first behaviour (now assert origin>live>explicit + kill-switch legacy).
- [x] `tsc --noEmit` clean · full `bun lint` chain clean.
- [x] **234 files / 3764 vitest tests green** — no regression across the gateway/streaming/reply surface.

## Risk
Fleet-wide reply-routing change, but: behaviour only changes when the model set an explicit topic that *contradicted* a framework anchor (the misroute) — otherwise identical. Multi-topic supergroups (marko, test-harness) are the only agents with threads; DM/single-channel agents are unaffected (no threads). Kill-switch + new telemetry for safe rollback/observation.


mekenthompson:
APPROVE

- `answer-thread-resolve.ts:526-548` — framework precedence (origin → live-by-presence → explicit → ended) is correct and matches the doc-comment at `:424-449`. Tier 2 keys off `liveTurnPresent` not `liveThreadId`, so a General live turn (thread undefined) still anchors the reply — the load-bearing part of the marko fix. No change needed.
- `answer-thread-resolve.ts:518-525` — legacy/kill-switch branch (`frameworkTopicAuthority === false`) preserves the old explicit-first chain verbatim. Clean rollback path. No change needed.
- `answer-thread-resolve.test.ts:241-254` — `specFramework`/`specLegacy` are independent reference encodings kept separate from the impl; the 54-row total-enumeration proof (`:258-372`) asserts both against the resolver. INV-ANCHOR (`:324-332`) proves output is independent of `explicitThreadId` whenever `originResolved || liveTurnPresent` — the model can never redirect an anchored reply. Strong verification, not a sample.
- `gateway.ts:573-577` — `liveTurnPresent: liveTurn != null` correctly mirrors `originResolved: originTurn != null`; `liveTurn` already in scope (used at the pre-existing `liveThreadId` line). `frameworkTopicAuthority` wired from the kill switch at `:565-566`. Correct.
- `gateway.ts:586-596` — `via` telemetry reordered to match active precedence (framework branch uses `liveTurn != null`, presence-based, consistent with the resolver); legacy branch unchanged. `EXPLICIT_OVERRIDDEN` tag (`:601-605`) surfaces every framework correction — no silent override.

Behavioural change is correctly scoped: routing only differs when the model set an explicit topic that contradicted a framework anchor; all other inputs are identical (proven by the dual-mode enumeration). changedFiles=4 matches the diff (resolver + its two test files + gateway). tsc/lint/3764 vitest green per PR body. Kill switch SWITCHROOM_REPLY_TOPIC_AUTHORITY=0 + new telemetry give safe rollback.

Rationale: deterministic, fully-enumerated decision-table change with an independent spec cross-check, a config-driven kill switch, and observability for every correction — exactly the right shape for a fleet-wide reply-routing change.

PR #2234, opened 2026-06-08
feat(ci): claude CLI regression coverage — flag contract + nightly latest canary + runbook

mekenthompson:
## Summary

Closes the gap behind "can we track latest claude as long as CI passes?" — **today, no.** A CLI bump is the one change where every *required* gate passes while the runtime can silently break: the build checks only prove the binary **installs and runs**, never that its surface/behaviour is still compatible. That's exactly how the 2.1.166 cross-session-messaging hardening slipped past green CI and needed a manual round-trip canary. This adds the missing layers.

## What's added
1. **Flag contract test** (`tests/claude-cli-contract.test.ts`) — asserts every CLI flag switchroom's agent boot depends on (from `start.sh.hbs`) still exists in the **installed binary's `--help`**, plus the `mcp` subcommand. Credential-free. **Self-skips green** when claude isn't on PATH (normal `vitest` job); **runs for real** where claude is present (local dev, the canary). The two hidden channel-load boot flags are covered by the canary's install smoke, not `--help`.
2. **Build-arg parameterization** (`Dockerfile.base` + `Dockerfile.hindsight`) — the pin is now a `CLAUDE_CODE_VERSION` build-arg **defaulting to the production version (2.1.168)**. A normal build is byte-identical + fully auditable; the canary overrides it to build on latest. Defaults kept in lockstep (#1978). Future bumps = one-line default change each.
3. **Nightly `latest-claude` canary** (`ci-claude-latest-canary.yml`) — informational, **not gating**. Resolves pinned-vs-latest, proves `claude@latest` installs in the base-image context (`node:22-trixie-slim`) **and** passes the flag contract — so a dropped flag / broken publish in the *next* version goes **red in Actions before the pin is bumped**. The literal answer to "is latest safe?"
4. **Runbook** (`docs/operators/claude-cli-updates.md`) — the update procedure, the regression-layer table, the hindsight manual-recreate recipe (`--tmpfs /run/claude-creds` uid=11000 + `--shm-size=2g`), and **how to wire the `uat-host` self-hosted runner** so `ci-uat` becomes the automatic **behavioural** gate (real round-trip) and make it required.

## What this does and doesn't get you
- ✅ Auto-detects nightly when latest **drops a flag**, **breaks install**, or has a **broken publish** — before you bump.
- ✅ Makes the pin a one-line, build-arg-able change.
- ⏳ The **behavioural** gate (a real inbound→reply turn on latest) still needs the `uat-host` runner + subscription creds — an operator creds-on-a-box step, fully documented. Until then the manual round-trip canary remains the behavioural check, now backed by the automatic flag/install canary.

## Test plan
- [x] contract test passes for real against the installed claude (9/9 required flags); self-skip path covered
- [x] both Dockerfile defaults = 2.1.168, build via the var, no stale hardcode; canary's pin-parse grep resolves 2.1.168
- [x] `tsc` clean; workflow YAML valid; `bunx vitest run` (the canary's exact command) green
- [x] CI `build-base`/`build-dependents`/`build-hindsight` rebuild on the build-arg default — identical to the prior hardcoded pin

## Risk
Low. The Dockerfile change is default-preserving (production build unchanged). The new test self-skips in normal CI. The canary is non-gating. No behavioural/runtime change.

### DIFF
--- a/.github/workflows/ci-claude-latest-canary.yml
+++ b/.github/workflows/ci-claude-latest-canary.yml
@@ -1,0 +1,93 @@
+name: ci-claude-latest-canary
+
+# Nightly "is the latest claude CLI safe to bump to?" canary.
+#
+# A claude version bump is the one change where every required gate (lint,
+# vitest-against-fixtures, image builds, `claude --version`) can pass while the
+# runtime silently breaks — because none of them exercise the NEXT version's
+# real surface. This workflow does, ahead of time: it pulls `@latest` nightly
+# and proves it (a) installs in the base-image context and (b) still exposes
+# every CLI flag switchroom's agent boot depends on. A failure here is the
+# early warning that the next bump needs attention — surfaced automatically
+# instead of discovered during a manual canary.
+#
+# Credential-free by design: no OAuth, no Telegram, no real turn — so it runs
+# on hosted runners with no setup. The BEHAVIOURAL half (a real round-trip on
+# latest) is the UAT gate; see docs/operators/claude-cli-updates.md for wiring
+# the `uat-host` self-hosted runner so ci-uat becomes that gate.
+#
+# NOT a required check — informational. Its job is to fail loudly in the
+# Actions tab when latest drifts, not to block PRs.
+
+on:
+  schedule:
+    # 06:17 UTC daily — off the hour to avoid the cron thundering-herd.
+    - cron: "17 6 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: claude-latest-canary
+  cancel-in-progress: true
+
+jobs:
+  latest-claude-smoke:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+      # ubuntu-latest ships node + npm (used for the global `claude` install
+      # below); bun is for the repo deps + vitest, matching the repo convention.
+      - name: Set up bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6  # v2
+
+      - name: Resolve pinned vs latest claude version
+        id: ver
+        run: |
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- `docker/Dockerfile.base:154,177` + `docker/Dockerfile.hindsight:44-45` — `ARG CLAUDE_CODE_VERSION=2.1.168` default matches the current hardcoded pin on main exactly (verified: both files install `claude-code@2.1.168` today), so a no-`--build-arg` build is byte-identical to the prior pin and stays auditable. Lockstep invariant (#1978) preserved. Correct.
- `tests/claude-cli-contract.test.ts:48-58` — REQUIRED_FLAGS verified complete against `profiles/_base/start.sh.hbs:673`: all 9 (--add-dir via SR_FLEET_ARG, --append-system-prompt, --model, --effort, --permission-mode, --fallback-model, --plugin-dir, --dangerously-skip-permissions, --continue) are present in the real exec line. The two excluded channel flags (--dangerously-load-development-channels / --channels) are genuinely hidden/experimental and present in start.sh.hbs but absent from --help — the exclusion comment is accurate. Contract is faithful.
- `tests/claude-cli-contract.test.ts:84-90` — self-skip via `describe.skipIf(!HAS_CLAUDE)` + a paired `skipIf(HAS_CLAUDE)` documentation suite means the file never reads as silently empty in the normal vitest job. Good defensive shape.
- `.github/workflows/ci-claude-latest-canary.yml:64` — `docker run --rm node:22-trixie-slim` is faithful to `docker/Dockerfile.base:24` FROM (verified node:22-trixie-slim), so install-in-base proves @latest in the same env agents ship. Version grep `CLAUDE_CODE_VERSION=[0-9.]+` -m1 correctly targets the new ARG line. Actions SHA-pinned (checkout v6.0.2, setup-bun v2). Credential-free, non-gating, concurrency-guarded. Sound.
- NIT `tests/claude-cli-contract.test.ts:48` — comment cites `start.sh.hbs:655-663` but the exec invocations are at lines 673-681 on main. Line-number drift in a doc comment; update or drop the exact range so it doesn't rot.

Rationale: test + CI + docs only; the sole runtime touch is the Dockerfile build-arg parametrization, which defaults byte-identically to the prior pin. Adds the missing regression layer (flag-contract + nightly @latest canary) for the one bump class where required checks pass while the runtime silently breaks. The single nit is non-blocking.

PR #2236, opened 2026-06-08
feat(ci): sandboxed self-hosted UAT runner (uat-host) + setup script

mekenthompson:
## Summary

Makes the **behavioural** CLI-bump gate automatic: `ci-uat` runs a real inbound→reply round-trip against `test-harness` on each relevant change, instead of relying on a manual operator canary. This is the piece that unlocks **"track latest claude as long as CI passes"** — the build/flag canaries (#2234) prove install + surface; this proves a real turn still works.

## Sandboxed by construction (switchroom is a PUBLIC repo)
A self-hosted runner is a code-execution surface for fork PRs and must **not** sit unprotected next to the vault/fleet. So the runner:
- runs as a **container**, non-root, `--security-opt no-new-privileges`, mem + pids capped
- **no docker socket, no host bind-mounts**
- **bridge network = egress only** — reaches Telegram + GitHub, **cannot touch the host's `127.0.0.1` fleet/vault**
- the host-mutating `jtbd-restart` scenarios self-skip (no sudo); only the network-only `fuzz-*` / `inbound-no-drop` round-trips run here — exactly the boundary we want

## Files
- `docker/Dockerfile.uat-runner` + `docker/uat-runner-entrypoint.sh` — the runner image (actions/runner v2.334.0 + bun on `node:22-trixie-slim`); config persists in the `uat-runner-data` volume → survives restart without re-register.
- `scripts/setup-uat-runner.sh` — one command: build, push the 4 `TELEGRAM_*` secrets + `UAT_GATE_ENABLED` from `.env`, set **fork-PR approval = `all_external_contributors`** (the critical public-repo lock), register + run.
- `docs/operators/claude-cli-updates.md` — the runner section now documents the sandboxed approach + a **security floor** (fork-PR approval, no socket, no host mounts — do not weaken).

## Validated live
The runner registers **online** (label `uat-host`), picks up `ci-uat`, and checkout/secrets/setup all pass; the fuzz round-trips execute against the live `test-harness`. **`ci-uat` is added to required checks only after a confirmed green run — not in this PR.**

## Security review focus
Please sanity-check the isolation claims in `setup-uat-runner.sh` + `Dockerfile.uat-runner`: no docker socket, no host mounts, non-root, fork-PR approval set to the strictest. That's the load-bearing posture for a runner on a public repo.

## Risk
Low for the repo itself — these are operator artifacts (Dockerfile + script + docs), not runtime code; no required-check change here. The real risk surface is the runner's isolation, addressed above.

### DIFF
--- a/docker/Dockerfile.uat-runner
+++ b/docker/Dockerfile.uat-runner
@@ -1,0 +1,34 @@
+# Sandboxed GitHub Actions self-hosted runner for switchroom UAT (label: uat-host).
+# Isolation by construction: runs as non-root, NO docker socket, NO host mounts,
+# egress-only networking. It can reach Telegram (to drive the test bot) + GitHub,
+# but cannot touch the host, the vault, the fleet, or docker. The host-mutating
+# jtbd-restart UAT scenarios self-skip here (no sudo) — exactly the boundary we
+# want: behavioural round-trip tests run; nothing on the host can be mutated.
+FROM node:22-trixie-slim
+
+ARG RUNNER_VERSION=2.334.0
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends \
+      curl ca-certificates git jq libicu76 unzip \
+ && rm -rf /var/lib/apt/lists/*
+
+# bun (UAT runs under bun, repo convention). Install as root, then copy the
+# real binary to a world-readable path the non-root runner can exec.
+RUN curl -fsSL https://bun.sh/install | bash \
+ && cp -L /root/.bun/bin/bun /usr/local/bin/bun \
+ && chmod 0755 /usr/local/bin/bun \
+ && rm -rf /root/.bun
+
+RUN useradd -m -d /home/runner runner
+WORKDIR /actions-runner
+RUN chown runner:runner /actions-runner
+USER runner
+
+RUN curl -fsSL -o runner.tar.gz \
+      "https://github.com/actions/runner/releases/download/v${RUNNER_VERSION}/actions-runner-linux-x64-${RUNNER_VERSION}.tar.gz" \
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- `docker/uat-runner-entrypoint.sh:3`: comment claims "`--ephemeral` keeps each job on a clean checkout" but the `config.sh` invocation (`:11-17`) passes no `--ephemeral` flag — and can't, since the design is a long-lived runner (`--restart unless-stopped` + persistent `uat-runner-data` volume that survives restart without re-register). Stale copy: drop the `--ephemeral` sentence so the comment doesn't promise per-job clean-checkout isolation the runner doesn't provide. Non-blocking (comment only; behavior is the intended persistent runner).
- `scripts/setup-uat-runner.sh:11`: the isolation claim is precise for the host's `127.0.0.1`, but a default-bridge container can still reach host services via the docker0 gateway IP (`172.17.0.1`). The boundary holds ONLY if fleet/vault bind strictly to `127.0.0.1`, not `0.0.0.0`. Worth a one-line note in the security floor; not a code change here.
- Verified the load-bearing posture: non-root (`Dockerfile.uat-runner:26-28`), `no-new-privileges` + mem/pids caps (`setup-uat-runner.sh:53-54`), no docker socket and no host bind-mounts (only mount is the named `uat-runner-data` volume, `:55`), and fork-PR approval forced to `all_external_contributors` (`:43-45`) BEFORE registration under `set -euo pipefail` — fail-closed if hardening errors. Secret push uses `printf '%s'` (no trailing newline) and skips cleanly when `.env` is absent.

Operator artifacts only (Dockerfile + script + docs), no runtime code, no required-check change in this PR. Isolation claims check out against the diff; the two items above are a stale comment and a doc nuance, neither blocks merge.

PR #2237, opened 2026-06-08
ci(uat): fast gating job + non-gating nightly fuzz; scope push trigger

mekenthompson:
Splits ci-uat into a FAST `uat-gate` (real inbound→reply turn + rapid-fire no-drop, ~minutes, the required-check candidate) and a thorough `uat-fuzz` (schedule/dispatch-only, non-required — the full pool exceeds a 35-min budget on the single runner and timeout-cancels). Also path-scopes the push trigger (was unfiltered → every merge ran the 35-min suite). Validated live: every fuzz scenario that ran on the new sandboxed runner PASSED — it's a time-budget split, not a regression. This PR touches ci-uat.yml so it triggers `uat-gate` on itself. `ci-uat`/`uat-gate` added to required checks only after a confirmed green+fast run, separately.

### DIFF
--- a/.github/workflows/ci-uat.yml
+++ b/.github/workflows/ci-uat.yml
@@ -43,6 +43,19 @@ on:
       - ".github/workflows/ci-uat.yml"
   push:
     branches: [main]
+    # Same path scope as the PR trigger — a docs/docker/scripts merge must not
+    # spin up the (slow, runner-bound) UAT. Pre-fix the push trigger had NO
+    # paths filter, so every main merge ran the 35-min fuzz suite and
+    # timeout-cancelled, tying up the single uat-host runner.
+    paths:
+      - "telegram-plugin/**"
+      - "src/agents/**"
+      - "telegram-plugin/uat/**"
+      - "vitest.uat.config.ts"
+      - ".github/workflows/ci-uat.yml"
+  schedule:
+    # Nightly thorough fuzz pass (the uat-fuzz job below is schedule/dispatch-only).
+    - cron: "41 4 * * *"
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

- `.github/workflows/ci-uat.yml:113-115`: `uat-gate`'s `if:` is only `vars.UAT_GATE_ENABLED == 'true'` with no event filter, so on the new nightly `schedule` trigger (`:55-56`) the fast gate runs alongside `uat-fuzz`. Both target `[self-hosted, uat-host]` and share the per-ref concurrency group (`cancel-in-progress: false`), so they serialize cleanly — harmless redundancy, not a bug. Optional: add `&& github.event_name != 'schedule'` to skip the gate on the nightly pass. Non-blocking.
- Verified the load-bearing references against main: `bun run test:uat jtbd-fast-trivial-dm` (`:108`) resolves to `telegram-plugin/uat/scenarios/jtbd-fast-trivial-dm.test.ts` (exists); `inbound-no-drop` (`:117`) → `inbound-no-drop-rapid-fire-dm.test.ts`; `fuzz-` and `jtbd-always-on-after-restart` likewise resolve to real scenario files. No dangling positional filter.
- Verified the push `paths` block (`:48-53`) is identical to the existing `pull_request.paths` on main — the PR's 'same path scope as the PR trigger' claim holds, and it correctly fixes the prior unscoped push trigger that ran the 35-min fuzz on every main merge.
- Job split is sound: gate timeout 18m (`:88`) for the two fast deterministic round-trips; fuzz timeout 50m (`:128`) gated to `schedule || workflow_dispatch` (`:122-124`) so the slow pool never blocks a PR. Secrets-gate + setup steps preserved in both jobs; `jtbd-restart` stays separated so a mutating self-skip can't stall the fuzz pool.

CI config only, no runtime code, no required-check change in this PR. Clean APPROVE.

PR #2238, opened 2026-06-08
feat(delivery): Google Drive resumable upload for files > 5MB (OneDrive parity)

mekenthompson:
## Summary
Adds a **resumable upload** path to the Google Drive provider so `deliver-file` handles large files there too — closing the parity gap a reviewer flagged on #2228.

## Why
The Google provider was multipart-only (reliable to ~5MB), but `DELIVERY_GUIDANCE` tells agents to use `deliver-file` for **"files they'll keep or edit, or > 50MB."** A large delivery on a Google-only agent failed (cleanly, but failed). OneDrive already handles this via a resumable session (`onedrive.ts`); this gives Google the same.

## Change
`uploadFile` dispatches on size:
- **≤ 5MB** → `uploadMultipart` (the existing single-request path)
- **> 5MB** → `uploadResumable`: POST `uploadType=resumable` for a session URL (`Location` header), then PUT 8 MiB chunks with `Content-Range` (308 = continue, 200/201 = done with the file resource).

## Test plan
- [x] `vitest run src/delivery/gdrive.test.ts` — 9 pass, incl. 2 new (>cap → resumable session with 3 correct-Content-Range chunks; missing session URL errors clearly)
- [x] `tsc --noEmit` clean
- [ ] **Canary** validates the real large upload once a Google-Drive-write agent exists (clerk's Google is read-only today). Cloud API isn't live-testable in CI; request shapes asserted with mocked fetch.

## Risk
Low — additive; the ≤5MB path (the common case, and the only one exercised live so far via OneDrive) is unchanged. Bounded to the agent's own `Switchroom/<agent>/` folder as before.

### DIFF
--- a/src/delivery/gdrive.test.ts
+++ b/src/delivery/gdrive.test.ts
@@ -5,6 +5,7 @@ import {
   uploadFile,
   createShareLink,
   deliverToGoogleDrive,
+  GDRIVE_MULTIPART_MAX_BYTES,
 } from "./gdrive.js";
 
 function route(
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- `src/delivery/gdrive.ts:193`: phase-2 PUT chunks correctly omit the `Authorization` header — the resumable session URL is pre-authenticated, so this is right (and matches OneDrive's anonymous-PUT-to-uploadUrl model). Verified-intentional, not a bug; flagging so nobody 'fixes' it by adding a bearer.
- `src/delivery/gdrive.ts:194`: comment says '256 KiB multiple per Drive's rules' on the `CHUNK = 8 * 1024 * 1024` line — technically true (8 MiB = 32 × 256 KiB) but reads oddly next to an 8 MiB constant. Optional: reword to 'multiple of 256 KiB, as Drive requires for non-final chunks'. Non-blocking.
- `src/delivery/gdrive.ts:198-214`: the loop advances `start` by a full `CHUNK` regardless of the 308 `Range` ack, and any non-2xx/non-308 chunk response (transient 5xx) throws and aborts the whole upload. This is 'resumable' only in the Drive-session sense, not crash/retry-resilient — fine for single-process delivery of a one-shot file, but worth knowing it won't recover a partially-acked chunk. Watch, not a blocker.
- Verified: init sets `X-Upload-Content-Type` (`:181`) and reads `location`/`Location` case-insensitively (`:188`); terminal detection on 200/201 sets `lastFile`, 308 continues, else throws (`:208-213`); null-guard on completion (`:216`). `uploadFile` (`:112-118`) dispatches `<= 5 MB` → `uploadMultipart`, else `uploadResumable`, so the existing small-file path is unchanged. Both new tests assert correctly: the 20 MiB case yields exactly 3 PUTs with `bytes 0-8388607/20971520` and `bytes 16777216-20971519/20971520`, and the no-Location case rejects with /no session URL/.

Clean implementation, well-scoped, both branches tested. The watch items are intentional simplifications, not defects.

PR #2239, opened 2026-06-08
fix(dashboard): resilient connections fetch + always-fresh quota (no manual refresh)

mekenthompson:
## Summary
Two dashboard-accuracy fixes from operator-reported symptoms:
- **Connections tab** showed no Microsoft/Google accounts (despite both connected) + "Failed to fetch agents".
- **Accounts tab** quota stayed empty until clicking "Refresh all quota".

## 1. Resilient connections fetch (`src/web/ui/index.html`)
`fetchConnections` used a bare `Promise.all`, so a single network blip on **any** of its 4 fetches (e.g. the web container restarting) rejected the whole batch → `renderConnections` never ran → the connected accounts "vanished". Now each fetch falls back independently (`.catch → default`), so Google/Microsoft/Notion render even if one endpoint blips. The 10s auto-refresh also tolerates one transient failure before flashing "Failed to fetch agents" over a healthy board.

## 2. Always-fresh quota (`src/auth/broker/server.ts`)
The 5h/7d quota only populated on the explicit "Refresh all quota" click (the cache started empty → "never probed"). Add a periodic `fleetQuotaProbeTick` that probes every account (a **1-token Haiku call** — negligible spend) and caches the snapshot `list-state` already serves, so the dashboard shows live quota **with no click**. Runs once at boot + on the consumer-probe interval; kill-switch `SWITCHROOM_DISABLE_FLEET_QUOTA_PROBE=1`. The cache-write is extracted into a shared `cacheQuotaSnapshot` helper (used by both the explicit op and the tick).

## Test plan
- [x] `vitest run src/auth/broker/server.test.ts` — 59 pass, incl. 2 new (tick populates `last_quota` for all accounts with no explicit probe; failed probe keeps the prior snapshot)
- [x] `tsc --noEmit` clean; frontend JS validated (`node --check`)
- [ ] Live after deploy: dashboard Connections shows both accounts through a refresh blip; Accounts quota fills in within a probe interval without clicking.

## Deploy note
Frontend change is baked into the `switchroom-web` image; broker change rides the `switchroom-broker`/auth-broker. Needs a release + web/broker recreate (the dashboard + brokers are separate compose projects the fleet roll doesn't auto-bounce — see the related memory).

## Risk
Low. Frontend: strictly more resilient. Broker: additive background probe (cheap, kill-switchable, never throws into the timer, never evicts a good snapshot on failure).

### DIFF
--- a/src/auth/broker/server.test.ts
+++ b/src/auth/broker/server.test.ts
@@ -1628,6 +1628,71 @@ describe("AuthBroker — probe-quota", () => {
       broker.stop();
     }
   });
+
+  it("fleetQuotaProbeTick populates last_quota for ALL accounts with no explicit probe (dashboard always-fresh quota)", async () => {
+    const h = makeHarness();
+    const config = makeConfig(h, {
+      active: "default",
+      fallback_order: ["default", "secondary"],
+      agents: { ziggy: {} },
+    });
+    seedAccount(h, "default", { expiresAt: 9_999_999_999_999 });
+    seedAccount(h, "secondary", { expiresAt: 9_999_999_999_999 });
+
+    const broker = new AuthBroker(config, {
+      home: h.home, stateDir: h.stateDir, socketRoot: h.socketRoot, disableRefreshLoop: true,
+      _testFetchQuota: async () => ({ ok: true, data: {
+        fiveHourUtilizationPct: 33,
+        sevenDayUtilizationPct: 12,
+        fiveHourResetAt: null,
+        sevenDayResetAt: null,
+        representativeClaim: null,
+        overageStatus: null,
+        overageDisabledReason: null,
+      } }),
+    });
+    const client = new AuthBrokerClient({ socket: join(h.socketRoot, "ziggy", "sock") });
+    try {
+      await broker.start();
+      // No probe-quota op called — just the background tick the timer would run.
+      await broker.fleetQuotaProbeTick();
+      const state = await client.listState();
+      for (const label of ["default", "secondary"]) {
+        const acct = state.accounts.find((a) => a.label === label);
+        expect(acct?.last_quota, `${label} should have cached quota`).not.toBeNull();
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `src/auth/broker/server.ts:451`: the new fleet-quota probe reuses `probeMs = resolveConsumerProbeIntervalMs(process.env)` (`:434`) for its cadence. So if an operator sets the consumer-probe interval to 0 to disable consumer probing, the dashboard's always-fresh quota silently disables too — even though it ships its own separate `SWITCHROOM_DISABLE_FLEET_QUOTA_PROBE` kill-switch implying independent control. Hidden coupling; consider a dedicated interval or a doc line. Non-blocking.
- `src/web/ui/index.html:507`: `safe(p, fallback)` correctly adds the missing `.catch` so one rejected fetch no longer rejects the whole `Promise.all` and blanks every connection tab (the real fix). Note the failing endpoint itself still falls back to its empty default, so that one provider's accounts blink out for a single ~10s refresh cycle while the others now survive — the two-strikes tolerance added to `fetchAgents` (`:443`) isn't applied here. Self-heals next refresh; minor.
- Verified `cacheQuotaSnapshot` (`:1230-1242`) is a faithful extraction of the prior inline block — identical fields, same `capturedAt: this.now()`, guarded `if (!result.ok) return` so a failed probe never evicts a good snapshot. The explicit-probe path (`:1209`) and the new fleet tick (`:1262`) both route through it, keeping the snapshot shape in one place.
- Verified `fleetQuotaProbeTick` (`:1253-1267`) iterates `listAccounts(this.home)`, skips accounts with no token, wraps `fetchQuotaImpl` in try/catch and `continue`s on error — it cannot throw into the `setInterval`. Boot tick is `void`-ed with a `.catch`, and `stop()` (`:504-507`) clears `fleetProbeTimer`. Both tests are meaningful: all-accounts population with no explicit probe, and no-eviction-on-503.

Clean fix, faithful refactor, both branches tested. The coupling on `probeMs` is the only thing worth a follow-up and it doesn't block merge.

PR #2240, opened 2026-06-09
feat(status): surface orphaned foreground sub-agents (extended autonomous work)

mekenthompson:
## Summary

When an agent keeps working **after its turn ends** (extended autonomous work — desirable), any **foreground** sub-agent it dispatches has no live parent turn to nest its status into. `onProgress` hit `if (currentTurn == null) return` and the work ran **completely invisible**.

Diagnosed on **finn**: a `reviewer` sub-agent ran for minutes with the operator seeing nothing; the registry showed it with `parent_turn_key = NULL` while the last turn had ended **12 minutes** earlier. Background workers already DM-fallback in this case (`origin?.chatId || fleetChatId || loadAccess().allowFrom[0]`); foreground sub-agents did not.

**Status visibility must not depend on the model holding a turn open.** A foreground sub-agent with no live turn now surfaces via the worker-activity feed (the same `🛠 Worker` edit-in-place message, routed to the dispatching chat or the owner DM via the existing fallback) — exactly like a background worker. In-turn foreground sub-agents still nest into the progress card, unchanged.

## Design

The routing is a **pure decision** (`telegram-plugin/gateway/subagent-status-surface.ts`):

```
resolveSubagentStatusSurface(input) → 'nest' | 'worker-feed' | 'legacy-relay' | 'skip'
```

- foreground + live turn → `nest` (unchanged)
- **foreground + NO live turn → `worker-feed`** (the fix; was an invisible `return`)
- background + feed on → `worker-feed`; background + feed off → `legacy-relay`
- orphaned foreground with the flag off → `skip` (exact pre-fix behaviour)

Both `onProgress` and `onFinish` route their foreground branch through it; the background path (and its handback) is untouched. Foreground results return inline as the Task tool result, so the orphan path finalizes the feed message and returns — **no** handback.

Kill switch `SWITCHROOM_ORPHAN_SUBAGENT_STATUS=0` restores the pre-fix behaviour.

## Determinism (proven, not sampled)

Per `feedback_prove_finite_fsm_not_sample`, `resolveSubagentStatusSurface` is proven by **total enumeration of its 4-boolean / 16-row input space**, with an independent spec encoding and the load-bearing invariants:
- **INV-ORPHAN-VISIBLE** — an orphaned foreground sub-agent is never `skip` when the flag and feed are on.
- **INV-KILL-SWITCH** — flag off ⟹ exactly the pre-fix `skip`.
- **INV-NEST-UNCHANGED / INV-BACKGROUND-UNCHANGED** — the existing paths don't move.

## Test plan
- [x] `subagent-status-surface.test.ts` — total-enumeration proof (16) + invariants + flag parser (14 tests).
- [x] No regression across the subagent/worker surface: worker-activity-feed, worker-feed-dispatch (incl. its 20k-row property sweep), subagent-watcher, subagent-tracker-hooks, streaming-orchestration, resolve-calling-subagent.
- [x] `tsc --noEmit` clean · full `bun lint` chain clean · gateway bundle contains the new logic.
- [x] **238 files / 3845 vitest + 77 bun tests green.**

## Risk
Behaviour changes only for a foreground sub-agent with **no live turn** (previously invisible → now a worker-feed message in the dispatching chat / owner DM). In-turn nesting and all background paths are byte-unchanged (proven by the invariants). Kill-switched. Multi-step autonomous agents (finn, etc.) are the beneficiaries.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -471,6 +471,10 @@ import {
 } from './resume-inbound-builder.js'
 import { applySubagentsSchema, getSubagentByJsonlId } from '../registry/subagents-schema.js'
 import { resolveWorkerFeedDispatch, type WorkerFeedDispatch } from './worker-feed-dispatch.js'
+import {
+  resolveSubagentStatusSurface,
+  isOrphanSubagentStatusEnabled,
+} from './subagent-status-surface.js'
 import { formatIdleFooter } from '../idle-footer.js'
 import { resolveCallingSubagent } from './resolve-calling-subagent.js'
 
@@ -20410,6 +20414,11 @@ void (async () => {
             // compose draft, so no answer-stream contention). The kill-switch
             // disables only the nesting; the parent's own feed is unaffected.
             const foregroundNestingEnabled = process.env.SWITCHROOM_FOREGROUND_SUBAGENT_NESTING !== '0'
+            // Orphaned-foreground status (2026-06-09): a FOREGROUND sub-agent
+            // with no live parent turn to nest into (dispatched outside a turn,
+            // or the turn ended while it kept running — extended autonomous
+            // work) is surfaced via the worker feed instead of vanishing.
+            const orphanStatusEnabled = isOrphanSubagentStatusEnabled(process.env.SWITCHROOM_ORPHAN_SUBAGENT_STATUS)
             const workerActivityFeed = createWorkerActivityFeed({
               bot: {
                 sendMessage: async (cid, text, sendOpts) => {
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- `subagent-status-surface.ts:54-66`: the pure `resolveSubagentStatusSurface` is logically equivalent to the independent `spec()` encoding in the test (`subagent-status-surface.test.ts:60-65`) — two separate encodings cross-checking each other, and the 16-row total enumeration (`:80-82`) asserts impl === spec on every input plus totality, determinism, and four named invariants. Correct and well-proven.
- `gateway.ts:20630-20641` (onFinish orphan-finish): correctly placed AFTER the nested-foreground branch (`:20590` `has(agentId)` → `return`), so no double-finish. `liveTurnPresent: false` is hardcoded here intentionally — this path is only reached when the sub-agent was NOT nested, which is exactly the 'was not nested' semantics the input doc (`:38-44`) documents. `workerActivityFeed.finish` is a documented no-op when no feed message was posted, so a foreground finish that never surfaced stays silent. Sound.
- `gateway.ts:20768-20800` (onProgress): both new call sites sit inside `if (!isBackground)` (onFinish split at `:20580`, onProgress at `:20768`), so the hardcoded `isBackground: false` faithfully matches the runtime context — no risk of mislabeling a background worker. The 'worker-feed' branch reuses the same `origin?.chatId || fleetChatId || loadAccess().allowFrom[0]` fallback chain background workers already use; `surface !== 'nest' → return` correctly preserves pre-fix 'skip' behaviour, and the trailing `if (turn == null) return` is sound defensive code since 'nest' implies a live turn.
- Verified scope: `workerFeedEnabled` (`:20410`) and `orphanStatusEnabled` (`:20421`) are declared in the same closure as the handlers; `loadAccess` (`:985`) and `resolveSubagentOriginChat` (`:1197`) are module-level. `isOrphanSubagentStatusEnabled` defaults ON, `=0` restores pre-fix invisibility — a clean kill switch (`subagent-status-surface.ts:68`).

New pure module is fully enumerated and proven; the gateway wiring routes orphaned foreground sub-agents through the existing worker feed without touching the nested-default or background paths. No blockers.

PR #2241, opened 2026-06-09
chore: release v0.14.91

mekenthompson:
Release v0.14.91 — surface orphaned foreground sub-agents (#2240). Reviewed (fresh-process APPROVE, mutation-verified) + CI-green on merge.

PR #2235, opened 2026-06-08
chore: release v0.14.90

mekenthompson:
Release v0.14.90 — framework owns the reply topic (#2233). Reviewed (fresh-process APPROVE) + CI-green on merge.

PR #2242, opened 2026-06-09
fix(ci): make the required `uat-gate` an always-running sentinel

mekenthompson:
## Summary

#2237 made `uat-gate` a **branch-protection-required** status check, but `ci-uat.yml` is **path-gated** (`telegram-plugin/**`, `src/agents/**`, `uat/**`, …). A PR that doesn't touch those paths never triggers the workflow → the required `uat-gate` context is **never produced** → GitHub blocks the merge forever waiting for it.

This **deadlocks every non-UAT PR**: releases (which only bump `package.json`), docs, docker, scripts. The v0.14.91 release PR (#2241) is the first release after #2237 and is stuck on exactly this. (Confirmed: `UAT_GATE_ENABLED=true`, #2235/v0.14.90 merged fine before #2237.)

## Fix — the repo's own sentinel pattern

Mirrors `ci-tests-plugin.yml` (`changes → bun-test-run → bun-test`):

- **Workflow triggers on every PR** — the `pull_request` paths filter is removed so the sentinel always reports.
- **`changes` job** (`dorny/paths-filter@fbd0ab8…`, same pin as the other workflows) detects the same behavioural surface the old filter gated on.
- **`uat-gate-run`** = the heavy, runner-bound round-trips, now gated on **both** a relevant path change **and** `UAT_GATE_ENABLED` — so docs/docker/release merges still don't spin up the slow self-hosted runner (operator intent preserved).
- **`uat-gate` sentinel** (ubuntu-latest, `if: always()`) reports the required context: **PASS** when the heavy run succeeded or was legitimately skipped (non-UAT PR / gate off), **FAIL** on a real UAT failure or a path-filter infra failure.

Net: real UAT gating is preserved (a true UAT failure still blocks); the path-gated-required-check deadlock is gone. The `push` trigger keeps its path filter (main merges aren't a required gate).

## Note
Because this PR touches `.github/workflows/ci-uat.yml`, `changes.relevant=true` for it, so `uat-gate-run` actually runs the live UAT on the self-hosted runner here — i.e. this PR also proves the gate still gates.

## Risk
CI-gating change. The sentinel's verdict logic is copied verbatim from the proven `bun-test` sentinel (fail on `failure`/`cancelled`, pass on `success`/`skipped`), so it cannot silently un-gate a real failure. Reversible.

### DIFF
--- a/.github/workflows/ci-uat.yml
+++ b/.github/workflows/ci-uat.yml
@@ -35,12 +35,13 @@ name: ci-uat
 on:
   pull_request:
     branches: [main]
-    paths:
-      - "telegram-plugin/**"
-      - "src/agents/**"
-      - "telegram-plugin/uat/**"
-      - "vitest.uat.config.ts"
-      - ".github/workflows/ci-uat.yml"
+    # NO paths filter: the workflow must run on EVERY PR so the required
+    # `uat-gate` SENTINEL always reports a status. A path-gated required check
+    # deadlocks any PR that doesn't touch the listed paths (a release PR only
+    # bumps package.json) — the context is never produced and the merge blocks
+    # forever (#2237 → release v0.14.91 stuck). The HEAVY, runner-bound UAT is
+    # still path-gated, now via the `changes` job below, so docs/docker/release
+    # PRs don't spin up the self-hosted runner.
   push:
     branches: [main]
     # Same path scope as the PR trigger — a docs/docker/scripts merge must not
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

- Verified the load-bearing fact: ruleset 16470166 ('main branch protection') lists `uat-gate` as a required status check, and the sentinel job id at `.github/workflows/ci-uat.yml:165` is `uat-gate` with no `name:` override — so it produces exactly the required context. The fix is correctly targeted; the deadlock is real and this resolves it.
- `.github/workflows/ci-uat.yml:172-186` (sentinel verdict): fail-closed and correct. Blocks on `changes` = failure/cancelled (`:178` — an untrustworthy path-filter skip), blocks on `uat-gate-run` = failure/cancelled (`:182` — a real UAT failure), passes only on success or a legitimate skip. A true UAT failure still blocks merge, so gating is preserved; this cannot silently un-gate.
- Verified the precedent claim: `ci-tests-plugin.yml` uses the same `changes → bun-test-run → bun-test` triad (`needs: [changes, bun-test-run]`, `if: always()`, same `c`/`r` result checks, same exit-1 conditions). The new sentinel is a faithful copy of that proven pattern.
- `.github/workflows/ci-uat.yml:113`: `uat-gate-run` correctly gates on BOTH `vars.UAT_GATE_ENABLED == 'true'` AND `needs.changes.outputs.relevant == 'true'`, preserving the operator intent (no self-hosted runner spin-up for docs/docker/release PRs) that the removed workflow-level path filter used to enforce. The `changes` filter even adds `.github/actions/setup-switchroom/**` — a real behavioural dependency the old filter missed. The `push` trigger keeps its path filter (`:43`), correct since main-merge isn't a required gate.

Watch (non-blocking): when `UAT_GATE_ENABLED` is false, `uat-gate-run` skips and the required `uat-gate` reports green with zero behavioural coverage — this is the intended 'invisible until opt-in' behaviour, but means the required check is signal-free in that state. Since the repo var is on, real gating is active.

Correct fix, mirrors a proven in-repo sentinel verbatim, targets the verified required-context name, fail-closed on both real failures and path-filter infra failures. Safe to merge.

PR #2243, opened 2026-06-09
ci(uat): remove the nightly fuzz cron (burns subscription quota)

mekenthompson:
Drops the `schedule: cron "41 4 * * *"` from ci-uat.yml — each fuzz scenario is a real claude turn on the fleet-active account, so an automatic nightly pass burns quota competing with the live fleet. The thorough fuzz pool becomes workflow_dispatch-ONLY (run deliberately during UAT). Fast `uat-gate` unaffected. UAT is disabled fleet-wide (UAT_GATE_ENABLED=false) regardless.

### DIFF
--- a/.github/workflows/ci-uat.yml
+++ b/.github/workflows/ci-uat.yml
@@ -54,9 +54,10 @@ on:
       - "telegram-plugin/uat/**"
       - "vitest.uat.config.ts"
       - ".github/workflows/ci-uat.yml"
-  schedule:
-    # Nightly thorough fuzz pass (the uat-fuzz job below is schedule/dispatch-only).
-    - cron: "41 4 * * *"
+  # No `schedule:` — the nightly fuzz cron was removed (2026-06-09): each fuzz
+  # scenario is a real claude turn on the subscription, and an automatic nightly
+  # pass burns quota that competes with the live fleet. The thorough fuzz pool is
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE

- `.github/workflows/ci-uat.yml:57-60` — removes the `schedule: cron "41 4 * * *"` trigger, replacing it with an explanatory comment. The only behavioral change: `uat-fuzz` no longer runs automatically nightly. Correct, given each fuzz scenario is a real claude turn on the fleet-active subscription account.
- `uat-fuzz` (line 192) guards on `github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'` — the `'schedule'` arm is now dead but harmless; `workflow_dispatch` still triggers the thorough pool on demand, so deliberate UAT runs are unaffected. No need to also strip the dead OR arm (non-blocking).
- `uat-gate` sentinel (line 166, `needs: [changes, uat-gate-run]`) is driven by pull_request/push/workflow_dispatch — untouched by removing `schedule:`. The required check stays satisfiable and the fast gate is unaffected, as the PR body claims.
- `uat-fuzz` is non-required (the required context is `uat-gate`, confirmed against branch protection in #2242), so dropping its only automatic trigger cannot leave a required check unsatisfiable.

Rationale: a one-line cron deletion that stops an automatic nightly quota burn; the thorough pool stays available on-demand and no required gate is affected.

PR #2244, opened 2026-06-09
feat(cron): cheap cron — deterministic polls + Sonnet cron-session (RFC + L1-L3)

mekenthompson:
## Summary

Cron fires are expensive because **every fire injects a turn into the agent's
live session** — paying the full standing-context cache-read tax + the agent's
model — even when it's a `*/10` poll that finds nothing. The two crons we
disabled 2026-06-09 (`clerk` telegram-capture `*/10`, `marko` lead-alert `*/15`)
are ~240 no-op fires/day, almost all wasted.

This lands the **RFC** (`docs/rfcs/cheap-cron-sessions.md`, reviewed by a 5-lens
adversarial pass) and the **L1–L3 core + plumbing**, all behind
`SWITCHROOM_CHEAP_CRON` (default **off** → no-op in prod until canary).

Three tiers, cheapest-first:
- **Tier 0** — declarative model-free poll; no change → `HEARTBEAT_OK` (zero
  model tokens), only a hit escalates. *This is the cost win.*
- **Tier 1** — a hit/light cron runs in a minimal-context `claude --model sonnet`
  cron session (gateway plumbing landed here; the session launch is the staged
  follow-up).
- **Tier 2** — `context: agent` injects into the live session, exactly as today.

Routing key = the **reactivated `ScheduleEntry.model`** + new `kind`/`context`.

## What's in this PR (built + tested)

- **L1** schema (`kind`/`context`/`poll`, un-deprecated `model`) + pure
  `resolveCronRouting` (flag-off = today's behaviour) — **total-enumeration**
  test (90 inputs) per the prove-by-enumeration standard.
- **L2** gateway routing to a derived `<agent>-cron` bridge — reuses the hardened
  single-bridge IPC machinery untouched (no `agentIndex`/buffer rekey) and gates
  the singleton status machinery off the cron identity (= §2.4 status-silence).
- **L3** Tier-0 **http-diff** engine + **SSRF/egress guard** (resolves the review
  BLOCK: https-only allowlist + host-pinned secret bindings + loopback/private/
  rebind rejection) + **write-ahead poll-state** (double-escalation guard) +
  first-run-no-escalate + `attemptFire` tier wiring.
- `cron.egress` operator-owned config block.

## Staged (next, canary-gated) — see RFC "Build status"

`main()` live-wiring (vault-broker secret-resolver — strict vault test
isolation), **L4** lazy Sonnet cron-session launch (2nd `claude`), **L5**
`schedule_add` cheap-default + `schedule report`, `telegram-reactions` poll.
With escalate-to-main, **Tier-0 delivers its full saving without L4**.

## Test plan

- [x] `tsc --noEmit` clean
- [x] L1 routing: 23 (total enumeration) · L2: 4 · L3: 45 engine/egress + 6 scheduler-integration
- [x] config suite 282 green; existing 22 scheduler tests green (back-compat)
- [ ] Pre-existing `idle-smoke.test.ts` fails in-worktree (symlinked node_modules) — unrelated, fails identically on base

## Compliance (pillar 3)

Tier 0 is model-free (outside the inference surface). Tier 1/2 are interactive
`claude` (no `-p`). No SDK/API key. Reviewed APPROVE on compliance.

## Risk

Flag default-off → dormant. SSRF guard is the load-bearing new security surface
(35 adversarial tests). The staged live-wiring touches the vault broker and a
2nd `claude` process — deliberately deferred to a focused canary PR.

### DIFF
--- a/docs/rfcs/cheap-cron-sessions.md
+++ b/docs/rfcs/cheap-cron-sessions.md
@@ -1,0 +1,379 @@
+# RFC — Cheap cron: deterministic polls + a per-agent Sonnet cron session
+
+**Status:** Draft, post-review (rev 2)
+**Author:** (agent-authored, operator-directed)
+**Targets:** `upstream/main` @ v0.14.91
+**Compliance pillar touched:** 3 (Claude-native, subscription-honest) — preserved by construction; see §7.
+
+---
+
+## Review verdict (5-lens adversarial, rev 1 → rev 2)
+
+| Lens | rev 1 | After rev-2 changes |
+|------|-------|---------------------|
+| Compliance (pillar 3) | **APPROVE** | unchanged — Tier 0 model-free; Tier 1/2 interactive, no `-p` |
+| Feasibility | REQUEST_CHANGES | §3.3 rewritten to own the full gateway refactor scope |
+| Resources/Ops | REQUEST_CHANGES | §2.2 runtime shape → **lazy session (B3)**; double-escalation §2.1 |
+| Security | **BLOCK** | §6.1 **SSRF/exfil hardening** added; §6.2 operator-commit boundary |
+| Completeness | REQUEST_CHANGES | §2.4 status isolation; §5 observability; §3.1 custom-model migration |
+
+The BLOCK was the `http-diff` SSRF/secret-exfil surface (§6.1). It is resolved at
+the design level here; the build must implement the egress guard + host-pinned
+secret bindings before any Tier-0 poll ships. Everything else was either
+"code-not-yet-written" (expected for a spec) or a genuine sharpening folded in
+below.
+
+---
+
+## 0. TL;DR
+
+Cron fires are expensive for one structural reason: **every fire injects a
+synthesized turn into the agent's live interactive session**, so it pays the
+full standing-context cache-read tax (system prompt + ~31k-token MCP schema
+surface + the whole running conversation) *and* runs at the agent's configured
+model — even when the fire is a `*/10` poll that finds nothing to do.
+
+Three tiers, cheapest-first:
+
+- **Tier 0 — deterministic poll (no model).** Poll-heavy crons run a *declarative,
+  operator-approved* poll in the scheduler process. No new data → `HEARTBEAT_OK`,
+  **zero model tokens.** Only a hit escalates.
+- **Tier 1 — per-agent cheap cron session.** Crons that need a model but not the
+  agent's persona/memory fire into a **minimal-context `claude --model sonnet`
+  session**, addressed by a new gateway session label `cron`. ~5–20× cheaper.
+- **Tier 2 — main session (status quo).** Crons that need the agent's full context
+  inject into the live session exactly as today.
+
+Routing key = the **dead-but-present `ScheduleEntry.model` field** (reactivated) +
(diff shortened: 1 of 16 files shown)


mekenthompson:
APPROVE

- \`src/scheduler/cron-routing.ts:64\` — flag-off (\`cheapCronEnabled:false\`) returns \`{tier:"main", session:"main"}\` unconditionally, so kind/model/context are inert and disabling the flag can never silently drop a cron. This is the load-bearing safety property and it holds; combined with default-off \`isCheapCronEnabled\` (line ~57), the PR is a no-op in prod until canary. Correct.
- \`src/scheduler/poll-engine.ts:104\` — \`checkHttpDiffEgress\` runs BEFORE any DNS/secret/fetch, and \`src/scheduler/poll-egress.ts:90-101\` enforces https-only + userinfo-reject + operator-owned host allowlist + IP-literal blocked-range check. The DNS-rebind pin re-checks the resolved IP at poll-engine.ts:112, and the fetch uses \`redirect:"error"\` + 5s timeout + 1MiB cap. This resolves the prior-review SSRF/exfil BLOCK. Solid.
- \`src/scheduler/poll-egress.ts:78\` — IPv4-mapped IPv6 (\`::ffff:a.b.c.d\`) correctly delegates to \`isBlockedV4\` on the embedded literal, so a loopback-via-mapped-v6 bypass is closed. Secret host-binding (\`checkSecretBinding\`) means an approved poll cannot ship a vaulted secret to a different host. Correct.
- \`src/scheduler/poll-state.ts\` — write-ahead (advance cursor + stage \`pendingEscalation\` in one atomic temp+rename) plus first-run-no-escalate are both encoded; a restart mid-escalation re-polls, sees the advanced cursor, and does not re-fire. Sound.
- \`telegram-plugin/gateway/gateway.ts:5478-5489, 5739, 5977, 6317-6338\` — cron identity gates the singleton status machinery off at register/onSessionEvent/onPtyPartial (early return) and onInjectInbound routes to \`resolveInjectTarget\` + buffers under the SAME cron target, so a fire landing mid cron-session-spawn drains to the cron bridge on register. \`markClaudeBusy\` is correctly skipped for cron fires. Consistent.
- NON-BLOCKING: \`src/scheduler/poll-engine.ts:166\` \`readCapped\` calls \`res.arrayBuffer()\` and checks \`byteLength\` AFTER buffering the whole body — an allowlisted-but-hostile endpoint could balloon memory before the cap rejects. Bounded by the 5s timeout and operator-owned allowlist, but a streaming/Content-Length pre-check would be tighter.
- NON-BLOCKING / CANARY WATCH: L4 (the 2nd \`claude\` cron-session launch) is staged, so with the flag ON a Tier-1 \`context:fresh\` fire routes to a \`<agent>-cron\` bridge that nothing spawns yet and buffers indefinitely. Documented in the PR body and canary-gated; the \`escalate-to-main\` path covers \`context:agent\`. Just confirm the canary only exercises poll/agent tiers until L4 lands.

Rationale: default-off feature flag, flag-off path is byte-identical to today, and the one genuinely new security surface (Tier-0 http-diff egress) is correctly fenced with the prior SSRF BLOCK resolved and 35+ adversarial tests; staged gaps are documented and dormant.

PR #2245, opened 2026-06-09
feat(cron): make Tier 0 live + observable (escalate-to-main, recovery, report)

mekenthompson:
## Summary

Follow-up to #2244 (the L1–L3 core, merged). That PR landed the routing,
gateway plumbing, and the Tier-0 engine + SSRF guard, but left the engine
**not wired into the scheduler boot** — dormant. This PR makes **Tier 0 live and
canary-able on `marko` without the second session**, behind the same
`SWITCHROOM_CHEAP_CRON` flag (default **off**).

With `escalate-to-main` (a poll entry with no `model`/`context` → Tier 2), a hit
fires into the live session exactly as today — so Tier 0 delivers its **full
cost saving** (marko's ~96 wasted no-op fires/day → ~0 model tokens) with zero
new session infrastructure.

## What's in this PR

- **`buildCheapCronHooks(config, env)`** — constructs the `CheapCronHooks` from
  `cron.egress` config + env + the **vault-broker secret resolver**, gated on the
  flag (off → `undefined` → `registerAgentSchedule` unchanged). Every external
  dep (broker resolver, fetch, DNS, poll-state) is **injectable with a real
  default**, so tests pass fakes and **never touch the real broker/network** (the
  vault test-isolation discipline — CLAUDE.md + the 2026-05-22 clobber incident).
- **`recoverPendingEscalations()`** at boot — the reviewer's lost-hit fix from
  #2244: a crash after a poll's write-ahead cursor advance but before delivery
  left a dangling `pendingEscalation`; this re-dispatches it once and clears it.
  Closes the window the prior PR's comment flagged as open.
- **`switchroom schedule report [agent]`** (§5 observability) — reads
  scheduler.jsonl and shows the saving: Tier-0 poll (free) vs Tier-1 cheap vs
  Tier-2 main fire counts + a relative cost-weight proxy. This is how the canary
  is measured. `--jsonl <path>` / `--since` / `--json`.

## Still staged (next PR)

**L4** the lazy Sonnet cron session (start.sh fork + trimmed `.mcp.json`), which
makes the *hit* path and non-poll crons cheap too; **L5** `schedule_add`
cheap-default params + escalation staging; the `telegram-reactions` poll (needs
the net-new internal gateway reactions verb) — it's schema-accepted but records a
poll-error (never a silent success) until then.

## Test plan

- [x] `tsc --noEmit` clean
- [x] 7 wiring/recovery tests (all fakes — no broker/network) + 5 report tests
- [x] full scheduler sweep 128 green; existing 22 scheduler tests green (back-compat: flag-off → no hooks)
- [ ] **Canary (post-merge):** flag-on `marko` only, re-enable lead-alert as a `kind: poll` http-diff (escalate-to-main), confirm no-op fires record `HEARTBEAT_OK` with zero model tokens via `schedule report`, and a seeded hit escalates + emails.

## Compliance / risk

Flag default-off → dormant. Tier 0 is model-free (outside the inference surface);
escalations are interactive `claude` (no `-p`). The secret resolver goes through
the agent's vault broker (ACL by socket identity) and the §6.1 egress guard
fences where each secret may go. No SDK/API key.

### DIFF
--- a/src/agent-scheduler/cheap-cron-wiring.test.ts
+++ b/src/agent-scheduler/cheap-cron-wiring.test.ts
@@ -75,9 +75,10 @@ describe("recoverPendingEscalations", () => {
     poll: { type: "http-diff", url: "https://api.brevo.com/x", method: "GET", secrets: [], diff_jsonpath: "$.a", state_key: "k" },
   };
 
-  it("re-dispatches a dangling pendingEscalation once and clears it", () => {
+  it("re-dispatches a dangling pendingEscalation once, clears it, and audits it", () => {
     const pollState = createMemoryPollStateStore({ k: { value: "9", pendingEscalation: "cursor 9 > 5 (1)" } });
     const dispatcher = capture();
+    const fires: Array<{ tier?: string; outputSummary: string }> = [];
     const n = recoverPendingEscalations({
       entries: [pollEntry],
       pollState,
@@ -86,11 +87,15 @@ describe("recoverPendingEscalations", () => {
       cheapCronEnabled: true,
       now: () => 1000,
       log: () => {},
+      sink: { recordFire: (r) => fires.push(r) },
     });
     expect(n).toBe(1);
     expect(dispatcher.calls).toHaveLength(1);
     expect(dispatcher.calls[0]!.msg.text).toBe("New: cursor 9 > 5 (1)");
     expect(pollState.get("k")?.pendingEscalation).toBeUndefined();
+    // recovered hit is audited so `schedule report` counts it (reviewer note #1)
+    expect(fires).toHaveLength(1);
+    expect(fires[0]).toMatchObject({ tier: "main", outputSummary: "recovered pending escalation (boot)" });
   });
 
   it("no dangling escalation → no dispatch", () => {
--- a/src/agent-scheduler/cheap-cron-wiring.test.ts
+++ b/src/agent-scheduler/cheap-cron-wiring.test.ts
@@ -1,0 +1,112 @@
+/**
+ * Live-wiring + boot recovery tests. All deps are faked — NO real vault
+ * broker or network is touched (vault test-isolation discipline).
+ */
+import { describe, expect, it } from "vitest";
+import type { SwitchroomConfig } from "../config/schema.js";
+import { buildCheapCronHooks } from "./cheap-cron-wiring.js";
+import { recoverPendingEscalations } from "./index.js";
+import { createMemoryPollStateStore } from "../scheduler/poll-state.js";
+import type { InboundDispatcher, InboundMessageWire, SchedulerEntry } from "../scheduler/dispatch.js";
+
+const CONFIG = {
+  cron: { egress: { allowed_hosts: ["api.brevo.com"], secret_bindings: { brevo_api_key: "api.brevo.com" } } },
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

- \`src/agent-scheduler/cheap-cron-wiring.ts:46\` — \`buildCheapCronHooks\` returns \`undefined\` when \`isCheapCronEnabled(env)\` is false, so \`registerAgentSchedule\` gets no hook and behaviour is exactly today's. Flag-off path is byte-identical; combined with #2244 this stays dormant in prod until the marko canary. Correct.
- \`src/agent-scheduler/cheap-cron-wiring.ts:30-44, 56-62\` — every external dep (broker \`resolveSecret\`, \`fetchImpl\`, DNS \`lookup\`, \`pollState\`) is injectable with a real default, so the 7 wiring/recovery tests pass fakes and never touch the real broker/network. Honours the vault test-isolation discipline. Good.
- \`src/agent-scheduler/cheap-cron-wiring.ts:78\` — the \`telegram-reactions\` poll type returns \`{hit:false, error:...}\` (poll-error, no escalation) rather than a silent success, matching the staged-but-safe claim. Correct.
- \`src/agent-scheduler/index.ts\` (\`recoverPendingEscalations\`) — only fires for \`kind:poll\` entries with a dangling \`pendingEscalation\`, re-dispatches once with \`templateEscalationPrompt(prompt, pendingEscalation)\`, and clears pending ONLY on \`delivered\` (else logs + retries next boot). This closes the lost-hit window my #2244 review flagged (crash after write-ahead cursor advance, before delivery). The at-most-once-deliver semantics match the live path. Sound.
- \`src/scheduler/schedule-report.ts\` — pure aggregation: legacy untiered rows count as \`main\` (back-compat), \`parseScheduleJsonl\` skips malformed lines and requires a numeric \`exitCode\`, \`-2\` is deferred vs \`<0\` errors, and the cost-weight is explicitly labelled a relative proxy, not dollars. Read-only observability, no risk.
- \`src/cli/agent-config-write.ts\` — \`schedule report\` reads in-container \`scheduler.jsonl\` via \`docker exec\` (or \`--jsonl\`), validates \`--since\`, exits non-zero with a clear message when the container read fails. Mirrors the existing inspection pattern. Fine.
- NON-BLOCKING (security, defense-in-depth): \`src/agent-scheduler/cheap-cron-wiring.ts:38\` \`defaultLookup\` resolves the host and the engine checks \`isBlockedAddress\` on that result, but the real \`fetch\` performs its OWN DNS resolution at connect time — the pinned IP and the connected IP are not bound to the same socket, so the rebind pin is TOCTOU and a multi-record / fast-flip host can still connect to an unvetted IP. The operator-owned allowlist is the real control (exploit needs DNS authority over an allowlisted host), so this is acceptable defense-in-depth, but the #2244 egress docstring slightly oversells it as a true rebind pin. Consider a custom \`lookup\` on the fetch agent (or undici dispatcher) to bind resolve-and-connect when L4 lands.

Rationale: default-off feature flag with a byte-identical off-path, correct closure of the prior lost-hit window, injectable deps preserving vault test isolation, and pure read-only observability; the one real gap (rebind-pin TOCTOU) is bounded by the operator allowlist and worth tightening but not a blocker for a dormant, canary-gated change.

PR #2246, opened 2026-06-09
feat(cron): L4 — per-agent cheap Sonnet cron session (dormant, flag-gated)

mekenthompson:
## Summary

The last cheap-cron build piece (after #2244 core + #2245 live-wiring). A
**second interactive `claude`** (no `-p`) in the agent container, dedicated to
Tier-1 (`context: fresh`) cron fires — registering to the gateway as the
`<name>-cron` bridge whose routing + status-silence already landed in #2244. It
gives the *hit* path and non-poll crons a cheap, minimal-context session instead
of the live Opus/Sonnet session.

**Dormant + flag-gated.** `cronSessionEnabled` is purely config-derived — true
iff the agent has a `context: fresh` entry, a brand-new field **no current fleet
agent has** — so the `start.sh` fork block renders **empty** and every current
agent's boot is byte-identical. The runtime `SWITCHROOM_CHEAP_CRON` flag is a
second gate inside the launcher.

## What's in this PR

- **`profiles/_base/cron-session.sh.hbs`** — the launcher. Interactive `claude`
  in its own tmux session/socket, `SWITCHROOM_AGENT_NAME=<name>-cron` (the
  switchroom-telegram bridge reads it from the process env → correct identity),
  own `CLAUDE_CONFIG_DIR` (separate transcript), shares the broker-managed OAuth
  creds via symlink (same subscription). Runtime kill-switch: `exit 78` (no
  respawn) when the flag is off.
- **`start.sh.hbs`** — one `_switchroom_supervise` fork, entirely inside
  `{{#if cronSessionEnabled}}`.
- **`scaffold.ts`** — computes `cronSessionEnabled` (via `scheduleNeedsCronSession`,
  reusing the L1 routing) for both the scaffold and reconcile render paths;
  renders `cron-session.sh` only when enabled.

## Load-bearing safety

The one risk here is the boot path. The structural test asserts **every**
cron-fork line lives inside a single `{{#if cronSessionEnabled}}` guard, and the
existing `start.sh`-render tests (gateway-env-order, reconcile, supervisor-backoff)
confirm the template stays well-formed. A test **caught a real bug** pre-merge: an
explanatory comment that contained a literal handlebars token would have broken
the `start.sh` render fleet-wide — fixed.

## Deliberately v1 (canary-time refinements, RFC §4)

- Reuses the agent's **main `.mcp.json`** (full surface). The bigger wins (fresh
  context + the cheap model) already hold; trimming the MCP to just
  switchroom-telegram is a follow-up.
- **No compose mem/pids bump** — RFC §4 says start at the default cap and bump
  only if the canary shows burst pressure.
- **End-to-end is dormant** — bridge registration / reply delivery / `/clear` for
  the second session need the live canary to validate (you deferred that).

## Test plan

- [x] `tsc --noEmit` clean
- [x] 15 L4 render/gate tests (incl. the byte-identical-when-off guard property)
- [x] existing start.sh-render + cli-contract + mcp-parity suites green (no regression)
- [ ] **Canary (post-merge, with Tier-0):** an agent with a `context: fresh` cron + flag-on; confirm the `<name>-cron` bridge registers, a fire delivers + replies, and the main session's card/feed stay untouched.

## Compliance / risk

Second session is interactive `claude` (no `-p`); shares the broker OAuth — no
SDK/API key. Flag + config double-gated → dormant fleet-wide.

### DIFF
--- a/profiles/_base/cron-session.sh.hbs
+++ b/profiles/_base/cron-session.sh.hbs
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+# Tier-1 cheap cron SESSION launcher — docs/rfcs/cheap-cron-sessions.md §2.2.
+#
+# A SECOND interactive `claude` (no -p — compliance pillar 3) in the agent
+# container, dedicated to cheap cron fires. It registers to the SAME gateway
+# as a DISTINCT bridge identity `<name>-cron` (start.sh's gateway sidecar
+# routes meta.session=cron fires here and gates all status surfaces off this
+# identity — see telegram-plugin/gateway/cron-session.ts). Minimal context:
+# its own CLAUDE_CONFIG_DIR (separate transcript) + a TRIMMED .mcp.json (only
+# switchroom-telegram) → it pays a fraction of the main session's schema +
+# cache-read tax, at the cheap cron model.
+#
+# Forked as a supervised sidecar by start.sh ONLY when {{name}} actually has a
+# Tier-1 (context: fresh) cron entry AND SWITCHROOM_CHEAP_CRON is on — so the
+# whole feature is inert for the rest of the fleet (the start.sh fork block
+# renders empty). This script is a no-op artifact on disk until then.
+set -u
+
+# Runtime kill-switch. The fork is baked into start.sh whenever {{name}} has a
+# context:fresh cron entry (a config property), but the session only actually
+# runs when SWITCHROOM_CHEAP_CRON is on at runtime. Exit 78 (EX_CONFIG) so the
+# supervisor does NOT respawn-loop when the feature is off — a clean idle.
+case "${SWITCHROOM_CHEAP_CRON:-}" in
+  1 | true | on | ON) : ;;
+  *)
+    echo "cron-session: SWITCHROOM_CHEAP_CRON off — not starting (exit 78, no respawn)" >&2
+    exit 78
+    ;;
+esac
+
+CRON_NAME="{{name}}-cron"
+CRON_CONFIG_DIR="{{agentDir}}/.claude-cron"
+MAIN_CONFIG_DIR="{{agentDir}}/.claude"
+# v1: reuse the agent's project .mcp.json. The switchroom-telegram BRIDGE reads
+# SWITCHROOM_AGENT_NAME from the process env (exported below) — NOT its .mcp.json
+# env block — so it registers as the cron identity even off the shared config.
+# The bigger wins (fresh context + the cheap model) already hold; trimming the
+# MCP surface to just switchroom-telegram is a documented canary-time follow-up.
+CRON_MCP_CONFIG="{{agentDir}}/.mcp.json"
+CRON_SOCKET="switchroom-${CRON_NAME}"
+
+mkdir -p "$CRON_CONFIG_DIR"
+
+# Share the broker-managed OAuth creds (same subscription) without sharing the
+# main session's transcript/settings. The broker is the sole writer of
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- \`profiles/_base/start.sh.hbs\` — the entire cron-session fork (the \`_switchroom_supervise cron-session\` block) lives inside a single \`{{#if cronSessionEnabled}}\` guard, so when the flag is false the block is absent and start.sh is byte-identical. The fork also guards on \`command -v claude\` + \`[ -f cron-session.sh ]\`. Structurally correct.
- \`profiles/_base/cron-session.sh.hbs\` — runtime kill-switch is a clean idle: \`exit 78\` (EX_CONFIG) when \`SWITCHROOM_CHEAP_CRON\` is off, chosen so the supervisor does not respawn-loop. Exports \`SWITCHROOM_AGENT_NAME=<name>-cron\` (the bridge reads it from process env → correct cron identity per #2244), uses its own \`CLAUDE_CONFIG_DIR\`, symlinks the broker-managed \`.credentials.json\` (same subscription, separate transcript), interactive \`claude\` (no -p). Sound.
- \`src/agents/scaffold.ts\` — \`cronModelQ: shellSingleQuote(DEFAULT_CRON_MODEL)\` rendered via triple-stache \`{{{cronModelQ}}}\`; DEFAULT_CRON_MODEL is a fixed constant and is shell-single-quoted, so no interpolation/injection risk. Both scaffold and reconcile paths compute the context and only write/chmod cron-session.sh when \`cronSessionEnabled\`. Good.
- \`src/scheduler/cron-routing.ts\` — \`scheduleNeedsCronSession\` returns true iff ≥1 entry resolves to \`session:'cron'\` (escalation routing for polls, direct for prompts), so an agent with only escalate-to-main polls or Tier-2 entries forks nothing. Pure, reuses the L1 routing. Correct.

- WATCH (claim accuracy, not a functional defect): \`scaffold.ts buildCronSessionContext\` hardcodes \`cheapCronEnabled: true\`, so \`cronSessionEnabled\` flips on for ANY entry that routes to the cron session — which per #2244 is \`context:fresh\` OR a known-cheap \`model:\` (sonnet/haiku) with context unset. The PR states dormancy rests only on "no agent has the brand-new \`context:fresh\` field", but \`model\` was reactivated from its formerly-DEPRECATED/IGNORED state, so any fleet agent carrying a stale \`model: sonnet\` on a schedule entry would render a non-byte-identical start.sh and fork an (idle, exit-78) sidecar. Functionally still dormant thanks to the runtime gate, but please confirm no fleet schedule entry has a leftover cheap \`model:\`, or the "byte-identical fleet-wide" guarantee is narrower than claimed.
- NIT: cron-session.sh is chmod \`0o700\` on the scaffold path but \`0o755\` on the reconcile path (\`scaffold.ts\`) — same file, two perms. Pick one (0o700 is the tighter, owner-only choice).
- NIT: the \`cron-session.sh.hbs\` header docstring says minimal context comes from "a TRIMMED .mcp.json (only switchroom-telegram)", but v1 sets \`CRON_MCP_CONFIG={{agentDir}}/.mcp.json\` (the full surface) — the inline comment correctly notes trimming is deferred, so the header oversells. Align the header with the v1 reality.

Rationale: dormant, flag+config double-gated second-session launcher with a clean exit-78 idle and a structural guard test; the lone substantive item is that the byte-identical-fleet-wide claim also depends on no stale cheap \`model:\` field, which the runtime gate de-risks to a cosmetic start.sh diff rather than any live cron.

PR #2247, opened 2026-06-09
feat(cron): L4 refinements — trimmed cron .mcp.json + compose mem bump

mekenthompson:
## Summary

The two canary-time refinements deferred from #2246, completing L4. Both are
**config-derived and no-op for the current fleet** (no agent runs a cron session
— it requires a brand-new `context: fresh`/cheap-`model:` schedule field).

### 1. MCP trimming
Scaffold now writes `<agentDir>/.claude-cron/.mcp.json` containing **only
`switchroom-telegram`** (the bridge the cron session needs to reply), reused
verbatim from the main server set. The cron session loads it via
`--mcp-config … --strict-mcp-config`, so it pays a fraction of the main
session's **~31k-token MCP schema tax** (no hindsight/perplexity/webkite/
agent-config/hostd). Identity still comes from the process-env
`SWITCHROOM_AGENT_NAME=<name>-cron` (not a baked `.mcp.json` field — the #2246
reviewer confirmed switchroom-telegram doesn't bake it). `cron-session.sh` points
at the trimmed config, falling back to the full `.mcp.json` for older scaffolds.
Written on both the scaffold and reconcile paths.

### 2. Compose mem/pids bump
`resolveResourceDefaults` gives a cron-session agent **+512M mem / +128 pids**
headroom for the second `claude` — but **only where the operator left the field
on defaults** (an explicit `resources.memory`/`pids_limit` always wins, never
bumped). RFC §4.

## Test plan

- [x] `tsc --noEmit` clean
- [x] 26 new tests: trimmed-config content (only switchroom-telegram, heavy
  servers gone, bridge entry verbatim), `parseMemToMib`, the bump/no-override
  matrix (default → bumped, explicit → untouched, partial → mixed), render points
  at the trimmed path
- [x] compose-generator (154) + scaffold/mcp-parity + reconcile suites green — no regression
- [ ] Validated end-to-end at the marko canary (still pending, with the rest of L4)

## Risk

No-op for the fleet (config-derived gate false everywhere). The bump never
overrides an operator sizing. The trimmed config falls back to the full one if
absent. Compliance unchanged (interactive claude, broker OAuth).

### DIFF
--- a/profiles/_base/cron-session.sh.hbs
+++ b/profiles/_base/cron-session.sh.hbs
@@ -31,12 +31,14 @@ esac
 CRON_NAME="{{name}}-cron"
 CRON_CONFIG_DIR="{{agentDir}}/.claude-cron"
 MAIN_CONFIG_DIR="{{agentDir}}/.claude"
-# v1: reuse the agent's project .mcp.json. The switchroom-telegram BRIDGE reads
-# SWITCHROOM_AGENT_NAME from the process env (exported below) — NOT its .mcp.json
-# env block — so it registers as the cron identity even off the shared config.
-# The bigger wins (fresh context + the cheap model) already hold; trimming the
-# MCP surface to just switchroom-telegram is a documented canary-time follow-up.
-CRON_MCP_CONFIG="{{agentDir}}/.mcp.json"
+# TRIMMED MCP config (scaffold writes .claude-cron/.mcp.json with ONLY
+# switchroom-telegram) → the cron session pays a fraction of the main session's
+# ~31k-token MCP schema tax. The switchroom-telegram BRIDGE reads
+# SWITCHROOM_AGENT_NAME from the process env (exported below), NOT its .mcp.json
+# env block, so it registers as the cron identity. Fall back to the full
+# .mcp.json if the trimmed file is missing (older scaffold) so it still boots.
+CRON_MCP_CONFIG="{{agentDir}}/.claude-cron/.mcp.json"
+[ -f "$CRON_MCP_CONFIG" ] || CRON_MCP_CONFIG="{{agentDir}}/.mcp.json"
 CRON_SOCKET="switchroom-${CRON_NAME}"
 
 mkdir -p "$CRON_CONFIG_DIR"
--- a/src/agents/compose.ts
+++ b/src/agents/compose.ts
@@ -27,6 +27,7 @@ import { createHash } from "node:crypto";
 import { existsSync, mkdirSync, readFileSync } from "node:fs";
 import { join } from "node:path";
 import type { SwitchroomConfig, AgentConfig, AgentBindMount } from "../config/schema.js";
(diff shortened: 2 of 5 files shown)


mekenthompson:
APPROVE

- \`src/agents/compose.ts resolveResourceDefaults\` — the cron-session bump applies mem ONLY when \`overrides?.memory === undefined\` and pids ONLY when \`overrides?.pids_limit === undefined && merged.pidsLimit !== undefined\`, so an explicit operator \`resources.memory\`/\`pids_limit\` is never overridden. The override-protection claim holds. Correct.
- \`src/agents/compose.ts\` (overrides refactor) — the old \`if (!overrides) return {...base}\` early-return became \`if (overrides) {...}\` then falls through to the cron bump, so an agent with NO explicit resources still reaches the headroom logic; \`bumpMem\` returns the original string when \`parseMemToMib\` fails, so a malformed mem value is passed through untouched rather than corrupted. Sound.
- \`src/agents/compose.ts parseMemToMib\` — regex \`^(\d+(?:\.\d+)?)\s*([gmk]?)b?$\` handles \`1.5g\`/\`256m\`/\`512\`/\`1gb\`, defaults unit to MiB, k→1/1024 MiB. Reasonable; bare integers read as MiB which matches docker’s default. Fine.
- \`src/agents/scaffold.ts maybeWriteTrimmedCronMcp\` — writes \`.claude-cron/.mcp.json\` with ONLY the switchroom-telegram entry reused verbatim, mode 0600, idempotent (rewrites only on content change), returns null when \`!cronSessionEnabled\` or no telegram bridge. Identity still comes from process-env \`SWITCHROOM_AGENT_NAME\`, not a baked field. Correct.
- \`profiles/_base/cron-session.sh.hbs\` — \`CRON_MCP_CONFIG\` now points at the trimmed \`.claude-cron/.mcp.json\` with a \`[ -f … ] ||\` fallback to the full \`.mcp.json\` for older scaffolds, so an un-rescaffolded agent still boots. This also resolves the header-oversells-"trimmed" nit I raised on #2246 — the docstring now matches reality. Good.
- NIT: \`maybeWriteTrimmedCronMcp\` never removes a stale \`.claude-cron/.mcp.json\` if \`cronSessionEnabled\` later flips false (agent drops its context:fresh entry). Harmless (cron-session.sh won’t run), but the file lingers — consider unlinking when the gate is false on reconcile.
- WATCH (carry-forward from #2246, not new): the bump + trim gate on \`scheduleNeedsCronSession(..., {cheapCronEnabled:true})\`, which is true for \`context:fresh\` OR a known-cheap \`model:\` entry — so a fleet agent with a stale cheap \`model:\` would get +512M/+128pids and a trimmed-config write. Both are override-safe and config-no-op for the current fleet, so no harm; same fleet-check-before-flag-on caveat as #2246.

Rationale: both refinements are override-safe, idempotent, fall back gracefully, and gate on the same config derivation that is false fleet-wide today; 26 new tests cover the trim content and the bump/no-override matrix. Completes L4 with no regression.

PR #2249, opened 2026-06-09
fix(restart): regenerate compose on agent restart so pin bumps apply

mekenthompson:
## Problem

The v0.14.92 fleet roll surfaced a real regression: editing `release.pin` in
`switchroom.yaml` and running the staggered `switchroom agent restart --wait
--force` loop **restarted every agent but left them on the old version**.

**Root cause:** `reconcileAndRestartAgent` re-emits the agent-dir scaffold
(`start.sh`, settings, hooks) but **not the fleet compose**. So the pin reached
`start.sh`, but the container recreated against the *old* image ref still in
`~/.switchroom/compose/docker-compose.yml`. The function's own docstring promises
*"operators edit switchroom.yaml → restart → done"* — for image/pin changes it
silently didn't, forcing a separate `switchroom apply` first (easy to forget;
it cost a wasted full-fleet bounce on the v0.14.92 roll).

## Fix

`agent restart` now **regenerates the compose from the current config right
before the recreate**, through a new shared `writeComposeFile()` that **both
`apply` and the restart path call** — so the two can never drift (apply's inline
compose-gen block is refactored to call it; behaviour-preserving).

- Only the named service is recreated (`--no-deps`), so regenerating the whole
  compose is safe — other agents converge as **they** restart (staggered roll).
- **Best-effort:** a write failure (e.g. non-sudo perms) **warns but never aborts**
  the restart — the bounce still works against the existing compose, exactly as
  before this fix.
- Logs the image drift (`vX → vY`) when the tag changes.
- Keeps the compose **operator-owned** under sudo (the chown `apply` previously
  did only via its whole-tree `restoreOperatorOwnership` sweep, which restart
  doesn't run — without this a root-mode restart would lock the operator out of
  non-sudo `docker compose` reads).
- Cron-only hot-reload path and the graceful path are handled correctly (cron-only
  skips regen since it never recreates; graceful regens then bounces in place).

## Test plan

- [x] `tsc --noEmit` clean
- [x] `writeComposeFile` unit (pin→image tag, changed/previousImageTag drift signals, idempotent re-write) — writes to a tmpdir, never `~/.switchroom`
- [x] 5 new `reconcileAndRestartAgent` cases: regen-**before**-recreate ordering, cron-only path skips regen, regen-failure-warns-but-restarts, graceful regens first
- [x] **apply (19) + compose-generator (154) + docker-fleet + agent suites green** — the apply refactor is behaviour-preserving
- [ ] Pre-existing `update.test.ts` failures (docker `status=127` in-worktree) are unrelated — fail identically on base

## Risk

Touches the load-bearing restart path. Mitigated: the compose write is
best-effort (never aborts restart), the shared writer makes apply/restart
byte-identical, and the full apply/compose suites confirm no behaviour change.
A pin bump now Just Works on `agent restart` — and `switchroom update --pin`
(which already ran `apply` internally) is unaffected.

### DIFF
--- a/src/cli/agent-reconcile.test.ts
+++ b/src/cli/agent-reconcile.test.ts
@@ -27,12 +27,20 @@ function mkDeps(overrides: Partial<ReconcileAndRestartDeps> = {}): ReconcileAndR
   restartAgent: ReturnType<typeof vi.fn>;
   gracefulRestartAgent: ReturnType<typeof vi.fn>;
   applyCronChangesHot: ReturnType<typeof vi.fn>;
+  writeComposeFile: ReturnType<typeof vi.fn>;
 } {
   return {
     reconcileAgent: vi.fn(() => ({ agentDir: "/tmp/a", changes: [] })),
     restartAgent: vi.fn(),
     gracefulRestartAgent: vi.fn(),
     applyCronChangesHot: vi.fn(() => ({ cronScripts: [], ipcSignalled: false })),
+    writeComposeFile: vi.fn(async () => ({
+      composePath: "/tmp/compose.yml",
+      imageTag: "v0.0.0",
+      bytes: 0,
+      changed: false,
+      previousImageTag: null,
+    })),
     ...overrides,
   } as never;
 }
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- **Fix is correct and self-heals the v0.14.92 regression.** `reconcileAndRestartAgent` now regenerates the fleet compose just before the container recreate (agent.ts:687-708), so a `release.pin`/image change in switchroom.yaml reaches the running container on a plain `agent restart` — restoring the function's own "edit switchroom.yaml → restart → done" docstring promise.
- **Shared writer eliminates drift.** `apply` and the restart path both call the new `writeComposeFile()` (write-compose.ts:58), so the two compose-gen paths can't diverge. The apply refactor (apply.ts:954-959) is behaviour-preserving: `generateCompose`/`resolveImageTag`/`resolveRelease` are removed from apply's imports and survive only in comments (tsc-clean), `mkdir`/`writeFile` stay because they're used elsewhere (apply.ts:309,433), and `operatorUid` is resolved identically (`resolveOperatorUid()`) inside the writer.
- **Wiring verified.** `configPath` is a real param of the function (agent.ts:561, `string | undefined`) passed as `switchroomConfigPath` (agent.ts:693); `DEFAULT_COMPOSE_PATH` is exported (apply.ts:71). Both match `WriteComposeOpts`.
- **Best-effort + ordering proven by tests.** The regen is wrapped in try/catch that warns-but-never-aborts (agent.ts:701-708); cron-only hot path returns before reaching it; the 4 new reconcile cases assert compose-before-recreate ordering, cron-only skip, failure-warns-restarts, and graceful-regens-first. The 3 write-compose cases cover pin→tag, previousImageTag drift, and idempotent changed=false.
- **chown is sound.** Writer chowns the compose to the operator when euid==0 (write-compose.ts:91-97), covering the gap that restart skips apply's whole-tree `restoreOperatorOwnership` sweep — without it a root-mode restart would lock the operator out of non-sudo `docker compose` reads.

Non-blocking watch: `writeFile` is non-atomic (no temp+rename) and the *whole* fleet compose is regenerated on *every* per-agent restart. Under the sequential `--wait` staggered roll this is fine, but a parallel roll could interleave one agent's write with another's `docker compose up` read. Not a regression (apply was non-atomic too) — just newly reachable from a per-agent-concurrent path. Worth a temp+rename if parallel rolls ever land.

Rationale: targeted fix to a load-bearing path, gated by best-effort semantics, with a shared writer that makes apply/restart byte-identical and full apply+compose suites green confirming no behaviour change.

PR #2250, opened 2026-06-09
fix(telegram): obligation sweep waits for in-flight background work

mekenthompson:
## Summary

The obligation ledger's escalate-grace is **45s** — far too short for
extended-autonomous sub-agent work. An agent that ack-firsts ("on it")
then delegates to a background worker or an orphaned foreground
sub-agent (#2240) ends its **foreground** turn in seconds, but the real
answer lands minutes later. The turn-in-flight machine (turn already
ended) doesn't see that work, so the sweep re-presents the obligation
twice and then escalates a false **"⚠️ I may have missed an earlier
message — re-send"** while the agent is genuinely researching.

This is the **gymbro card on 2026-06-10**: a Liven-research request was
acked + delegated; the v0.14.91 rollout restart then killed the
in-flight work; and because the represent budget had already been burnt
by the false sweeps, the hydrated obligation **escalated instead of
re-presenting (resuming)** the research.

## Why

`isMachineInTurn()` / `claudeBusyKeys` track only the foreground turn.
`countRunningBackgroundWorkers()` counts only `background=1` workers, so
an **orphaned foreground** sub-agent (`background=0`) is invisible to
both. Neither existing signal covers post-turn extended-autonomous work
— so the ledger's safety net fires a false alarm on genuine work.

## What

Make the sweep aware of in-flight autonomous sub-agent work via a
**unified, bounded** signal (`agentHasInFlightBackgroundWork`):
- running background worker → `countRunningWorkers()` (dispatch-time DB), OR
- orphaned/extended foreground sub-agent → fresh `turn-active.json` mtime
  (the watcher touches it on foreground sub-agent JSONL growth; the two
  signals are complementary — background activity deliberately does not
  touch the parent marker).

While that holds, an obligation younger than
`OBLIGATION_BACKGROUND_WORK_GRACE_MS` (default **20m**, measured from
`openedAt`) is skipped — neither re-presented nor escalated.

**Bounded BY CONSTRUCTION:** the ceiling is a hard wall-clock deadline,
so a stuck/leaked worker signal cannot suppress escalation forever — the
ledger FSM still terminates. This also fixes the restart case: with the
false represents suppressed, the represent budget survives, so after a
restart kills the work the hydrated obligation **re-presents (resumes
the research)** instead of escalating.

## Test plan

- [x] `obligation-ledger.test.ts` — 7 new cases: bg-grace skip, ceiling
      override (bounded), kill switch, composition with trailing grace,
      oldest-eligible, **represent-budget-preserved → resume-after-restart**.
- [x] `obligation-determinism.test.ts` — new proof: **3000 schedules
      with background-work grace perpetually active** → ceiling forces
      every obligation to its correct terminal (no livelock, no silent
      loss; grace only delays). Identical terminals to the no-grace run.
- [x] `turn-active-marker.test.ts` — `readTurnActiveMarkerAgeMs`:
      absent→null, fresh→small, back-dated→~age (injected clock).
- [x] `tsc --noEmit` clean · `check-bot-api-wrapping.sh` clean ·
      `check-plugin-references.mjs` clean.
- [x] Full related bun suite (obligation/escalation/drain/marker/
      delivery): 157 pass.

## Risk

Low / flag-gated. Kill switch
`SWITCHROOM_OBLIGATION_BACKGROUND_WORK_GRACE_MS=0` → exact pre-fix
behaviour. The new bound only ever DELAYS the existing represent/escalate
ladder (bounded by the 20m ceiling); it cannot drop a message or loop
forever — proven by the extended determinism enumeration. Marker read is
a single `statSync`, never throws (null on failure).

JTBD: **you hold the leash** / **always available** — the status/obligation
safety net must not cry "did I miss this?" while the agent is genuinely
mid-research.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -429,6 +429,7 @@ import {
   touchTurnActiveMarker,
   removeTurnActiveMarker,
   sweepStaleTurnActiveMarker,
+  readTurnActiveMarkerAgeMs,
   TURN_ACTIVE_MARKER_FILE,
 } from './turn-active-marker.js'
 import {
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

- **Correct fix for the real bug (premature escalation during genuine post-turn work).** `oldestEligible` now skips an obligation while `backgroundWorkActive && now - o.openedAt < backgroundGraceMs` (obligation-ledger.ts:194-195), and crucially the sweep takes the `action:'none'` path **without** calling `markRepresented` — so the represent budget is preserved across the work window and a restart-killed obligation **resumes (re-presents)** instead of escalating. That's the gymbro bug, and the dedicated test proves it (obligation-ledger.test.ts:398-415).
- **Bounded by construction — no new unbounded liveness dependency.** The skip flips eligible the instant `now - openedAt >= backgroundGraceMs` regardless of work state (obligation-ledger.ts:194). The determinism proof exercises the worst case: 3000 schedules with `backgroundWorkActive` PERPETUALLY true, asserting `steps < 10_000` and terminals **identical to the no-grace run** (obligation-determinism.test.ts:278-319). Grace only ever delays.
- **Both signals real and complementary.** `agentHasInFlightBackgroundWork` (gateway.ts:5375-5380) ORs `countRunningWorkers() > 0` (gateway.ts:3091 — dispatch-time DB count of running background workers, INSERTed before the parent turn ends) with a fresh `turn-active.json` mtime via `readTurnActiveMarkerAgeMs` (turn-active-marker.ts:220 — catches the orphaned/extended FOREGROUND sub-agent #2240 that background activity deliberately doesn't mark). Verified `statSync`/`join`/`TURN_ACTIVE_MARKER_FILE` are imported in turn-active-marker.ts.
- **Double-gated kill switch.** `SWITCHROOM_OBLIGATION_BACKGROUND_WORK_GRACE_MS=0` short-circuits at both the gateway (`OBLIGATION_BACKGROUND_WORK_GRACE_MS > 0 &&`, gateway.ts:5395) and the ledger (`backgroundGraceMs > 0 &&`, obligation-ledger.ts:194) → exact pre-fix behaviour. Marker read never throws (catch → null = not working), clock injectable.
- **Oldest-eligible still fair.** A young in-work obligation doesn't block an ancient one past its ceiling (obligation-ledger.test.ts:388-396).

Non-blocking watch: the 90s `TURN_ACTIVE_MARKER_FRESH_MS` window means a plain foreground turn that just ended (marker freshly touched, no sub-agent) grants up to ~90s of background-grace on top of the 45s trailing grace, since the marker can't distinguish 'turn just ended' from 'orphaned sub-agent still touching it'. The in-code comment (gateway.ts:5371-5374) owns this as 'a harmless small extra grace, bounded by the ceiling either way' — correct, just be aware the effective post-turn minimum grace is ~90s, not 45s, whenever the marker is fresh.

Rationale: targeted, flag-gated fix to a false-alarm path, with a genuine bounded-termination proof (3000-schedule enumeration) confirming it can only delay the represent/escalate ladder, never drop a message or livelock.

PR #2254, opened 2026-06-09
fix(quota-watch): fleet dedup + staleness gate + never send unverified numbers

mekenthompson:
## Summary

Hardens the quota-watch notifier against the 2026-06-09 incident: a fleet bounce released days-stale recovery latches on **all 11 agent gateways at once**, flooding the operator with **26 byte-identical 🟢 "Quota back in healthy range" messages in 16 minutes** — recoveries announced 19h (pixsoul) and 3.4 days (ken.outlook) after the fact. Content was technically true; volume and timing made it read as wrong.

Three compounding mechanisms → three fixes, each independently kill-switched:

| # | Mechanism | Fix | Kill-switch |
|---|---|---|---|
| 1 | No fleet dedup: 11 independent watchers × allowFrom fan-out = up to 26 sends per edge | New broker `claim-notification` op — first claimant per **(account, transition, chat)** in a 30-min window sends; the rest advance local state silently. Per-chat keys keep the audience identical. Claims persist across broker restarts; synchronous check-and-set ⇒ exactly one grant per window by construction. Gateways **fail open** on claim error (skewed-rollout safety). | `SWITCHROOM_QUOTA_WATCH_FLEET_DEDUP=0` |
| 2 | No staleness check: classification off the broker's `last_quota` cache (capturedAt dropped in `reviveLastQuota`); latches only release at restart probe-storms → synchronized floods | Thread `capturedAt` → `AccountSnapshot.capturedAtMs`. Cached snapshots >60 min old carry **no opinion** (skip). Recovery edges on the **first post-boot tick** or whose 🟡 warning is **>6h old** reconcile **silently** (latch advances, nothing sent). Warnings (🟡) still notify on any tick — level-state, still actionable. | `SWITCHROOM_QUOTA_WATCH_MAX_STALE_MS=0`, `SWITCHROOM_QUOTA_WATCH_LATE_RECOVERY_MS=0` |
| 3 | Probe-failure fallthrough: a failed pre-send validation probe sent the notification from arbitrarily-stale cache — the only path that could ship outright FALSE numbers (latent in this incident) | Failed/missing fresh probe **defers** the notification to the next 15-min tick (state untouched → re-evaluated and re-probed). | `SWITCHROOM_QUOTA_WATCH_SEND_ON_PROBE_FAIL=1` (legacy) |

The fleet all-exhausted alert (#2215) routes through the same claim gate. With #2239's broker fleet-probe keeping caches ≤10 min warm, genuine recoveries still notify promptly on normal ticks — the reconciliation rules only swallow archaeology.

**Replay of the incident under this PR:** wave 2 (boot-tick recoveries, 19h/3.4d stale) → **0 messages** (silent reconcile on every agent). Wave 1 (genuine 🟡 80% crossing) → **1 message per chat** instead of 13.

## Why

`reference/vision.md` outcome 2 (*you hold the leash — awareness + control, not a feed to babysit*): 26 duplicate stale notifications is the opposite of awareness. JTBD `track-plan-quota-live` anti-pattern explicitly warns against quota noise the user tunes out.

## Test plan

- [x] `telegram-plugin/tests/quota-watch.test.ts` — 46 pass under **both** vitest and `bun test` (no bun-incompatible APIs)
- [x] **Total enumeration** of the decision FSM: 144 cells (prev-health × current-health × staleness × boot-tick × warning-age) asserted against an independently-computed table, plus a terminal-state invariant: any notify/reconcile re-run against its own output state is a skip — the FSM cannot loop (`feedback_prove_finite_fsm_not_sample`)
- [x] `src/auth/broker/server.test.ts` — 65 pass; 6 new claim-op tests in the isolated tmpdir harness: grant/deny across agents, key independence, window expiry re-grant, restart persistence, 24h prune, concurrent-claim race (exactly one grant)
- [x] `tsc --noEmit` clean; `check-plugin-references.mjs` clean; `check-bot-api-wrapping.sh` clean (no new raw `bot.api` callsites — claim gate wraps existing wrapped sends)
- [x] Full vitest + `bun test tests/`: failure set **byte-identical** to a pristine upstream/main baseline worktree (pre-existing environmental failures only; 0 new)

## Risk

- **Version skew (new gateway + old broker):** `claim-notification` is rejected at the protocol layer → gateway fails **open** → behavior identical to today (duplicates). No flag day.
- **Old gateway + new broker:** op simply unused.
- **Claim-then-send-fails loss window:** if the claim winner's Telegram send fails (already swallowed best-effort today), that chat misses one advisory until the next genuine edge. Accepted: quota advisories are non-critical, `/auth` shows live state on demand, and the pre-PR behavior already tolerated send failures.
- **Rollback:** the 4 env kill-switches restore pre-PR behavior per mechanism, with ONE deliberate residual: boot-tick recovery suppression has no env knob (it is the core incident fix and is safe-direction — one fewer advisory, never a wrong or lost warning). Under full kill-switch rollback, a recovery edge landing exactly on a gateway's first post-boot tick still reconciles silently; all other ticks behave exactly as pre-PR.
- **Flap damping (intended):** the 30-min claim window also collapses a genuine re-crossing of the SAME edge fleet-wide within 30 min to one notification.

Incident forensics: 4-lens multi-agent investigation + adversarial verification, 2026-06-10 session (root cause, refuted hypotheses, and log evidence recorded in the session memory).

### DIFF
--- a/src/auth/broker/client.ts
+++ b/src/auth/broker/client.ts
@@ -241,6 +241,11 @@ export interface RefreshAccountData {
   expiresAt?: number;
 }
 
+export interface ClaimNotificationData {
+  /** True for the first claimant of the key inside the window. */
+  granted: boolean;
+}
+
 export interface AddAccountData {
   label: string;
   expiresAt?: number;
@@ -471,6 +476,25 @@ export class AuthBrokerClient {
       : { v: PROTOCOL_VERSION, id: randomUUID(), op: "mark-exhausted" };
     const data = await this.send(req);
     return data as MarkExhaustedData;
+  }
+
+  /**
+   * Fleet notification-dedup claim. Returns `granted: true` only for
+   * the first caller of `key` within `windowMs` — the winner sends the
+   * notification, everyone else stays silent. Callers should FAIL OPEN
+   * on error (send anyway): a broker that predates this op rejects the
+   * request at the protocol layer, and duplicated notifications beat
+   * silently dropped ones during a skewed rollout.
+   */
+  async claimNotification(key: string, windowMs: number): Promise<ClaimNotificationData> {
+    const data = await this.send({
+      v: PROTOCOL_VERSION,
+      id: randomUUID(),
+      op: "claim-notification",
+      key,
+      windowMs,
+    });
+    return data as ClaimNotificationData;
   }
 
   async refreshAccount(account: string): Promise<RefreshAccountData> {
--- a/src/auth/broker/protocol.ts
+++ b/src/auth/broker/protocol.ts
@@ -341,6 +341,36 @@ export const ProbeQuotaRequestSchema = z.object({
   timeoutMs: z.number().int().positive().max(60_000).optional(),
 });
 
+/**
+ * Fleet-wide notification dedup claim (#E4 follow-up — quota-watch
+ * de-duplication). Every agent gateway independently runs the same
+ * watchers (quota-watch, fleet all-exhausted) over the same shared
+ * account pool, so a single transition used to fan out one Telegram
+ * message PER AGENT (×11 on a full fleet). The broker — already the
+ * fleet's shared singleton — arbitrates: the first caller to claim a
+ * `key` within `windowMs` is granted (and should send); everyone else
+ * is denied (and should stay silent but still update local state).
+ *
+ * Keys are caller-defined, e.g.
+ * `quota-watch:<account>:<transition>:<chatId>` — per-chat keys keep
+ * the audience identical to the pre-dedup behaviour (every chat any
+ * agent would have notified still gets exactly one copy).
+ *
+ * ACL: same posture as `list-state` — no identity restriction. A
+ * denied claim is `granted: false`, never an error. Claims are
+ * persisted (`notification-claims.json`) so a broker restart inside
+ * the window does not re-open the gate.
+ */
+export const ClaimNotificationRequestSchema = z.object({
+  v: z.literal(PROTOCOL_VERSION),
+  op: z.literal("claim-notification"),
+  id: z.string().min(1),
+  /** Dedup key. Caller-namespaced (e.g. "quota-watch:<acct>:<edge>:<chat>"). */
+  key: z.string().min(1).max(512),
+  /** Deny subsequent claims for the same key for this long. */
+  windowMs: z.number().int().positive().max(86_400_000),
+});
+
 export const RequestSchema = z.discriminatedUnion("op", [
   GetCredentialsRequestSchema,
   ListStateRequestSchema,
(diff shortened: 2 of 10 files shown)


mekenthompson:
APPROVE

Verified against the real diff and head-SHA files; all three mechanisms are coherent and the failure-mode handling is honest. Findings (none blocking):

- `src/auth/broker/server.ts:1325-1346` (opClaimNotification) — check-and-set is fully synchronous between read and write, so the concurrent-claim guarantee holds on the single-threaded event loop. Prune iterates an `Object.entries` snapshot, so deleting during iteration is safe. Correct.
- `src/auth/broker/protocol.ts:71-79` — `claim-notification` has no identity ACL (same posture as `list-state`). Any agent on the socket could pre-claim keys and suppress fleet notifications. Acceptable in a trusted-fleet model and consistent with existing ops; worth a line in the protocol doc comment if the threat model ever tightens. Non-blocking.
- `telegram-plugin/gateway/gateway.ts:451-462` — fail-open on claim error is the right default for skewed rollouts; worst case is pre-PR duplicate behavior, never dropped alerts.
- `telegram-plugin/gateway/gateway.ts:584` (`if (decision.kind !== 'notify') continue`) — the "differing-transition notify" gap in the re-evaluation branch (fresh probe returns notify with a different transition → cached-data message would send) is unreachable: `entered-throttling` requires prev=healthy and `recovered-to-healthy` requires prev=throttling, and prev is fixed across the re-eval, so the fresh re-eval can never produce the opposite transition. No latent unverified-numbers path.
- `telegram-plugin/quota-watch.ts:832-838` — staleness gate fires only when `capturedAtMs` is set; verified at head that `buildSnapshotsFromCachedState` (auth-snapshot-format.ts:664) sets it from `lq?.capturedAt` and the enrichment path clears it on fresh probes (gateway.ts:589). Live-probe snapshots are never gated. Correct.
- `telegram-plugin/quota-watch.ts` `FLEET_ALL_EXHAUSTED_KEY` — confirmed it exists at head (quota-watch.ts:312), so the fleet alert's claim key composes correctly.
- Claim-before-send loss window (winner claims, then Telegram send fails → chat misses one advisory until next genuine edge) is acknowledged in the PR risk section and acceptable for non-critical advisories.
- Tests: 144-cell total enumeration of the decision FSM with an independently computed expectation table plus a terminal-state invariant, and broker claim tests covering grant/deny, window expiry, restart persistence, prune, and the concurrent race. This is the right level of coverage for an incident fix.

Rationale: each mechanism is independently kill-switched, defaults degrade to pre-PR behavior on any error path, and the only behavior removed is verified noise (stale recoveries and N-1 duplicates). Safe to merge.

PR #2255, opened 2026-06-10
fix(telegram,cli): fallback failure notice + honest quota columns in auth list

mekenthompson:
## Summary

Two follow-ups from the 2026-06-09/10 quota-status forensics — sibling of #2254 (which fixed the quota-watch notifier itself):

**1. Fallback broken-promise fix.** The ⚠️ model-unavailable card says *"Auto-failover in progress — see the announcement below"* before the dispatcher's outcome is known. When the dispatcher errors (broker unreachable, `listState`/`markExhausted` throw), no announcement ever arrived — the 06-06→07 incident sent **12 cards whose promised announcement never came** while every dispatch errored `set-active requires admin`. That root cause was fixed (#2206), but any future dispatcher error reproduced the broken promise. Now `doFireFleetAutoFallback`'s error paths broadcast a failure notice (pure builder `renderFallbackFailureNotice`, unit-tested for shape + HTML-escaping) naming the trigger agent, the error, and the manual recovery verbs. **Invariant: card promised an announcement → an announcement always arrives, success or failure.** Kill-switch: `SWITCHROOM_FLEET_FALLBACK_FAILURE_NOTICE=0`.

**2. Honest quota columns.** `switchroom auth list` and the Telegram legacy fleet-snapshot table (which renders exactly when the live probe FAILED — i.e., when cached-data disclosure matters most) showed only the broker exhausted flag: a 99%-utilized account read `available —`, and an agent verbally parroting the table under-reported. Both tables now carry a `QUOTA 5h·7d` column from the broker's cached `last_quota` with the snapshot age as staleness disclosure (`16%·1% (3m ago)`), and `no data` (never probed) is distinct from `—` (not exhausted). Display-only change.

## Why

`reference/vision.md` outcome 2 (awareness + control): a status card that promises an event that never comes, and a status table that hides 99% utilization, both break the operator's trust in the surfaces they're supposed to act on. JTBD `track-plan-quota-live`.

## Test plan

- [x] `renderFallbackFailureNotice` — shape + HTML-escaping of broker error strings (angle brackets/quotes), under vitest AND bun
- [x] `formatQuotaUtilCell` ×2 (CLI + Telegram twin) — no-data vs not-exhausted distinction, rounding, age coarsening, the exact 99%-invisible regression case
- [x] Existing `auth-command-format2` / `vernacular` suites pass (46/46) — Format 2 (live-probe) path untouched
- [x] `tsc --noEmit` clean · `check-bot-api-wrapping.sh` clean (new send is `swallowingApiCall`-wrapped) · `check-plugin-references.mjs` clean

## Risk

- Failure notice fires only on dispatcher **error** paths (no-client + catch), which previously logged-only. Volume bound: a dedicated **30-min per-gateway cooldown** (`evaluateFallbackFailureNotice`, storm-shape regression test: 60s `quota_wall_detected` re-fires for 1h → ≤2 notices). The `fleetFallbackGate` window does NOT bound this (it arms only on successful swaps) — reviewer-caught, fixed in the second commit. All four `runFleetAutoFallback` outcome paths (switched / all-blocked / no-eligible / no-old-active) already produced announcements and are unchanged.
- Table changes are render-only; no behavior, no wire changes. Worst case: a wide row wraps on narrow phones (it's inside `<pre>` — same as today's table).

### DIFF
--- a/src/cli/auth.ts
+++ b/src/cli/auth.ts
@@ -244,9 +244,36 @@ function formatQuotaReset(state: AccountState): string {
   return `${hours}h ${mins}m`;
 }
 
+/**
+ * Cached-utilization cell. Honesty fix (2026-06-09 incident follow-up):
+ * the table previously rendered ONLY the broker exhausted flag, so a
+ * 99%-utilized account read "available —" — and an agent verbally
+ * parroting `switchroom auth list` under-reported. Shows the broker's
+ * cached 5h·7d utilization with its age; "no data" (never probed) is
+ * distinct from "not exhausted". The (age) suffix is the staleness
+ * disclosure — this is the broker cache, not a live probe.
+ */
+export function formatQuotaUtilCell(
+  a: { last_quota?: { fiveHourUtilizationPct: number; sevenDayUtilizationPct: number; capturedAt: number } | null },
+  now: number = Date.now(),
+): string {
+  const lq = a.last_quota;
+  if (!lq) return "no data";
+  const ageMs = Math.max(0, now - lq.capturedAt);
+  const mins = Math.floor(ageMs / 60_000);
+  const ageStr =
+    mins < 1 ? "just now"
+    : mins < 60 ? `${mins}m ago`
+    : mins < 1440 ? `${Math.floor(mins / 60)}h ago`
+    : `${Math.floor(mins / 1440)}d ago`;
+  const five = Math.round(lq.fiveHourUtilizationPct);
+  const seven = Math.round(lq.sevenDayUtilizationPct);
+  return `${five}%·${seven}% (${ageStr})`;
+}
+
 function printAccountsTable(state: ListStateData): void {
   console.log(
-    chalk.bold("  ACCOUNT                           STATUS       EXPIRES    QUOTA-RESET"),
+    chalk.bold("  ACCOUNT                           STATUS       EXPIRES    QUOTA 5h·7d          QUOTA-RESET"),
   );
   for (const a of state.accounts) {
     const marker =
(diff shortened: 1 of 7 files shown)


mekenthompson:
APPROVE

Verified against the real diff and head-SHA files. Findings (none blocking):

- `src/cli/auth.ts:18-34` (formatQuotaUtilCell) — structural param matches the broker's `AccountState.last_quota?: LastQuotaSnapshot | null` with `capturedAt: number` (verified at head, src/auth/broker/client.ts:157,173). Negative-age clock skew guarded by `Math.max(0, ...)`. Correct.
- `telegram-plugin/gateway/auth-command.ts:130-139` — Telegram twin guards skew differently (`age > 0 ? ... : '0s'`) and uses `formatRelativeMs`, already in scope in this file. Cell content is numeric so the lack of HTML-escaping inside the aligned table is safe; the label remains the only escaped field, unchanged.
- Twin duplication: CLI and Telegram each carry their own `formatQuotaUtilCell` with slightly different age formats ("3m ago" coarse vs "1m 30s ago" precise). Intentional per the test files, but a shared helper would prevent the two drifting further. Non-blocking.
- `telegram-plugin/gateway/gateway.ts:14828-14839` (broadcastFleetFallbackFailure) — kill-switch checked first, sends wrapped in `swallowingApiCall` (verified present at head), fire-and-forget `void` so a Telegram failure can't abort the dispatcher's return path. Both error paths (no-client at gateway.ts:14851, catch at gateway.ts:14917) now broadcast, closing the card's "see the announcement below" promise on every outcome.
- `telegram-plugin/auto-fallback-fleet.ts:72-87` (renderFallbackFailureNotice + escFailureHtml) — pure builder, full HTML entity escaping including quotes, tested for the angle-bracket broker-error case. Correct.
- Volume risk is bounded: failure notices fire only on dispatcher error paths previously log-only, inside the existing fleetFallbackGate dedup window, as the PR states.
- Tests cover the exact regression (99%-utilized non-exhausted account no longer invisible), the no-data vs not-exhausted distinction, and HTML-escaping of broker errors.

Rationale: display-only table change plus a deterministic promise-closure on fallback failure, both kill-switchable or render-only; no wire or behavior changes beyond the new notice.

PR #2257, opened 2026-06-10
feat(memory): hindsight bank-health in doctor + dashboard Memory tab

mekenthompson:
## Summary
Per-agent memory-bank ingest health in both operator surfaces — built for the incident class hit live on 2026-06-10: conversations were **retained but never fact-extracted** (consumer OAuth quota wall / shm exhaustion), so agents silently remembered nothing for days while every existing health surface stayed green. A document with `memory_unit_count == 0` is the durable fingerprint of that failure; this PR makes it impossible to miss.

- **`src/memory/bank-health.ts`** — REST inspection client over hindsight's `/v1/default/banks/*` surface (stats, documents with per-doc extracted-fact counts, mental models with refresh timestamps). Injectable fetch, never throws, 15s timeouts.
- **doctor** — per-bank checks in the hindsight section: **FAIL** on recent (≤30d, >1KB) zero-fact documents with the exact `/reprocess` curl in the fix string; **WARN** on stale (>7d) mental models or a stuck pending queue; **OK** line summarises docs/facts/newest-activity/models. Banks probed concurrently; agents sharing a `memory.collection` dedupe into one row.
- **dashboard** — new **Memory** tab (`GET /api/memory-health`) rendering the same truth per bank: status dot, conversations/facts, latest activity with age, extraction-gap banner, per-model refresh age.

## Why
JTBD: a-standing-team-that-knows-you — memory loss is the worst silent failure this product can have, and it has now happened twice (2026-06-04, 2026-06-06→10) with zero observability. Doctor and the dashboard now agree on one truth with the same thresholds.

## Test plan
- [x] `tests/memory.bank-health.test.ts` — 13 tests: REST client (healthy / gapped / HTTP-fail), helpers (age windows, stale fallback to created_at, tiny-doc exemption), doctor check (fail+fix wording, shared-bank dedupe, inspection-error → warn not fail), dashboard handler (per-bank statuses, unreachable → empty).
- [x] Adjacent suites green: tests/web.test.ts (75), tests/doctor.test.ts (53), cli.doctor.memory*.
- [x] `tsc --noEmit` clean, check-plugin-references clean.
- [x] **Live-validated mid-incident**: doctor correctly flagged the 9 banks holding 153 zero-fact documents queued for reprocess, went green per-bank as the queue drained, and the tiny-doc exemption removed the one false positive (a fleet-wide 515-byte stub from 2026-05-20).

## Risk
Low. Doctor/dashboard are read-only observability; new code never throws (degrades to warn rows); no agent-runtime or compliance-boundary surface touched.

### DIFF
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -1038,7 +1038,7 @@ export async function checkBankIngestHealth(
 
 function hindsightDocReprocessUrl(mcpUrl: string, bankId: string, docId: string): string {
   const base = mcpUrl.replace(/\/mcp\/?$/, "").replace(/\/$/, "");
-  return `${base}/v1/default/banks/${encodeURIComponent(bankId)}/documents/${docId}/reprocess`;
+  return `${base}/v1/default/banks/${encodeURIComponent(bankId)}/documents/${encodeURIComponent(docId)}/reprocess`;
 }
 
 async function checkHindsight(config: SwitchroomConfig): Promise<CheckResult[]> {
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -28,6 +28,7 @@ import { getSlotInfos, type SlotInfo } from "../auth/accounts.js";
 import type { AgentConfig, SwitchroomConfig } from "../config/schema.js";
 import { loadManifest, detectDrift, type DriftProbers } from "../manifest.js";
 import { probeHindsight, isHindsightEnabled, fetchHindsightToolsList } from "../memory/hindsight.js";
+import { inspectBankHealth, staleMentalModels, recentUnextracted, ageDays } from "../memory/bank-health.js";
 import { checkHindsightContainerHealth, classifyToolContract } from "./doctor-memory.js";
 import { isDockerMode, runDockerChecks } from "./doctor-docker.js";
 import { runAuthBrokerChecks } from "./doctor-auth-broker.js";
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

Verified against the real diff and head files. Findings (none blocking):

- `src/cli/doctor.ts:39` — bank-id derivation (`agentConfig.memory?.collection ?? agentName`) inlines the same logic `getCollectionForAgent` implements (verified identical at head, src/memory/hindsight.ts:83-89), while the dashboard handler calls the helper (src/web/api.ts:430). Functionally equivalent today; using the helper in doctor too would remove the duplication. Non-blocking.
- `src/memory/bank-health.ts:196,264` — documents fetched with a single `?limit=500` page and no explicit sort param. If the endpoint returns oldest-first and a bank exceeds 500 docs, `newestDocumentAt` and the recent-gap scan could miss the newest documents. Live-validated against real banks per the PR, so current behavior is presumably newest-first or banks are small, but worth pinning `sort`/`order` in a follow-up if the API supports it. Non-blocking.
- `src/memory/bank-health.ts:203-224` (getJson) — AbortController with cleared timeout on both paths, never throws, distinct Timeout vs HTTP vs network reasons. Correct.
- `src/memory/bank-health.ts:346-357` (recentUnextracted) — the 1000-byte floor excludes trivial stubs (the live-validated 515-byte false positive) while real conversations sit well above it; ordering preserved from the oldest-first sort so `gaps[0]` in doctor's fix string (doctor.ts:79-86) is genuinely the oldest gap. Correct.
- `src/cli/doctor.ts:45-49` — banks probed via Promise.all; inspection failure degrades to warn (doctor.ts:53-59), never fail, so a flaky REST surface can't fake an extraction outage. Matches the dashboard's identical degradation (api.ts:441-443). Correct.
- `src/web/ui/index.html:546-592` — all interpolated strings (bank, agents, statusDetail, model names) go through escapeHtml; numeric fields raw. Tab wiring complete (switchTab list, fetch dispatch). Correct.
- `src/web/server.ts:493-495,504-506` — route + handler use freshConfig(), consistent with adjacent endpoints.
- Tests cover the doctor fail path with the exact reprocess fix string, shared-bank dedupe, inspection-error → warn, dashboard status parity, and unreachable → empty. Both surfaces assert against the same fixtures, which is what keeps doctor and dashboard agreeing.

Rationale: read-only observability with never-throw degradation on every path, built directly against a live-validated incident fingerprint; the two findings are dedup and pagination hygiene, not correctness today.

PR #2258, opened 2026-06-10
feat(memory): detect mental models corrupted by LLM-failure messages

mekenthompson:
## Summary
Follow-up to #2257, from a live incident minutes after it merged: mental-model refreshes that run during an OAuth quota wall persist the literal error string (`You're out of extra usage · resets 3am (UTC)`) as the model's **content**, which is then injected into every agent turn until the next successful refresh. Five lawgpt models + carrie/kdogg user-profiles were corrupted this way with zero observability.

- `corruptedMentalModels()` in bank-health: fingerprints tiny (<300 char) content matching quota/limit failure phrasing, or empty content on a refreshed model. Model summaries now carry `contentLength` + `contentHead` (200 chars).
- doctor: per-bank **FAIL** naming the corrupted models (checked above the extraction-gap gate).
- dashboard Memory tab: fail status + red banner listing the corrupted models.

## Test plan
- [x] 15 tests in tests/memory.bank-health.test.ts incl. the exact live incident strings, empty-after-refresh vs never-refreshed, >=300-char real content immune, doctor + dashboard end-to-end.
- [x] tests/web.test.ts (75) + tests/doctor.test.ts (53) green; tsc clean.
- [x] Live-validated: doctor flags exactly the 3 real corrupted banks (lawgpt: Goodfellow Timeline + Case State; carrie + kdogg user-profile), healthy banks unaffected.

## Risk
Low — read-only observability refinement; heuristic tuned against live data with negative cases pinned (real content mentioning 'limit' at length is immune).

### DIFF
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -28,7 +28,7 @@ import { getSlotInfos, type SlotInfo } from "../auth/accounts.js";
 import type { AgentConfig, SwitchroomConfig } from "../config/schema.js";
 import { loadManifest, detectDrift, type DriftProbers } from "../manifest.js";
 import { probeHindsight, isHindsightEnabled, fetchHindsightToolsList } from "../memory/hindsight.js";
-import { inspectBankHealth, staleMentalModels, recentUnextracted, ageDays } from "../memory/bank-health.js";
+import { inspectBankHealth, staleMentalModels, corruptedMentalModels, recentUnextracted, ageDays } from "../memory/bank-health.js";
 import { checkHindsightContainerHealth, classifyToolContract } from "./doctor-memory.js";
 import { isDockerMode, runDockerChecks } from "./doctor-docker.js";
 import { runAuthBrokerChecks } from "./doctor-auth-broker.js";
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- `src/memory/bank-health.ts:92-100` (corruptedMentalModels) — two-part fingerprint is sound: empty content flags only when the model HAS refreshed (`lastRefreshedAt !== null`), so never-refreshed models stay green; the phrase match requires BOTH tiny content (<300) and a failure pattern, with the >=300 length immunity pinned by a test using real content that mentions "usage limit". Matches the live incident strings exactly.
- `src/memory/bank-health.ts:76` — `contentHead` is 200 chars but the length gate is 300, so a failure phrase sitting in chars 200-299 of a ~250-char corrupted blob would be missed. Real failure messages put the phrase at the start (both incident strings do), so this is theoretical; aligning the head to 300 would remove the gap. Non-blocking.
- `src/memory/bank-health.ts:95` — `resets \d` and `rate.?limit` could in principle match a short genuine note ("standup resets 9am"), but only under 300 chars, and the cost of a false positive is one doctor fail line pointing at a refresh, not data loss. Acceptable precision/recall trade for an incident-class detector. Non-blocking.
- `src/cli/doctor.ts:25-35` — corruption check ordered above the extraction-gap gate, so a bank with both fails on the corruption (the higher-severity, turn-injected failure) first. The dashboard handler orders identically (src/web/api.ts:136-141), keeping the two surfaces in agreement.
- `src/web/ui/index.html:158-160` — corrupted names go through escapeHtml before interpolation. Correct.
- Tests pin the exact live incident strings, the empty-refreshed vs empty-never distinction, both negative cases, and doctor + dashboard end-to-end against the same fixture.

Rationale: read-only detector for a turn-corrupting failure observed live, with both false-positive classes explicitly tested; the two notes are precision tuning, not correctness.

PR #2259, opened 2026-06-10
feat(telegram): /model command to show or switch the Claude model

mekenthompson:
## Summary

Adds a first-class `/model` Telegram command:

- **`/model`** (bare) — shows the cascade-configured model + switch options. Deliberately **never injects the bare verb**: claude's no-arg `/model` renders an interactive picker modal that Telegram can't drive (no arrows, no Esc), which would wedge the pane — the `/rate-limit-options` class of TUI wedge.
- **`/model <alias|full-id>`** — types claude's own `/model <name>` REPL command into the agent's tmux pane via the existing allowlisted inject primitive (`src/agents/inject.ts` — `/model` is already on the allowlist) and relays the captured response. Session-scoped; the reply spells out that persistence is `model:` in switchroom.yaml + restart.

## Why

Operators currently switch models via `/inject /model` (which opens the undriveable picker) or by editing switchroom.yaml + restart. A typed `/model sonnet` from mobile is the missing one-tap path, and it stays Claude-native (pillar 3): the unmodified CLI's own REPL command, no API/SDK, no new model callsite.

## Design notes

- **Arg shape-gated twice** (parser + handler seam): single token, alphanumeric start, model-id charset incl. `[1m]`-variant brackets, ≤100 chars — no whitespace/control smuggling can ride into `send-keys`.
- Auth gate matches `/inject` (`isAuthorizedSender`, not admin-gated — identical power to the existing `/inject /model`).
- Parser/handler split mirrors `auth-command.ts` so logic is unit-testable without booting the bot.
- Not restricted to known aliases — full ids pass through and claude validates, so new models work without a switchroom release.
- Registered in slash menu (19/20 cap), `/commands` catalogue, and help text; existing welcome-text invariant tests pin the sync.

## Test plan

- [x] `bun test tests/model-command.test.ts` — 26 new assertions: parse shapes, injection-smuggling rejection, picker-wedge guard (show/help never inject), exact inject string, outcome shaping (ok / ok_no_output / session_missing / throw)
- [x] `bun test tests/welcome-text.test.ts` — menu/help sync invariants green
- [x] Full `bun test` — 0 non-UAT failures (UAT scenarios need live Telegram creds, excluded from gating CI)
- [x] `tsc --noEmit`, `check-plugin-references.mjs`, `check-bot-api-wrapping.sh` all clean

## Risk

Low — new command surface only; the inject primitive, allowlist, and routing are untouched. Known limitation (same as `/inject`): an inject landing mid-turn sits in claude's input buffer until the turn ends.

### DIFF
--- a/src/cli/agent.ts
+++ b/src/cli/agent.ts
@@ -9,7 +9,7 @@ import { resolveAgentsDir, loadConfig } from "../config/loader.js";
 import { isHindsightEnabled } from "../memory/hindsight.js";
 import type { SwitchroomConfig } from "../config/schema.js";
 import { withConfigError, getConfig, getConfigPath } from "./helpers.js";
-import { scaffoldAgent, reconcileAgent, buildSettingsHooksBlock, detectHooksDrift } from "../agents/scaffold.js";
+import { scaffoldAgent, reconcileAgent, buildSettingsHooksBlock, detectHooksDrift, SWITCHROOM_DEFAULT_MAIN_MODEL } from "../agents/scaffold.js";
 import { writeComposeFile } from "./write-compose.js";
 import { bareClonePath } from "../repos/bare-clone.js";
 import { removeAgentWorktree } from "../repos/agent-worktree.js";
@@ -787,10 +787,16 @@ export function registerAgentCommand(program: Command): void {
             const agentConfig = config.agents[name];
             const status = statuses[name];
             const sched = schedulerStates[name];
+            // Cascade-resolved effective model — the same value scaffold
+            // bakes into settings.json / `--model` (falls back to the
+            // baked default when unset). Consumed by the gateway's
+            // /status and /model surfaces.
+            const resolved = resolveAgentConfig(config.defaults, config.profiles, agentConfig);
             return {
               name,
               status: status?.active ?? "unknown",
               uptime: formatUptime(status?.uptime ?? null),
+              model: resolved.model ?? SWITCHROOM_DEFAULT_MAIN_MODEL,
               extends: agentConfig.extends ?? "default",
               topic_name: agentConfig.topic_name,
               topic_emoji: agentConfig.topic_emoji,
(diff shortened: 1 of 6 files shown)


mekenthompson:
APPROVE

Verified against the real diff and head files. Findings (none blocking):

- `telegram-plugin/gateway/model-command.ts:93` (MODEL_ARG_RE) — single-token, alphanumeric-start, no whitespace/control charset including the `[1m]` brackets, 100-char cap. Combined with the parser's `split(/\s+/)` multi-token rejection (model-command.ts:113-115) and the re-gate at the handler seam (model-command.ts:182-184), nothing un-shaped reaches `send-keys`. Smuggling shapes (`$(...)`, backticks, newline, pipe, leading dash) are all pinned by tests. Correct.
- `telegram-plugin/gateway/model-command.ts:108-123` — the bare and help forms structurally cannot reach `deps.inject` (only `kind === 'set'` flows past the early returns), which is the picker-wedge guard; tests assert zero inject calls on both paths.
- `src/agents/inject.ts:60` — confirmed at head: `/model` is on INJECT_COMMANDS, and the new test pins it so the set path can't silently break if the allowlist is pruned.
- `telegram-plugin/gateway/gateway.ts:13930-13939` — auth gate is `isAuthorizedSender`, identical power to the existing `/inject /model` path, so no privilege change. `switchroomExecJson` is synchronous (gateway.ts:13496 at head), matching the sync `getConfiguredModel` signature.
- `telegram-plugin/gateway/model-command.ts:185` — model arg is HTML-escaped in every reply rendering, and inject errors/messages pass through `escapeHtml`. Correct.
- `telegram-plugin/welcome-text.ts:452,463,471-472` — help text, command catalogue, and slash menu all updated together; the existing welcome-text invariant tests pin the sync.
- One behavioral note: the inject primitive's known mid-turn limitation (command sits in claude's input buffer until the turn ends) applies here and the ok_no_output reply (model-command.ts:209-217) explains exactly that, pointing at `/inject /status` to verify. Honest handling rather than a silent success claim.

Rationale: new command surface only, double shape-gated before touching the pane, structurally incapable of opening the undriveable picker, riding an untouched allowlisted primitive with the same auth posture as the path it replaces.

PR #2262, opened 2026-06-10
test(uat): /model command DM scenario (show, switch, bad-name)

mekenthompson:
## Summary
UAT scenario for the `/model` command shipped in v0.15.2 (#2259) — the exact checks used for the rollout canary, kept as a permanent gate.

- bare `/model` → 'Configured:' reply with switch affordances + switchroom.yaml caveat (proves no picker inject)
- `/model sonnet` → switch routed, session-only note (same-value, leaves the harness agent unchanged)
- `/model bogus-model-name-xyz` → still a reply, never silence

## Test plan
- [x] Ran live against test-harness on v0.15.2: 3/3 green (real Telegram DMs via mtcute driver)
- [x] tsc --noEmit clean

## Risk
None — uat/** is excluded from gating CI; self-skips without driver credentials.

### DIFF
--- a/telegram-plugin/uat/scenarios/jtbd-model-command-dm.test.ts
+++ b/telegram-plugin/uat/scenarios/jtbd-model-command-dm.test.ts
@@ -1,0 +1,91 @@
+/**
+ * UAT — `/model` Telegram command (PR #2259, shipped v0.15.2).
+ *
+ * Serves: `reference/vision.md` outcome 2 (you hold the leash) — the
+ * operator can see and switch the agent's Claude model from Telegram
+ * without SSH. Session-scoped switch via claude's own `/model <name>`
+ * REPL verb injected into the tmux pane.
+ *
+ * Three assertions against a real agent over real Telegram:
+ *
+ *   1. Bare `/model` → shows the configured model (never opens claude's
+ *      interactive picker — the reply must come from the gateway, fast,
+ *      containing "Configured:").
+ *   2. `/model <valid-name>` → switch is injected; reply relays claude's
+ *      response and carries the session-only persistence note.
+ *   3. `/model bogus-name` → still a reply (claude's inline error is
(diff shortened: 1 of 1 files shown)


mekenthompson:
APPROVE — test-only PR, all asserted strings verified against the merged #2259 handler at head.

- jtbd-model-command-dm.test.ts:43-51 — every assertion in the bare-/model test matches a real gateway string: `Configured:` (model-command.ts:122), the `/model opus · /model sonnet · /model haiku` affordances (model-command.ts:123 via MODEL_ALIASES at :34), and the switchroom.yaml persistence caveat (PERSIST_NOTE, model-command.ts:96). A regression in any of these reply lines fails the test, which is exactly what a UAT gate should pin.
- jtbd-model-command-dm.test.ts:68-71 — the switch test's alternation (`/model sonnet|no response captured|Session-only`) covers all three real reply shapes from the set path: relayed response with PERSIST_NOTE (:96), and the empty-capture explanation (:163). Same-value switch (sonnet on a sonnet agent) correctly avoids mutating the harness agent.
- Nit (non-blocking): jtbd-model-command-dm.test.ts:86 — the bad-name test matches `/\S/`, which any bot message satisfies (e.g. a late message from the previous test's session). Anchoring on `not a valid model name` (the literal helpText reason at model-command.ts:71/:100/:134) would make the never-silence assertion actually about /model. As written the test can pass vacuously.
- Nit (non-blocking): the switch test accepts the `no response captured` branch, so a switch that never lands still goes green — fine as a routing-only check, but worth remembering this scenario does not prove the model actually changed.

Rationale: 91-line test-only addition, uat/** excluded from gating CI, harness API (spinUp/sendDM/expectMessage/tearDown) confirmed against uat/harness.ts at head, and the author ran it 3/3 green live on v0.15.2.

PR #2263, opened 2026-06-10
feat(telegram): /model dashboard — live picker-driven menu + quota brief

mekenthompson:
## Summary

`/model` (bare) becomes a model dashboard: **live current model + brief quota + inline-keyboard menu** of exactly the options claude's own `/model` picker offers.

- **No hardcoded models**: options are discovered at request time by driving the CLI's picker over tmux (open → parse → Esc). When Anthropic ships a new model, it appears in the menu the moment the installed CLI offers it.
- **Tap to switch**: re-opens the picker fresh, matches the row **by label** (callback carries a label hash, never a stale index), arrow-keys to it, verifies the cursor row, presses `s` (claude's session-only apply).
- **Quota brief**: `29% / 5h · 33% / 7d` via the existing broker probe → local-headers ladder; never blocks the menu.
- Typed `/model <name>` form unchanged; kill-switch `SWITCHROOM_MODEL_MENU=0` reverts to v1 static text.

## Why session-only (`s`) instead of Enter

Enter persists claude's own settings default — which switchroom's scaffold overwrites on every reconcile. `s` keeps `model:` in switchroom.yaml the single persistent truth (the dashboard says so).

## Safety invariants (pinned by tests)

- **Esc-always**: the picker is never left open — every non-selected exit path sends Escape with verified dismissal (the /rate-limit-options wedge class).
- **Never select blind**: cursor-row label re-verified before `s`; pane drift aborts via Esc.
- **Busy gate**: `currentTurn != null` refuses (driving the picker mid-turn would type '/model' into the prompt as user text).
- **mdl:\* callbacks** carry the strict `allowFrom` gate like every other mutating callback family.

## Parser grounded in reality

`parseModelPicker` is validated against a **verbatim capture of claude 2.1.170's picker** taken on the live test-harness (cursor `❯`, current `✔`, wrapped descriptions, effort row, footer). Layout drift fails discovery gracefully (static v1 fallback + reason) — it cannot mis-select.

## Test plan

- [x] `tests/model-picker.test.ts` — 15 tests: parser on real fixture (+ scrollback anchoring, future extra options), Esc-always invariant, Up/Down/zero-delta navigation, cursor-verify abort, label-not-offered abort
- [x] `telegram-plugin/tests/model-command.test.ts` — 13 new: menu render (current ✅, quota, 64-byte callback cap), busy gates, stale-tag never-selects, no-op on current, failure surfacing, refresh
- [x] Full `bun test`: 0 non-UAT failures; `tsc --noEmit`, plugin-references, bot-api-wrapping all clean
- [x] UAT scenario updated to accept both v1/v2 shapes

## Risk

Medium-low: new tmux interaction is read-mostly (open/parse/Esc) with verified dismissal and a hard timeout; failure mode is a graceful static fallback. Kill-switch reverts the new surface entirely.

### DIFF
--- a/src/agents/inject.ts
+++ b/src/agents/inject.ts
@@ -31,6 +31,7 @@
 
 import { execFile, execFileSync } from "node:child_process";
 import { promisify } from "node:util";
+import { withPaneLock } from "./pane-lock.js";
 
 const execFileAsync = promisify(execFile);
 
@@ -365,13 +366,19 @@ export async function injectSlashCommand(
   const settleMs = opts.settleMs ?? 2000;
   const timeoutMs = opts.timeoutMs ?? 5000;
 
-  return injectSlashCommandWith(makeTmuxRunner(tmuxBin), {
-    socket,
-    session,
-    command: command.trim(),
-    settleMs,
-    timeoutMs,
-  });
+  // Serialize against every other drive on this pane (other injects,
+  // the /model picker driver) — see src/agents/pane-lock.ts. An
+  // inject's Enter landing while another drive's modal is open would
+  // act as that modal's confirm key.
+  return withPaneLock(`${socket}:${session}`, () =>
+    injectSlashCommandWith(makeTmuxRunner(tmuxBin), {
+      socket,
+      session,
+      command: command.trim(),
+      settleMs,
+      timeoutMs,
+    }),
+  );
 }
 
 /**
--- a/src/agents/model-picker.ts
+++ b/src/agents/model-picker.ts
@@ -1,0 +1,353 @@
+/**
+ * Claude CLI `/model` picker driver — discovery + selection over tmux.
+ *
+ * The Telegram `/model` menu must list whatever models the *installed*
+ * claude CLI offers, without hardcoding names that go stale when
+ * Anthropic ships new models. The only durable source of that list is
+ * claude's own `/model` picker, so this module drives it:
+ *
+ *   discoverModels()  open picker → parse the rendered option rows →
+ *                     ALWAYS Esc out (verified) → return the options.
+ *   selectModel()     open picker → parse → arrow-key the cursor to the
+ *                     target row (matched by label, never by stale
+ *                     index) → press `s` (session-only apply) → capture
+ *                     claude's confirmation. Esc + bail on any mismatch.
+ *
+ * Validated against claude 2.1.170's picker (2026-06-10, test-harness):
+ *
+ *     Select model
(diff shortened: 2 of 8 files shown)


mekenthompson:
APPROVE — the safety invariants claimed in the PR body are real in the code and pinned by tests; all cross-module imports verified at head.

- src/agents/model-picker.ts:287-345 — selectModel's never-select-blind invariant holds: target matched by label on a fresh parse (:303), cursor-marker required (:308), post-navigation re-capture re-verifies the cursor row's label before `s` (:323-331), and the finally block Escapes on every non-selected path (:343). tests/model-picker.test.ts:1133-1153 pin both abort paths asserting `s` was never sent.
- src/agents/model-picker.ts:231-252 — dismissPicker verifies dismissal by re-capture with one retry, and the tail-text heuristic correctly handles the post-Esc 'Kept model as …' line below the stale render. Esc-always pinned at tests/model-picker.test.ts:1095-1100 (picker never rendered → Escape still sent).
- telegram-plugin/gateway/gateway.ts:18483-18506 — mdl:* callbacks carry the allowFrom gate before any picker driving; callback identity is an 8-hex label hash (model-command.ts:593-599), and a stale tag re-renders instead of selecting (model-command.ts:700-705, pinned at model-command.test.ts:842-849).
- Imports verified at head: makeTmuxRunner/TmuxRunner (src/agents/inject.ts:210,225), formatQuotaLine (telegram-plugin/quota-check.ts:208), probeQuotaForBootCard/resolveAgentDirFromEnv/currentTurn all present in gateway.ts.
- Nit (non-blocking): no in-flight mutex around picker driving. Two near-simultaneous bare /model sends or menu taps each run openPicker against the same pane and their send-keys interleave (one's Down/Esc lands in the other's modal). The cursor-verify + Esc-always invariants mean the worst case is a verified abort, never a wrong selection, but a cheap module-level 'picker busy' latch would avoid the confusing double-failure UX.
- Nit (non-blocking): isBusy is checked before discover (model-command.ts:629, :685) but a turn can start in the gap — openPicker then types '/model' as user text into the prompt. The Esc in dismissPicker clears the input box so it self-heals, just worth knowing the gate is advisory, not atomic.
- Nit: busyReply (model-command.ts:601) takes a deps argument it never uses.

Rationale: read-mostly tmux interaction with verified dismissal, hard timeout, label-hash callback identity, kill-switch fallback, and 28 new tests including a verbatim 2.1.170 fixture — the failure modes degrade to a static menu, never to a wrong model or a wedged pane.

PR #2264, opened 2026-06-10
feat(agents): root agent auto-provisions the docker CLI on first boot

mekenthompson:
## Summary

Follow-up to the root debugging agent (#2261). A root agent (`root: true`) has `/var/run/docker.sock` mounted and its `CLAUDE.md` + `docs/root-agent.md` promise `docker ps/logs/exec` for fleet debugging — but the shared agent image deliberately omits the docker client, so on bring-up the binary had to be installed by hand.

This makes it automatic: `start.sh` fetches a version-pinned static `docker` client into the persistent `$HOME/.local/bin` on a root agent's **first boot**.

## Why this, not bake-into-image

The docker client is **~38 MB**. Baking it into the shared agent image is a ~25% size increase on **every** agent (10+), for a feature used by typically **one** root agent — and the docs already flag the heavy agent image as a roll/cache pain point. So instead:

- **gated** on the in-container `SWITCHROOM_AGENT_ROOT` marker (emitted only for `root: true` — see `compose.ts`), so non-root agents are completely untouched
- **idempotent** — skips when `$HOME/.local/bin/docker` is already present → one-time fetch, persists across restarts via the `/state` bind mount
- **non-fatal** — a fetch failure warns and the agent boots fully functional minus docker, retried next restart

Aligns with the persistent-HOME Layer-1 model (`pip --user`, `npm -g`) already used for agent self-provisioned tools.

## What changed
- `profiles/_base/start.sh.hbs` — the gated provisioning block, placed right after the stray-claude prune and before the `CLAUDE_CONFIG_DIR` launch anchor.
- `tests/scaffold.regenerate-start-sh.test.ts` — pins the gate, idempotency guard, non-fatal warn, and ordering-before-claude.
- `docs/root-agent.md`, `CHANGELOG.md`.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `scaffold.regenerate-start-sh.test.ts` green (+1)
- [x] Validated live during overlord bring-up: the same static client (27.3.1) installed to `$HOME/.local/bin` reaches the daemon and lists the fleet (`docker ps`).

## Risk
Low. Behaviour for every non-root agent is byte-identical (the block is a runtime `if` on `SWITCHROOM_AGENT_ROOT`, false for them). For root agents the only new behaviour is a one-time, non-fatal network fetch on first boot. No image change, no impact on rolls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## unreleased — root agent auto-provisions the docker CLI
+
+Follow-up to the root debugging agent (#2261). A root agent has
+`/var/run/docker.sock` mounted but the shared agent image omits the
+~38 MB docker client (inert for every non-root agent). On a root agent's
+first boot, `start.sh` now fetches a version-pinned static `docker`
+client into the persistent `$HOME/.local/bin` — gated on
+`SWITCHROOM_AGENT_ROOT`, idempotent (skips when present), and non-fatal
+(a fetch failure leaves the agent fully functional minus docker, retried
+next restart). `docker ps/logs/exec/inspect` then work out of the box,
+no manual install. See `docs/root-agent.md`.
+
 ## v0.15.2 — memory-bank health, /model command, root debugging agent (#2257, #2258, #2238, #2259, #2261)
 
 Built mid-incident on 2026-06-10: every active agent had silent memory
--- a/docs/root-agent.md
+++ b/docs/root-agent.md
@@ -70,6 +70,15 @@ host (`docker run -v /:/x …`), so the mounts and uid only make that
 reach *ergonomic*, not *new*. Mem/CPU/PID limits and the tmpfs `/tmp`
 are unchanged.
 
+**The `docker` CLI.** The shared agent image deliberately omits the
+~38 MB docker client (it's inert for the other agents, which have no
+socket). So on a root agent's **first boot**, `start.sh` fetches a
+version-pinned static `docker` client into `$HOME/.local/bin` (which
+persists across restarts via the `/state` bind mount) — gated on the
+`SWITCHROOM_AGENT_ROOT` marker, idempotent, and non-fatal (if the fetch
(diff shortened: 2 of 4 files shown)


mekenthompson:
APPROVE — gate, idempotency, and non-fatality all check out, and the gating env var is real at head.

- profiles/_base/start.sh.hbs:65 — the gate is correct: `SWITCHROOM_AGENT_ROOT` is emitted only for `a.root === true` (verified at head, src/agents/compose.ts:1646-1648), so the block is a runtime no-op for every non-root agent. The `! -x` presence check makes it one-shot across restarts via the persistent $HOME.
- profiles/_base/start.sh.hbs:70-79 — non-fatal is structurally sound: the entire curl→tar→mv→chmod chain is the `if` condition, so a failure takes the WARN branch instead of tripping any errexit, and cleanup at :80 runs on both paths.
- Should-fix (non-blocking): no checksum verification on the fetched binary. This tarball becomes a uid-0-executable in a container holding docker.sock and `/host` rw — the highest-privilege surface in the fleet. The version is already pinned, so pinning the known sha256 for 27.3.1 per-arch (`echo "<sha>  $HOME/.cache/docker-cli.tgz" | sha256sum -c`) costs three lines and removes the trust-the-CDN-and-TLS assumption. Fine to land as-is given HTTPS + vendor CDN, but this is the one container where supply-chain paranoia is proportionate.
- Nit: profiles/_base/start.sh.hbs:67 — `uname -m` is used unmapped in the URL. Correct for x86_64/aarch64 (download.docker.com uses those literal names); any other arch just takes the graceful WARN path, so this is fine, merely worth a comment.
- tests/scaffold.regenerate-start-sh.test.ts:95-117 — pins the gate string, idempotency guard, non-fatal WARN, and the before-CLAUDE_CONFIG_DIR ordering; the assertions match the template byte-for-byte.

Rationale: 78-line additive change, runtime-gated to the single root agent, fails open to a functional boot, and validated live during the overlord bring-up; the only real ask is the checksum pin as a follow-up.

PR #2265, opened 2026-06-10
fix(auth-broker): make live quota authoritative over stale exhaustion marks

mekenthompson:
## Summary
The 2026-06-10 fleet failover outage was a **stale-data-over-live-truth inversion**. The broker judged an account eligible-or-not purely by the persisted `exhausted_until` mark, never consulting the live quota probe already cached beside it. One misfired `+7d` weekly mark on the healthy primary (pixsoul, live 7d=20%) stranded the **whole fleet** — every failover returned `no-eligible-target` — and it survived `auth use`/`refresh` **and** a broker recreate because nothing clears a mark on a healthy probe. The one failover that fired sent hindsight onto an account whose mark had *expired* (looked eligible) while its **live 5h was 100%**.

Operator directive: **rely on accurate live status, never stale JSON, especially >24h old.** Encoded as *most-recent-signal-wins*.

## Changes
- **New pure module `src/auth/broker/account-eligibility.ts`** (17 unit tests):
  - `isAccountBlocked` — a fresh (≤24h) snapshot that is *newer than the mark* is authoritative (walled→blocked, healthy→not); otherwise the mark governs.
  - `snapshotShouldClearMark` — self-heal a stale mark on a clearly-healthy probe (both windows <80%); a real weekly wall (7d≥99.5%) is never cleared.
  - `clampMarkExpiry` — override a >5h mark **only** when live data *disproves* the weekly wall; absence of evidence trusts the caller, so the legit #2218 gateway-parsed weekly reset still holds.
- **Wired into the three broker decision points**: `isAccountExhausted` (shared serving/failover predicate), `nextHealthyAccount`, and the `listState` `exhausted` field (stops the dashboard reporting `exhausted=true` while live util is healthy — the contradicting-fields bug).
- Marks carry `marked_at` for recency; legacy marks (no `marked_at`) are overridden by any fresh probe (safe direction — that's how a bogus legacy mark gets ignored).
- **Self-heal** wired in `cacheQuotaSnapshot`; consumer tick now caches its probe so its accounts get the same warmth + self-heal + clamp corroboration.
- Boot reprobe already present (#2239) → a broker recreate is now a real reset, not a stale-mark re-affirmation.

## Test plan
- [x] `account-eligibility.test.ts` — 17 tests incl. both incident shapes, legacy-mark, >24h staleness, newer-mark-wins, clamp-only-on-contradiction.
- [x] `server.test.ts` — **Incident A** (bogus future mark on live-healthy account → served + self-healed) and **Incident B** (failover skips a live-walled candidate whose mark expired) as full broker integration tests.
- [x] 289 broker + 67 server tests green; existing #2218 weekly-durability test still passes; `tsc --noEmit` clean.
- [x] 42 failures in the full `tests/` sweep are **pre-existing on upstream/main** (bundler/CLI suites needing build artifacts) — identical count with my changes stashed.

## Risk
Surgical change to the failover heart, but the predicate is a pure, exhaustively-tested function and the fallback-to-mark path preserves today's behavior whenever live data is absent/stale. Kill path: no new flags — behavior degrades to the old mark-only logic when `lastQuotaCache` is empty. Deploy = broker recreate.

### DIFF
--- a/src/auth/broker/account-eligibility.test.ts
+++ b/src/auth/broker/account-eligibility.test.ts
@@ -44,11 +44,11 @@ describe("snapshot predicates", () => {
 });
 
 describe("isAccountBlocked — live truth is authoritative over the mark", () => {
-  it("INCIDENT A (pixsoul): bogus future mark, but fresh healthy probe → NOT blocked", () => {
+  it("INCIDENT A (healthy primary): bogus future mark, but fresh healthy probe → NOT blocked", () => {
     const mark: ExhaustionMark = { exhausted_until: NOW + 7 * 24 * 60 * 60 * 1000, marked_at: NOW - 60_000 };
     expect(isAccountBlocked({ mark, snapshot: snap(4, 20, 30_000), now: NOW })).toBe(false);
   });
-  it("INCIDENT B (outlook): stale PAST mark, but fresh walled probe → blocked", () => {
+  it("INCIDENT B (walled secondary): stale PAST mark, but fresh walled probe → blocked", () => {
     const mark: ExhaustionMark = { exhausted_until: NOW - 60_000, marked_at: NOW - 3 * 60 * 60 * 1000 };
     expect(isAccountBlocked({ mark, snapshot: snap(100, 27, 30_000), now: NOW })).toBe(true);
   });
@@ -91,7 +91,7 @@ describe("snapshotShouldClearMark — self-heal", () => {
 
 describe("clampMarkExpiry — only override a long mark the live data DISPROVES", () => {
   it("clamps a >5h mark when a fresh probe CONTRADICTS the weekly wall (7d healthy)", () => {
-    // The pixsoul misfire shape: a +7d mark proposed while live 7d=20%.
+    // The healthy-primary misfire shape: a +7d mark proposed while live 7d=20%.
     const proposed = NOW + 7 * 24 * 60 * 60 * 1000;
     expect(clampMarkExpiry({ proposedUntil: proposed, now: NOW, shortMs: FIVE_H, snapshot: snap(4, 20, 0) }))
       .toBe(NOW + FIVE_H);
--- a/src/auth/broker/account-eligibility.test.ts
+++ b/src/auth/broker/account-eligibility.test.ts
@@ -1,0 +1,119 @@
+import { describe, it, expect } from "vitest";
+import {
+  isAccountBlocked,
+  snapshotShouldClearMark,
+  clampMarkExpiry,
+  snapshotFresh,
+  snapshotWalled,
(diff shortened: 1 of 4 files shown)


mekenthompson:
APPROVE

- src/auth/broker/account-eligibility.ts:224-239 — `isAccountBlocked` correctly implements most-recent-signal-wins: fresh (≤24h) snapshot newer than the mark decides; otherwise mark governs. Legacy marks (`marked_at` absent → 0) are overridden by any fresh probe, which is the safe direction. Verified the wiring at all three decision points (server.ts:1043 `isAccountExhausted`, server.ts:2313-2325 `nextHealthyAccount`, server.ts:1124-1130 `listState.exhausted`).
- src/auth/broker/server.ts:1265-1290 — `lastQuotaCache` entries carry `capturedAt` + both utilization pcts at head (server.ts:1241-1249 pre-change), so they structurally satisfy `QuotaSnapshot`; the extra fields are fine under structural typing. Self-heal in `cacheQuotaSnapshot` deletes + persists atomically via `persistQuota` (server.ts:2432, atomic write).
- account-eligibility.ts:158 — `WALL_PCT = 99.5` matches `EXHAUSTION_PCT` in consumer-quota-sensor.ts:24 as the comment claims. Verified.
- account-eligibility.ts:198 — `snapshotFresh` includes a +60s clock-skew guard rejecting future-dated snapshots; test-pinned (account-eligibility.test.ts:38).
- Watch (non-blocking): server.ts:1499-1511 — the mark-exhausted RPC clamp can defeat a GENUINE gateway-parsed weekly reset when the cached probe is up to 24h old and shows 7d just under the wall (e.g. a 3h-old probe at 7d=95% while the account walls now). The mark clamps to 5h and gets re-asserted on the next wall, so it self-corrects with some churn rather than stranding — acceptable, but worth a fresh-probe-on-mark follow-up if the churn shows up in audit logs.
- Nit: account-eligibility.ts:282-283 — `clampMarkExpiry` only checks the 7d window for contradiction. Correct for the weekly-wall case it targets, but a long mark proposed for a 5h wall would also survive; harmless today since long untils only come from weekly parsing.
- Tests: the two incident-shape integration tests (server.test.ts:310-361) drive the real broker through `fleetQuotaProbeTick` with a per-account quota seam and pin both outage shapes; the 17-case unit suite covers staleness, legacy marks, newer-mark-wins, and the #2218 no-evidence-trusts-caller path (account-eligibility.test.ts:110-118).

Rationale: the root-cause inversion (mark-only eligibility) is fixed by a pure, exhaustively-tested predicate; fallback-to-mark preserves old behavior when live data is absent, so the blast radius on the failover heart is well-contained.

Issue #910, opened 2026-05-10
Container HOME=/state/agent/home breaks ~/.switchroom/... tilde paths in cron yaml prompts


mekenthompson:
## Summary

\`src/agents/compose.ts\` line 516 sets \`HOME=/state/agent/home\` for agent containers (justified — the numeric UID has no /etc/passwd entry, so default HOME breaks gh/git/pip/etc.). But yaml \`schedule\` cron prompts widely use \`~/.switchroom/agents/<name>/...\` and \`~/.switchroom/skills/...\` paths. Inside the container, \`~\` expands to \`/state/agent/home\`, which contains only \`analytics-id\` and \`logs/\` — none of the agent dir, skills, or credentials live there.

## Forensic evidence (gymbro, switchroom 0.7.7)

\`\`\`bash
$ echo \$HOME ; ls -d ~/.switchroom/agents/gymbro
/state/agent/home
ls: cannot access '/state/agent/home/.switchroom/agents/gymbro': No such file or directory
\`\`\`

The bind-mount of the agent dir lives at \`/home/<user>/.switchroom/agents/<name>/\` AND \`/state/agent/\` (dual mount in compose.ts:576-580), neither of which is under \`\$HOME\`.

Affected yaml prompts include all gymbro cron entries that reference \`~/.switchroom/agents/gymbro/.claude/skills/food-log/scripts/food-log\`, \`~/.switchroom/agents/gymbro/workspace/memory/health-log/\$(date +%Y-%m-%d).md\`, etc. Same problem in any user yaml that uses tilde paths in cron prompts.

## Suggested fix (pick one)

1. **Make HOME match the host path**: set \`HOME=/home/<user>\` and ensure \`/etc/passwd\` is populated for the UID (a passwd shim is the standard k8s/container approach). Then \`~/.switchroom/...\` resolves to the dual-mounted location and matches operator intuition.
2. **Symlink \`\$HOME/.switchroom\` → \`/home/<user>/.switchroom\`** at start.sh time, so tilde paths reach the right place even with the remapped HOME.
3. **Document the constraint and lint yaml**: refuse to scaffold agents whose schedule prompts use tilde paths; require absolute \`/home/...\` paths instead. (Worst UX, easiest to ship.)

(1) or (2) is invisible to existing users and breaks no existing yaml. (3) requires every user to rewrite their schedule blocks.

## Related

- switchroom/switchroom#907 (bind-mount + cron regression umbrella).

[...]

PR #2267, opened 2026-06-10
fix(agents): pre-create $HOME/.switchroom symlink for admin/root agents

mekenthompson:
## Summary

Fixes a first-boot bug where a **fresh admin (or root) agent silently fails to answer Telegram DMs**: the gateway connects and buffers inbound messages, but the gateway↔claude **bridge never registers**, so the buffer never drains. The agent looks alive (typing indicators, `getMe` ok) but never replies.

## Root cause — a first-boot ordering race

The admin audit-log bind mounts (`vault-audit.log`, `host-control-audit.log` in `compose.ts`) target paths **under `$HOME/.switchroom`**. On a fresh agent's first `docker up`, docker materialises `$HOME/.switchroom` as a **real directory** to host those mount points — **before** `start.sh` runs. `start.sh`'s `#910` symlink step then refuses to clobber the now-real dir (by design), so `$HOME/.switchroom` stays a real directory.

The bridge MCP server locates the gateway socket at `$HOME/.switchroom/agents/<name>/telegram/gateway.sock`. With `$HOME/.switchroom` a real dir (not the symlink to the host `~/.switchroom`), that path doesn't resolve to where the gateway actually bound its socket (`/state/agent/telegram/gateway.sock`) → the bridge logs `gateway sidecar not running` and never registers.

- **Non-admin agents** dodge this — they have no mounts under the link, so `start.sh`'s runtime symlink wins on first boot.
- **Existing admin agents** (klanker/carrie) work only because their symlink was created on an early boot and **persisted on the host**.

This surfaced when standing up the **first root agent** (#2261) — it buffered DMs for hours without answering. The live instance was repaired by hand (replace the real dir with the symlink + restart → bridge registered immediately); this PR prevents recurrence.

## Fix

`scaffold` pre-creates the host-side `$HOME/.switchroom` symlink (→ the real `~/.switchroom`) when it lays down the persistent HOME — **before the container ever boots** — so docker resolves the audit binds *through* the symlink (the working klanker shape) instead of creating the conflicting real dir.

- **Host-apply-only** — guarded by `isContainerContext()` + `process.env.HOME` (in-container reconcile has the wrong HOME and the link already exists).
- **Create-only** — never clobbers an existing symlink or a real dir (an already-booted agent is left to the live `start.sh`/operator path; can't safely `rm` live mount points anyway).

## Test plan
- [x] `tsc --noEmit` clean
- [x] `scaffold.persistent-home.test.ts` (+2): symlink created → host `~/.switchroom`; idempotent across re-scaffold (no clobber).
- [x] Full `scaffold.test.ts` + `compose-generator.test.ts` green (362 tests) — the new host-scaffold symlink step doesn't regress any existing scaffold behaviour.
- [x] Validated live: the same symlink-replacement on the running root agent made `bridge registered — agent=overlord` appear within seconds and the buffered DMs were answered.

## Risk
Low. Host-apply-only + create-only, so: non-admin agents are unaffected (their boot already worked), existing agents are untouched (link already exists), and the only new behaviour is laying down a symlink for fresh agents that docker would otherwise turn into a bridge-breaking real dir.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ client into the persistent `$HOME/.local/bin` — gated on
 next restart). `docker ps/logs/exec/inspect` then work out of the box,
 no manual install. See `docs/root-agent.md`.
 
+Also fixes a first-boot bridge break for admin/root agents: the admin
+audit-log bind mounts (`vault-audit.log`, `host-control-audit.log`) target
+paths under `$HOME/.switchroom`, so on a fresh agent's first boot docker
+materialised `$HOME/.switchroom` as a real directory before `start.sh`
+could lay down the `#910` symlink the gateway↔claude bridge relies on —
+the bridge then never registered and the agent silently buffered every
+Telegram DM without answering. `switchroom apply` now pre-creates that
+symlink on the host (create-only, never clobbers an existing link/dir)
+before first boot, so docker resolves the binds through it. Surfaced by
+the first root agent; would have hit any freshly-scaffolded admin agent.
+
 ## v0.15.2 — memory-bank health, /model command, root debugging agent (#2257, #2258, #2238, #2259, #2261)
 
 Built mid-incident on 2026-06-10: every active agent had silent memory
--- a/docs/root-agent.md
+++ b/docs/root-agent.md
@@ -79,6 +79,17 @@ persists across restarts via the `/state` bind mount) — gated on the
 fails the agent still boots, and it retries next restart). After first
 boot, `docker ps/logs/exec/inspect` just work.
 
+**The `$HOME/.switchroom` symlink.** A root agent is also an admin agent,
+so it gets the admin audit-log bind mounts (`vault-audit.log`,
+`host-control-audit.log`) at paths under `$HOME/.switchroom`. Because the
+gateway↔claude bridge locates its socket through `$HOME/.switchroom`,
+`switchroom apply` pre-creates that symlink on the host (pointing at the
+real `~/.switchroom`) *before* the container's first boot — otherwise
+docker would materialise `$HOME/.switchroom` as a real directory to host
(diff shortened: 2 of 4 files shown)


mekenthompson:
APPROVE

- src/agents/scaffold.ts:3339-3376 — root cause verified at head: start.sh.hbs:196 only creates the #910 symlink when `$HOME/.switchroom` is absent or already a link, and compose.ts:1739 binds the admin audit log to `/state/agent/home/.switchroom/vault-audit.log`, so docker does materialise a real dir there before start.sh on a fresh admin agent. Pre-creating the link host-side is the right fix shape.
- scaffold.ts:3354 — `isContainerContext()` with default args (agent-config.ts:93) checks `SWITCHROOM_CONTAINER` + `/.dockerenv`; correct guard for host-apply-only. `lstatSync`/`symlinkSync` already imported (scaffold.ts:9,14), `created` tracking consistent with existing usage.
- scaffold.ts:3356-3370 — create-only via `lstatSync` (catches dangling links too, so a broken existing link is NOT repaired — acceptable, that's the operator path), best-effort try/catch matches the non-fatal design.
- Nit: the link is seeded for EVERY agent, not just admin/root. Harmless (it's exactly what start.sh would do at runtime) and simpler than plumbing the admin flag in, but the comment could say so explicitly.
- Nit: tests/scaffold.persistent-home.test.ts:120-163 mutate `process.env.HOME` without containerContext stubbing — fine because the test host is not a container, but the suite would behave differently under a containerized CI. Pre-existing pattern in this file, not new risk.
- CHANGELOG.md:9-19 and docs/root-agent.md:31-40 accurately describe the fix and its scope.

Rationale: ordering race confirmed against the real start.sh guard and compose bind targets; the fix is create-only, host-only, non-fatal, and test-pinned for both creation and idempotency.

PR #2268, opened 2026-06-10
fix(profiles): correct root agent CLAUDE.md guidance (paths, apply, no-tap)

mekenthompson:
## Summary

Reviewing the **rendered** guidance against the first live root agent (overlord) surfaced three inaccuracies in the `{{#if root}}` block of `profiles/default/CLAUDE.md.hbs`. All verified against the running container.

## The three fixes

1. **Wrong log path.** The block said host logs are at `/host/var/log/switchroom/` — that path **doesn't exist**; it's the *in-container* path. From the root agent's `/host` view, per-agent logs are at `/host-home/.switchroom/logs/<agent>/`, or `docker logs switchroom-<agent>` for container stdout. The agent would `ls` the documented path and get nothing.

2. **`switchroom apply` doesn't work from inside the container.** The block said "edit `switchroom.yaml` then `switchroom apply`." But `~/.switchroom/compose/` isn't mounted in the container, so apply can't regenerate the compose file. Replaced with the honest path: edit the yaml + `docker restart` for boot-re-read config; hand a full compose-regenerating `apply` to the operator as a host op.

3. **Contradiction with the shared "Admin surface" block.** Because `root` implies `admin`, the root agent also renders the admin section — which says *"every mutating verb pauses for an operator approval card... a human must tap approve"* — directly contradicting the root tier's standing/no-tap model. (The `hostd` MCP socket those verbs need isn't even mounted for a root agent.) Added an explicit line: the root block **supersedes** the approval-card section; the root agent works through its own `docker` + `/host`, not hostd's gated verbs.

Also: notes the `docker` CLI is auto-provisioned (#2264) so the examples just work, and fixes the same per-agent-log-path example in `docs/root-agent.md`.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `scaffold.persona.test.ts` gains accuracy pins for all three (supersedes line, correct log path + absence of the wrong one, apply-is-a-host-op); 6 tests green.
- [x] Each claim verified against the live overlord container (log path exists / wrong one doesn't; compose dir unmounted; hostd socket absent).

## Risk
None — template/docs/test only. Affects only agents with `root: true` (one today). The live overlord will be re-rendered (`agent restart`, which reconciles CLAUDE.md) after merge.

### DIFF
--- a/docs/root-agent.md
+++ b/docs/root-agent.md
@@ -16,7 +16,8 @@ its schedules like any other agent.
 Reach for a root agent when you find yourself SSHing into the switchroom
 host as root to:
 
-- read another agent's logs (`docker logs`, `/var/log/switchroom/...`),
+- read another agent's logs (`docker logs switchroom-<agent>`, or
+  `~/.switchroom/logs/<agent>/`),
 - `docker exec` into a wedged peer to see what it's stuck on,
 - inspect or edit `~/.switchroom/switchroom.yaml` and re-`apply`,
 - look at host-level state (Coolify, nginx, systemd, disk) that an
--- a/profiles/default/CLAUDE.md.hbs
+++ b/profiles/default/CLAUDE.md.hbs
@@ -144,16 +144,25 @@ You're NOT `admin: true`. If asked to restart agents / read peer logs / exec int
 
 You are the **root debugging agent** — a privilege tier above `admin`. You run as **uid 0 in a container with the host's docker socket and filesystem mounted**, so you have standing, un-tapped root over this host. You exist so the operator can debug the fleet by DMing you instead of opening an SSH root shell. Use that power deliberately.
 
+**This supersedes the "Admin surface" section above.** That section
+describes the `hostd` approval-card flow for ordinary admin agents — a
+human taps Allow before each verb. It does **not** apply to you: your root
+tier is standing and un-tapped, and you work through your own `docker` +
+`/host`, not hostd's gated verbs (which aren't wired into your container).
+Ignore the approval-card model — you are the safety boundary.
+
 What you can reach directly from your shell (no approval card — that's the point):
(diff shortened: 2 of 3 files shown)


mekenthompson:
APPROVE

- profiles/default/CLAUDE.md.hbs:147-152 — the supersedes paragraph correctly resolves a real contradiction: `root: true` implies the admin block renders (test pins `## Admin surface` present), and the admin block's tap-to-approve model genuinely conflicts with the root tier's standing access.
- Log-path fix verified at head: compose.ts:1822 binds host `~/.switchroom/logs/<agent>` → container `/var/log/switchroom`, so `/host/var/log/switchroom/` indeed does not exist on the host filesystem and `/host-home/.switchroom/logs/<agent>/` is the correct host-side path. The old guidance pointed at nothing.
- Apply-is-a-host-op fix is consistent with compose generation living on the host (the compose dir is not in the root agent's mount set, compose.ts:1706-1717 — only docker.sock, `/host-home/.switchroom`, and `/host` are emitted for `root: true`).
- Nit: CLAUDE.md.hbs:150 says hostd's gated verbs "aren't wired into your container" — that's true on the live instance but conditional in general: compose.ts:1779 mounts the hostd socket for EVERY agent (root included) when `host_control.enabled: true` and the daemon dir exists. The sentence would be more accurate as "aren't part of your tier's mount set unless host_control is enabled". Non-blocking; the operative guidance (use your own docker, ignore the approval-card model) is right either way.
- tests/scaffold.persona.test.ts:159-168 — the three accuracy pins assert the rendered output, including the negative pin on the dead path. Good shape; a future regression re-introducing `/host/var/log/switchroom/` fails the suite.
- docs/root-agent.md:9-11 matches the corrected path.

Rationale: all three corrections check out against the real compose mounts and template at head; the only imprecision found is a conditional edge in a parenthetical.

PR #2269, opened 2026-06-10
test: fix two host-dependent test failures (isolation + timeout)

mekenthompson:
## Summary
Two tests pass in CI (clean runner) but fail on a host with a live `~/.switchroom/` — found running the full `src/`+`tests/` suite on the operator box. Test-only changes, no product code.

## Fixes
- **`src/cli/update.test.ts` (2 tests)** — "runs the steps in order" + "hostd install … when host_control enabled" pinned `hostControlEnabled` but not `webServiceManaged`, so `runUpdate` fell through to the host's real `switchroom.yaml`, saw `web_service.managed: true`, and added a 7th `refresh-web` step → `expected length 6 got 7` (plus EACCES overlay-loader noise from reading production agent dirs). Pinned `webServiceManaged: false` to match the existing `hostControlEnabled: false` isolation. This was a CLAUDE.md hard-rule violation (tests must never read production `~/.switchroom/`).
- **`tests/run-hook.test.ts` (2 flaky tests)** — the coalesce + auto-resolve E2E cases cold-start `bun dist/cli/switchroom.js` 4–5 times (~1s each); under full-suite parallel load that exceeds vitest's **5s default testTimeout** → an intermittent *timeout* (not a logic failure; verified flaky: failed/passed/failed across runs). Bumped those two spawn-heavy cases to 30s; 5/5 clean after.

## Test plan
- [x] `update.test.ts` 41/41; `run-hook.test.ts` 13/13 across 5 consecutive runs.
- [x] Full `src/`+`tests/` suite: **9193 passed, 0 failed**, 46 skipped (with `dist/` built).
- [x] No product code changed — only test isolation + timeouts.

## Risk
None to runtime — test-only. Both fixes make the suite host-independent (the CI behavior is unchanged; this just makes local/operator-box runs match).

### DIFF
--- a/src/cli/update.test.ts
+++ b/src/cli/update.test.ts
@@ -786,6 +786,11 @@ describe("runUpdate", () => {
         // switchroom.yaml and the call count would depend on whether
         // the developer running tests has hostd enabled.
         hostControlEnabled: false,
+        // Likewise pin web_service.managed off so the refresh-web step is
+        // skipped — otherwise the host's real switchroom.yaml leaks in and a
+        // dev/operator box with managed: true adds a 7th call (the EACCES
+        // overlay-loader noise + "length 6 got 7" on the live host).
+        webServiceManaged: false,
       });
       expect(code).toBe(0);
       // 6 calls total:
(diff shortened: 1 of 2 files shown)


mekenthompson:
APPROVE

- src/cli/update.test.ts:789-793, 845-847 — `webServiceManaged: false` is the correct isolation pin: verified at head that `runUpdate` falls back to `loadConfig().web_service?.managed` when the option is absent (update.ts:357-366), which is exactly the host-config leak described. The seam is pre-existing (update.ts:85) and already used at update.test.ts:37, so this brings the two unpinned tests in line with the file's own pattern.
- tests/run-hook.test.ts:149,172 — 30s timeouts on the two multi-fork cases only; the single-fork tests stay on the default, so the suite still catches a genuine hang in the common cases. The comment documents why, which is what stops someone "simplifying" it away later.
- Test-only; no product code touched (confirmed: diff spans two test files).

Rationale: both fixes are minimal, match existing patterns in their files, and remove host-state dependence rather than papering over a real failure.

PR #2270, opened 2026-06-10
fix(telegram): /model picker never leaves a dead menu — keep buttons + clear toast

mekenthompson:
## Summary
Operator hit "/model → tap a model → nothing happens" on overlord. Diagnosed live: the picker **drive itself works** (Down + `s` on an idle pane → "Set model to Fable 5 for this session only"). The failure is **UX**, in three places:

1. **Tapped mid-turn** → the picker correctly refuses, but the gateway edited the menu into a button-less "⏳ try again" line → menu collapsed, looked dead. **Now:** a mid-turn tap is an **instant toast-only refusal** (sync `isBusy()`, no drive) that leaves the menu message untouched — buttons survive, tap again when idle.
2. **The toast was always a generic "Working…"** even though the handler computed a rich `answer`. **Now:** ack "Switching…", and every outcome re-renders the menu **with a banner** (✅ set / ❌ failed: reason / ℹ️ already default) — a failure keeps the buttons so the operator retries instead of re-running `/model`. Same "status line + interactive menu" shape as the other dashboards.
3. **The `✔` marks claude's _default-for-new-sessions_, not the live session model** (overlord launched `--model claude-fable-5` but the ✔ sat on "Opus 4.8", so the menu said "Now: Opus" while `/status` said Fable). Relabelled **"Now:" → "Default (new sessions):"** so it stops claiming to be the running model.

New `menuWithBanner()` helper + `toastOnly` outcome flag. No change to the picker *driver* (it was correct) or the kill-switch.

## Test plan
- [x] `model-command.test.ts` updated/extended — busy = `toastOnly` (gateway leaves menu intact), failure/success/already-default all keep the keyboard; 29 pass under **both** bun and vitest.
- [x] `model-picker.test.ts` 19 green (driver untouched).
- [x] `tsc`, `check-plugin-references`, `check-bot-api-wrapping` all clean (uses `ctx.*`, no raw bot.api).
- [ ] Live canary on overlord (deploy + tap each outcome) — pending.

## Risk
Low — UX-only in the callback handler; the picker drive, auth gate, and kill-switch are unchanged. Worst case degrades to the prior edit behavior; `SWITCHROOM_MODEL_MENU=0` still disables the family.

### DIFF
--- a/src/build-info.ts
+++ b/src/build-info.ts
@@ -2,8 +2,8 @@
 // Tracked in git so `tsc --noEmit` works on a fresh clone before `npm run build`.
 // Values are refreshed every time `npm run build` runs.
 
-export const VERSION: string = "0.14.72";
-export const COMMIT_SHA: string | null = "0e840d59";
-export const COMMIT_DATE: string | null = "2026-06-06T00:39:32Z";
-export const LATEST_PR: number | null = 2183;
+export const VERSION: string = "0.15.3";
+export const COMMIT_SHA: string | null = "af652154";
+export const COMMIT_DATE: string | null = "2026-06-10T09:15:23Z";
+export const LATEST_PR: number | null = 2266;
 export const COMMITS_AHEAD_OF_TAG: number | null = 0;
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -18409,15 +18409,29 @@ bot.on('callback_query:data', async ctx => {
         .catch(() => {})
       return
     }
-    // Answer the callback IMMEDIATELY — the select path drives the
-    // picker up to three times (multi-second); leaving the tap
-    // spinning invites a double-tap, which queues a second drive
-    // behind the pane lock and confuses the user. The final state is
-    // conveyed by the message edit (a callback can only be answered
-    // once).
-    await ctx.answerCallbackQuery({ text: 'Working…' }).catch(() => {})
+    const modelDeps = buildModelDeps()
+    // Mid-turn refusal is INSTANT (a sync isBusy() check, no picker drive),
+    // so handle it before the "Working…" ack: toast WHY and leave the menu
+    // message untouched (buttons intact) so the operator taps again when
+    // idle. Editing the menu into a button-less "try again" line was the
+    // "nothing happened" report — the menu looked dead.
+    if (modelDeps.isBusy()) {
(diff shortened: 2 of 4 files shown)


mekenthompson:
APPROVE

- `telegram-plugin/gateway/gateway.ts:18412-18443` — sync `isBusy()` short-circuit before the ack, toast-only refusal, menu untouched. Verified at head: `buildModelDeps().isBusy` is `currentTurn !== null` (gateway.ts:13950), so the check is genuinely synchronous and the comment's claim holds. The `toastOnly` early-return after `handleModelMenuCallback` is correct defence-in-depth.
- `telegram-plugin/gateway/model-command.ts:370-398` — `menuWithBanner()` keeps the keyboard on success / already-default / failure. Confirmed `buildModelMenu` (model-command.ts:262-303) never hard-fails: discover failure falls back to v1 static text, so the "falls back to the banner alone" doc comment is accurate (keyboard-less but not broken).
- `model-command.ts:281-301` — "Now:" → "Default (new sessions):" relabel is the right fix; claude's ✔ row genuinely tracks the new-session default, not the live session model.
- Nit (non-blocking), `model-command.ts:373` — `answer: deps.escapeHtml(result.confirmation)`: `answerCallbackQuery` toasts are plain text, so a confirmation containing `&`/`<` would render as `&amp;`/`&lt;`. Moot today (confirmations are ASCII prose) and the gateway already burned the one-answer-per-callback on "Switching…", so this `answer` is dead on the webhook path anyway — but drop the escape for hygiene.
- Nit (non-blocking), `model-command.ts:349-360` — on discover failure the handler calls `menuWithBanner`, whose `buildModelMenu` immediately re-runs `deps.discover` (a multi-second picker drive that just failed). Cheap retry is defensible, just noting the doubled drive on the failure path.
- Edge case, documented as acceptable: if a turn starts while `select()` is in flight, `menuWithBanner`'s inner `buildModelMenu` hits the `isBusy()` guard and returns `busyReply` (no keyboard) under the success banner. Self-corrects on re-running /model.
- Tests (`model-command.test.ts:316-251`) pin all four outcomes including `toastOnly` and keyboard preservation; the driver and kill-switch are untouched as claimed.

Rationale: UX-only change in the callback handler; every claim in the PR body checks out against head, the previously-dead-menu paths now all preserve buttons, and the behavioral contract is test-pinned.

PR #2271, opened 2026-06-10
fix(telegram): /model applies the ✔ row to the live session + /status reflects it

mekenthompson:
Follow-up to #2270 (the menu-UX fix, already merged). Two related fixes for the operator report *"tapped Default, nothing changed, and /status still says Fable"*:

## 1. Tapping the ✔ row now actually switches the live session
The picker short-circuited on `target.current` — but claude's `✔` marks the **default for new sessions**, a *different axis* from the live session model (set by `--model` at launch). An agent launched on Fable tapping **"Default (Opus)"** is a legitimate session switch, but the `✔` check skipped it → **nothing happened**. Now it always drives the selection; claude harmlessly answers "Kept model as X" if it's already the session model. **This is the exact action the operator took.**

## 2. /status now reflects the live session model
`/status` read switchroom's **configured** model while `/model` drives claude's live **session** model — so they silently disagreed after a switch (status said Fable, the session was Opus). Per the operator's call (keep switches **session-only**), `/status` now mirrors the live override:
- a successful switch returns `selectedModel` (parsed from claude's confirmation via `sessionModelFromConfirmation`),
- the gateway holds it in `activeSessionModelOverride` — **in-memory, clears on restart**, matching the session-only scope (claude's session also reverts on restart),
- `formatAgentLine` renders `model: <configured> · live session: <override>` when active.

## Test plan
- [x] New: `ModelCallbackOutcome.selectedModel`, `AgentMetadata.sessionModel`, `sessionModelFromConfirmation()`.
- [x] Tests: parser cases (paren / Switched-to / Kept / empty), `formatAgentLine` session line present/absent, `selectedModel` on success, ✔-row-still-switches. 92 bun + vitest green; `tsc`, plugin-refs, bot-api clean.
- [x] Live-canaried on overlord (hand-patched bundle; gateway boots clean).

## Known limitation (noted, not blocking)
The override is a best-effort mirror of pane state — an operator who attaches via tmux and types `/model` directly bypasses the gateway, and reverting to the exact session model shows "Kept" (override unchanged). Both narrow; the override clears on restart regardless.

## Risk
Low — parsing + in-memory state + render; no model/API/SDK callsites. `SWITCHROOM_MODEL_MENU=0` still disables the family.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -13767,6 +13767,14 @@ function buildAgentAudit(agentName: string): AgentAudit | undefined {
 // broker's fleet-wide `ListStateData` payload via
 // `buildAuthSummaryFromBroker`, with billingType pulled from the
 // agent's `.claude.json` (the broker doesn't track plan tier).
+/**
+ * Live session-model override set by the `/model` picker (session-only). Held
+ * in gateway memory so it clears on restart, the same point at which claude's
+ * session reverts to the configured model — keeping `/status` honest without
+ * a persisted store. Null when no session switch is active.
+ */
+let activeSessionModelOverride: string | null = null
+
 async function buildAgentMetadata(agentName: string): Promise<AgentMetadata> {
   type AgentListResp = {
     agents: Array<{
@@ -13795,6 +13803,7 @@ async function buildAgentMetadata(agentName: string): Promise<AgentMetadata> {
   return {
     agentName,
     model: a?.model ?? null,
+    sessionModel: activeSessionModelOverride,
     extendsProfile: (a?.extends ?? a?.template) ?? null,
     topicName: a?.topic_name ?? null,
     topicEmoji: a?.topic_emoji ?? null,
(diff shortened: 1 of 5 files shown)


mekenthompson:
APPROVE

- `telegram-plugin/gateway/model-command.ts:373-380` — removing the `target.current` short-circuit is the correct fix: the ✔ row is the new-session default, not the live session model, so tapping it is a legitimate switch. Test pins it (`model-command.test.ts:124-136`, `calls.select` now `["Sonnet"]`).
- `telegram-plugin/gateway/gateway.ts:13767-13778, 13899` — `activeSessionModelOverride` is a module-level global, which is safe here: `buildAgentMetadata` has exactly one call site, `buildAgentMetadata(getMyAgentName())`, and the gateway is per-agent. Verified at head.
- `model-command.ts:96-102` — `sessionModelFromConfirmation` regex handles all the tested shapes ("Set model to X for this session only", "Switched to X", paren-suffix trimming, "Kept …" → null). Correct.
- Nit (non-blocking), `model-command.ts:85` — on a "Kept model as X" confirmation the parser returns null and the `?? target.label` fallback records the override anyway. The PR body's known-limitation section says "reverting to the exact session model shows 'Kept' (override unchanged)", but the code DOES update the override to the tapped label in that case. The behavior is fine (arguably better — the session really is on that model), but the body text and code disagree; worth a one-line doc fix so future readers trust the comment.
- Nit (non-blocking), `model-command.ts:97` — the `\s*\(` cut means "Set model to Opus 4.8 (1M context)…" records the override as "Opus 4.8", dropping the 1M variant, while a picker label fallback would keep it. Cosmetic; /status would show the base name.
- `telegram-plugin/welcome-text.ts:205, 217-221` — `sessionModel` rendering is empty-string-safe and test-pinned both present and absent (`welcome-text.test.ts:176-186`).

Rationale: both halves of the operator report are fixed at the right layer (always drive the selection; mirror the result into /status via in-memory state that expires exactly when the session does), the global override is provably single-agent-scoped, and the behavior is test-pinned under both runners.

PR #2273, opened 2026-06-11
perf(hooks): inject-on-change gating for turn-pacing and workspace-dynamic UserPromptSubmit hooks

mekenthompson:
## Problem

The switchroom-owned UserPromptSubmit hooks re-inject static or unchanged content on every user turn:

- `hook:turn-pacing` injects the full ~3 KB `<turn-pacing>` directive verbatim on every message (inline printf in scaffold.ts).
- `hook:workspace-dynamic` re-injects all dynamic workspace files (HEARTBEAT.md etc., ~4 KB) every turn even when nothing changed.

Measured on a live fleet (26-prompt session): 290 KB of turn-pacing + 103 KB of workspace re-injection in one transcript, ~3k wasted tokens per user turn. The UX cost is concrete: agents hit auto-compaction 2-3x sooner (mid-conversation amnesia) and pay a larger cold-cache read on every bursty Telegram turn (slower time-to-first-token).

## Change

Inject-on-change semantics, default ON, per-agent escape hatch `channels.telegram.inject_on_change: false`.

- New `bin/turn-pacing-hook.sh`: emits the full directive on the first turn of a session or when the directive hash changes; silent otherwise. Per-session state under `$TELEGRAM_STATE_DIR/.hook-state/`.
- `bin/workspace-dynamic-hook.sh`: per-session content-hash gating; a file's section re-emits only on new session or content change. HEARTBEAT.md additionally only injects on heartbeat-shaped prompts (or when it changed).
- Fail-open everywhere: any state-handling error emits the full content; corrupted state never drops context.
- `src/agents/scaffold.ts` wires the new script and flag; `src/config/schema.ts` adds the config key.

## Tests

- 9 new tests for turn-pacing gating, 8 for workspace-dynamic gating (first-turn emit, suppression, content-change re-emit, new-session re-emit, heartbeat gating, corrupted-state fail-open, legacy mode).
- Full vitest suite green, `tsc --noEmit` clean, `bash -n` on both scripts.

## Impact

~7 KB (~1.8k tokens) saved per user turn from turn 2 onward, per agent, fleet-wide. On the measured session: ~180 KB less transcript, which roughly triples turns-until-compaction for chat-heavy agents.

### DIFF
--- a/bin/turn-pacing-hook.sh
+++ b/bin/turn-pacing-hook.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+# UserPromptSubmit hook that emits the per-turn pacing directive.
+#
+# Wired into the agent's .claude/settings.json hooks.UserPromptSubmit by
+# scaffold.ts when context_efficiency inject_on_change is enabled (the
+# default). Replaces the inline `printf` that previously re-injected the
+# static directive on every single turn, causing ~3 KB of redundant context
+# per turn (measured ~290 KB over a 26-prompt session).
+#
+# Environment variables (set in the hook command by scaffold.ts):
+#
+#   TURN_PACING_DIRECTIVE - The full directive text to inject. Scaffold.ts
+#                           embeds the directive in the hook command string.
+#   TURN_PACING_HASH      - sha256 of the directive text (hex). Used to
+#                           detect when the directive changed between
+#                           scaffold regenerations.
+#
+# Inject-on-change semantics: on every fire:
+#   1. Read session_id from stdin JSON.
+#   2. Check per-session state file under $TELEGRAM_STATE_DIR/.hook-state/.
+#   3. If session matches AND directive hash matches recorded state → emit
+#      nothing (the model already has the directive in context).
+#   4. Otherwise emit the directive and record the new state.
+#
+# Fail open: any error in state management falls back to emitting the full
+# directive so context is never silently dropped.
+#
+# State files: $TELEGRAM_STATE_DIR/.hook-state/turn-pacing.<session_id>
+# Each file contains: <directive_hash>\n
+# Stale state files (different session_id) are cleaned up on each new-
+# session write (keeps at most 5 files to avoid accumulation).
+
+set -u
+
+DIRECTIVE="${TURN_PACING_DIRECTIVE:-}"
+DIRECTIVE_HASH="${TURN_PACING_HASH:-}"
+
+# If directive or hash are missing, emit whatever we have and exit.
+if [ -z "$DIRECTIVE" ]; then
+  exit 0
+fi
+
+emit_full() {
+  printf '%s\n' "$DIRECTIVE"
+}
+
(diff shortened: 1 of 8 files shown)


mekenthompson:
REQUEST_CHANGES

- BLOCKER — `docker/Dockerfile.agent:174-179`: the image COPYs each bin hook script explicitly (`run-hook.sh`, `timezone-hook.sh`, `user-profile-refresh-hook.sh`, `workspace-stable-hook.sh`, `workspace-dynamic-hook.sh`) and this PR does not add `bin/turn-pacing-hook.sh` to the list. With `inject_on_change` defaulting to true, the scaffolded hook command is `bash "/opt/switchroom/bin/turn-pacing-hook.sh"` (`src/agents/scaffold.ts:4490`) against a path that won't exist in the container — the script's internal fail-open can't save you when the script itself is missing, so the turn-pacing directive is silently never injected, fleet-wide, on every turn. Add the COPY line (the existing `chmod +x /opt/switchroom/bin/*.sh` then covers it) and ideally extend `tests/docker/dockerfile-agent-bakes.test.ts` to pin the bin-hook list against the scaffold's references so the next new hook can't repeat this.
- Design gap (should fix, could be follow-up) — `bin/turn-pacing-hook.sh:94-99` and the ws-dynamic session gating: suppression state is keyed on (session_id, content hash), but auto-compaction keeps the session_id while evicting the injected content from context. Post-compaction, the pacing directive and workspace content are gone and never re-emitted for the rest of the session — the exact failure the PR is trying to delay becomes permanent-until-restart. A `PreCompact` hook that deletes `$TELEGRAM_STATE_DIR/.hook-state/*.<session_id>` would close this cleanly.
- `bin/workspace-dynamic-hook.sh:215-218` — the fast-path comment says HEARTBEAT suppression "is handled by the HEARTBEAT-aware render path below", but the fast path does `cat "$BODY_FILE"; exit 0` before that path is ever reached, so a cached body re-emitted on session change includes HEARTBEAT regardless of prompt. Behavior is acceptable (it's the emit-everything case), but the comment claims a mechanism that never runs on this path — fix the comment.
- Nit — `bin/turn-pacing-hook.sh:108-116` caps stale `turn-pacing.*` state files at 5, but the ws-dynamic side (`ws-dynamic.*`, `ws-heartbeat.*` writes at workspace-dynamic-hook.sh:298-305, 266-269) has no equivalent cleanup; long-lived agents accumulate one pair per session indefinitely.
- Nit — the no-jq fallback `grep -o '"prompt":"[^"]*"'` (workspace-dynamic-hook.sh:173) truncates at the first escaped quote in the prompt, which can miss a "heartbeat" keyword and strip HEARTBEAT.md on a genuine heartbeat turn. jq is present in the agent image so this is dev-only exposure.

Rationale: the gating design, fail-open posture, escape hatch, and tests are all sound, but the new hook script never reaches the container image, which turns the default-on feature into a fleet-wide silent loss of the turn-pacing directive — one COPY line away from approvable.

PR #2274, opened 2026-06-11
chore: remove HEARTBEAT.md workspace file and injection

mekenthompson:
## Summary

- **No built-in heartbeat scheduler exists** in switchroom — `HEARTBEAT.md` was a ~1.6 KB template injected into *every* agent turn via the workspace-dynamic `UserPromptSubmit` hook, with no corresponding cron. It was pure token overhead on every turn for every agent.
- Removes the template (`profiles/default/workspace/HEARTBEAT.md.hbs`), its constant and type union member (`DEFAULT_HEARTBEAT_FILENAME` in `bootstrap-types.ts`), its entry in `DYNAMIC_BOOTSTRAP_FILENAMES` (`workspace.ts`), its hot-reload classification in `scaffold.ts`, and the entire `_strip_heartbeat_if_needed` gating function plus all associated heartbeat session-state bookkeeping from `bin/workspace-dynamic-hook.sh` (added in #2273 — now superseded).
- Updates docs (`workspace-files.md`, `cli-reference.md`, `reference/onboarding-gap-analysis.md`) and shell script comments to remove `HEARTBEAT.md` references.
- **`HEARTBEAT_OK` sentinel is retained** — it is used by operator cron prompts (e.g. credential canary jobs) for silent-reply suppression and is unrelated to the file injection.

## Test plan

- [ ] `npx tsc --noEmit` — clean (zero errors)
- [ ] `bash -n bin/workspace-dynamic-hook.sh` — clean syntax
- [ ] `npx vitest run` — all pre-existing passing tests continue to pass; three HEARTBEAT gating tests replaced by a single regression test asserting no `HEARTBEAT.md` section is ever emitted by the hook
- [ ] Existing agents with a stale `HEARTBEAT.md` on disk are unaffected — `writeIfMissing` won't reseed (template is gone); operator can delete the stubs manually


mekenthompson:
APPROVE

- bin/workspace-dynamic-hook.sh — `_strip_heartbeat_if_needed` and all ws-heartbeat session-state bookkeeping removed cleanly; `bash -n` passes; zero HEARTBEAT references remain in the hook. PROMPT_TEXT stdin parsing removed too since heartbeat gating was its only consumer.
- src/agents/bootstrap-types.ts:26-36 / src/agents/workspace.ts:104-106 — constant, type-union member, and `DYNAMIC_BOOTSTRAP_FILENAMES` entry all removed; grep at head SHA 7ce4ef1 confirms zero remaining `DEFAULT_HEARTBEAT_FILENAME` usages anywhere in src/ or tests/.
- HEARTBEAT_OK sentinel correctly retained — remaining hits are sentinel-only (profiles/default/workspace/CLAUDE.md.hbs:95, src/agent-scheduler/index.ts:324, telegram-plugin silent-reply paths), unrelated to file injection as the PR body claims.
- Nit: tests/workspace-dynamic-hook-inject-on-change.test.ts:180-204 — the replacement regression test shims `switchroom workspace render` with a payload that already excludes HEARTBEAT, so it only proves the hook adds nothing back, not that the real renderer dropped the file. The renderer-side guarantee rests on the DYNAMIC_BOOTSTRAP_FILENAMES change alone; a unit test on `loadDynamicBootstrapFiles` asserting HEARTBEAT.md is not read would close that gap.
- Nit: src/agents/bootstrap-budget.test.ts:139 — stale literal "Read HEARTBEAT.md. Reply HEARTBEAT_OK." in the preserve-exact-prompt test; harmless (it's just an opaque string) but worth rewording in a follow-up.
- Watch: existing agents keep a stale HEARTBEAT.md on disk and reconcile won't delete it (writeIfMissing only seeds). The PR body says operator deletes manually — fine, but until then every live agent carries a dead file whose instructions ('this file is read on every turn') are now false.

Rationale: pure-overhead removal with no behavioral loss — there was never a scheduler behind the file, the sentinel survives, and the removal is complete and consistent across hook, renderer, scaffold classification, templates, docs, and tests.

PR #2275, opened 2026-06-11
fix(compose): config bind source must be a host path, never a container path

mekenthompson:
## Summary
**Root cause of the 2026-06-11 fleet outage.** When `apply` (or hostd's `/update apply`) runs **inside a container**, `getConfigPath()` returns the in-container config location `/state/config/switchroom.yaml` (set via `SWITCHROOM_CONFIG`), and the compose generator emitted it verbatim as a bind-mount **source**:

```
- /state/config/switchroom.yaml:/state/config/switchroom.yaml:ro
```

That path doesn't exist on the host, so **docker auto-created an empty directory** there. The brokers then read a directory as their config → **auth-broker EISDIR**, **vault-broker SQLiteError** (couldn't resolve its grants-DB path) — both crash-looped, so all 12 agents were stuck **"Created"** behind their `depends_on` health gate. Whole fleet down.

## Fix
A bind **source** must live on the host filesystem; `switchroomConfigPath` (where the *running generator* reads its own config) isn't guaranteed to be one. New `resolveConfigMountSource()` sanitises it: any `/state/…` container-only path is replaced with the canonical host config `<homePrefix>/.switchroom/switchroom.yaml` — the same `homePrefix` every other mount already uses, correct whether the generator runs on the host or in a container. A genuine custom **host** path passes through untouched; `undefined` still emits no config mount (back-compat). Resolved once, used at all three emission sites (vault-broker, kernel, auth-broker).

Defence **at the generator** so no caller — host CLI, hostd, or a future in-container path — can poison the mount again.

## Test plan
- [x] `resolveConfigMountSource` unit cases: undefined → undefined; host path → unchanged; `/state/config/switchroom.yaml` and any `/state/…` → canonical host path.
- [x] `generateCompose` end-to-end: a container-path `switchroomConfigPath` **never** becomes a bind source and the config mounts from the canonical host path; a genuine host path is preserved.
- [x] 165 compose-generator + 277 docker/agent tests green; `tsc` clean; build clean.
- [x] **Live-validated by the recovery itself**: regenerating compose from the host shell produced the correct `/home/kenthompson/.switchroom/switchroom.yaml` source and the fleet came back healthy.

## Risk
Low — pure path resolution, no IO. Only rewrites a source that could never have worked (a container path as a host bind source); host paths and the no-mount back-compat case are untouched.

## Follow-up (separate)
Re-pin the fleet to the context-diet build (#2273/#2274) only after this ships — that build is what regenerated the broken compose via the in-container apply path; the fleet is rolled back to v0.15.3 meanwhile.

### DIFF
--- a/src/agents/compose.ts
+++ b/src/agents/compose.ts
@@ -641,6 +641,44 @@ export function normalizeBindMountPath(p: string): string {
   return out;
 }
 
+/** The in-container path the config is mounted AT (also the SWITCHROOM_CONFIG
+ *  value for containerised services). Never a valid host bind SOURCE. */
+export const CONTAINER_CONFIG_PATH = "/state/config/switchroom.yaml";
+
+/**
+ * Resolve the HOST bind-mount source for the switchroom.yaml config mount.
+ *
+ * `switchroomConfigPath` is the path the *running generator* reads its config
+ * from. On the host that's the real host path; but when `apply` runs INSIDE a
+ * container (the hostd `/update apply` path) it's the in-container path
+ * `/state/config/switchroom.yaml` — which is NOT a host path. Emitting it as a
+ * bind source makes docker auto-create an empty host directory, so the broker
+ * reads a directory and dies with EISDIR (the 2026-06-11 fleet outage).
+ *
+ * A bind source must live on the host filesystem. Any path under `/state/`
+ * (the container-only state tree, including the config-mount target itself)
+ * can never be a host source, so it's replaced with the canonical host config
+ * `<homePrefix>/.switchroom/switchroom.yaml` — the same `homePrefix` every
+ * other mount uses, correct whether the generator runs on the host or in a
+ * container. A genuine custom HOST path is passed through untouched. Returns
+ * `undefined` when no config path was given (back-compat: no config mount).
+ * Pure — no IO.
+ */
+export function resolveConfigMountSource(
+  switchroomConfigPath: string | undefined,
+  homePrefix: string,
+): string | undefined {
+  if (!switchroomConfigPath) return undefined;
+  // `/state/...` is the in-container state mount — never a host source.
+  if (
+    switchroomConfigPath === CONTAINER_CONFIG_PATH ||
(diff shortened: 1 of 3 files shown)


mekenthompson:
APPROVE

- src/agents/compose.ts:666-679 — `resolveConfigMountSource` is correct: `undefined` → no mount (back-compat), `/state/…` → canonical `<homePrefix>/.switchroom/switchroom.yaml`, genuine host path untouched. Pure, no IO, matches the test matrix.
- src/agents/compose.ts:836-838 — sanitising once at the top of `generateCompose` and shadowing the local means every downstream emission gets the resolved value. Verified at head: vault-broker (1020), kernel (1192), auth-broker (1307), AND the per-agent emitter via the param threaded at 1329 → emitAgentService → line 2043. That's four emission sites covered, not the three the PR body claims — the fourth (agent services) is covered by the same shadowing, which is exactly why generator-level defence beats per-site fixes.
- src/agents/compose.ts:804 — `homePrefix` may be the literal `${HOME}` when no `homeDir` is passed; the rewritten source then relies on compose env interpolation, same as every other mount on that prefix. Consistent, no new behaviour.
- tests/docker/compose-generator.test.ts:105-149 — unit matrix plus two end-to-end `generateCompose` assertions (container path never a bind source; host path preserved). The negative regex at :136 pins the exact outage line shape.
- Nit: src/build-info.ts:79-87 — generated build-stamp churn (`COMMIT_SHA`/`LATEST_PR`/`COMMITS_AHEAD_OF_TAG`) bundled into a fix PR. Harmless but noise in the diff; consider excluding build-stamp refreshes from feature/fix commits.
- Edge accepted: a host that genuinely keeps its config under a real `/state/` directory would get silently rewritten. Documented in the docstring and vanishingly unlikely; fine as-is.

Rationale: minimal, generator-level fix for the 2026-06-11 fleet outage with the root cause precisely identified, full test coverage of both the bug shape and back-compat, and live validation by the recovery itself. Safe to merge.

PR #2276, opened 2026-06-11
fix(compose): translate container config path to host path in apply/restart

mekenthompson:
## Summary

**Incident:** 2026-06-11, rollback to v0.15.3. During a fleet update to sha-18e9d15 (PRs #2273/#2274), `switchroom apply` was run from inside the overlord agent container. The regenerated compose file had `/state/config/switchroom.yaml` as the bind-mount **source** — the container-internal path — instead of the host path. Docker auto-created an empty directory at that path on the host. vault-broker and auth-broker crash-looped with `EISDIR` on every start; the entire fleet was stuck in "Created". Recovery: rollback to v0.15.3.

## Root cause

`findConfigFile()` returns `process.env.SWITCHROOM_CONFIG` first. Inside an agent container `SWITCHROOM_CONFIG=/state/config/switchroom.yaml`. The callers in `apply.ts` and `write-compose.ts` passed this verbatim as `switchroomConfigPath` to `generateCompose()`, which emitted it directly as the Docker bind-mount source:

```yaml
- /state/config/switchroom.yaml:/state/config/switchroom.yaml:ro  # BUG: source is a container path
```

This was a latent pre-existing bug made load-bearing by PR #2249 (`fix(restart): regenerate compose on agent restart`). Before #2249, `agent restart` didn't regenerate compose, so only a manual `switchroom apply` from the host would trigger it. After #2249, any in-container `agent restart` (or apply) also triggers the broken path.

## Fix

Three targeted changes:

**`src/agents/compose.ts`** — emit `SWITCHROOM_HOST_HOME` into every agent container env, baked at apply time to `homeDir` (the host operator's home directory). This gives the in-container CLI the information needed to reverse the `/state/config` → `~/.switchroom` bind-mount mapping.

**`src/cli/write-compose.ts`** — add `resolveHostSwitchroomConfigPath(rawPath)`:
- On the host (non-docker): returns the path unchanged.
- Inside a container with a `/state/config/…` path: translates to `$SWITCHROOM_HOST_HOME/.switchroom/<filename>`.
- Inside a container with `/state/config/…` but no `SWITCHROOM_HOST_HOME` (pre-fix fleet): throws a clear, actionable error rather than emitting a broken compose. The error tells the operator to run `switchroom apply` once from the host.
- Also uses `SWITCHROOM_HOST_HOME` as `homeDir` inside containers so all volume source paths resolve to the host filesystem.

**`src/cli/apply.ts`** — call `resolveHostSwitchroomConfigPath` when resolving `switchroomConfigPath` in the apply action handler.

## Tests

6 new unit cases in `write-compose.test.ts`:
- Host passthrough (no-op)
- Host passthrough for `/state/config/…` paths (operator edge case — not in docker)
- Container translation: `/state/config/switchroom.yaml` → `/home/kenthompson/.switchroom/switchroom.yaml`
- Non-`/state/config/` path inside container (operator explicit `--config`, left unchanged)
- Guard: throws clear error when in container with `/state/config/` path but no `SWITCHROOM_HOST_HOME`
- Full `writeComposeFile` integration: asserts HOST path in bind-mount source, not container path

## Test results

```
✓ src/cli/write-compose.test.ts (9 tests)        — 9/9 pass (6 new)
✓ src/cli/apply.test.ts (19 tests | 1 skipped)   — 18/18 pass
✓ tests/docker/compose-generator.test.ts (159)   — 159/159 pass
tsc --noEmit clean
```

## JTBDs / principles check

- **Outcome:** Always-on — this is a correctness fix that prevents the fleet from being bricked by an in-container apply.
- **Docs test:** the error message is self-contained and actionable.
- **Defaults test:** works on a fresh install (pre-fix containers get the guard; post-fix containers get the translation).
- **Consistency test:** follows the existing `SWITCHROOM_RUNTIME=docker` / `isDockerRuntime()` pattern.


mekenthompson:
APPROVE

- src/cli/write-compose.ts:240-265 — `resolveHostSwitchroomConfigPath` logic is sound: host context passes through, in-container `/state/config/<file>` maps to `$SWITCHROOM_HOST_HOME/.switchroom/<file>` (filename preserved, so legacy `clerk.yaml` names survive), non-prefix container paths (explicit `--config` bind mounts) are trusted, and the no-HOST_HOME case throws an actionable error instead of writing a poisoned compose. Fail-loud beats silent-rewrite for the pre-fix fleet.
- src/cli/write-compose.ts:295 — `homeDir: process.env.SWITCHROOM_HOST_HOME || homedir()` is the quietly load-bearing line: it fixes ALL volume sources for in-container regeneration, not just the config mount. Note this is something the sibling PR #2275 does NOT do — #2275's generator-level rewrite uses `homePrefix`, which for an in-container apply is the CONTAINER home (e.g. `/root`), still not a valid host path. This PR is the complete fix; #2275 is defence-in-depth that only becomes fully correct once this homeDir change lands. Coordinate the merge order / rebase accordingly.
- src/agents/compose.ts:1626-1628 — baking `SWITCHROOM_HOST_HOME` from `hostHomeForChecks` is self-sustaining: post-fix regeneration passes host-home as `homeDir`, so the env var stays correct across cycles. Nit: the comment at compose.ts:1623-1625 says "Emitted unconditionally" immediately above a conditional `if (hostHomeForChecks)` — reword.
- src/cli/apply.ts:1349-1351 — `parentOpts.config ?? findConfigFile()` feeds a `string`-typed param; verified at head that `findConfigFile()` returns `string` (loader.ts:102, throws when nothing found), so no undefined can reach the translator. Type-safe.
- src/cli/write-compose.test.ts:99-191 — env stash/restore is correct, and the matrix covers host passthrough, host `/state/config` edge, container translation, explicit-mount passthrough, the throw guard, and a full `writeComposeFile` integration asserting the host path lands in the bind source. The negative assertion at :189 pins the exact outage line.
- Watch: both this PR and #2275 modify src/agents/compose.ts (different regions, unlikely textual conflict, but semantic overlap on the same outage). After both merge, the #2275 generator rewrite becomes a near-dead backstop behind this PR's throw guard — fine as belt-and-braces, but worth a comment cross-referencing the two so future readers don't think one is redundant.

Rationale: correct and complete fix for the 2026-06-11 outage at the actual root (container-context path resolution), with a fail-loud guard for the pre-fix fleet, full test coverage, and the homeDir fix that the generator-level sibling PR misses. Safe to merge.

PR #2466, opened 2026-06-20
fix(self-improve): break the review-gate infinite re-fire loop (#2462)

### LINKED ISSUE #2462: Self-improve gate infinite re-fire loop: review-turn guard never matches channel-wrapped transcript + gate ingests its own output
## Summary

The self-improvement turn-end gate (`self-improve-stop.mjs`) enters an **infinite re-fire loop**. Once a learning signal trips on a session, every injected `self_improve_review` turn re-trips the gate, injecting another review turn, forever. Observed live on agent `clerk` 2026-06-20 ~05:15–05:32 UTC: the same `[directive-not-bound]` signal re-fired 30+ times with the restate count monotonically climbing (2× → 14×, and a second cluster 3× → 25×), each iteration prepending another `evidence:` to its own evidence string.

This is in the slice merged today (#2452/#2453/#2454).

### Bug 1 — re-fire guard never matches (the loop)

`src/cli/self-improve-stop.ts` (built: `/opt/switchroom/hooks/self-improve-stop.mjs:368-372`):

```js
const lastUser = [...messages].reverse().find((m) => m.role === "user");
if (lastUser && lastUser.text.startsWith("[self-improvement review]")) {
  clearReviewContext(resolveStateDir());
  process.exit(0);                       // <-- intended: don't re-fire on a review turn
}
```

The guard assumes the injected review prompt appears verbatim in the transcript starting with `[self-improvement review]`. It does not. The gateway wraps every injected inbound in a `<channel source="..." source="self_improve_review" ...>…</channel>` envelope before it lands in the transcript, so the stored user text starts with `<channel `, **not** `[self-improvement review]`. `startsWith` is always false → the guard never trips → the gate re-evaluates on every review turn and re-fires. `clearReviewContext` is therefore dead code in practice.

### Bug 2 — the gate ingests its own output (the climbing count)

`runGate()` scans the last `SCAN_WINDOW = 40` transcript messages (`:225`, `:291`) with **no filter on `meta.source`**, so the injected review prompts are scanned as if they were ordinary user turns. The review prompt embeds the prior evidence line:

```
      evidence: <prior sample>
```

`directiveStatements()` re-parses that line as a fresh "standing rule," and because the sample-growth picks the longest variant (`:143-144`), each round prepends another `evidence:` — producing the telltale `evidence: evidence: evidence: …` and incrementing the cluster count every iteration. With `repetitionThreshold = 2` (`:17`) the count is permanently ≥ threshold, so even if Bug 1 were fixed the detector would still feed on itself.

### Secondary

- The detector counts the **review prompt's own boilerplate** as a restated rule — the second cluster's evidence is literally the instruction text `"Never auto-create a cron or a new skill, never edit a shared/other-agent skill."` from rule 4 of the prompt. Instruction text must be excluded from correction detection.
- The guard uses *last* user message only; when real operator messages arrive in the same batch as a review inbound (happened here — operator's "Raise a bug" interleaved), `lastUser` can be the operator turn, another way the guard misses.

## Impact

- Runaway token burn: one tripped signal = unbounded review turns until the gateway is bounced or the session ends.
- Operator-facing confusion: the agent looks perpetually "busy"; the thumbs-up/progress-card ack semantics break (operator asked "are you doing anything?").
- The actual lesson bound correctly on the **first** pass (directive created), so every subsequent fire is pure waste.

## Repro

1. On any agent with self-improve enabled, produce a turn whose transcript contains a directive-like statement restated ≥ `SWITCHROOM_SELF_IMPROVE_THRESHOLD` (2) times (e.g. a cron prompt that states a standing rule twice).
2. Gate trips, injects a `self_improve_review` inbound.
3. That inbound lands in transcript wrapped as `<channel …>[self-improvement review]…`.
4. Next Stop: guard at `:369` fails the `startsWith` check; gate re-scans, now also counting the injected prompt's embedded evidence; re-fires. Loop.

## Proposed fix

1. **Fix the guard to match real transcript format.** Detect the review marker anywhere after the channel envelope, or check `meta.source === "self_improve_review"` on the inbound rather than string-prefixing the wrapped text. Strip the `<channel …>` wrapper before the `startsWith`.
2. **Exclude self-injected turns from the scan.** In `readTranscript`/`runGate`, drop any message whose text carries the `self_improve_review` source marker (or the `[self-improvement review]` body) so the gate can never ingest its own prompts. This kills the climbing-count amplifier independently of Bug 1.
3. **Don't count instruction boilerplate** — exclude the review-prompt rule lines from `directiveStatements`/correction detection.
4. **Add a per-session re-fire breaker** as defense-in-depth: persist `triggering_session` + a fire counter in review-context and refuse to re-fire for the same session beyond N (e.g. 1) within a window, even if detection misfires.
5. Add a regression test that drives the Stop hook with a **channel-wrapped** review inbound in the transcript (not the bare `[self-improvement review]` string) and asserts the gate does NOT re-fire. The existing test used the unwrapped string, which is why this slipped through.

## Mitigation now

Set `SWITCHROOM_SELF_IMPROVE=0` (kill switch, `:23-24`) and bounce the gateway, or lower the blast radius by filtering at the gateway. The agent cannot self-disable (env/gateway-level, not in the agent config overlay).

## Forensic artifacts

- Hook: `/opt/switchroom/hooks/self-improve-stop.mjs` (390 lines, built bundle).
- Guard: `:368-372`. Scan window: `:225`,`:291`. Detection: `:120-168`. Sample growth: `:143-144`. Threshold default 2: `:17`. Enable predicate: `:23-24`. Inject envelope w/ `meta.source`: `:310-326`.
- Live transcript: agent `clerk`, session `5b2f80bd-3f4b-4e4f-a830-dd8dc314a158`, 2026-06-20 05:15–05:32 UTC.

mekenthompson:
Closes #2462.

> **Heads-up for review:** PR #2465 (also open, same author) fixes the same issue with layers 1, 2 and a wrapped-form regression test. **This PR overlaps it deliberately and adds the one layer #2465 lacks: the per-session re-fire breaker (defense-in-depth).** Pick whichever you prefer — merge this and close #2465, fold the breaker into #2465, or cherry-pick. I did not enable auto-merge.

## RCA

The self-improve Stop hook tried to avoid reviewing its own injected review turn with:

```ts
lastUser.text.startsWith("[self-improvement review]")
```

But the gateway wraps every injected inbound in a `<channel source="self_improve_review" ...>...</channel>` envelope before it lands in the transcript. The `[self-improvement review]` banner is therefore **never at offset 0**, so `startsWith` never matched, the guard never tripped, and the gate re-injected a review on every Stop — an **infinite re-fire loop** (observed live on agent `clerk`, 2026-06-20 ~05:15–05:32 UTC).

Compounding it: each re-fire left the prior review prompt inside the 40-message scan window, so the directive detector re-counted the review's own embedded `evidence:` line as a freshly restated rule — the count climbed unbounded (`evidence: evidence: evidence: ...`).

## The four-layer fix

1. **Robust review-turn guard.** New `isReviewInjectedText(text)` in `review-prompt.ts` matches **both** the channel-wrapped form (the envelope's `source="self_improve_review"` attribute, via the `REVIEW_SOURCE` constant — not hardcoded twice) **and** the `[self-improvement review]` banner as a substring (not `startsWith`). The hook now checks **any** recent user message, not just the last, so a batched real+review inbound still trips the guard.
2. **Drop self-injected turns from the gate scan.** The hook filters review turns out of the window before `runGate`, so the review prompt's embedded evidence can never re-count as a signal — kills the climbing-count amplifier independently of layer 1.
3. **Per-session re-fire breaker (defense-in-depth).** A sibling state file keyed by `session_id` (`hasFiredForSession` / `markFiredForSession` in `review-context.ts`, bounded to the last 200 sessions) ensures one tripped session injects **at most one** review, even if the guard logic ever regresses.
4. **Regression tests** using the exact **channel-wrapped** transcript shape that broke it, plus the bare form and the breaker.

No change to the cost guarantee or the real-signal path: a clean turn still returns immediately; a genuine operator correction still fires the review exactly once.

## Files changed

- `src/cli/self-improve-stop.ts` — robust guard, scan-window filter, per-session breaker wiring
- `src/self-improve/review-prompt.ts` — `REVIEW_MARKER` + `isReviewInjectedText()`
- `src/self-improve/review-context.ts` — `hasFiredForSession` / `markFiredForSession`
- `tests/self-improve-stop-hook.test.ts` — channel-wrapped regression + breaker tests
- `tests/self-improve-review-prompt.test.ts` — `isReviewInjectedText` unit coverage
- `tests/self-improve-review-context.test.ts` — breaker unit coverage

## Test evidence

- **Targeted suites:** `self-improve-stop-hook`, `self-improve-review-context`, `self-improve-review-prompt`, `self-improve-gate`, `self-improve-apply-guard` — **49 passed**.
- **Channel-wrapped regression proven against old code:** reverting the guard to the original `startsWith` (keeping the new test) makes the wrapped-form test **fail** — the hook re-injects an `inject_inbound` envelope whose evidence is literally the wrapped review (`pushed back 3×`, the `evidence: <channel ...>` amplifier). With the fix it **passes** (no injection).
- **Typecheck:** `tsc --noEmit` — clean (exit 0).
- **Lint:** `npm run lint` — clean (all 8 checks pass).
- **Build:** `bun run build` — clean.
- Full `vitest run` has 2 pre-existing failing files (`tests/host-control/server.test.ts`, `tests/install/static-binary.test.ts`) that fail identically on a pristine `main` checkout — environment/infra tests unrelated to this change.

### DIFF
--- a/src/cli/self-improve-stop.ts
+++ b/src/cli/self-improve-stop.ts
@@ -31,11 +31,17 @@ import { join } from "node:path";
 import { homedir } from "node:os";
 
 import { runGate } from "../self-improve/gate.js";
-import { buildReviewPrompt, REVIEW_SOURCE } from "../self-improve/review-prompt.js";
+import {
+  buildReviewPrompt,
+  REVIEW_SOURCE,
+  isReviewInjectedText,
+} from "../self-improve/review-prompt.js";
 import { selfImproveEnabled } from "../self-improve/config.js";
 import {
   writeReviewContext,
   clearReviewContext,
+  hasFiredForSession,
+  markFiredForSession,
 } from "../self-improve/review-context.js";
 import type { TurnMessage } from "../self-improve/types.js";
 
@@ -231,21 +237,36 @@ function main(): void {
 
   // Review-turn detection FIRST — BEFORE the gate check, and independent of
   // it. A turn we injected is itself a transcript whose most-recent user
-  // message starts with the "[self-improvement review]" banner. On such a
-  // turn we (a) clear the review-context marker so the apply-guard stops
-  // enforcing on later NORMAL turns, and (b) never recurse by reviewing a
-  // review. This MUST run before `!gate.tripped` returns: a well-behaved
-  // review (the review prompt + a benign skill edit) trips NO learning
-  // signal, so if the clear sat after the gate check the marker would leak
-  // after every real review — wrongly blocking later normal-turn edits and,
-  // worse, letting a normal-turn edit auto-apply against a STALE benchmark.
-  const lastUser = [...messages].reverse().find((m) => m.role === "user");
-  if (lastUser && lastUser.text.startsWith("[self-improvement review]")) {
+  // message is the review prompt. On such a turn we (a) clear the
+  // review-context marker so the apply-guard stops enforcing on later NORMAL
+  // turns, and (b) never recurse by reviewing a review. This MUST run before
+  // `!gate.tripped` returns: a well-behaved review (the review prompt + a
+  // benign skill edit) trips NO learning signal, so if the clear sat after
+  // the gate check the marker would leak after every real review.
+  //
+  // The injected inbound returns CHANNEL-WRAPPED, so its transcript text is
+  //   <channel source="self_improve_review" ...>[self-improvement review] …</channel>
+  // NOT the bare banner. The old `startsWith("[self-improvement review]")`
+  // never matched the wrapped form → infinite re-fire loop (issue #2462).
+  // `isReviewInjectedText` matches the channel envelope's REVIEW_SOURCE and
+  // the banner substring, catching both forms. We also no longer rely on the
+  // *last* user message only: if ANY recent user message is a review we
+  // injected, this is (or follows) a review turn — guard it.
+  if (messages.some((m) => m.role === "user" && isReviewInjectedText(m.text))) {
     clearReviewContext(resolveStateDir());
     process.exit(0);
   }
 
-  const gate = runGate(messages);
+  // Layer 2: drop any self-injected review turns from the gate scan, so the
+  // review prompt's own embedded evidence can never re-count as a signal
+  // (the climbing-count amplifier in issue #2462). Belt-and-suspenders with
+  // the guard above — even if a review slips past detection, its content is
+  // never fed back into runGate.
+  const scanMessages = messages.filter(
+    (m) => !(m.role === "user" && isReviewInjectedText(m.text)),
+  );
+
+  const gate = runGate(scanMessages);
 
   // Cost guarantee: no signal → return immediately, zero added work.
   if (!gate.tripped) process.exit(0);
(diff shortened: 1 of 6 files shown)


mekenthompson:
Superseded by #2465 (already merged). #2465 fixes the loop with layers 1+2 and a stricter, anchored review-turn matcher (rejects mid-text banner mentions; tested). This PR's only unique addition is the per-session breaker — optional defense-in-depth that carries a same-session second-correction suppression edge, and its isReviewInjectedText uses a loose includes() substring that can false-positive on operator messages quoting the banner. Closing in favor of the merged fix. If we later want the breaker, port just hasFiredForSession/markFiredForSession onto a fresh branch off main rather than merging this (it conflicts and regresses the matcher).

Issue #1198, opened 2026-05-13
compose: agent containers run without TZ — in-container cron fires in UTC regardless of resolved timezone


mekenthompson:
## Symptom

`clerk`'s morning-brief cron, configured `0 8 * * *` and intended for `Australia/Melbourne`, fired at 08:00 UTC on 2026-05-13 — 18:00 Melbourne, not 08:00. The UserPromptSubmit hook warned `SWITCHROOM_TIMEZONE unset; systemd unit may be stale, run \`switchroom systemd install\``.

## Root cause

`src/agents/compose.ts:1057+` emits the agent service's `environment:` block. It does not emit `TZ`, `SWITCHROOM_TIMEZONE`, or any `/etc/timezone`-equivalent volume mount.

Confirmed live on this host:

```
$ docker exec switchroom-clerk env | grep -iE '^TZ=|TIMEZONE|SWITCHROOM_TIME'
(empty)
$ docker exec switchroom-clerk cat /etc/timezone
Etc/UTC
$ docker exec switchroom-clerk node -e 'console.log(process.env.TZ, Intl.DateTimeFormat().resolvedOptions().timeZone)'
undefined UTC
```

`src/agent-scheduler/index.ts:95` registers cron expressions via `node-cron.schedule(expr, handler)` with no third-argument timezone option. node-cron falls back to `process.env.TZ` → undefined → system zone (UTC inside the container). `0 8 * * *` matches **08:00 UTC**, not 08:00 Melbourne.

## Why the wiring is missing

`src/config/timezone.ts:resolveTimezone()` exists and cascades agent → profile → switchroom → `/etc/timezone` → UTC. Pre-#906 (Phase 2a Docker migration), `generateUnit` baked `TZ=<resolved>` and `SWITCHROOM_TIMEZONE=<resolved>` into the systemd `[Service]` block. #906 removed `generateUnit` along with the entire systemd dual-path, but the equivalent wiring was never added to the compose generator.

Result: `resolveTimezone` is now dead code for the production path. Its only effect is the CLI warning at `scaffold.ts:2993` and the SOUL.md template (`profiles/default/workspace/USER.md.hbs:15`). Cron, `Date.now()` arithmetic in handlers, and the per-turn time hint all see UTC.

## Repro

Standing fleet. Any agent with `timezone: Australia/Melbourne` (or any non-UTC zone) at any cascade layer:

```bash
docker exec switchroom-<agent> printenv TZ      # empty
[...]

PR #2489, opened 2026-06-21
fix(compose): bind-mount zoneinfo onto /etc/localtime for per-agent timezone

mekenthompson:
## Problem

The agent container's `/etc/localtime` stays at the base image's **UTC** regardless of the operator's declared timezone, so any tool that reads `/etc/localtime` directly (ignoring `TZ`) reports the wrong zone.

`TZ` / `SWITCHROOM_TIMEZONE` (#1198) already cover everything that honors the `TZ` env var — glibc `date(1)`, Node `Intl`, Python, cron. The gap is **statically-linked Go binaries and some Java runtimes**, which read `/etc/localtime` directly. Cosmetic for ~95% of cases; real for any Go-based tooling in the container.

## Why the in-entrypoint fix doesn't work

An `ln -sf /usr/share/zoneinfo/$TZ /etc/localtime` in `start.sh` was the obvious suggestion, on the assumption start.sh runs as root. It doesn't — and that premise is the bug:

- `Dockerfile.agent` sets `USER 10001`; compose sets `user: "<uid>:<uid>"` → the entrypoint runs **unprivileged**.
- The agent service is `read_only: true` with `cap_drop: ALL` and `no-new-privileges`.

So `/etc` is read-only to the agent and `ln -sf` hard-fails — exactly the reported symptom.

## Fix

A docker-daemon bind mount in `compose.ts`, applied by the daemon (root) **before** the entrypoint and exempt from the read-only rootfs:

```yaml
- /usr/share/zoneinfo/<a.timezone>:/etc/localtime:ro
```

- Source derived from the already-resolved per-agent `a.timezone` (validated IANA, resolver fallback `"UTC"`) — automatic fleet-wide, no per-agent hardcoding.
- `existsSync`-guarded because `docker compose up` hard-fails on a missing bind source; when the host lacks the zone file we skip the mount and fall back to the env-var path. (Absolute system path, so a plain `existsSync` rather than the home-relative `conditionalMountPresent` helper used by the skills/fleet mounts.)
- Same absolute path inside the container, so a plain bind with no `homePrefix` rewrite.

## Tests

New `tests/docker/compose-generator-localtime.test.ts` (3 cases): mount emitted `:ro` onto `/etc/localtime`; uses the operator-declared zone (not hardcoded); omitted when the host lacks the zoneinfo file.

Rebased on latest `main` (d44b71b9). `tsc --noEmit` clean; `tests/docker/` + timezone suites green (336 passed, 38 skipped).

### DIFF
--- a/src/agents/compose.ts
+++ b/src/agents/compose.ts
@@ -27,6 +27,7 @@ import { createHash } from "node:crypto";
 import { existsSync, mkdirSync, readFileSync, lstatSync, readlinkSync } from "node:fs";
 import { join, isAbsolute, dirname, resolve } from "node:path";
 import type { SwitchroomConfig, AgentConfig, AgentBindMount } from "../config/schema.js";
+import { isValidTimezone } from "../config/schema.js";
 import { scheduleNeedsCronSession } from "../scheduler/cron-routing.js";
 import { applyDefaultTier } from "../scheduler/tier-selector.js";
 import { resolveAgentConfig } from "../config/merge.js";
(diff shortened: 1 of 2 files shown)


mekenthompson:
Fresh separate-process review (opus, isolated): **APPROVE**. One non-blocking nit applied before merge — `resolveTimezone`'s server-detection fallback returns its string unvalidated, so the bind source now gates on `isValidTimezone(a.timezone)` as well as `existsSync` (keeps a stray `..` out of the path regardless of how the zone resolved), comment tightened, and a traversal-rejection test added. tsc clean; docker suite 286 passed.

Issue #2527, opened 2026-06-22
DM agents leave an early-ack 👀 reaction but send no reply — looks 'done' while silent (v0.15.56)


mekenthompson:
## Symptom (operator report, repeated)

> "Twice now you are failing to tell me status of what is happening and marking original question with thumbs up so it feels like you are done and not sending anything further. We should have fixed this with the latest release."

The user sends a DM, sees an early-ack reaction appear on their message, then the agent works for minutes (or ends the turn) **without ever sending a reply**. The reaction reads as "handled / done", so the user is left staring at a thumbs-up/eyes emoji with no actual answer. Happened twice in one session (v0.15.56).

## Mechanism

`maybeEarlyAckReaction(ctx, from)` — `telegram-plugin/dist/gateway/gateway.js:62103` (called at :62095):

- On a **private** chat, for an allowed sender, when no turn is already active, it sets a 👀 reaction (`👀`) on the inbound message and fires a `typing` action.
- That is the **only** guaranteed visible artifact of "I received this." It is never cleared or escalated.

The gap: nothing converts either (a) a long silent work stretch or (b) **a turn that ends with zero `reply`/`stream_reply` tool calls** back into a visible message. If the model finishes a turn without calling the reply tool, the user's entire experience is the 👀 reaction — which looks like completion.

## Concrete evidence (this session, v0.15.56)

Overlord was asked to fix marko's OneDrive connection. After the user's "still 403" message, overlord spent ~6 minutes diagnosing + restarting marko's container with **no interim reply** — only the early-ack reaction was visible. The user sent "Status?" twice during that window. v0.15.56's execution-discipline floor did not prevent it.

## Asks

1. **Add debug logging around the reply/reaction path** so we can detect the failure mode in logs:
   - log every `maybeEarlyAckReaction` fire (chatId, msgId, emoji)
   - log every `reply`/`stream_reply` tool invocation with its turn id
   - **log a warning when a turn ends with an early-ack reaction set but zero reply-tool calls** (the smoking gun)
   - log turn start/end boundaries so silent-gap duration is measurable
2. **Consider a backstop**: on turn-end with no reply sent, either emit a fallback message or escalate the 👀 to a visible "still working / no answer produced" state, so an early-ack can never masquerade as a completed answer.
3. Confirm whether the v0.15.56 execution-discipline change is reaching the model in DM turns (it evidently isn't fully binding here).

## Environment
- switchroom v0.15.56 (current fleet)
[...]

PR #2532, opened 2026-06-22
fix(telegram): role-aware terminal honesty on bridge crash (#2527 follow-up)

mekenthompson:
## Summary

Follow-up to #2530. Auditing **every** reaction-finalize path (asked: do the resume/disconnect terminal paths inherit the #2527 fix?) surfaced one real gap: **`disconnect-flush.ts:110`** sweeps all in-flight reactions to **👍 unconditionally** when the claude bridge crashes/restarts mid-turn. That's the false-done bug in its **crash/restart variant** — a user turn killed mid-flight, never delivered, painted 👍 — and it's against the job spec ("crashes/restarts must not read as done"). #2530 only made the main `turn_end` handler honest.

## Fix
- **disconnect-flush.ts** — optional `resolveDisposition(key)` dep; the bridge-death sweep finalizes **😐 (`undelivered`)** for a user turn killed undelivered, **👍** otherwise. Defaults to `'done'` (back-compat + the "never leave a stuck working emoji" guarantee is preserved). Both `Ctrl` generic constraints widened to admit `'undelivered'`.
- **gateway.ts** — wires `resolveDisposition` through the **same pure `decideTerminalReason`** the main gate uses, keyed on the in-flight `currentTurn`. No worker-hold carve-out (a crashed bridge has nothing still working); unclassifiable keys → `'done'`. If a restart resumes and the turn then delivers, the resume's reply re-finalizes 👍; if it never delivers, 😐 stays honest.

## Audit result (the other paths are fine)
- `8260` reply executor → 👍 = delivered (correct)
- `11197` NO_REPLY/HEARTBEAT_OK → 👍 = legitimate silence (correct)
- `11494` turn-flush backstop → a flush sets `finalAnswerDelivered=true` (correct)
- `reaction-defer.ts:84` → downstream of an already-decided `'done'` (worker-hold; accepted)
- **`disconnect-flush.ts:110` → the one gap, fixed here**

## Test plan
- [x] 3 new `gateway-disconnect-flush.test.ts` cases (default-done back-compat; role-aware undelivered/done split; every-key-still-finalized)
- [x] `npm run lint` clean (incl. strict plugin-references tsc — caught + fixed a second generic-constraint site)
- [x] Full `vitest run telegram-plugin/` — 4303 pass / 0 fail
- [x] kill-switch `SWITCHROOM_TG_TERMINAL_HONESTY=0` still reverts (via `decideTerminalReason(enabled:false)`)

## Risk
Low — additive, default-`'done'` (no behaviour change unless the resolver fires), only repaints undelivered user turns on a crash, and kill-switchable. disconnect-flush is incident-scarred (5min-restart-wedge), but the map-clearing/sweep behaviour is untouched — only the finalize *reason* changes for undelivered user turns.

### DIFF
--- a/telegram-plugin/gateway/disconnect-flush.ts
+++ b/telegram-plugin/gateway/disconnect-flush.ts
@@ -27,7 +27,7 @@
  * needing to spin up the whole gateway.
  */
 
-export interface DisconnectFlushDeps<Ctrl extends { finalize: (reason?: 'done' | 'error') => void }, Stream extends { isFinal: () => boolean; finalize: () => Promise<void> }> {
+export interface DisconnectFlushDeps<Ctrl extends { finalize: (reason?: 'done' | 'undelivered' | 'error') => void }, Stream extends { isFinal: () => boolean; finalize: () => Promise<void> }> {
   /** The disconnecting client's agentName. `null` ⇒ anonymous (never registered). */
   agentName: string | null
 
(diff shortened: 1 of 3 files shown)


mekenthompson:
Closing — fresh-process review (correctly) blocked this. The premise is wrong: a bridge-crash-killed turn is normally AUTO-RESUMED, but the disconnect-flush sweep deletes the reaction controller and the resume turn never traverses handleInbound (no new controller), so a 😐 painted here gets STUCK even when the resume successfully delivers — making the common case worse, not better. The disconnect-flush 👍 is correct-by-design (the resume re-answers; #2530's main turn_end gate already handles a real undelivered turn-end where the controller is live and re-paintable). No code change needed on this path. Audit outcome: the crash/restart terminal path is correctly covered by the resume mechanism, not by terminal-reaction honesty.

PR #2587, opened 2026-06-26
fix(telegram): surface post-answer background sub-agent activity (restore #2547 visibility under determinism gate)

mekenthompson:
## Outcome

Background sub-agent/workflow activity that runs AFTER a substantive answer was delivered now surfaces as an activity card below the prior reply. Operators see their agents' post-answer cleanup/sub-agent work again (the regression reported as 'background agent stuff seems gone' after v0.16.x).

Idle thinking-gaps and pure silence after the answer remain suppressed — the reply-is-last and idle-gap-suppression invariants from #2557/#2564 are fully preserved.

## Root Cause

Three compounding issues from the June 25 determinism wave:

1. **Dormant post-answer branch** (`feedHeartbeatTick`): the branch guarded by `turn.finalAnswerDelivered` reached the threshold check then hit a bare `return` with no implementation. No card was ever opened or maintained post-answer.

2. **`POST_ANSWER_LIVENESS_MS` defaulted to 0**: `parsePostAnswerLivenessMs(undefined)` returned `0`, so `POST_ANSWER_LIVENESS_MS <= 0` fired immediately and the branch returned before any detection logic ran.

3. **Lever 1 applied a blanket block** (`feed-open-gate.ts`): `mayOpenActivityCard` blocked ALL producers after `finalAnswerEverDelivered`, with no exception for genuine real tool activity occurring post-answer.

## Deterministic-Signal Approach

This fix drives off state deltas, not wall-clock age alone:

- **New `finalAnswerDeliveredAt: number | null`** field on `CurrentTurn`, stamped alongside `finalAnswerEverDelivered = true` at every site. Provides a deterministic anchor for "was this tool label before or after the answer?"

- **`feedHeartbeatTick` post-answer branch** now: gates on `lastToolLabelAt > finalAnswerDeliveredAt` (a tool label rendered genuinely after the answer). Idle liveness (no new label since answer) never sets this — idle-gap suppression is preserved by construction.

- **`POST_ANSWER_LIVENESS_MS` default changed to 6 000 ms** (matching `FEED_HEARTBEAT_MIN_STALE_MS` — after one heartbeat interval of post-answer real-tool activity, a card surfaces). Operators who prefer silent post-answer behaviour set `SWITCHROOM_POST_ANSWER_LIVENESS_MS=0`.

[...]

PR #2588, opened 2026-06-26
fix(telegram): restore conversational-turn narrative card with reply-supersede ordering (revisit Lever 5 / #2545)

mekenthompson:
## Outcome

Conversational turns where the agent emits narrative text blocks (type `text`, zero tool calls) now surface a visible activity card in Telegram — restoring the behavior from #2545. Operators experiencing "I only see tool updates" on conversational turns will see the agent's narration in a card again.

## Root cause

PR #2557's **Lever 5** in `mayOpenActivityCard` (`telegram-plugin/gateway/feed-open-gate.ts`) suppressed the narrative producer from opening a card on 0-tool turns:

```ts
// OLD (suppressed all narrative opens on 0-tool turns)
if (input.producer === 'narrative' && input.labeledToolCount === 0) return false
```

This was added to prevent a "triplication" reorder: a narrative card would open (message_id N), then the reply would send (message_id N+1), then the card would sit above the reply in Telegram — correct order, but then a Stop-hook re-prompt caused a second card (G2+G3 in the determinism doc). The fix over-suppressed: it killed the card entirely rather than fixing the G2/G3 half of the triplication.

## The fix: Lever 2 already handles the reorder

The reply-is-last ordering is guaranteed by **Lever 2** (`clearActivitySummary` in `executeReply`, `gateway.ts` line 8332–8357), which edits the narrative card **in-place BEFORE the reply chunks send**. The card keeps its lower message_id; the reply (higher message_id) is structurally last. No new mechanism needed.

Fix: remove the Lever 5 open-suppression in the **pre-answer case**. The gate now:

- **Pre-answer** (`finalAnswerEverDelivered === false`): narrative producer may OPEN on a 0-tool turn. Lever 2 supersedes/finalizes the card before the reply.
- **Post-answer** (`finalAnswerEverDelivered === true`): Lever 1 already blocks all producers — Lever 5 was unreachable here anyway.

The `labeledToolCount` field remains in `FeedOpenInput` (used by the R4 path: tool turns still accumulate narration and open once a tool label arrives). Lever 5 is now documented as "inert" rather than removed, since the post-answer case is covered by Lever 1.

## Why it's deterministic

The decision is keyed only on `finalAnswerEverDelivered` (sticky latch, set once, never cleared) and `producer` — both deterministic inputs. No timers as truth. Same inputs → same gate verdict on every run.

## Tests

- Replaced the old "0-tool narrative must NOT open" assertion (pinned the old Lever 5 block) with the revised contract: **"0-tool narrative DOES open pre-answer"** (restores #2545).
- Added a **supersede/no-reorder test**: pins that `finalAnswerEverDelivered === true` blocks all producers (narrative, tool, liveness) — meaning no card can open after the reply sends, so the reply-is-last invariant holds. This is the gate-side complement of Lever 2's in-place edit.
- All existing lever 1, lever 4, #2547 post-answer real activity, and #2141 ack-then-work tests remain GREEN.
- Fix B's new post-answer-background-liveness tests (from `fix/telegram-post-answer-bg-liveness`) remain GREEN (confirmed: same 1036 pass / 2 pre-existing unrelated vault failures as base branch).

**Test counts:** 21 pass (was 19; +2 new tests), 0 fail, 0 changes to other test files.

## Stacking note

This PR is stacked on **#2587** (`fix/telegram-post-answer-bg-liveness`) and should merge AFTER it. Both touch `feed-open-gate.ts`; this PR's diff is clean on top of that branch.

## References

- Restores: #2545 (surface JSONL text-block narrative)
- Regressed by: #2557 (Lever 5 addition)
- Stacks on: #2587 (post-answer bg liveness)
- Related: #2547 (post-answer real activity exception)

---

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/telegram-plugin/gateway/feed-open-gate.ts
+++ b/telegram-plugin/gateway/feed-open-gate.ts
@@ -24,18 +24,32 @@
  * (non-substantive) does NOT trip it and the #2141 ack-then-work feed still
  * opens.
  *
- * ## Lever 5 base case — narrative-SHOW alone must not OPEN (§9 lever 5, G1)
+ * ## Lever 5 — narrative-SHOW OPEN on 0-tool turns (#2545 / #2557 / #2587)
  *
- * The triplication: a 0-tool conversational turn opened a full card from
- * narration text alone. Rule: the narrative-SHOW producer (A) may only EDIT an
- * already-open card, never OPEN one. Only a tool label (producer B) or the
- * liveness timer (producer C, a genuine ≥12s thinking-gap) may OPEN a card.
+ * ### Pre-answer: OPEN is allowed (restores #2545)
  *
- * So a narrative-SHOW-triggered OPEN is refused while the turn has surfaced ZERO
- * tool labels (`labeledToolCount === 0`). Once a tool label arrives
- * (`labeledToolCount > 0`) the card opens and the accumulated narration renders;
- * a turn that starts conversational then dispatches a tool DOES open (R4). The
- * liveness path (producer C) is exempt — it owns the genuine thinking-gap open.
+ * When `finalAnswerEverDelivered === false`, a narrative producer on a 0-tool
+ * conversational turn MAY open a card. Reply-is-last ordering is preserved by
+ * Lever 2 (`clearActivitySummary` in `executeReply`) — that path edits the card
+ * in-place BEFORE the reply chunks send, so the card keeps its lower message_id
+ * and the reply (higher message_id) is structurally last on screen regardless of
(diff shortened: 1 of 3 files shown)


mekenthompson:
Completed and superseded by #2589. This PR had Lever 5 removal but left the 120-char clip in place. #2589 raises it to 200 chars (STATUS_LINE_MAX) and also delivers Fix 2 (post-answer background-agent liveness).

mekenthompson:
Superseded by #2589.

PR #2596, opened 2026-06-27
fix(auth): demote out_of_credits from serve-block to informational overage flag

mekenthompson:
## Summary
Demotes `out_of_credits` overage flag from a serve-block to an informational annotation. The fleet runs on Pro Max subscription quota; overage credits are a separate billing mechanism switchroom doesn't use. An account at 0% util with `overageDisabledReason === "out_of_credits"` was being blocked from serving and excluded from failover — defeating the failover pool.

## Why
Jun 20 incident (PR #2455): pixsoul@gmail.com at 5h=0%/7d=2% showed 🔴 BLOCKED and never received requests. Root cause: `out_of_credits` was in `SERVE_BLOCKING_OVERAGE_REASONS`, treated like a util wall (≥99.5%). The structurally identical `org_level_disabled` reason was correctly benign — the asymmetry was the defect.

## Test plan
- [x] `vitest` auth/quota suites green (10467 pass overall)
- [x] `tsc --noEmit` clean
- [x] overage-allowlist-drift guard updated in both packages
- Note: failover safety preserved via the existing mark-exhausted-on-429 path.

## Risk
Low — pure semantic demotion; util-wall still governs serve-blocking. Supersedes #2513, reverses #2455 with a correct gate.

### DIFF
--- a/src/auth/broker/account-eligibility.test.ts
+++ b/src/auth/broker/account-eligibility.test.ts
@@ -1,12 +1,14 @@
 import { describe, it, expect } from "vitest";
 import {
   isAccountBlocked,
+  hasNoOverageHeadroom,
   isOverageServeBlocking,
   snapshotShouldClearMark,
   clampMarkExpiry,
   snapshotFresh,
   snapshotWalled,
   snapshotClearlyHealthy,
+  OVERAGE_EXHAUSTED_REASONS,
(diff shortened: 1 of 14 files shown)


mekenthompson:
Both commits (fix + test cleanup) are already fully merged into main via subsequent PRs. Closing as redundant.

PR #2700, opened 2026-06-30
feat(voice): local GPU STT sidecar — gateway consume + token injection (PR-B2)

mekenthompson:
## Summary

PR-B2 of the staged voice feature: the **consuming side** of the local GPU STT sidecar plus the apply-time plumbing the PR-B1 compose stanza assumed.

- **Gateway** transcribes inbound Telegram voice notes via the in-fleet `voice-sidecar` (faster-whisper) when the host's PR-B1 verdict is `engine: local`. New `transcribeViaSidecar` (POSTs audio → sidecar `/stt` with the `X-Voice-Token` shared secret, mirroring `transcribeViaWhisper`'s return-contract) + `maybeTranscribeVoiceLocal`. Engine selection reads `loadHostCapabilities()`; the local path needs no `provider === 'openai'` gate (it has no API key). Telegram download factored into a shared `downloadVoiceBytes`. Any failure falls back to the legacy `(voice message)` envelope — voice-in stays non-critical.
- **Token injection** (the open question in the spec): the sidecar shared secret lives in the vault at `voice/sidecar-token`, resolved by **both** ends from there. The gateway resolves it at use-time via the broker (`materializeSidecarToken`, mirroring `materializeVoiceKey`). At `switchroom apply` on a `local` verdict, the token is **seeded** in the vault if absent (random 256-bit, operator-attested put — same recovery path as `provisionLiteLLMKeys`) and **written to the compose-adjacent `.env`** (0600) so docker's native `${VOICE_SIDECAR_TOKEN}` interpolation populates the sidecar's env. The secret never lands in the generated YAML. A `cloud` verdict removes any stale token `.env`.
- **Singleton reconcile + doctor** manage `voice-sidecar` as a verdict-gated conditional singleton (`resolveSingletonServices`): tracked for image-drift on a `local` host, never flagged missing/unhealthy on a `cloud` host.

### Token-injection decision (#2) + rationale
The LiteLLM precedent keeps the secret out of compose by having `start.sh` fetch it from the vault at the **agent** container's boot. That doesn't transfer: the `voice-sidecar` container is a pure-Python faster-whisper server with **no switchroom CLI and no broker socket** — it reads `VOICE_SIDECAR_TOKEN` straight from its env (`server.py:65`). So the cleanest vault-sourced injection that keeps the secret out of the YAML is the **standard docker-compose `.env` interpolation file**, written 0600 at apply time and regenerated each apply. The token's source of truth is still the vault; both ends agree on `voice/sidecar-token`. Documented in `src/cli/voice-sidecar-token.ts`.

## Why
Serves `reference/jobs/talk-to-agents-from-anywhere.md`. Advances vision **#3 (subscription-honest** — local STT means zero third-party STT key) and **#4 (always-available** — no cloud reachability dependency for voice on a GPU host).

## Test plan
- [x] `tsc --noEmit` clean
- [x] `vitest run tests/docker/compose-generator.test.ts src/agents/singleton-reconcile.test.ts src/cli/voice-sidecar-token.test.ts` → 211 passed (both compose branches: local emits service+GPU+`0.0.0.0:18900:8126`+healthcheck+`voice-model-cache`; cloud emits neither service nor volume; singleton verdict gating)
- [x] `bun test telegram-plugin/tests/voice-transcribe-sidecar.test.ts voice-transcribe.test.ts` → 32 passed (sidecar transcription mocked-fetch + token resolution mocked-broker)
- [x] `scripts/check-bot-api-wrapping.sh` clean (no allowlist drift)
- [x] `node scripts/check-plugin-references.mjs` clean

## Risk
Low. Strictly additive on the `local` verdict path; the `cloud` (default) path is unchanged — compose, gateway routing, and singleton management all behave exactly as before when no GPU verdict exists. End-to-end against a real GPU sidecar isn't exercised here (no GPU in CI); that's UAT/canary territory. Do NOT auto-merge — orchestrator merges after CI + review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,36 @@
 # Changelog
 
+## unreleased — Voice PR-B2: local GPU STT sidecar (gateway consuming side + token injection)
+
+Wires the gateway's voice-in path to the local `voice-sidecar` GPU
+speech-to-text service when the host's PR-B1 voice verdict is `engine: local`.
+On such a host an inbound Telegram voice note is transcribed entirely
+on-box (faster-whisper) — no third-party STT key (vision #3,
+subscription-honest) and no cloud reachability dependency (vision #4,
+always-available). On a `cloud` verdict the existing OpenAI Whisper path is
+unchanged. Either way, any transcription failure falls back to the legacy
+`(voice message)` envelope — voice-in stays a non-critical UX enhancement.
+Serves `reference/jobs/talk-to-agents-from-anywhere.md`.
+
+- Gateway: new `transcribeViaSidecar` (POST audio → sidecar `/stt` with the
+  `X-Voice-Token` shared secret, mirroring `transcribeViaWhisper`'s
+  return-contract) and `maybeTranscribeVoiceLocal`. Engine selection reads
+  the persisted `loadHostCapabilities()` verdict; the local path needs no
+  `provider === 'openai'` gate (it has no API key). The Telegram download
+  is now a shared `downloadVoiceBytes` helper used by both paths.
+- Token: the sidecar's shared secret is resolved by BOTH ends from the vault
+  at `voice/sidecar-token`. The gateway resolves it at use-time via the
+  broker (`materializeSidecarToken`, mirroring `materializeVoiceKey`). At
+  `switchroom apply` (local verdict only), the token is seeded in the vault
+  if absent and written to the compose-adjacent `.env` (0600) so docker's
+  `${VOICE_SIDECAR_TOKEN}` interpolation populates the pure-Python sidecar's
+  env — the secret never lands in the generated YAML. A `cloud` verdict
+  removes any stale token `.env`.
+- Singleton reconcile + doctor now manage `voice-sidecar` as a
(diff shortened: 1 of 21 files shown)


mekenthompson:
Reviewer (fresh-process) returned **APPROVE** — all five focus checks pass, no blockers. Addressed the one worthwhile nit (corrected the `emitVoiceSidecarService` token-resolution comment in compose.ts).

CI follow-ups during the run:
- Fixed a `bun:test`-in-vitest lint/shard failure by excluding `voice-transcribe-sidecar.test.ts` from vitest (runs under `bun test`, like its cloud twin).
- Fixed the `e2e` `ci-image-matrix` failure: `Dockerfile.voice` (committed here) is neither in the build matrix nor opted out. **Decision:** opted it out of `DOCKER_MATRIX_OPT_OUT` with a reason rather than adding it to `build-dependents` — the voice image extends a **CUDA runtime base** (not switchroom-base, so it doesn't fit the BASE_IMAGE-threading dependents matrix), is **amd64-only**, and ships **only on a `local` GPU verdict**. A dedicated CUDA-aware build/publish job (build-hindsight-style) is the right home and is left as a **follow-up** — flagging here for the orchestrator.

PR #2758, opened 2026-07-03
fix(telegram): sub-agent handback survives Claude Code split-line writes

mekenthompson:
## Summary

**Job:** `know-what-my-agent-is-doing` (visibility outcome — "at any moment during a turn the user can see what the agent is up to").

switchroom has **no** stream-json path — the status/liveness surface is derived by scraping Claude Code's on-disk per-session JSONL transcript (`session-tail.ts`). Those shapes are undocumented and version-fragile.

A recent Claude Code release (confirmed on the fleet build **2.1.199**) changed assistant-message persistence: one logical message is now split across **multiple JSONL lines sharing one `message.id`** (one content-block per line), and the terminal `stop_reason` (`end_turn`) is stamped on **every** split line — including the leading `[thinking]` line that precedes the `[text: final answer]` line.

`projectSubagentLine` treats `stop_reason === 'end_turn'` as the sub-agent terminal (background `Agent` workers never write `system/turn_duration`). In the new shape it fired on the **thinking-only first line**, marking the sub-agent `done` and handing back stale/empty text **before** the real handback `[text]` line was projected (`onProgress` is `state==='running'`-gated, so the late text is dropped). Result: **background/sub-agent work stopped surfacing in the thread** — intermittently, because non-final split lines carry `end_turn` on some sessions/versions and `null` on others.

## Evidence (real 2.1.199 transcript)
```
{"type":"assistant","message":{"id":"msg_01PMsHNF...","stop_reason":"end_turn","content":[{"type":"thinking",...}]}}
{"type":"assistant","message":{"id":"msg_01PMsHNF...","stop_reason":"end_turn","content":[{"type":"text","text":"<final answer>"}]}}
```

## Why the main-agent path survives
`projectTranscriptLine` derives the main-agent terminal from the separate `system/turn_duration` line, not `stop_reason`. `lastInMessage`/`blockIndex` (projection artifacts the split makes technically wrong) are unused downstream, and the reducer's preamble→tool_use pairing is stream-stateful. The break is specific to the sub-agent terminal.

## Fix
Only treat `end_turn` as terminal on a line that carries the **answer surface** (a non-empty `text` block, or a real non-`Agent`/`Task` `tool_use`). The terminal + handback ride the following content line, which also carries `end_turn` and fires correctly **after** the text event. Degrades gracefully:
- legacy single-line `[thinking, text](end_turn)` shape has a text block → fires exactly as before;
- a genuine thinking-only end (no answer) still terminates via `turn_duration` / capped-reaper / watcher stall nets.

## Why existing tests missed it
`session-tail.test.ts` builds assistant fixtures as a **single line** with all blocks in `content[]` (the old shape); nothing fed the split-`message.id` shape. And there was **no contract test** pinning the upstream event stream — a format shift lands silently while CI stays green. The fuzzed `turn-liveness-invariant.test.ts` fuzzes turn *shape*, not transcript *wire shape*, so it can't see a projector that mis-derives events from the new format.

## Test plan
- [x] `session-tail.test.ts`: no premature terminal on a thinking-only `end_turn` split line; handback text emitted before exactly one `turn_end` across a split message; legacy single-line shape still fires.
- [x] **`claude-code-event-contract.test.ts` (new canary):** pins every Claude Code transcript wire shape the projector consumes (queue-operation, thinking / tool_use / tool_result / text blocks, `system/turn_duration`, split `message.id`, sub-agent kickoff). An upstream format change now **fails CI loudly** instead of silently blanking the status card.
- [x] All three new assertions **fail on pre-fix code, pass after** (revert-proof).
- [x] Broader suite green: subagent-watcher, handback-gaps, turn-liveness-floor/invariant, silence-poke, session-tail-capped, subagent-handback-decision (156 tests).
- [x] `tsc --noEmit` clean; `check-plugin-references.mjs` clean.

## Risk
Low and localized to `projectSubagentLine`'s terminal gate. Behaviour on the legacy shape is unchanged (pinned by a test). Live DM+channel background-worker UAT (`jtbd-worker-activity-feed-dm`/`-channel`) owns the end-to-end proof; those exercise the real surface but can't inject a specific wire shape, so the split-shape guarantee lives in the projector-level + contract tests here.

### DIFF
--- a/telegram-plugin/hooks/tool-label-pretool.d.mts
+++ b/telegram-plugin/hooks/tool-label-pretool.d.mts
@@ -1,0 +1,12 @@
+/**
+ * Type declarations for the PreToolUse hook module so TypeScript surfaces
+ * (tool-activity-summary.ts's `describeToolUse`) can delegate to
+ * `computeLabel` — the ONE status-vocabulary composer — without tsc
+ * resolution errors. The runtime module is plain ESM (hooks are executed
+ * directly by Claude Code, so they cannot be TS); main() is guarded behind
+ * an is-main check, making the import side-effect-free.
+ */
+export function computeLabel(
+  toolName: string,
+  input: Record<string, unknown> | undefined | null,
+): string | null
--- a/telegram-plugin/hooks/tool-label-pretool.mjs
+++ b/telegram-plugin/hooks/tool-label-pretool.mjs
@@ -76,6 +76,16 @@ function safeBasename(p) {
   }
 }
 
+/** Hostname only, `www.` stripped — the WebFetch label form ("Reading example.com"). */
+function urlHost(u) {
+  if (!u || typeof u !== 'string') return ''
+  try {
+    return new URL(u).hostname.replace(/^www\./, '')
+  } catch {
+    return u
+  }
+}
+
 function urlHostPath(u) {
   if (!u || typeof u !== 'string') return ''
   try {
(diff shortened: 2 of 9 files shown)


mekenthompson:
**Second commit pushed (65afebeb): single status-vocabulary composer.**

Follow-up scope from the same job spec:

1. **Agent-type coverage audit** — the projector/watcher path is agent-type-agnostic by construction: `projectSubagentLine(line, agentId, state)` takes no type input; the only tool-name gates are `Agent`/`Task` (the dispatch tool, identical for every `subagent_type`). Dispatch metadata (`agent_type`, `background`) is recorded generically by `subagent-tracker-pretool.mjs` for any Agent/Task call. Fleet evidence (335 real dispatch metas): workflow-subagent 107, worker 92, reviewer 51, researcher 49, Plan 14, Explore 13, general-purpose 6, fork 3 — all through the same hook + `subagents/agent-*.jsonl` layout. Transcript sweep across all types: every `end_turn` terminal carries a text-bearing `end_turn` line (the fixed terminal fires for all); zero types rely on the thinking-only line as the sole terminal carrier. `workflow-subagent` transcripts never write `end_turn` at all (they end mid-`tool_use`; the watcher's `workflows/` scanning + stall nets own their terminal — pre-existing design, unaffected by this fix).

2. **Status-wording unification** — the per-tool activity copy was composed by two drifted tables (mjs sidecar `computeLabel` for the live feed vs TS `describeToolUse` for flush/nested/worker-card lines). Now one composer: `describeToolUse` delegates to `computeLabel`; `status-vocabulary-unification.test.ts` sweeps a full corpus asserting identical labels + suppressions on both entry points, so a re-fork fails CI (verified by mutation).

Full plugin vitest suite green (296 files / 4820 tests), tsc + plugin-refcheck clean, hook smoke-tested standalone.

PR #2979, opened 2026-07-10
fix(telegram): harden progress/status surface — dropped handbacks, frozen cards, stale replays

mekenthompson:
## What & why

Full adversarially-validated review of the work-visibility / progress-card subsystem (the thing that makes any agent or worker's activity visible across sub-levels). Two Opus reviewers found the issues, two more adversarially cross-examined each finding against real code, Sonnet workers implemented, and an independent Opus pass reviewed each lane's diff.

**Headline:** delegated background-worker results could silently vanish (no recap, no handback, no reply) when the sub-agent's identity link didn't stitch. That's the fix that matters; the rest is hardening.

## Findings scorecard (after adversarial validation)

| # | Sev | Status | Fix |
|---|-----|--------|-----|
| 1(+2) | HIGH | CONFIRMED | Dropped background handback → degrade to in-memory `background`; deliver on resultText |
| 3 | MED | CONFIRMED (narrow, legacy) | Fuzzy backfill refuses ambiguous links |
| 4 | MED | CONFIRMED | Card frozen "running" on 429 cooldown → deferred terminal edit re-driven by heartbeat |
| 5 | MED | CONFIRMED | Killed-before-restart worker resurrected as "completed" → require post-boot growth |
| 7 | LOW | CONFIRMED | Typing flicker under parallel tools → ref-count lanes |
| 8 | LOW | PARTIAL | Retry budget bleeds across dark turns → age-bound the record |
| 9a/9b | LOW | CONFIRMED | Unbounded id set (LRU cap) + stale comment |
| 6 | — | FALSE POSITIVE | (withdrawn — transcript-file conflation) |

## Tests

All fixes carry regression tests **verified to fail against HEAD** (not green-by-construction). 224 vitest + 15 bun green, `tsc --noEmit` clean.

## Reviewer follow-up nits (non-blocking, not in this PR)

- #1: add a guard to suppress the rare foreground double-handback when a live parent turn still holds the agent (deliver-over-drop was the deliberate call).
- #1: add an integration test pinning the real `gateway.ts` background-flip line (currently only the resolver level is pinned).
- #5: give-up is permanent — a genuinely-live worker silent >15min immediately post-boot is classified historical (spec-accepted tradeoff).

## Deploy note

Gateway-touching. Per build/restart discipline: merge first, then staggered fleet restart (not all at once) on confirmation.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -27276,7 +27276,7 @@ void (async () => {
               // Gated to background completions: foreground sub-agents
               // need nothing here, and 'orphan' is a stale historical-at-
               // boot row, not a fresh completion the user is waiting on.
-              onFinish: ({ agentId, outcome, description, resultText, toolCount, durationMs }) => {
+              onFinish: ({ agentId, outcome, description, resultText, toolCount, durationMs, background: entryBackground }) => {
                 // Reaction promotion: if the parent turn already ended
                 // with this (or another) worker still running, its 👍 was
                 // deferred (held on ✍️/⚡). Now that a worker finished,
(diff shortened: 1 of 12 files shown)


mekenthompson:
Rebased onto main (12 commits, incl. #2960). **Fix #4 (frozen card on 429 cooldown) dropped** — #2960 already landed the same fix (pendingFinish staging + heartbeat re-drive) plus a resurrection guard ours lacked, so our #4 was a strict duplicate; took main's worker-activity-feed.ts wholesale. Remaining fixes (#1 dropped-handback, #3, #5, #7, #8, #9a/b) are all distinct from #2960 and verified to coexist — our onFinish background-flip composes with #2960's resolveWorkerFeedChat (a recovered background handback now also gets their robust chat routing). 226 vitest + 15 bun green, tsc clean post-rebase.

PR #3058, opened 2026-07-11
feat(hindsight): redirect direct mental-model writes to mental_model_propose

mekenthompson:
## Summary

Adds a PreToolUse hook `src/cli/hindsight-mental-model-pretool.ts` that DENIES the four gated hindsight mental-model WRITE tools (`create_mental_model`, `update_mental_model`, `delete_mental_model`, `refresh_mental_model`) with a deterministic, instructive message:

> Mental-model writes are operator-approved. Use `mcp__switchroom-telegram__mental_model_propose` instead — it posts an approval card and persists on approval.

Every other tool — all hindsight READ tools (`get_mental_model`, `list_mental_models`, `recall`, `retain`, `reflect`, …) and every non-hindsight tool — passes through untouched. The hook is registered in `scaffold.ts` (the actual settings.json PreToolUse wiring path — `settings.json.hbs` carries no PreToolUse hooks), matcher-scoped to `^mcp__hindsight__`, bundled via `build.mjs` and shipped to `/opt/switchroom/hooks/` by `Dockerfile.agent`, mirroring `skill-validate-pretool`.

Also rewrites the profile CLAUDE.md templates (`default`, `coding`, `health-coach`, `executive-assistant`) that previously instructed agents to call `create_mental_model` directly — they now recommend `mcp__switchroom-telegram__mental_model_propose`, keeping the explanation of what mental models are. `grep -r "create_mental_model" profiles/` now shows only don't-call-directly guidance.

## Why

The four write tools are deliberately un-preapproved (#2911 least-privilege). A direct call falls through to a TUI permission prompt that wedges the turn (and, until the watchdog card-race fix lands, gets auto-denied). Our own profile docs manufactured these prompts by telling agents to call the tools directly. This makes the redirect mechanical (hook deny is instant, agent self-corrects in-turn) and fixes the docs. Job spec: `reference/jobs/approve-what-my-agent-can-touch.md`.

## Test plan

- New unit test `tests/hindsight-mental-model-pretool.test.ts` (driven through `bun` against source, mirroring `skill-validate-pretool.test.ts`): denies each of the four write tools with the exact redirect message; allows hindsight read tools and arbitrary non-hindsight tools (including a lookalike `mcp__other__create_mental_model`); fails open on empty/non-JSON/unknown-shape stdin.
- `tests/scaffold.test.ts`: new assertion that the hook is registered with matcher `^mcp__hindsight__`, bundled path, node invocation, timeout 10.
- `tests/reconcile-hooks-drift.test.ts`: drift assertion that the hook is present with the hindsight matcher.
- Scoped run: `vitest run tests/hindsight-mental-model-pretool.test.ts tests/reconcile-hooks-drift.test.ts tests/scaffold.test.ts` → 223 passed.
- `npm run lint` → clean (tsc + all 8 guard scripts).
- Bundled `.mjs` smoke-tested: deny tool emits the block JSON, read tool emits nothing.
- Live test-harness verification (deploy + direct-call scenario) is deferred to the operator — no deploys from this batch.

## Risk

Low. The hook is a pure `tool_name` gate with no src/ imports, fail-open on any stdin error (a redirect hook must never fail a legitimate call). Only the four exact write-tool names are blocked; the `^mcp__hindsight__` matcher keeps the stdin-parse cost off non-hindsight tools. Docs-only changes otherwise.

Closes #2974

### DIFF
--- a/docker/Dockerfile.agent
+++ b/docker/Dockerfile.agent
@@ -234,6 +234,16 @@ RUN chmod 0755 /opt/switchroom/hooks/notion-write-pretool.mjs
 COPY dist/cli/skill-validate-pretool.mjs /opt/switchroom/hooks/skill-validate-pretool.mjs
 RUN chmod 0755 /opt/switchroom/hooks/skill-validate-pretool.mjs
 
+# Hindsight mental-model write redirect (#2974). A PreToolUse hook scoped
+# to `^mcp__hindsight__` that DENIES the four mental-model WRITE tools with
+# an instructive message pointing at mcp__switchroom-telegram__mental_model_propose,
+# so a direct call self-corrects in the same turn instead of wedging on a
+# permission prompt. Read tools pass through. Not in the security-plugin:
+# it's a routing nudge toward the sanctioned approval path, not a
+# load-bearing safety control (the least-privilege gate in scaffold.ts is).
+COPY dist/cli/hindsight-mental-model-pretool.mjs /opt/switchroom/hooks/hindsight-mental-model-pretool.mjs
+RUN chmod 0755 /opt/switchroom/hooks/hindsight-mental-model-pretool.mjs
+
 # Agent self-improvement Stop gate (RFC reference/rfcs/agent-self-improvement.md
 # slice 1). A cheap deterministic per-turn triage; on a learning signal
 # it injects a forked review turn via the gateway socket (inject_inbound,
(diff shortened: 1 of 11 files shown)


mekenthompson:
Coordinator adversarial review (batch K2 of #3051) — APPROVE. Verified: hook block shape matches skill-validate-pretool ({decision:block}); fail-open on stdin errors consistent with siblings; four write tools denied, reads pass; tests exercise the real hook process and assert stdout outcomes; scaffold registration pinned by tests (matcher ^mcp__hindsight__, run-hook wrap, timeout 10); build.mjs bundling + Dockerfile.agent COPY mirrored from skill-validate-pretool; profile docs no longer instruct direct create_mental_model. Registration in scaffold.ts instead of settings.json.hbs is the correct deviation (that template carries no PreToolUse hooks). Live test-harness verification deferred to operator per batch constraints.

PR #3060, opened 2026-07-11
fix(gateway): retry rate-limited operator-approved mental-model persist (#2975 Stage 1)

mekenthompson:
## Summary

Stage 1 (gateway-only backstop) of #2975. When an operator taps **Approve** on a mental-model proposal, the gateway persists it via `config_propose_edit` on the agent's hostd socket. hostd counts that persist against the agent's 3/hour `config_propose_edit` bucket and can deny it with `E_RATE_LIMITED` — silently losing an already-approved change (observed: klanker 2026-07-06).

This adds a gateway-only retry backstop:

- `resolveMentalModelProposal` now distinguishes a `rate_limited` dispatch result. The gateway adapter parses hostd's structured `fix.retry_after` (server.ts ~2270-2280) into an epoch-ms window-open time.
- On rate-limit it schedules **exactly ONE** bounded re-dispatch at that time (single-shot `setTimeout`, clamped to the 32-bit ceiling; the resolver never re-schedules).
- The retry re-runs the same closure over the captured `built.diff`, so register + dispatch use **byte-identical** diff bytes — preserving the forge-resistance contract (mental-model-propose-resolve.ts:54-63) that hostd's config-approval callback auto-resolves only on an exact byte-match.
- The proposal card flips to `⏳ Approved — applying at HH:MM (config-edit rate window)` so the operator sees it isn't lost.
- If the single retry ALSO fails, a loud operator-events notification (`mental-model-persist-failed`, mirroring `always-allow-persist-failed`) fires as a new message — no silent loss.

Stage 2 (hostd `checkPreApproved` bypass) is a deliberate separate follow-up and is NOT in this PR.

## Why

An approved operator action must never be silently dropped. The 3/hour throttle exists to cap approval-card spam; the approved-persist path posts zero cards and consumes zero operator attention, yet was subject to the same bucket. Until Stage 2 removes the collision at the hostd layer, this backstop guarantees the approved change either lands at the window or fails loudly.

Cites job specs `reference/jobs/get-better-the-longer-they-run.md` and `reference/jobs/approve-what-my-agent-can-touch.md` — the operator's approval is the source of truth and must be honored durably.

## Test plan

- `vitest run telegram-plugin/tests/mental-model-propose-resolve.test.ts` — 8 passed. New cases (mocked clock, manual scheduler):
  - bucket exhausted → approve → `E_RATE_LIMITED` → asserts one retry scheduled at `retry_after`, card flipped, retry fires, persist succeeds, correlation re-registered with **byte-identical** diff, no loud failure.
  - retry also fails → operator-events notification emitted exactly once, no second retry, agent woken with failed inbound.
  - no scheduler wired → rate-limited persist fails loudly instead of silently dropping.
- `vitest run telegram-plugin/tests/operator-events.test.ts telegram-plugin/tests/mental-model-propose-card.test.ts` — 75 passed.
- `npm run lint` — clean (tsc + all 8 guard scripts).

## Risk

Low, gateway-scoped. No hostd/protocol change. The retry timer is gateway-memory-only — a gateway restart between schedule and fire loses the retry (documented in-code; acceptable for Stage 1, eliminated by Stage 2). Bounded to one retry, so no unbounded loop.

Part of #2975 (Stage 1).

### DIFF
--- a/telegram-plugin/gateway/callback-query-handlers.ts
+++ b/telegram-plugin/gateway/callback-query-handlers.ts
@@ -83,6 +83,7 @@ import { getAuthBrokerClient } from './auth-broker-client.js'
 import { chatKey } from './chat-key.js'
 import { tryHostdDispatch, hostdRequestId } from './hostd-dispatch.js'
 import type { HostdRequest } from '../../src/host-control/protocol.js'
+import type { OperatorEvent } from '../operator-events.js'
 import type { InboundMessage } from './ipc-protocol.js'
 import type { SweepableCardStore } from './approval-card-stores.js'
 import type { SweepableStore } from './pending-state-stores.js'
@@ -389,6 +390,13 @@ export interface CallbackQueryHandlersDeps {
   vaultKeyRegex: RegExp
   /** MENTAL_MODEL_PROPOSE_TTL_MS (config-driven approval-card lifetime). */
   mentalModelProposeTtlMs: number
+  /**
+   * Emit an operator-visible event through the gateway's single funnel
+   * (`emitGatewayOperatorEvent`: cooldown + record + broadcast). Used by the
+   * #2975 Stage-1 backstop to loudly surface an approved mental-model persist
+   * whose rate-window retry also failed.
+   */
+  emitOperatorEvent: (event: OperatorEvent) => void
 }
 
 // Freshness throttle for the /auth dashboard ↻ refresh button — one live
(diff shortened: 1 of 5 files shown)


mekenthompson:
Coordinator adversarial review (batch K2 of #3051) — APPROVE (Stage 1). Verified: dispatchOnce closure guarantees register+dispatch over identical captured diff bytes (forge-resistance contract held on the retry, asserted by test: retryDiff === firstDiff, re-register call 2 with same bytes); exactly one retry (scheduleRetry called once, retry path never re-schedules; second rate_limit is terminal-loud); no-scheduler fallback fails loudly not silently; card flip to applying-at-HH:MM leaves the retry to resolve the turn (scheduled_retry outcome skips the final card overwrite); 32-bit setTimeout clamp; operator-event kind mirrors always-allow-persist-failed. Stage-1 restart caveat documented in-code as specced. spec.name rendering matches existing card lines. No findings. Stage 2 (hostd checkPreApproved bypass) follows as a separate PR.

PR #3065, opened 2026-07-11
feat(hooks): Stop-hook guard for narrate-without-execute dispatch claims (#396)

### LINKED ISSUE #396: telegram-plugin: deterministic narrate-without-execute prevention via Claude Code hooks
## Bug

When the model emits `mcp__switchroom-telegram__reply` with text announcing in-flight work (\"Dispatching a worker…\", \"On it…\", \"Kicking off…\") but does NOT invoke the dispatch tool (Agent / Bash / Task) in the same response, the turn ends with a misleading message in Telegram and zero queued work. Telegram user has no signal the work didn't start.

## Forensic evidence (chat `8248703757`, 2026-04-29)

| ts (UTC) | event |
|---|---|
| 21:19:59 | user msg 414: \"Yes\" (confirming work on PR #385 fix) |
| 21:20:44 | assistant msg 417 via `reply`: \"On it. Worktree switchroom-fix-admin-gating still exists from last night… **Dispatching a worker** to triage the red Buildkite, fix it, merge #385, file the Fix 2 issue, update gymbro, and validate /auth end-to-end.\" |
| 21:20:44 → 22:09:53 | ~49 min silence — no Agent call; `TaskList` returned \"No tasks found\"; transcript shows exactly ONE tool call in that turn (the `reply` itself) |
| 22:09:53 | user msg 418: \"status?\" |
| 22:13:?? | forensic confession msg 425: model treated narration as execution; `reply` is terminal — turn ended after the reply, no second tool call possible |

## Root cause

1. `reply` is end-of-turn semantics. Once it returns, the tool loop exits unless other calls were queued in the SAME response.
2. Model emitted only `reply` (no parallel Agent call). Template completion bias drove the response toward \"acknowledge + plan + close.\"
3. No deterministic gate prevents `reply` from announcing work that wasn't actually queued.

## Why Telegram-specific

At plain CLI, plain assistant text reaches the user — narrate-without-execute is benign because the next turn's dispatch is visible. In Telegram-tmux only `reply`/`stream_reply` output reaches the user; plain assistant text in tmux is invisible. So a misleading `reply` followed by turn-end leaves the Telegram user thinking the agent is working when it's idle.

### Hook A — PreToolUse on switchroom-telegram tools

Ship `telegram-plugin/hooks/narrate-audit.ts` (or `.sh`).

Logic:
1. Read PreToolUse JSON on stdin (`tool_name`, `tool_input`, `transcript_path`).
2. Match `tool_name in {mcp__switchroom-telegram__reply, mcp__switchroom-telegram__stream_reply}`.
3. Regex-match `tool_input.text` against dispatch verbs: `\\b(dispatch(ing)?|kick(ing)? ?off|on it|spinning up|starting (a |the )?worker|launching|firing off)\\b` (case-insensitive). Conservative seed list, expand with telemetry.
4. If matched: read `transcript_path`, scan turn-since-last-user-message for any `Agent`, `Bash`, or `Task*` tool call.
5. If no dispatch tool call found: exit with `{\"permissionDecision\": \"deny\", \"reason\": \"You announced dispatching work but did not invoke Agent/Bash/Task in this turn. Either invoke the dispatch tool first (same response), or rephrase the message to not promise immediate work.\"}`.
6. Otherwise: allow.

### Hook B — Stop backstop

Same matcher at Stop event. If a dispatch promise was sent without a corresponding tool call, return `{\"decision\": \"block\", \"reason\": \"...\"}` to force continuation. Catches paraphrase escapes Hook A misses. Note: Stop is post-send, the misleading message is already in Telegram — Stop only patches via a follow-up.

### Wiring (switchroom provisioning)

Switchroom already writes `~/.claude/settings.json` per agent. Add:

```json
{
  \"hooks\": {
    \"PreToolUse\": [
      {
        \"matcher\": \"mcp__switchroom-telegram__(reply|stream_reply)\",
        \"hooks\": [{\"type\": \"command\", \"command\": \"node \$SWITCHROOM_TELEGRAM_PLUGIN/hooks/narrate-audit.js pre\"}]
      }
    ],
    \"Stop\": [
      {
        \"hooks\": [{\"type\": \"command\", \"command\": \"node \$SWITCHROOM_TELEGRAM_PLUGIN/hooks/narrate-audit.js stop\"}]
      }
    ]
  }
}
```

Reconcile flow keeps hooks installed across upgrades. `switchroom doctor` warns on missing hooks. Fail-closed on script error (deny by default) to avoid silent regression.

## Acceptance criteria

- [ ] `narrate-audit.{ts,js}` committed under `telegram-plugin/hooks/`
- [ ] Unit test: dispatch-verb reply + no prior Agent → deny + reason
- [ ] Unit test: dispatch-verb reply + prior Agent in transcript → allow
- [ ] Unit test: non-dispatch reply → allow
- [ ] Unit test: malformed transcript → fail-closed (deny)
- [ ] Switchroom provision writes hook entries
- [ ] `switchroom doctor` reports missing hooks
- [ ] Live verification: agent emits `reply` with \"Dispatching…\" without Agent call → blocked, retries

## Residual non-determinism (acknowledged)

- Regex paraphrasing (e.g., \"I'll handle that\"). Conservative deny-list, expand on telemetry.
- Stop is post-send. PreToolUse prevents; Stop is backstop only.
- User-level `settings.local.json` can disable. Doctor detects.

## Related

- Hindsight directive already added (priority 10, \"Never narrate dispatching without dispatching\") — soft layer.
- Pairs with companion issue: tool description hardening (separate issue).
- Companion issue: Hindsight corpus curation to prevent anti-pattern propagation (separate issue).

mekenthompson:
## Summary

Adds a deterministic Stop hook that catches the "narrate a dispatch without executing it" defect (#396): a reply that tells the user work was dispatched ("Dispatching a worker now to fix this") while the turn contains **no** `Agent`/`Task` tool call and **no** backgrounded `Bash`. When that happens the hook re-prompts the model exactly once to actually dispatch, or send a correction reply.

- New pure module `telegram-plugin/hooks/dispatch-claim-scan.mjs` (mirrors the `silent-end-scan.mjs` extraction pattern) + Stop hook `telegram-plugin/hooks/dispatch-claim-stop.mjs`.
- Registered as one new Stop entry after `silent-end-interrupt` in `telegram-plugin/hooks/hooks.json` **and** the hand-wired `src/agents/scaffold.ts` stop-hooks list (both registration sites, kept in lockstep). Scaffold/manifest pin tests updated.

Implements the approved design in the #396 "Design/investigation report (2026-07-11, main @ 8fbd9574)" comment, which supersedes the 2026-04 Hook A/Hook B body proposal.

## Why

The #1664 silent-end Stop gate blocks turns with **no** qualifying final reply. A narrate-without-execute turn HAS a qualifying reply, so it sails through — the user sees a confident hand-off that never happened. This closes that gap deterministically instead of relying on model discipline. Cites `reference/jobs/know-what-my-agent-is-doing.md` (hold-the-leash: if the user can't trust what the card says the agent is doing, they stop using it).

**Stop-hook only — no PreToolUse deny** (per the design): the correct pattern is reply-then-dispatch in the same response; a pre-send deny fires before the Agent call exists and would block legitimate flows. A wrong Stop block costs one invisible re-prompt; a wrong deny kills a real answer.

**Registry-check decision:** the design's belt-and-braces step 4 offered either querying the subagent-tracker `registry.db` for a dispatch row, or skipping it and relying on transcript-only detection with past-tense exclusion. I took the **transcript-only** option. The `Agent`/`Task` tool_use blocks for THIS turn are already in the same JSONL slice the hook walks, so a second sqlite round-trip from a dependency-free `.mjs` hook is redundant and awkward. The registry's real value was clearing PAST-tense references to earlier dispatches; I get the same by dropping pure past-tense verb forms (`dispatched`, `launched`, `kicked off`, `delegated`, `spawned`) from the commitment regex, so "the worker I dispatched earlier" never trips the gate. Documented in the module header.

## Test plan

- `telegram-plugin/tests/dispatch-claim-scan.test.ts` (23 cases, vitest — pure module, matches the sibling `silent-end-interrupt-stop-scan.test.ts` style): (a) claim+no dispatch → block; (b/b2) Agent/Task in turn → allow; (c) backgrounded Bash → allow, (c2) foreground Bash → block; (d/d2) conditional offer → allow; (e) ordinary reply → allow; (f) NO_REPLY-only → allow; (g) sidechain Agent doesn't count as dispatch, (g2) sidechain reply doesn't count as claim; (i/i2/i3) empty/anchor-less/malformed transcript → fail open; past-tense earlier-turn reference → allow; (h) retry budget: first Stop blocks, second same-turn Stop fails open, new turn resets.
- `tests/scaffold.test.ts` + `tests/telegram-plugin-manifest.test.ts` updated and green (210 tests).
- `npm run lint` clean (tsc + all 8 guards).

## Risk

Low. Fails open on every error path (missing/unreadable/anchor-less/malformed transcript, state-file failure) consistent with every sibling reply-path hook; the 1-retry budget bounds the block direction to exactly one invisible re-prompt per turn. No PreToolUse deny, so no legitimate reply is ever blocked pre-send.

Closes #396

### DIFF
--- a/src/agents/scaffold.ts
+++ b/src/agents/scaffold.ts
@@ -4937,6 +4937,18 @@ export function buildSettingsHooksBlock(p: HooksBlockParams): Record<string, unk
       timeout: 5,
       async: false,
     });
+    // Narrate-without-execute guard (#396): re-prompt once when a reply
+    // claims a dispatch the turn never actually made (no Agent/Task call,
+    // no backgrounded Bash). Sync so it can block the stop; fail-open.
+    stopHooks.push({
+      type: "command",
+      command: wrap(
+        "hook:dispatch-claim-stop",
+        `node "${join(DOCKER_HOOKS_PATH, "dispatch-claim-stop.mjs")}"`,
+      ),
+      timeout: 5,
+      async: false,
+    });
     // Reaper for the PreToolUse tool-label sidecar files (#783).
     // Runs on every Stop; idempotent age + count rotation.
     stopHooks.push({
--- a/telegram-plugin/hooks/dispatch-claim-scan.mjs
+++ b/telegram-plugin/hooks/dispatch-claim-scan.mjs
@@ -1,0 +1,259 @@
+/**
+ * Pure helpers for the dispatch-claim Stop hook (#396) — extracted so
+ * unit tests can exercise the scan logic without spawning the .mjs
+ * subprocess (mirrors the `silent-end-scan.mjs` extraction pattern).
+ *
+ * The defect (#396): the model sends a reply that TELLS the user it has
+ * dispatched work ("Dispatching a worker now to fix this") but the turn
+ * contains no `Agent`/`Task` tool call and no backgrounded `Bash` — the
+ * narration was never backed by an actual dispatch. The silent-end Stop
+ * gate (#1664) does not catch this: that turn HAS a qualifying final
+ * reply, so it passes untouched.
+ *
+ * Mechanism (Stop-hook only — the design comment on #396 drops the
+ * 2026-04 PreToolUse deny; a pre-send deny would fire before the Agent
+ * call exists and block legitimate reply-then-dispatch flows):
+ *   1. Walk the current turn's transcript slice (same turn-anchor walk as
+ *      `silent-end-scan.mjs:scanTurnForFinalReply`), skipping
+ *      `isSidechain:true` lines. A sub-agent's own reply/dispatch lines
+ *      leak into the parent transcript with that marker; they are neither
+ *      the parent's claim NOR the parent's dispatch, so they are skipped
+ *      for BOTH detections.
+ *   2. Collect the text of qualifying `reply`/`stream_reply` tool_use
+ *      calls (reuse `REPLY_TOOLS`).
+ *   3. Detect whether the turn actually dispatched: any non-sidechain
+ *      `Agent`/`Task` tool_use, OR a `Bash` tool_use with
+ *      `run_in_background: true`.
+ *   4. Block only when a reply text makes a first-person dispatch
+ *      commitment AND the turn dispatched nothing.
+ *
+ * Registry-check decision (design comment "belt-and-braces" step 4):
+ * The design offered two options — (a) query the subagent-tracker
+ * `registry.db` for a dispatch row in this turn window, or (b) skip the
+ * registry and rely on transcript-only detection, in which case
+ * "past-tense phrasing must be excluded by the regex". We take (b). The
+ * transcript already carries the ground truth for THIS turn's dispatches
+ * (the `Agent`/`Task` tool_use blocks are in the same JSONL slice we
+ * walk), so a second sqlite round-trip from a dependency-free .mjs hook
+ * would be redundant and awkward. The registry's real value in the design
+ * was clearing PAST-tense references to earlier-turn dispatches; we
+ * achieve the same by EXCLUDING pure past-tense verb forms from the
+ * commitment regex (see DISPATCH_COMMIT_RE below) so "the worker I
+ * dispatched earlier" never trips the gate in the first place. This keeps
+ * the hook self-contained and deterministic (no cross-process DB read on
+ * the reply path).
+ */
+
+// Same two MCP tools whose payload is the model's free-text answer text
+// reaching the user (verified complete in silent-end-scan.mjs's header).
+const REPLY_TOOLS = new Set([
+  'mcp__switchroom-telegram__reply',
+  'mcp__switchroom-telegram__stream_reply',
+])
+
+// Dispatch tool names — Claude Code emits the sub-agent dispatch tool under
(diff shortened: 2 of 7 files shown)


mekenthompson:
Coordinator adversarial review (batch K2 of #3051) — APPROVE. Verified against the approved design comment: Stop-only (no PreToolUse deny), sentence-scoped commitment regex with interrogative/conditional exclusion, past-tense verb forms excluded as the documented transcript-only substitute for the registry belt-and-braces (accepted deviation — equivalent outcome, no sqlite dependency on the reply path), sidechain lines skipped for both claim and dispatch detection, backgrounded-Bash counts as a dispatch, 1-retry budget keyed on per-turn signature with fail-open exhaustion, fail-open on all error paths matching siblings. Registration mirrored in hooks.json + scaffold.ts and pinned by tests. Residual FP cost bounded to one invisible re-prompt by design. No findings.

Issue #2566, opened 2026-06-25
inject: v2.1.185 dialog commands (/status /cost /usage /model /memory) leave a modal open → next Telegram message is swallowed


mekenthompson:
## Summary
Discovered while verifying the `/clear` chrome-leak fix (#2538 / the anchored-path follow-up). On the bundled **Claude Code v2.1.185**, several inject-allowlisted slash commands no longer print inline text — they open **interactive dialogs/pickers**. `injectSlashCommand` (src/agents/inject.ts) never dismisses them, so the modal is left open on the pane and the user's **next message is silently swallowed** by the dialog.

## Affected commands (allowlist, v2.1.185 behavior)
- `/status` → Settings/Status dialog (`Esc to cancel`)
- `/cost` → usage dialog
- `/usage` → usage dialog
- `/model` → model picker (untested but same class)
- `/memory` → memory picker (untested but same class)
- `/hooks` → likely a dialog too (untested)

`/clear` and `/compact` are NOT dialogs (handled by the chrome-leak fix). `/effort` is already correctly in INJECT_BLOCKED for the same modal class (#2471).

## Reproduced live (test-harness, v2.1.185)
1. `send-keys -l '/status'` + Enter → Settings dialog opens, pane shows `Esc to cancel`.
2. `send-keys -l 'ping while dialog open'` (a normal message) → **the text vanished**; the dialog stayed open and the message was eaten. It did NOT become a prompt.

`injectSlashCommand` polls capture-pane until settle, runs diffPane, returns — it sends **no Escape/dismissal** after capture (confirmed in source). So the modal lingers until some later keystroke happens to dismiss it (producing a confusing `⎿ Settings dialog dismissed` artifact), and any plain message in between is lost.

## Severity
Silent user-message loss / pane wedge. Worse than the cosmetic `/clear` chrome leak. The allowlist metadata predates v2.1.185 turning these into modals.

## Proposed fix
1. Add a metadata flag to `INJECT_COMMANDS` (e.g. `dialog: true` / `dismissWithEscape: true`) for the dialog commands.
2. After capturing output, if the command is a dialog command, `send-keys Escape` to dismiss the modal and restore a clean `❯` prompt before returning.
3. Regression test: assert inject sends Escape for a dialog command and that the pane returns to an idle prompt (no lingering `Esc to cancel`).
[...]

Issue #2471, opened 2026-06-20
Agent session wedges on blocking /effort confirmation modal; watchdog does not self-heal


mekenthompson:
## Summary

`overlord` wedged its Claude Code TUI session live today on a blocking `/effort` confirmation modal. Observed from its docker logs:

- A blocking interactive modal was on screen: **"Change effort level? 1. Yes, switch to high / 2. No, go back"** — waiting on a `1`/`2` keypress that nothing in the agent-driven loop sends.
- A stack of queued `/effort high` slash-command messages had piled up (repeated duplicates).
- **"Stop hook error occurred"** and **"running stop hooks 0/6"** — the Stop hook errored, so the turn never cleanly ended (no clean-shutdown marker, no handoff).
- Session showed **~27 min "Manifesting", ~112k tokens**, with the manifest timer **resetting to 0s repeatedly** — spinning, not progressing.

Net effect: the session is alive and emitting terminal activity, but makes zero semantic progress and never self-recovers.

## Root-cause hypothesis

`/effort` is a Claude Code TUI slash command that opens an interactive confirmation modal ("Change effort level? 1. Yes / 2. No"). In an agent/headless-driven session the model loop cannot tap the modal's keyboard choice itself, so the modal blocks indefinitely. Queued duplicate `/effort high` injections compound it (each re-arms the modal), and because the turn never ends, the Stop hook chain throws ("0/6").

This is the same class of blocking-TUI wedge that #2064/#2065 (v0.14.33) already partially addressed for `AskUserQuestion` — but `/effort` was not covered.

### Why the existing guards do NOT catch this

**1. Inject allowlist (`src/agents/inject.ts`).** `INJECT_COMMANDS` allows only `/cost /status /usage /hooks /memory /model /clear /compact`; `INJECT_BLOCKED` lists `/login /logout /exit /quit`. **`/effort` is in neither** — so `injectSlashCommand` / `validateInjectCommand` would *reject* it with `not_allowed`. The fact that duplicate `/effort high` commands reached the pane and stacked up means they arrived via a path that bypasses this validation (raw `tmux send-keys`, operator-typed, or another injection route) with **no dedupe and no blocking-modal guard**.

[...]

Issue #725, opened 2026-05-05
Epic: tmux as agent supervisor (replace script -qfc; enable slash-command passthrough)


mekenthompson:
## Summary

Replace `script -qfc` with `tmux` as the supervisor for every long-running Claude Code agent process. This is the foundation for #163 (`/remotecontrol` and similar Telegram-driven slash-command passthrough) and incidentally fixes the cgroup-escape footgun documented at `src/agents/systemd.ts:91-98`.

## Why now

- Issue #163's implementation sketch describes a generic Claude-slash-command injection primitive (`injectClaudeCommand`) that **does not exist** in the codebase. There is no path today that puts a literal `/foo` into the running Claude REPL. `/restart` and `/reauth` are CLI restarts of the systemd unit, not REPL injections.
- Switchroom **already uses tmux** for the OAuth flow (`src/auth/manager.ts:666` — `tmux new-session 'claude setup-token'`; `:804-805` — `tmux send-keys -l <code>` then `Enter`). The hardest send-keys-into-Ink case (masked password input) is in production and stable. Extending the same pattern to the long-running agent is incremental, not novel.
- Current `script -qfc` wrapper detaches Claude from the unit's cgroup, which has previously caused a 12-day zombie process (#361). Foreground tmux with `Delegate=yes` keeps the cgroup intact.

### Phase 1 — Supervisor swap only (no inject feature)

**Goal:** every agent runs under tmux instead of `script -qfc` with zero functional regressions.

- [ ] `src/agents/systemd.ts:33-46` — replace ExecStart from `script -qfc 'expect ... start.sh' service.log` to `tmux -L switchroom-${name} -f ${tmuxConf} new-session -A -d -s ${name} -x 400 -y 50 'expect -f autoaccept.exp start.sh'` plus `ExecStartPost=tmux pipe-pane -o 'cat >> service.log'` and `ExecStop=tmux kill-session`. Add `Type=forking`, `Delegate=yes`. Keep `KillMode=control-group`, `KillSignal=SIGTERM`, `SendSIGKILL=yes`, `TimeoutStopSec=15`.
- [ ] Ship a switchroom-managed `tmux.conf` (`set -g default-terminal \"xterm-256color\"`, `set -g history-limit 100000`, `set -g status off`). `-e TERM=…` does not pin TERM under tmux 3.x — `default-terminal` config is required.
[...]

PR #3068, opened 2026-07-11
feat(inject): expand allowlist + arg-gate for sensitive verbs (#730)

### LINKED ISSUE #730: inject: expand allowlist + tighten arg-gate for sensitive verbs (#725 followup)
Followup to epic #725 (Phase 2 inject feature). Tracks expanding `INJECT_ALLOWLIST` to cover more useful read-only Claude Code slash commands, expanding `INJECT_BLOCKLIST` defensively, and tightening the validator so bare-verb-only commands can't be invoked with mutating arguments.

## Currently in allowlist (8)

`/cost`, `/status`, `/model`, `/clear`, `/compact`, `/memory`, `/hooks`, `/usage`

## Recommend adding (read-only, useful remotely)

- `/help` — list available slash commands. Sanity probe + discovery.
- `/context` — context window usage % (separate signal from `/cost`).
- `/doctor` — Claude Code health check.
- `/diff` — pending uncommitted file changes in agent's session.
- `/todo` — current in-process TODO list.
- `/agents` — list configured sub-agents (read-only display).
- `/release-notes` — Claude version info.

## Discuss before adding

- `/config` — bare invocation reads; subcommands mutate. Need arg-gate (see below).
- `/pr-comments` — fetches GH PR comments. Network-blocking; output highly variable. May overflow the 3000-byte truncation cap.
- `/export` — writes session transcript. Useful but mutates filesystem.

## Recommend expanding blocklist (defensive — distinct error code)

`/upgrade`, `/init`, `/mcp`, `/permissions`, `/install-github-app`, `/add-dir`, `/terminal-setup`, `/privacy-settings`, `/bug`

## Skip entirely

`/ide`, `/vim`, `/resume`, `/review` — need local interactivity or session args.

## Subtle issue: bare-verb match passes mutating args

Today `validateInjectCommand()` in `src/agents/inject.ts:103` lowercases the first whitespace-split token and checks the allowlist. So:

- `/model` → allowed (display) ✓
- `/model claude-sonnet-4-6` → allowed ❌ (mutates active model)

Same gap on `/memory edit`, `/clear <flag>`, etc. Tighten by adding a per-command "args allowed?" boolean alongside the allowlist, e.g.:

```ts
const INJECT_ALLOWLIST: ReadonlyMap<string, { argsAllowed: boolean }> = new Map([
  ["/cost", { argsAllowed: false }],
  ["/model", { argsAllowed: false }],   // bare display only — block setter form
  ["/memory", { argsAllowed: false }],  // bare display only — block /memory edit
  ["/help", { argsAllowed: false }],
  // ...
])
```

Then validator rejects if there's anything after the bare verb for a `argsAllowed: false` entry. Where args are intentional (e.g. `/clear` accepting `--keep-todos` if that's a real flag), set true.

## Tests to add

- Each new allowlist entry has a positive validator test
- Each new blocklist entry returns `blocked` code (not `not_allowed`)
- Arg-gate: `/model claude-sonnet-4-6` returns a clear error like `"args not permitted for /model"`
- Arg-gate doesn't break existing tests where args are inherent (none currently — current allowlist all bare-verb)

## Out of scope for this issue

- Idle-gate (mid-tool-call queueing) — separate followup, see `inject.ts:14-22` FUTURE GAP comment
- ttyd debug overlay — Phase 3 of epic #725

mekenthompson:
## Summary

Followup to epic #725 (Phase 2 inject). Tightens `validateInjectCommand`
so bare-verb allowlist matches can't smuggle mutating setter forms into
the pane, expands the read-only allowlist, and grows the defensive
blocklist.

- **Arg-gate:** new `argsAllowed: boolean` on `InjectCommandMeta`.
  `validateInjectCommand` rejects anything after the bare verb with a new
  `args_not_allowed` code and an actionable message
  (`args not permitted for /memory — inject the bare verb only`). Before,
  `/model claude-sonnet-4-6`, `/memory edit`, `/clear --flag` all passed
  the first-token check and were typed verbatim into the pane.
- **Allowlist additions (all `argsAllowed: false`):**
  - `/help` — `dialog: true`. Opens a tabbed help modal ("Esc to cancel"),
    rides the #2566 Escape-dismiss path.
  - `/context` — inline (no `dialog` flag). Renders the context-window
    usage breakdown and returns straight to `❯`.
  - `/release-notes` — `dialog: true`. Opens a version PICKER
    ("Enter to confirm · Esc to cancel"), same shape as `/memory`;
    inject never selects a row, so the Escape dismiss is side-effect-free.
- **Blocklist additions (`blocked`, not `not_allowed`):** `/upgrade`,
  `/init`, `/mcp`, `/permissions`, `/install-github-app`, `/add-dir`,
  `/terminal-setup`, `/privacy-settings`, `/bug`.

Every allowlist/dialog classification was verified with a **disposable
TUI probe against the installed `claude` v2.1.205** (the repo's grounding
method), not guessed.

## Why

Job spec `reference/jobs/operate-the-fleet-from-telegram.md` (drive an
agent's REPL from Telegram) and `reference/jobs/approve-what-my-agent-can-touch.md`
(the surface an operator can trigger must be safe by construction). The
first-token-only allowlist check let a remote surface mutate live model
state and open unanswerable modals that wedge the pane (the #2471 class).

## Deviation from the issue (important)

The issue suggested `/model { argsAllowed: false }` ("block the setter
form"). **That would break `/model`.** Since #2566 the dedicated driver
(`telegram-plugin/gateway/model-command.ts`) injects `/model <alias|id>`
WITH an argument as the sanctioned, gated, recorded session-switch path,
and `model-command.test.ts` "inject allowlist contract" pins that `/model`
stays injectable. `/model` is therefore the ONLY entry with
`argsAllowed: true`, with a code comment explaining why. A regression
test asserts `/model claude-sonnet-4-6` is accepted while every other
entry rejects args.

## Skipped commands (verified, not added — with reasons)

- `/agents` — verified safe (inline text, no modal) but its only output
  is a static "the /agents wizard has been removed" notice (no agent
  list) → near-zero remote value.
- `/doctor` — in v2.1.205 it is NOT a recognized read-only slash command:
  the probe showed it routing to the model as a prose turn (a "Puttering…"
  spinner). Injecting it would start an unwanted model turn.
- `/diff`, `/todos` — not confirmed as recognized read-only commands; the
  probe run was contaminated by `/doctor`'s model turn, so no clean
  verification. Left pending live TUI re-verification.
- `/config`, `/pr-comments`, `/export` — the issue marks these
  discuss-first; not added.

## Test plan

- `vitest run src/agents/inject.test.ts telegram-plugin/tests/model-command.test.ts telegram-plugin/gateway/inject-handler.test.ts src/web/hermes-adapter.test.ts` → **228 passed**.
- New tests: positive validator test per new allowlist entry; dialog-flag
  assertions; each new blocklist entry returns `blocked`; arg-gate rejects
  `/cost some-arg` / `/memory edit` / `/clear --flag`; `/model <id>`
  accepted (contract); every non-`/model` entry forbids args.
- `npm run lint` → clean (tsc + all 8 guards).

## Risk

Low. Pure validator/allowlist logic; no tmux/runtime path change. The one
behavior change (allowlisted-command-with-args now rejected) is the intent
and only affects non-`/model` verbs, none of which have a caller that
passes args (grepped gateway + hermes + tests). `/model`'s driver path is
unchanged and contract-pinned.

Closes #730

### DIFF
--- a/src/agents/inject.test.ts
+++ b/src/agents/inject.test.ts
@@ -30,8 +30,48 @@ describe("validateInjectCommand", () => {
     }
   });
 
-  it("accepts an allowed command with trailing args (verb-only check)", () => {
-    expect(validateInjectCommand("/cost some-arg")).toBe("/cost");
+  it("#730 arg-gate — rejects trailing args for a bare-verb-only command", () => {
+    // Was previously accepted (verb-only match); the arg-gate now rejects it.
+    for (const cmd of ["/cost some-arg", "/memory edit", "/clear --flag"]) {
+      const err = (() => {
+        try {
+          validateInjectCommand(cmd);
+          return null;
+        } catch (e) {
+          return e;
+        }
+      })();
+      expect(err, `${cmd} must be rejected`).toBeInstanceOf(InjectError);
+      expect((err as InjectError).code).toBe("args_not_allowed");
+      expect((err as InjectError).message).toMatch(/args not permitted/i);
+    }
+  });
+
+  it("#730 arg-gate — /model KEEPS args allowed (sanctioned set path, #2566)", () => {
+    // The dedicated /model driver injects `/model <alias|id>` WITH an arg as
+    // the sanctioned session-switch path. Gating it off would break /model.
+    expect(validateInjectCommand("/model claude-sonnet-4-6")).toBe("/model");
+    expect(validateInjectCommand("/model opus")).toBe("/model");
+  });
+
+  it("#730 arg-gate — every non-/model allowlist entry forbids args", () => {
+    for (const [verb, meta] of INJECT_COMMANDS.entries()) {
+      if (verb === "/model") {
+        expect(meta.argsAllowed, "/model must allow args").toBe(true);
+        continue;
+      }
+      expect(meta.argsAllowed, `${verb} must forbid args`).toBe(false);
+      const err = (() => {
+        try {
+          validateInjectCommand(`${verb} extra`);
+          return null;
+        } catch (e) {
+          return e;
+        }
+      })();
+      expect(err, `${verb} extra must be rejected`).toBeInstanceOf(InjectError);
+      expect((err as InjectError).code).toBe("args_not_allowed");
+    }
   });
 
   it("is case-insensitive on the verb", () => {
@@ -50,6 +90,33 @@ describe("validateInjectCommand", () => {
       })();
       expect(err).toBeInstanceOf(InjectError);
       expect((err as InjectError).code).toBe("blocked");
+    }
+  });
+
+  it("#730 — new defensive blocklist entries return code 'blocked' (not not_allowed)", () => {
+    const added = [
+      "/upgrade",
+      "/init",
+      "/mcp",
+      "/permissions",
+      "/install-github-app",
+      "/add-dir",
+      "/terminal-setup",
+      "/privacy-settings",
+      "/bug",
+    ];
+    for (const cmd of added) {
+      expect(INJECT_BLOCKED.has(cmd), `${cmd} must be blocklisted`).toBe(true);
+      const err = (() => {
+        try {
+          validateInjectCommand(cmd);
+          return null;
+        } catch (e) {
+          return e;
+        }
+      })();
+      expect(err, `${cmd} must throw`).toBeInstanceOf(InjectError);
+      expect((err as InjectError).code, `${cmd} → blocked`).toBe("blocked");
     }
   });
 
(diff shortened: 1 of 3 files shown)


mekenthompson:
Coordinator adversarial review (batch K2 of #3051) — APPROVE. Verified: arg-gate defaults hard-false with /model the sole argsAllowed:true entry (correct deviation from the issue — the #2566 model-command driver's sanctioned set path injects '/model <alias>' and model-command.test.ts pins the contract); new args_not_allowed code distinct from not_allowed/blocked; allowlist additions (/help, /context, /release-notes) each classified by live TUI probe on claude v2.1.205 with dialog flags matching observed modal behavior; /doctor correctly rejected (routes to a model turn — would burn a turn, not display); blocklist additions all return the specific blocked reason. Skips (/diff, /todos, /agents, /config, /pr-comments, /export) documented with reasons — acceptable scope; the security-relevant arg-gate and safe expansions land now. No findings.

PR #3074, opened 2026-07-11
fix(gateway): clear stale stacked pins in DM chats (#3026)

mekenthompson:
## Summary
DM chats accumulated stale bot-authored pins that never cleared across gateway restarts. This adds a DM-only stale-pin sweep that runs once per DM chat per boot (and lazily on first inbound from a DM after boot): `unpinAllChatMessages` to clear the whole stack, then re-pin the live tracked cards from the in-memory pin claims (silent, `disable_notification`).

New pure module `telegram-plugin/gateway/dm-pin-sweep.ts` (chat-id classification, store→DM-chat-id collection, DI sweeper with once-per-boot guard); wired into `gateway.ts` at the mutex-gated boot orchestrator (`runBootPinCleanupAndDmSweep`) and at `handleInbound` after the gate passes.

## Why
Two independent gaps made DM pins un-sweepable:
1. `shouldSweepChatAtBoot` skips positive (DM) IDs — the Bot API returns `400 chat not found` for a user who never messaged the bot, so DM chats were never boot-swept. Ken saw stale pins in every DM on 2026-07-11 (fleet on v0.18.9).
2. Even when reachable, `getChat()` exposes only the NEWEST pin; DMs stack pins and the Bot API has no list-pins method, so older orphans are invisible to a probe-based (unpin-one-by-id) sweep forever.

`unpinAllChatMessages` clears the whole stack in one call — safe in a DM where virtually every pin is bot-authored. Groups/supergroups (negative IDs) keep the existing probe path unchanged; an unpin-all there would nuke human pins.

Design details honored:
- Rate-limit aware: unpin-all + re-pins route through the existing `robustApiCall` wrapper (a 429 flood-wait is retried then degrades; the sweeper absorbs the reject — never crashes, never unhandled).
- At most once per DM chat per boot via an in-memory guard (claimed before the first await, so concurrent boot+first-inbound triggers dedupe).
- Never runs on the losing side of the startup mutex: `dmPinSweepEligible` is set true only inside the won-lock boot branches, mirroring `statusPinBootCleanup`'s gate.
- Silent throughout: no messages sent, unpin/re-pin only.

## Test plan
- New `telegram-plugin/gateway/dm-pin-sweep.test.ts` (13 tests, vitest) asserts OUTCOMES: unpin-all called exactly once for a DM with tracked pins followed by re-pin of exactly the live ids; DM with no tracked pins → unpin-all once, no re-pin; groups never get unpin-all; not-eligible does nothing; second sweep no-ops; concurrent triggers fire unpin-all once; 429/unpin-all reject degrades without throwing and still re-pins; re-pin failure absorbed; no unhandled rejection escapes a fire-and-forget failing sweep.
- Updated the two structural mutex-gating tests in `status-pin-boot-recovery.test.ts` to track the new gated orchestrator (invariant preserved: cleanup only runs post-lock).
- Local: `tsc --noEmit` clean, `npm run lint` clean (bot-api-wrapping + no-pii-secrets included), boot-recovery + slot-banner + dm-pin-sweep suites green (22 tests).

## Risk
Low. DM-only behavior change; groups untouched. Worst case a benign extra unpin-all/re-pin in a DM (idempotent, silent). Fire-and-forget with full error absorption — cannot affect turn flow or crash the gateway.

Job spec: `reference/jobs/know-what-my-agent-is-doing.md`.

Fixes #3026

### DIFF
--- a/telegram-plugin/gateway/dm-pin-sweep.test.ts
+++ b/telegram-plugin/gateway/dm-pin-sweep.test.ts
@@ -1,0 +1,181 @@
+import { describe, expect, it, vi } from 'vitest'
+import {
+  isDmChatId,
+  collectDmChatIdsFromStores,
+  createDmPinSweeper,
+  type DmPinSweeperDeps,
+} from './dm-pin-sweep'
+
+// Synthetic chat IDs only (check-no-pii-secrets): positive = DM, negative =
+// group/supergroup.
+const DM_A = '5000001'
+const DM_B = '5000002'
+const GROUP = '-1002000000001'
+
+describe('isDmChatId', () => {
+  it('true for positive integer DM ids', () => {
+    expect(isDmChatId(DM_A)).toBe(true)
+    expect(isDmChatId('1')).toBe(true)
+  })
+  it('false for group/supergroup (negative), zero, non-integer, non-numeric', () => {
+    expect(isDmChatId(GROUP)).toBe(false)
+    expect(isDmChatId('-1')).toBe(false)
+    expect(isDmChatId('0')).toBe(false)
+    expect(isDmChatId('1.5')).toBe(false)
+    expect(isDmChatId('abc')).toBe(false)
+    expect(isDmChatId('')).toBe(false)
+    expect(isDmChatId('Infinity')).toBe(false)
+  })
+})
+
+describe('collectDmChatIdsFromStores', () => {
+  it('collects distinct DM chat ids across all three stores and drops non-DMs', () => {
+    const ids = collectDmChatIdsFromStores({
+      statusPins: [{ chatId: DM_A }, { chatId: GROUP }, { chatId: DM_A }],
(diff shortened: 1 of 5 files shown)


mekenthompson:
Reviewer blocker addressed in 927d628d: the DM sweep's re-pin source now unions live in-memory claims with unexpired `tool:` store rows via a new pure helper `unexpiredStoreRepinIds` (read LIVE from status-pins.json, so both the boot sweep — where in-memory maps are empty — and first-inbound sweeps preserve deliberate agent pins per the #3001 contract). Expired and work-scoped rows are still swept; ids are deduped before re-pin. Added the three requested outcome tests (unexpired tool row re-pinned after unpinAll; expired row not re-pinned; first-inbound preserves + dedupes). Scoped suites 50/50 green, npm run lint clean. No auto-merge enabled.

PR #3075, opened 2026-07-11
feat(hostd): pre-approved config_propose_edit bypasses rate limit (#2975 Stage 2)

mekenthompson:
## Summary

Stage 2 (hostd bypass) of #2975. An operator-approved mental-model persist (and the "🔁 Always allow" path) reuses `config_propose_edit` and was counted against the proposing agent's 3/hour config-edit bucket — an already-approved change could be silently throttled. Stage 1 (PR #3060) added a gateway-only retry backstop; Stage 2 removes the collision at the hostd layer so the approved persist applies immediately.

- **`ApprovalGateway.checkPreApproved(agentName, unifiedDiff)`** — a new read-only query. `SocketApprovalGateway` sends a `check_pre_approved` message over the **existing hostd→gateway approval-gateway socket** (the same connection `requestApproval` uses) and reads back `pre_approved_result.preApproved`.
- **Gateway answers by EXACT diff byte-match.** `isDiffPreApproved` (new pure module `telegram-plugin/gateway/pre-approval-check.ts`) peeks the gateway's existing correlation maps (`pendingMentalModelCorrelations` + `pendingAlwaysAllowCorrelations`) — the same forge-resistance gate as the `tryAutoResolve` auto-resolve hook, **minus the single-use delete**. It never mutates the store, so a peek can't consume the consent the real `request_config_approval` spends later.
- **`handleConfigProposeEdit` consults `checkPreApproved` BEFORE `checkConfigEditRate`.** On an affirmative answer it skips **both** the rate-limit enforcement and its bucket count (never calls `checkConfigEditRate`, so no timestamp is recorded), and marks the `requested` audit row `pre_approved: true`.
- **Dedupe atomicity preserved.** The pre-approval query is async, so the pre-approval + rate-limit + apply stages were moved into one deduped in-flight promise (`runConfigProposeRateThenApply`), registered synchronously — identical concurrent proposals still collapse onto one card + one apply.

## Why

An operator's approval is the source of truth and must be honored durably (`reference/jobs/approve-what-my-agent-can-touch.md`) — a system that gets better the longer it runs must not silently drop an approved change (`reference/jobs/get-better-the-longer-they-run.md`). The 3/hour throttle exists to cap approval-card spam; the approved-persist path posts zero cards and consumes zero operator attention, so it should never have been subject to the bucket. Stage 1 guaranteed the change lands at the window or fails loudly; Stage 2 lets it land immediately.

## Fail closed (mixed-version rolls)

The query is a read-only optimisation with a hard 2s cap and no state mutation. Every non-affirmative condition resolves `false` and falls through to today's exact rate-limited behaviour (then Stage 1's gateway retry covers an approved-but-throttled persist):

- no reachable gateway socket;
- an **old gateway that doesn't implement `check_pre_approved`** — the IPC server replies `preApproved: false` when no handler is wired;
- a malformed / non-JSON / unknown-type / mismatched-correlationId reply;
- socket error/close before an answer;
- query timeout.

A hostd built with the check but wired to a gateway/stub lacking `checkPreApproved` is the same fail-closed default (the method is optional on the interface).

## Security

The `check_pre_approved` message travels **hostd → gateway only** over the approval-gateway socket. No agent-facing hostd verb emits it, so the agent process can't forge it; the gateway handler additionally rejects a query whose `agentName` isn't this gateway's own agent, and matches strictly on exact diff bytes.

## Test plan

- `vitest run src/host-control/approval-gateway.test.ts` — adapter unit: `checkPreApproved` true/false, and fail-closed on no-answer/timeout, unknown-message, garbled reply, socket close, no-socket, mismatched correlationId.
- `vitest run tests/host-control/config-propose-edit-rate-limit.test.ts` — server outcomes: exhausted bucket + pre-approval → applies immediately (no `E_RATE_LIMITED`, audit `pre_approved: true`); bypass does NOT consume the bucket (next unapproved proposal still hits the limit exactly as before); fail-closed with no `checkPreApproved` → `E_RATE_LIMITED` with structured `fix.retry_after`.
- `vitest run telegram-plugin/tests/pre-approval-check.test.ts` — pure predicate: true only for a byte-exact registered pair (mental-model + always-allow), forge-resistant on a one-byte diff, agent mismatch → false, and **never mutates the store** on hit or miss.
- `bun test telegram-plugin/tests/ipc-server-check-pre-approved.test.ts` — real `createIpcServer` over a tmp UDS: wire-shape validation, handler invoke/relay, and the no-handler fail-closed default (added to the bun runner + vitest exclude per the two-runner boundary).
- `npm run lint` — clean (tsc + all 8 guard scripts).

Pre-existing, environment-only local failures (verified identical on clean `origin/main`, not touched by this diff): `config-propose-edit-idempotency.test.ts` apply/reconcile and `config-propose-edit.test.ts` `mount --bind` cases fail in this sandbox for lack of privileges — CI is the authority for those paths.

## Risk

Low. New IPC is read-only and additive; the bypass only fires on a byte-exact operator-consented correlation. Everything else fails closed to today's behaviour, so an old gateway during a mixed-version roll is safe. No change to the ordinary (un-consented) rate-limit path.

Closes #2975 (Stage 2; Stage 1 was PR #3060)


mekenthompson:
Coordinator adversarial review (batch K2 of #3051) — APPROVE (Stage 2). Verified: bypass consults checkPreApproved BEFORE checkConfigEditRate and on true never calls it (no enforcement, no bucket timestamp — bucket-not-consumed asserted in config-propose-edit-rate-limit.test.ts); fail-closed at three layers (SocketApprovalGateway adapter resolves false on no-socket/timeout-2s/malformed/unknown/close/correlation-mismatch, hostd defensive catch, ipc-server fail-closed default when handler unwired) with mixed-version behavior identical to today + Stage 1 retry; gateway predicate extracted pure (pre-approval-check.ts), read-only — the single-use correlation delete stays with the real auto-resolve; exact byte-match forge-resistance asserted incl. one-byte-diff; audit row pre_approved:true. The synchronous dedupe-registration restructure correctly closes the async-query window (worker caught the double-card regression and added coverage). Worst-case agent-originated check_pre_approved is a boolean read, not a bypass — hostd issues its own query. No findings. Local mount-bind/idempotency failures reproduced on clean main (env-only); CI is authority.

PR #3092, opened 2026-07-11
feat(telegram-plugin): deterministic outbound send gate with token buckets (flag-off) — #3084 PR 1/3

mekenthompson:
## Summary

PR **1 of 3** in the Telegram flood-safety series (issue #3084). Introduces a deterministic token-bucket outbound send gate wired at the `robustApiCall` layer so **every** Bot API call the plugin makes routes through one scheduler — no call site can bypass it. Feature-flagged and **default-OFF**; when off the gate is a pure passthrough with zero behaviour change.

Design: `part3-design.md` §1 (central token-bucket send gate) and §4 (same-message edit hygiene). Audit input: `part2-audit.md` §3 (no global/per-chat token bucket; per-surface throttles don't compose into a ceiling).

## Why

Each agent's gateway drives many outbound surfaces (reply chunks, answer/draft stream edits, typing, worker-feed edits, reactions, cards). Each has its own local throttle but nothing composes them into a global or per-chat ceiling, so during a busy turn the per-surface throttles add up with no cap and trip a per-bot-token flood ban (429 `retry_after` ~hours, #3084). The #1 documented ban trigger is rapid repeated `editMessageText` on the same message — exactly switchroom's progress/worker/answer-card pattern.

## What changed

- **`telegram-plugin/send-gate.ts`** (new) — nested token buckets:
  - Global **25/sec** (headroom under Telegram's ~30/sec).
  - Per-chat **1/sec sustained, burst 3**.
  - Per-group **18/min** (headroom under 20), keyed on chat type (group/supergroup only).
  - Per-message edit floor **1.5s** with **last-write-wins coalescing** — an edit arriving while one is pending replaces the pending payload rather than queuing behind it; the next edit carries full state.
  - **No-op edit skip** — hash the rendered payload per `message_id`; identical → dropped before the API (Telegram 400s "message is not modified" today, still costing flood budget).
  - **Injectable clock** (`now()` + `sleep()`); no inline `Date.now()`/`setTimeout()` in the scheduling logic. Token consumption is a synchronous critical section so concurrent admissions never double-spend.
  - **Stats getter** exposes `sent`/`queued`/`coalesced`/`dropped` counters + live bucket fill for the observability PR (PR 3).
- **`telegram-plugin/retry-api-call.ts`** — `RetryCallOpts` gains optional `chatType` / `messageId` / `editPayload` so call sites can key the buckets. The retry policy itself ignores them (additive, backward compatible).
- **`telegram-plugin/gateway/gateway.ts`** — compose the gate around the retry wrapper at the single `robustApiCall` definition. Everything already routing through `robustApiCall` / `swallowingApiCall` / the boot-card adapter now transits the gate when enabled.

## Feature flag

`SWITCHROOM_TELEGRAM_SEND_GATE=1` enables it (follows the `SWITCHROOM_*=== '1'` gateway flag convention). **Default OFF** → `gate()` passes straight through to the retry wrapper; buckets never engage; no timers; identical behaviour to today.

## Scope note (1 of 3)

- **PR 1 (this):** send gate + buckets + no-op-edit skip, pure gateway, flag-off.
- **PR 2:** priority classes + shedding + degraded mode wired to the #2923 flood breaker; structured `flood_wait` error on the MCP reply path.
- **PR 3:** observability + operator alert (builds on the counters exposed here).

When the flag flips on in PR 2/3, call sites will start passing `chatType`/`messageId`/`editPayload` to fully engage the per-group bucket and edit coalescing; PR 1 wires the plumbing and the global/per-chat buckets that key off the `chat_id` call sites already pass.

## Test plan

- New vitest suite `telegram-plugin/send-gate.test.ts` (12 tests) driven by a **deterministic fake clock** — asserts outcomes, not code paths:
  - global bucket admits burst then paces the rest by refill (3rd call runs at t=500ms under 2/sec);
  - per-chat burst-3-then-1/sec with **independent** other chats;
  - per-group ceiling keyed on chat type (3rd group send at t=30s under 2/min); private chats never engage the group bucket;
  - edit coalescing sends **only the last payload** (v1 immediate, v2/v3 within floor → only v3 fires at t=1.5s);
  - no-op skip drops an identical repeat and doesn't clobber a genuinely-different queued edit;
  - flag-off passthrough (no delay, counters untouched, edit floor bypassed).
- Existing `retry-api-call.test.ts` (37 tests) still green.
- `npm run lint` (tsc --noEmit + all 8 guard scripts, incl. `check-bot-api-wrapping`) clean.

Scoped tests + lint locally per repo protocol; CI is the full-suite authority.

## Risk

Low. Default-off feature flag; when off the code path is `if (!enabled) return fn()`. No raw `bot.api.*` calls added (lint guard clean). Additive `RetryCallOpts` fields.

---

## Adversarial review follow-up (commit ba9fa88b)

All findings from the DO-NOT-MERGE review fixed; flag remains default-OFF and flag-off passthrough is unchanged. The send gate now uses a **serialized per-message driver** (one send per `message_id` in flight at a time) as the durable mechanism behind H1/M3/M4.

| Finding | Fix | Test |
|---|---|---|
| **H1** coalescing error path orphaned the wrong pending edit | Driver rejects the closure-local `p`, never `state.pending`; a newer edit that arrived mid-flight survives and sends on the next loop | `rejects the failing edit and still sends a newer edit that arrived mid-flight (H1)` |
| **H2** unbounded `perMessage` map | TTL sweep of idle states + hard LRU cap (`messageStateTtlMs`, `maxMessageStates`); never evicts pending/running/window-open states; `stats().messageStates` exposes size | `evicts idle message states after the TTL`, `enforces a hard LRU cap on live message states` |
| **M1** failed edit poisoned the no-op skip | `lastHash` set only AFTER a successful send | `retries an identical payload after a FAILED edit (M1)` |
| **M2** buckets permitted ~2x sustained rate as burst | Split burst capacity from sustained rate (`globalBurst` default 5, `perGroupBurst` default 2); rate = budget | `bounds burst to capacity and holds sustained rate at the budget over windows`, `holds a per-group ceiling near budget/min, not ~2x` |
| **M3** same-tick double-send on the immediate path | Per-message serialization — second same-tick edit coalesces behind the first | `serializes two same-tick edits: one now, one after the floor (M3)` |
| **M4** parallel edits while a send sleeps on `retry_after` | Edits arriving during an in-flight send coalesce (last-write-wins); exactly one follow-up fires | `coalesces edits behind an in-flight send sleeping on retry_after (M4)` |
| **L2** `hashPayload(undefined)` threw | Fixed sentinel for undefined/non-stringifiable payloads | `does not throw when editPayload is undefined; treats a repeat as a no-op` |
| **L3 / §7** no restart-proof flood hook | Added `initialWindows` + `bootRamp` config and `openFloodWindow(scopeKey, untilTs)`; buckets consult a per-scope "suppressed until" window so PR 2 can load/persist `flood-wait.json` and re-open windows without refactoring this module | `an open flood window blocks admission until untilTs`, `re-opens flood windows passed at construction`, `starts the global bucket at a fraction of capacity during the boot ramp` |
| **L1** edit hygiene dormant in PR 1 | Scope acknowledgement only — call sites start passing `messageId`/`editPayload`/`chatType` in **PR 2**, so the per-group bucket, edit floor, coalescing, and no-op skip engage in production then. No code change in PR 1 beyond the above. | n/a |

Scoped `send-gate.test.ts` (24 tests) + `retry-api-call.test.ts` (37) green; `npm run lint` clean. CI is the full-suite authority.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -189,6 +189,7 @@ import {
   isPhotoDimensionRejectError,
   isFloodWaitActiveError,
 } from '../retry-api-call.js'
+import { createSendGate, sendGateEnabledFromEnv } from '../send-gate.js'
 import { classifyPhotoFile, rerouteResultSuffix } from '../photo-precheck.js'
 import { installTgPostLogger, withTgPostTags } from '../shared/bot-runtime.js'
 import {
@@ -5331,7 +5332,7 @@ const PHOTO_EXTS = new Set(['.jpg', '.jpeg', '.png', '.gif', '.webp'])
 // Declared ABOVE the typing indicator because the typing sends now route
 // through the same policy (#3084) — see nonEssentialApiCall below.
 const FLOOD_STATE_PATH = floodStatePath(STATE_DIR)
-const robustApiCall = createRetryApiCall({
+const rawRobustApiCall = createRetryApiCall({
   log: (line) => process.stderr.write(line),
   onFloodWait: makeFloodWaitRecorder(FLOOD_STATE_PATH),
   // #3094: while a LONG per-bot ban is open, don't issue the call at all.
@@ -5341,6 +5342,20 @@ const robustApiCall = createRetryApiCall({
   // fires thousands of requests into the open window and extends the ban.
   floodWaitRemainingMs: makeFloodWaitProbe(FLOOD_STATE_PATH),
 })
+
+// #3084 PR 1/3 — deterministic outbound send gate (token buckets + per-message
+// edit floor + no-op-edit skip). Wrapped HERE at the robustApiCall layer so
+// every Bot API call routed through the standard retry policy also transits
+// one scheduler (no call site can bypass it). Feature-flagged, default OFF:
+// when SWITCHROOM_TELEGRAM_SEND_GATE !== '1' the gate is a pure passthrough
+// and the retry policy behaves exactly as before. Composes with #3094's
+// pre-call flood gate and #3097's non-essential drop policy — those decide
+// whether a call happens at all; this paces the calls that do.
+const sendGate = createSendGate({ enabled: sendGateEnabledFromEnv() })
+const robustApiCall = <T>(
+  fn: () => Promise<T>,
+  opts?: Parameters<typeof rawRobustApiCall<T>>[1],
+): Promise<T> => sendGate.gate(() => rawRobustApiCall(fn, opts), opts)
 
 // Fire-and-forget wrapper for outbound surfaces that previously had
 // `.catch(() => {})` directly on `bot.api.*` calls. Resolves to undefined
--- a/telegram-plugin/retry-api-call.ts
+++ b/telegram-plugin/retry-api-call.ts
@@ -44,6 +44,22 @@ export interface RetryCallOpts {
    * call sites to self-document. Future: surface in retry logs / metrics.
    */
   verb?: string
+  /**
+   * Chat type of the destination (private / group / supergroup / channel).
+   * Consumed by the outbound send gate (#3084, send-gate.ts) to key the
+   * per-group bucket. Optional and additive — the retry policy ignores it.
+   */
+  chatType?: 'private' | 'group' | 'supergroup' | 'channel'
+  /**
+   * For edit calls: the target message id. Enables the send gate's
+   * per-message edit floor + last-write-wins coalescing. Retry policy ignores.
+   */
+  messageId?: number
+  /**
+   * For edit calls: the rendered payload the send gate hashes to skip no-op
+   * edits (identical → dropped) and to coalesce. Retry policy ignores it.
+   */
+  editPayload?: unknown
 }
 
 export interface RetryObserver {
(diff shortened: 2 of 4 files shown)


mekenthompson:
Applied the LOW residual fixes from the re-review (commit `4e2714ed`), all in flag-OFF-dormant code:

- **M2 residual** — default `globalBurst` 5 → 4 so worst-case sliding-window admissions are `4 + 25 = 29 < 30/s` (real headroom vs. burst 5 hitting exactly the ceiling). M2 test bounds + comments updated; added an explicit `maxAny < 30` assertion.
- **N1** — the coalesce (last-write-wins) check now runs *before* the no-op skip in `handleEdit`, so a later edit reverting to the on-screen payload replaces a queued distinct edit instead of being dropped while the stale edit renders. The driver's own no-op re-check drops the coalesced revert. Rewrote the no-op-while-queued test to assert corrected LWW; added a plain-no-op test.
- **N2** — `openFloodWindow` for a `msg-edit:` scope now goes through `maybeEvict` / LRU-cap accounting like normal state creation. Added a test.
- **N3** — documented the never-settling-`fn` liveness assumption (`robustApiCall`'s own timeouts bound `p.fn()`; no watchdog needed).
- **N4** — the non-edit path increments `counters.sent` only after a successful `fn()`, mirroring the edit path; a throwing send is no longer counted. Added a test.

Scoped local run: `send-gate.test.ts` (27) + `retry-api-call.test.ts` (37) = 64 pass; `npm run lint` clean. Not merging — leaving that to CI + operator.

PR #3097, opened 2026-07-11
fix(telegram): cap typing pings per chat so they can't earn a flood ban

mekenthompson:
## Summary

Makes the Telegram typing indicator structurally incapable of earning a per-bot flood ban (#3084).

- **New `telegram-plugin/typing-emitter.ts`** — one shared per-chat-key emission floor (`TYPING_FLOOR_MS = 3500`, deliberately just under the 4s refresh so a loop's own on-time refresh is never eaten by timer jitter). **Both** typing loops (`startTypingLoop`/`typingIntervals` and `turnTypingLoop`) plus the two one-shot inbound pings now emit through it, so the floor holds *across* them — they target the same chat key and neither can out-shout the other. A **cold start still pings instantly** (the UX intent); only *redundant* restarts are dropped.
- **Typing sends stop bypassing the retry policy.** They no longer call `bot.api.sendChatAction` raw with `.catch(() => {})`. They route through a new `nonEssentialApiCall` (`createRetryApiCall({ maxRetries: 1, sleep: noop })`) which still fires `onFloodWait`, so the #2923 flood circuit breaker finally observes 429s from the single largest source of them. A typing ping is **never retried** — retrying a non-essential send is what feeds a ban, and we are not going to sleep out a 16,739s `retry_after` on a typing indicator.
- **No typing at all while a flood window is open.** `suppressNonEssentialSendMs` gained its second callsite (it had exactly one: `boot-card.ts`). Typing is non-essential by definition; sending into an open window cannot succeed and can extend the ban.
- Fixed the two comments that encoded the wrong assumption ("the redundant `typing` pings ... are harmless — same action, and `sendChatAction` is cheap"). That assumption is what earned the ban.

## Why

On 2026-07-11 06:24 UTC `overlord` earned `429 retry_after=16739s` — a **4.6-hour** per-bot-token ban that killed every outbound reply until 11:03 UTC (#3084).

Measured outbound mix for that agent that day:

| call | count |
|---|---|
| `sendChatAction` | **8,729** (55% of all outbound volume) |
| `editMessageText` | 6,443 |
| `sendRichMessage` | **203** (actual messages delivered) |

8,729 typing pings to deliver 203 messages, bursting at 200-300/min into a **single DM**.

Root cause: `startTypingLoop()` fires an action *immediately* on every (re)start, and the tool-use wrapper restarts the loop on **every tool call**. So the ping rate equalled the agent's **tool-call rate**, not one per 4s — the 4000ms interval was decorative. With parallel subagents hammering tools this hit ~5/sec. A second loop (`turnTypingLoop`) with its own interval map targets the same chat key and also fires immediately on start, compounding it.

**This is the gap PR #3092 (`feat/telegram-send-gate`) does NOT cover.** That token-bucket send gate is composed around `robustApiCall` — and typing called `bot.api` **raw**, so the gate would never have seen 55% of outbound volume. This PR does not depend on or merge into #3092; it just stops typing from bypassing the wrapper, so it benefits automatically when #3092 lands. Same story for the #2923 breaker, whose `onFloodWait` hook only fires inside `createRetryApiCall` — it was blind to 429s from the biggest emitter.

Job spec: `reference/jobs/talk-to-agents-from-anywhere.md` (**always-available**; "Bad looks like" — the surface going dead). The typing indicator itself serves `know-what-my-agent-is-doing`, and its Good behaviour is preserved: "typing…" still lands the instant a turn begins, and the emitter's floor sits under Telegram's ~5s action expiry so the chat never goes dark mid-turn (asserted in the tests).

## Test plan

New `telegram-plugin/tests/typing-emitter.test.ts` (vitest, **fully deterministic fake clock**, no real timers), 12 tests asserting outcomes:

- **The regression:** 50 loop restarts across 2s emit **exactly 1** chat action (was 50).
- **The compound case:** the *real* `createTurnTypingLoop` running alongside a faithful reproduction of the tool-use loop (stop → restart → immediate fire → re-arm interval), driven by 120 tool calls over 60s into one DM — the shape that produced the ban. Asserts total emissions ≤ `ceil(60s / floor) + 1`, that no two emissions on a key are inside the floor, **and** that no gap exceeds Telegram's 5s action expiry (the chat stays lit).
- Cold start still pings immediately; the floor releases on schedule; an on-time 4s refresh is never dropped.
- Two chats — and two topics in one chat — are independent lanes.
- **Nothing** is emitted while a flood window is open; emission resumes when it closes.
- A 429 **is recorded** to the flood breaker and the ping is **not retried** (`fn` called exactly once); the recorded window then suppresses further typing end-to-end.
- Flood state is read at most once per floor per chat (no fs read-storm under a restart burst).

Ran green:
- `vitest run telegram-plugin/tests` — **358 files / 5844 tests passed**.
- `npm run lint` — clean (tsc + all 8 guards, including `check-bot-api-wrapping`; the one remaining raw call carries a drift-proof `// allow-raw-bot-api:` marker, no line-range allowlist edits needed).

## Risk

Low-moderate, UX-surface-touching.

- **Ping volume drops ~20x** by design (worst case ~17/min per chat, was 200-300/min). The visible risk is a *dark* chat, and it is bounded: the floor (3.5s) is strictly under both the refresh cadence (4s) and Telegram's action expiry (~5s), and a cold start is never delayed. The 5s-gap assertion in the loop-integration test pins this.
- Behaviour change on failure: a failing typing ping is now **logged and dropped** instead of silently swallowed (a 429 additionally trips the breaker). One extra stderr line per failed ping — rare, and it is exactly the signal that was missing for months.
- The 401 → exponential-backoff restart path is preserved unchanged.
- No change to reply/edit delivery paths. Not an invariant crossing; deterministic mechanism, no model-dependent behaviour.

Closes #3084.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -187,6 +187,7 @@ import {
   createSwallowingRetryApiCall,
   retryWithThreadFallback,
   isPhotoDimensionRejectError,
+  isFloodWaitActiveError,
 } from '../retry-api-call.js'
 import { classifyPhotoFile, rerouteResultSuffix } from '../photo-precheck.js'
 import { installTgPostLogger, withTgPostTags } from '../shared/bot-runtime.js'
@@ -5362,17 +5363,25 @@ const swallowingApiCall = createSwallowingRetryApiCall(
  * which only learns about 429s through `createRetryApiCall`'s `onFloodWait`
  * hook — was blind to 429s from the biggest source of them.
  *
- * `maxRetries: 1` + a no-op sleep gives exactly one attempt: a 429 still runs
- * `onFloodWait` (recording the window to the breaker), then the call gives up
- * immediately instead of sleeping out a 4-hour retry_after. The swallowing
- * wrapper logs and drops the failure.
+ * `maxRetries: 1` + a no-op sleep gives exactly one attempt at the API, whatever
+ * the outcome: a 429 still runs `onFloodWait` (recording the window to the
+ * breaker), and then the call ends — a SHORT ban falls out of the loop as
+ * `max retries exceeded`, a LONG one throws #3094's `FLOOD_WAIT_ACTIVE`. Either
+ * way the ping is dropped, never slept and never retried. That is the whole
+ * point: a retried non-essential send is what feeds a ban.
+ *
+ * Composes cleanly on top of #3094 (`bbe14471`): that PR bounded the in-process
+ * flood sleep and added a pre-call gate for LONG open windows. This wrapper
+ * pre-dates neither mechanism nor fights them — it sits under the emitter's own
+ * flood gate, which short-circuits earlier still (a typing ping during a ban
+ * never even reaches the retry layer). Defense in depth, in that order.
  */
 // No `log` hook: retryApiCall's flood line reads "waiting Ns", which would be a
-// lie here — we never wait, we drop. And the error the caller finally sees is
-// `retryApiCall: max retries exceeded` (retry-api-call.ts replaces the original
-// on give-up), which would hide the retry_after — the single number an operator
-// needs during a flood ban. So the honest 429 line is written HERE, in the
-// onFloodWait hook, where the real retry_after is still in hand.
+// lie here — we never wait, we drop. And the error the caller finally sees names
+// neither the code nor the retry_after (retryApiCall replaces the original on
+// give-up) — and retry_after is the single number an operator needs during a
+// ban. So the honest 429 line is written HERE, in the onFloodWait hook, where
+// the real value is still in hand.
 const recordTypingFloodWait = makeFloodWaitRecorder(FLOOD_STATE_PATH)
 const nonEssentialApiCall = createRetryApiCall({
   maxRetries: 1,
@@ -5466,11 +5475,15 @@ const typingEmitter = createTypingEmitter({
           typingRetryTimers.set(key, retry)
           return
         }
-        // A 429 already logged its retry_after (and hit the breaker) in
-        // onFloodWait above; by the time it lands here retryApiCall has
-        // replaced it with the opaque give-up error, so don't double-log a
-        // line that names neither the code nor the retry_after.
-        if (msg.includes('max retries exceeded')) return
+        // A flood-wait already logged its retry_after (and hit the breaker) in
+        // onFloodWait above. What lands here is either retryApiCall's opaque
+        // give-up error (short ban) or #3094's FLOOD_WAIT_ACTIVE marker (long
+        // ban / pre-call gate) — neither names the retry_after, so don't
+        // double-log a strictly less informative line. Both are DROPS: a
+        // typing ping is never retried, and the marker is caught here (this is
+        // an onRejected handler), so it can't surface as an unhandled rejection
+        // out of a fire-and-forget ping.
+        if (isFloodWaitActiveError(err) || msg.includes('max retries exceeded')) return
         // Everything else is DROPPED, never retried — but logged, so the
         // largest outbound emitter is no longer silent (#3084).
         process.stderr.write(
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -5367,12 +5367,23 @@ const swallowingApiCall = createSwallowingRetryApiCall(
  * immediately instead of sleeping out a 4-hour retry_after. The swallowing
  * wrapper logs and drops the failure.
  */
-// No `log` hook: retryApiCall's flood line reads "waiting Ns", which would be
-// a lie here — we never wait, we drop. The callsite logs the drop itself.
+// No `log` hook: retryApiCall's flood line reads "waiting Ns", which would be a
+// lie here — we never wait, we drop. And the error the caller finally sees is
+// `retryApiCall: max retries exceeded` (retry-api-call.ts replaces the original
+// on give-up), which would hide the retry_after — the single number an operator
+// needs during a flood ban. So the honest 429 line is written HERE, in the
+// onFloodWait hook, where the real retry_after is still in hand.
+const recordTypingFloodWait = makeFloodWaitRecorder(FLOOD_STATE_PATH)
 const nonEssentialApiCall = createRetryApiCall({
   maxRetries: 1,
   sleep: async () => {},
-  onFloodWait: makeFloodWaitRecorder(FLOOD_STATE_PATH),
+  onFloodWait: (retryAfterSec) => {
+    recordTypingFloodWait(retryAfterSec)
+    process.stderr.write(
+      `telegram gateway: 429 flood-wait on a NON-ESSENTIAL send (retry_after=${retryAfterSec}s) — ` +
+        `recorded to the flood breaker; the send is DROPPED, not retried (#3084)\n`,
+    )
+  },
 })
 
 // ─── Typing indicator ─────────────────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


mekenthompson:
## REQUEST_CHANGES addressed — both blockers fixed (`d3989f8`)

### BLOCKER 1 — the chat goes dark: fixed with the coalesced catch-up

Reviewer is right, and the mechanism is exactly as described: both loops re-arm a **fixed-cadence** interval, so a tick eaten by the floor was simply *lost* — pushing the next emission out to `floorMs + refreshMs` = 7500ms, past Telegram's ~5s action expiry. I took the preferred fix, not the `floor <= 1000` alternative: the catch-up keeps the tight cap **and** bounds the gap, where the alternative only bounds the gap while tripling the emission ceiling.

`typing-emitter.ts`: a floor-dropped emit arms **one coalesced catch-up per chat key** at `claimedAt + floorMs` (a second drop inside the window does not arm a second timer). Worst-case gap is now bounded by the floor (3.5s); the emission cap is untouched, because a catch-up can only fire *after* the window has opened. `schedule`/`cancel` are injected deps — no inline `setTimeout` in schedulable logic.

Lifecycle, so it can neither leak nor resurrect typing after a reply lands:
- **Canonical turn-end** (`stopTurnTypingLoop`) → `cancelPending`. Deliberately NOT in `stopTypingLoop` — the tool loop stops on every tool result, which is precisely the churn the catch-up exists to cover.
- **Open flood window** → the catch-up is cancelled, not deferred into the ban.
- **Shutdown drain** (`stopTurnTypingLoops`) → `reset()` cancels every armed timer.
- A successful send cancels the pending catch-up, which is what prevents a catch-up from re-arming itself into a self-sustaining heartbeat.

### BLOCKER 2 — the compound test couldn't fail: replaced with absolute bounds + an adversarial sweep

Confirmed: `Math.ceil(60_000 / TYPING_FLOOR_MS)` is `Infinity` at `floor = 0`. The bound was written in terms of the thing under test. Replaced with bounds derived from the **outcome**, not the implementation:

- `MAX_EMISSIONS_PER_TURN = 20` (60s at the enforced floor is ~17; the pre-fix code emits ~136)
- `MAX_DARK_MS = 5000` (Telegram's action expiry — `know-what-my-agent-is-doing`: "stays present from receipt to turn end")

Three tests now carry them, over the REAL `createTurnTypingLoop` + a faithful tool-loop driver:
1. **The reviewer counterexample** — `phase=3500 dur=3400 n=6`, asserted never dark.
2. **Adversarial sweep** — 576 schedules over tool-call phase (0-4000) x duration (200-4200) x gap (150/900/2600) x count (3/6/12). Asserts worst dark gap and worst emission count across ALL of them, and that the floor held on every one.
3. **The 120-tool-call storm** — the incident shape, `<=20` emissions over 60s.

### Mutation results (the part that matters)

| mutation | outcome |
|---|---|
| `TYPING_FLOOR_MS = 0` (floor disabled) | **RED** — storm test: `expected 136 to be less than or equal to 20`; sweep: `worst emission count 40`. 9/16 tests fail. |
| catch-up call removed (this PR's Blocker-1 fix disabled) | **RED** — sweep: `worst dark gap 7450ms on {"phase":1500,"dur":3400,"gap":150,"count":12}`; counterexample test also red. 4/16 fail. |

**Adversarial max-gap with the fix in place: 3500ms** (= the floor, as designed), across all 576 schedules. Worst emission count: 18.

### Nits

- Removed the `// allow-raw-bot-api:` marker — correct, the guard's PATTERN never matches `sendChatAction` (chat actions take no `message_thread_id`). Replaced with a comment saying *why* it's routed through the retry module anyway (that's where `onFloodWait` lives).
- The 429 drop log now happens in `onFloodWait`, where the real `retry_after` is still in hand, and the give-up path no longer double-logs the opaque `max retries exceeded`. The loudest diagnostic on the largest emitter now names the code and the retry_after.

### Verification

- `vitest run telegram-plugin/tests` — **358 files / 5848 tests passed**
- `npm run lint` — clean (tsc + all 8 guards)

Auto-merge still off.

mekenthompson:
## Blocker fixed + rebased onto #3094 — CI is GREEN on `ef886b84`

### 1. `bun-test` blocker — the guard now bites in BOTH runners

Reviewer is right and the framing is right: under bun those five tests **errored on startup**, so the entire #3084 guard existed in one runner only. `vi.useFakeTimers()` / `vi.setSystemTime()` are gone from the file — no `vi` timer API anywhere.

Replaced with a **virtual clock + event queue**. The emitter already takes injected `now` / `schedule` / `cancel`, so it needs no timer mocking at all; `createTurnTypingLoop` reaches for the global `setInterval`, so `withVirtualTimers(clock, body)` swaps the global for the queue for the span of one simulation and restores it in a `finally`. Deterministic, runner-agnostic, and strictly better than fake timers here. Also added an assertion that **nothing fires in the 30 s after turn end** (no zombie typing), which the old harness couldn't express.

**Both runners, on the test file:**
- `vitest run telegram-plugin/tests/typing-emitter.test.ts` → **17/17 pass**
- `bun test telegram-plugin/tests/typing-emitter.test.ts` → **17 pass, 0 fail**

**Mutations, re-run in BOTH runners (post-rebase):**

| mutation | vitest | bun |
|---|---|---|
| `TYPING_FLOOR_MS = 0` | RED — 9 fail, `expected 136 to be less than or equal to 20` | RED — 9 fail, same assertion |
| catch-up removed | RED — 4 fail, `worst dark gap 7450ms on {"phase":1500,"dur":3400,"gap":150,"count":12}` | RED — 4 fail, same |

### 2. Rebased onto #3094 (`bbe14471`) — kept BOTH mechanisms

Conflicts in `gateway.ts` resolved deliberately:

- **`robustApiCall` keeps #3094's `floodWaitRemainingMs: makeFloodWaitProbe(...)` pre-call gate.** The block simply moves above the typing section, since typing now depends on it.
- **The emitter keeps its own `isSuppressed` gate.** It short-circuits *earlier*: a typing ping during a ban never reaches the retry layer, so it costs neither a flood-state read per restart nor a wasted call. Nothing was ripped out — defense in depth, in that order.

**Re-traced `nonEssentialApiCall` (`maxRetries: 1`, `sleep: noop`) on #3094's changed retry loop:**
- No `floodWaitRemainingMs` passed → `openWindowMs()` is 0 → no pre-call short-circuit; the emitter's gate is what does that job for typing.
- 429 → `onFloodWait` fires (window recorded) → `delayMs > maxFloodSleepMs` → throws `FLOOD_WAIT_ACTIVE`. **Exactly one `fn()` call, zero sleep, no hot loop.**
- Short 429 (under the 120 s ceiling) → `onFloodWait` → no-op sleep → loop exits → `max retries exceeded`. Still exactly one `fn()` call.
- The marker **cannot leak as an unhandled rejection**: the typing send's `onRejected` handler recognizes it via `isFloodWaitActiveError` and drops it quietly (it also stops double-logging a line that names neither the code nor the retry_after — the honest 429 line with `retry_after` is written in `onFloodWait`).
- New test pins this composition: `composes with #3094: a long ban surfaces FLOOD_WAIT_ACTIVE, still ONE call`.

### 3. CI — actual state, not local green

`gh pr checks 3097` on head `ef886b84`: **zero failing checks.** `mergeStateStatus: CLEAN`.

| check | state |
|---|---|
| `lint` / `lint-run` | pass |
| `bun-test` / `bun-test-run` | **pass** (was fail) |
| `vitest` + all 4 shards | pass |
| `images-ok` | pass |
| `uat-gate` | pass (heavy run path-skipped) |

Local, for the record: `vitest run telegram-plugin/tests` → 360 files / 5907 tests pass; the full bun CI list → 7379 tests, 0 fail.

Auto-merge NOT enabled.

PR #3100, opened 2026-07-11
feat(auth): Telegram-native OAuth loopback relay for Google/Microsoft account add (#2582)

### LINKED ISSUE #2582: Admin agents: deterministic Telegram-only mode + native OAuth relay (stop falsely asserting 'can't do on Telegram')
## Problem

Admin/root agents (e.g. `overlord`) too readily assert that an operator task **cannot be done over Telegram** and fall back to "run it at the keyboard / SSH port-forward" — **without first validating the claim**. This is a fabricated-boundary failure: the agent reports a limit it never tested.

### Concrete incident (2026-06-26)

Operator asked overlord to re-auth a Google account for `carrie` (`switchroom auth google account add pixsoul@gmail.com --replace --write`). overlord initially responded that the flow is browser/loopback-only and therefore **not doable from Telegram**, recommending the keyboard or an SSH port-forward.

That was wrong. After the operator pushed back ("doesn't it give a URL/code? this needs to be doable from telegram"), overlord actually tested it and a **Telegram-only relay worked end-to-end**:

1. Run the loopback flow **inside the agent container that already holds the Google grant** (carrie), not overlord. The OAuth client secret `google/client-secret` is an **admin-only vault key** and cannot be posture-minted to an agent via the approval card — so "run it in the agent that holds the grant" is the correct pattern, not granting it to the root agent.
2. The CLI prints the consent URL with `redirect_uri=http://127.0.0.1:<port>` and holds a loopback listener.
3. Relay the consent URL to the operator over Telegram. Operator approves on their phone; the redirect to `127.0.0.1:<port>` fails to load there, but the address bar holds `?code=...&state=...`.
4. Operator pastes that redirect URL back over Telegram.
5. Agent feeds it to the waiting listener (`curl` the loopback port inside the container). Token mints + registers.

Net: no SSH, no keyboard. The boundary the agent asserted did not exist.

## Why it matters

- Google retired device-code (`invalid_scope` for Drive) and OOB, leaving loopback as the only flow — which *looks* un-relayable but isn't. Any agent hitting this will repeat the false "can't be done on Telegram" assertion unless we make the capability deterministic.
- The grounding rule already says "don't assert a limit you haven't tested." This needs a **deterministic mechanism**, not just a prompt nudge.

### 1. First-class Telegram-native OAuth relay (deterministic)
Build the loopback-relay into the gateway as a real capability so it's not improvised per-incident:
- Detect a loopback OAuth flow (CLI emits a `127.0.0.1:<port>` consent URL).
- Surface the consent URL as a Telegram card/message.
- Capture the operator's pasted redirect URL, validate `state`, and deliver `code` to the waiting listener automatically.
- Run in the **agent container that holds the relevant grant**; never require minting an admin-only key to the root agent.

### 2. "Telegram-only mode" + capability validation
An explicit mode where an admin agent **attempts and validates** an operator task over Telegram before ever recommending keyboard/SSH:
- Before asserting "can't do X over Telegram," the agent must run a deterministic capability probe (e.g. actually start the flow and inspect what it emits) — not assume.
- If a true blocker exists (e.g. admin-only key, no grant), surface the *specific* blocker with evidence, plus the minimal operator action — rather than a blanket "do it at the keyboard."

## Reference: validated working mechanism
- Loopback flow run via `docker exec <agent-container> switchroom auth google account add ...` (agent holds the grant).
- Listener fed via `curl 'http://127.0.0.1:<port>/?state=...&code=...'` from inside that container.
- `google/client-secret` is admin-only (cannot be posture-minted to an agent) — confirmed by the failed approval-card mint on overlord.

## Acceptance
- An admin agent can complete a Google (and ideally Microsoft) account re-auth entirely over Telegram, deterministically, with no SSH/keyboard.
- The agent never asserts "not doable over Telegram" without a tested, specific blocker.

mekenthompson:
## Summary

Admin agents can now complete a Google or Microsoft account add/re-auth **entirely over Telegram** — no SSH port-forward, no keyboard. This makes the incident-validated 2026-06-26 relay a first-class, deterministic capability so agents stop falsely asserting "OAuth can't be done over Telegram."

Flow: the gateway runs the CLI loopback flow in the agent container that holds the grant, relays the consent URL to chat, the operator approves on their phone (the redirect to `127.0.0.1:<port>` fails to load — expected), then pastes the full address-bar URL (`?code=...&state=...`) back into chat. The gateway validates `state` and hands the code to the still-waiting loopback listener.

## In scope
- New gateway module `telegram-plugin/gateway/auth-loopback-relay.ts` (pending-flow map, TTL reaper reuse, paste intercept), modeled on `auth-add-flow.ts`.
- New chat verbs (extend `auth-command.ts`): `/auth google add <email> [--replace] [--write]`, `/auth microsoft add <email> [--replace] [--org-mode]`, and `/auth <provider> cancel`. Anthropic `/auth add` untouched.
- Gateway wiring: state-validated single-use paste-back intercept (code redacted from chat history via `redactAuthCodeMessage`), reaper sweep, shutdown cleanup that kills in-flight CLI children.
- Docs updated (`docs/auth.md`, `docs/google-workspace.md`, `docs/microsoft-workspace.md`) to document the Telegram-native path, replacing/augmenting the SSH-port-forward framing.

## Out of scope (deliberate, per spec)
- A **generic capability-probe registry** (part b's runtime mechanism) — greenfield, no primitive exists; the deterministic relay + docs satisfy the issue's acceptance. Noted here explicitly.
- Cross-container relay orchestration and any Anthropic `/auth add` changes.

## Design choice: CLI subprocess (piped stdout), not in-process `runLoopbackOAuth`
The gateway **can** import `src/` modules (`auth-add-flow.ts` already imports `src/auth/manager.js`), so an in-process drive was on the table. I chose to drive the CLI subprocess because:
- `runLoopbackOAuth` (`src/drive/oauth.ts:418`, `src/microsoft/oauth.ts:412`) only returns a `TokenResponse`. The large, security-sensitive **registration** surface (scope selection, vault client-secret resolution, `buildGoogleCredentials`, broker `addAccount`, `--replace` semantics) lives in `src/cli/auth-google.ts` / `auth-microsoft.ts`. Reproducing it in the gateway would duplicate that surface and drift from the CLI. Driving the CLI keeps it the single source of truth.
- Unlike `claude setup-token` (writes to `/dev/tty`, needs a tmux pty), the switchroom CLI prints the consent URL via `console.log` to stdout (`src/cli/drive.ts:357`). A plain piped child process captures it — **no tmux required**. This is exactly the incident-validated mechanism (run in the agent container that holds the grant, feed the listener a loopback GET).

## Accepted residual lows (deliberate)
Two narrow gaps are accepted rather than widened: (a) a code-bearing but malformed paste (one that fails `parseLoopbackRedirect`, e.g. mangled by a phone keyboard) falls through un-consumed and un-redacted — the gateway's secret-scrub hooks are the backstop there; (b) the same applies to a paste arriving after the flow's TTL has expired (stale path). Widening the consume-gate to catch these would reintroduce review finding 1 (consuming/deleting unrelated messages).

## Security invariants held
- `state` validated in-process before the loopback GET, AND re-validated by the CLI's own listener (defence in depth).
- The `code`/tokens are never logged or echoed; the pasted redirect is scrubbed (`scrubSensitive`) and redacted from chat history.
- Single-use: a submitting/completed flow rejects further pastes; TTL sweep drops abandoned flows and kills their child.
- No `claude -p`, no API keys.

## Test evidence
- New `telegram-plugin/tests/auth-loopback-relay.test.ts` — 25 tests, all green. Asserts outcomes: valid paste completes the mocked exchange (code reaches the right loopback port); wrong-`state` paste is rejected and the flow stays pending; second paste after completion is rejected (single-use); bad paste rejected without touching the listener; parser valid/invalid/missing-state/wrong-port/provider-error cases; verb parser + flag handling.
- Related suites re-run green: `auth-add-flow`, `auth-command-vernacular`, `secret-detect-oauth-code` (reaper-structure pin) — 140 tests passed.
- `npm run lint` clean (tsc + all 8 guards, including `check-bot-api-wrapping` — no new raw `bot.api` calls).

Local scope only; CI is the full-suite authority.

Closes #2582

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -302,6 +302,23 @@ it, and unblock the fleet — entirely from chat:
 
 `/auth cancel` aborts an in-flight `/auth add`.
 
+**Google / Microsoft accounts add over Telegram too (issue #2582).** The
+Drive/Graph loopback OAuth flow is fully relayable — no SSH port-forward,
+no keyboard:
+
+- `/auth google add <email> [--replace] [--write]`
+- `/auth microsoft add <email> [--replace] [--org-mode]`
+
+The agent runs the loopback flow in its own container, relays the consent
+URL to chat, and you approve on your phone. The redirect to `127.0.0.1`
+fails to load — expected — so you paste the full address-bar URL
+(`?code=...&state=...`) back into chat; the gateway validates `state` and
+hands the code to the waiting listener. The pasted URL is redacted from
+history on completion. `/auth google cancel` / `/auth microsoft cancel`
+abort. Run it against the agent that should hold the grant
+(`google/client-secret` is admin-only and can't be minted to the root
+agent).
+
 ### Full surface
 
 | Chat command | Equivalent CLI verb |
@@ -309,6 +326,8 @@ it, and unblock the fleet — entirely from chat:
 | `/auth show [<agent>]` | `switchroom auth show [<agent>]` |
 | `/auth list` | `switchroom auth list` |
 | `/auth add <label>` | `switchroom auth add <label> --from-oauth` (chat-native OAuth flow) |
+| `/auth google add <email>` | `switchroom auth google account add <email>` (Telegram-native loopback relay) |
+| `/auth microsoft add <email>` | `switchroom auth microsoft account add <email>` (Telegram-native loopback relay) |
 | `/auth cancel` | (chat-only: aborts an in-flight `/auth add`) |
 | `/auth use <label>` | `switchroom auth use <label>` |
 | `/auth rotate` | `switchroom auth rotate` |
--- a/docs/google-workspace.md
+++ b/docs/google-workspace.md
@@ -140,11 +140,14 @@ If you'd rather do it by hand (the wizard automates exactly this):
    > flows are dead ends for Drive: **device-code** returns
    > `invalid_scope` for Drive scopes (Google does not allow Drive on
    > device flow), and **OOB** was retired by Google in 2022. On a
-   > **headless server** you complete the single browser step over an
+   > **headless server** you can complete the single browser step over an
    > SSH port-forward — `switchroom auth google account add` prints the
    > exact URL and `localhost` port; you `ssh -L <port>:127.0.0.1:<port>`,
-   > open the URL, approve. Same Desktop+loopback shape the
-   > `examples/personal-google-workspace-mcp/` host MCP uses.
+   > open the URL, approve. Or skip SSH entirely and run it from Telegram
+   > with `/auth google add <email>` (issue #2582) — the agent relays the
+   > consent URL to chat and you paste the redirect back. Same
+   > Desktop+loopback shape the `examples/personal-google-workspace-mcp/`
+   > host MCP uses.
 2. **Vault the secrets** so they never land in YAML:
 
    ```bash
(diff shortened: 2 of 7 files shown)


mekenthompson:
Adversarial review complete (batch O6 coordinator). Round 1: 6 findings (1 med intercept false-positive deletion, 5 lows) — all fixed in c2d54df2 and verified by re-review. Round 2 surfaced one new low (child leak on attempts-exhausted) — fixed in 97dce2ff, diff verified directly: idempotent cancelLoopbackFlow before entry delete, module test pins the leak-then-kill outcome. Two residual lows accepted and documented in the PR body (secret-scrub hooks as backstop). Approving; auto-merge armed on CI green.

PR #3112, opened 2026-07-11
feat(telegram-plugin): send-gate observability + operator flood alerts — #3084 PR 3/3

mekenthompson:
## Summary
PR 3 of 3 for the Telegram flood-safety series (#3084), closing part3-design §6: **observability + operator alerting**. PRs 1 (#3092) and 2 (#3106) built the send gate — token buckets, priority shedding, degraded mode, restart-proof scoped flood windows — but a live ban was **silent** until a user noticed dead air. This PR surfaces it.

- **Periodic stats logging** (`createStatsLogger`): a low-frequency one-line summary of `sendGate.stats()` to the gateway-supervisor stderr log, emitted only when a counter changed **and** the interval (60s) elapsed — no log spam. Silent when the flag is off.
- **Flood-window snapshots + operator alert** (`createFloodWindowObserver`): a snapshot line on every window open/close, plus operator alerting for prolonged (>60s) windows.
  - Deliverability-aware, so the alert never fights the window it reports. **Today** a real 429 opens a coincident `global` window (the #3111 over-suppression), so the operator chat is always covered while a ban is active and every alert is **deferred**: the operator gets **one** "flood ban cleared, was banned from X to Y" card per incident, **coalescing all scopes** (global + chat + group) into a single card rather than 2–3 near-identical ones.
  - The **immediate-delivery** branch (a `chat:<other>` ban while the operator chat is clear alerts right away) is present and unit-tested but **forward-looking**: it only goes live once #3111 makes `onFloodWait` scope-precise. It is not claimed as live behavior today.
  - **At-most-once** anchored by a persisted `alertedAt` marker on the window record, so a restart mid-window never re-alerts.
  - Windows already open at boot log a single "loaded N open flood window(s)" line, not a per-scope OPENED snapshot.
- **`/status`**: a **Send gate** block (queued / shed / expired / fail-fast / dropped + open flood windows with clear-in seconds) when the flag is on; omitted entirely when off.
- **M1 doc nit fixed** (flood-circuit-breaker.ts): the 5-min conservative fail-safe window also fail-fasts CRITICAL sends (5min > the 60s `criticalFailFastMs` ceiling), so the comment no longer understates the blast radius as "a few suppressed cosmetic sends".

## Why
part3-design §6: "On an open flood window > 60s: one operator alert — currently a ban is silent until the user notices dead air." Advances the **always available** vision outcome.

## Test plan
- New `send-gate-observability.test.ts` — 17 tests, injectable-clock driven, outcome-asserting:
  - stats-line shape + rounding; `countersChanged`; logger only-when-changed + interval + flag-off no-op.
  - window open/close snapshot (post-boot); **boot-loaded windows log one "loaded" line, no OPENED spam**; observer flag-off no-op.
  - forward-looking immediate chat-scoped alert at-most-once (#3111); **deferred global alert sent on close**; **one incident (global+chat+group) coalesced into ONE cleared card listing every scope**; msg-edit scopes never alerted.
  - **restart safety**: `markFloodWindowAlerted` persists `alertedAt`, `readFloodWindows` + extend-merge preserve it, and a restart with `alertedAt` on disk does not re-alert.
  - `/status` renders the block when on, omits it when off.
- Scoped local: `send-gate-observability` (17) green. `npm run lint` clean (tsc + all guards, incl. `check-bot-api-wrapping`).

## Risk
Low. Everything is behind `SWITCHROOM_TELEGRAM_SEND_GATE` (default **off**): flag off ⇒ `stats().enabled` is false, both `tick()`s are no-ops, and the 15s timer isn't even scheduled. The operator alert is a `critical` send through `robustApiCall`. Deferred (filed, not forced into this PR): draft-stream edit-path wiring — #3110; scope-precise onFloodWait / global over-suppression — #3111 (which also makes the immediate-alert path live).

### DIFF
--- a/telegram-plugin/flood-circuit-breaker.ts
+++ b/telegram-plugin/flood-circuit-breaker.ts
@@ -380,6 +380,12 @@ export interface FloodWindowRecord {
   retryAfterSrc: string
   /** Epoch ms the window was (re)recorded. */
   observedAt: number
+  /**
+   * Epoch ms an operator alert was sent for THIS window (#3084 PR 3, §6). The
+   * at-most-once anchor: persisted so a restart mid-window never re-alerts.
+   * Unset until the observer alerts.
+   */
+  alertedAt?: number
 }
 
 /** Sibling file holding the scoped-window array. */
@@ -400,8 +406,17 @@ export function floodWindowsPath(stateDir: string): string {
  * conservative window — the exact opposite of `flood-wait.json`'s essential-send
  * probe, which must fail OPEN so a marker problem never gags the user's reply.
  * The asymmetry is deliberate: this file drives boot-time shedding, not the
- * user's reply, so failing safe here costs at most a few suppressed cosmetic
- * sends, while failing open resends straight into a ban (H2/M1, #3106 posture).
+ * user's reply, so failing safe here is the right call, while failing open
+ * resends straight into a ban (H2/M1, #3106 posture).
+ *
+ * Note the blast radius is NOT only cosmetic: this conservative window is
+ * `FLOOD_WINDOWS_CORRUPT_SUPPRESS_MS` (5 min), which exceeds the send gate's
+ * `criticalFailFastMs` (60s) ceiling — so while it is open a CRITICAL reply also
+ * fail-fasts with a structured `FLOOD_WAIT_ACTIVE` (a retryable error carrying
+ * `untilTs`, NOT a silent drop and NOT a hang) for up to 5 min after each boot.
+ * That only fires with the send gate flag ON and a genuinely corrupt/unreadable
+ * persisted file; the fail-fast is a real signal to the caller, so the posture
+ * is defensible — but it is fail-fast criticals, not merely suppressed cosmetics.
  */
 export const FLOOD_WINDOWS_CORRUPT_SUPPRESS_MS = 5 * 60_000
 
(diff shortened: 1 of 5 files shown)


mekenthompson:
Pushed fixes for the #3112 review (commit 680958bd):

- **M1** — coalesced flood close-alerts: all scopes of one 429 incident (global + chat + group, grouped by an ~untilTs bucket) now collapse into **one** operator "flood ban cleared" card listing every scope, instead of 2–3 near-identical cards. Kept the immediate-delivery branch but documented it (module header + inline + its test) as **forward-looking for #3111** — unreachable today because the recorder opens a coincident global window — so nothing overstates immediate delivery as live behavior. PR body updated to match.
- **L3** — pre-existing windows loaded at boot now log a single `loaded N open flood window(s) at boot` line on the first tick instead of a per-scope `OPENED` snapshot.
- **L1 / L2** — added code comments marking the send-first at-most-once ordering (L1) and the in-memory owed/pending-close state (L2) as deliberate documented tradeoffs; no behavior change.
- **Nit** — dropped the redundant `.replace(.'000','')` in `fmtTs`.

Tests: `send-gate-observability.test.ts` now 17 (was 15) — added a multi-scope-incident → exactly-one-card test and a boot-loaded → no-OPENED-spam test; the snapshot test now opens its window post-boot. Scoped vitest green, `npm run lint` clean.

PR #3159, opened 2026-07-12
feat(agents): foreground turn-hog PreToolUse gate for Bash

mekenthompson:
## Summary

Deterministic PreToolUse gate that DENIES effectively-unbounded **foreground** Bash commands with an instructive message: re-run the exact command with `run_in_background: true` (and act on the completion notification), or use a bounded alternative. Same family as the destructive-git detector (#3150): a conservative deny on known-bad shapes, not a heuristic classifier.

**What it gates** (foreground only — `run_in_background: true` always passes):
- `tail -f` / `tail -F` / `--follow` (incl. combined short flags like `-fn200`)
- `watch <cmd>`
- `sleep N` with N > 30s — bare or as a segment of a compound command (`npm run build && sleep 300`)
- sleep-in-a-loop (`while|until|for ... do ... sleep ... done`)
- `gh pr checks --watch`, `gh run watch`
- `docker logs -f/--follow`, `kubectl logs -f/--follow`, `journalctl -f` (incl. `-xef`)

**Deny message** (principles.md #1 — tells the model exactly what to do next, one line, no doc links):

> Foreground-hog gate: `tail -f/--follow` would hold the session foreground for its full duration. Re-run this exact command with `run_in_background: true` and act on the completion notification, or use a bounded alternative (e.g. `tail -n 200` instead of `tail -f`).

## Why conservative

False negatives are acceptable; false positives are not. Every rule is anchored on the command token of each foreground segment — generic build/test/deploy commands (`npm test`, `cargo build`, `vitest run`, `docker build && docker push`) are never matched, `--follow`/`-f` only counts for the four follower commands (so `git log --follow` passes), `watch` only as a command (so `watchman`, `stopwatch.ts`, quoted text pass), and shell-backgrounded segments (`tail -f x &`) pass because they don't hold the turn. Wrappers (`bash -c "..."`, `timeout 600 tail -f`) are deliberate false negatives, not parsed.

## Kill-switch (principles.md #2 — default ON, escape hatch not opt-in)

`SWITCHROOM_FOREGROUND_HOG_GATE=0|false|off|no` disables (trimmed, case-insensitive); unset or any other value keeps it enabled. Follows the repo's default-on kill-switch convention (`sendGateEnabledFromEnv`, #3153 shape).

## Sub-agent scoping decision

Main-session-only scoping is **not expressible at the hook layer**: Claude Code fires PreToolUse for sub-agent (Task) tool calls with the parent's session_id and no sidechain marker in the payload (documented at `telegram-plugin/hooks/tool-label-pretool.mjs` — "Claude Code does not pass sub-agent agent_id directly to the hook"), and sub-agents inherit the same settings.json hooks. Trade-off, made explicit: the gate applies inside workers too. The pattern list is therefore limited to shapes that are wrong even inside a worker — an unbounded follower, a `--watch` poll, or a > 30s dead foreground sleep should be backgrounded or bounded there as well (fleet dev protocol: no foreground watches over 30s, anywhere), and every gated shape has a trivial compliant alternative (`run_in_background: true` or a bounded flag).

## Registration (one more hook in the existing shape, no new mechanism)

- `src/cli/foreground-hog-pretool.ts` — the hook (deny = exit 0 + `{"decision":"block","reason"}`, fail-OPEN on protocol errors; exact template: `hindsight-mental-model-pretool`)
- `scripts/build.mjs` — bundled to `dist/cli/foreground-hog-pretool.mjs`
- `docker/Dockerfile.agent` — shipped to `/opt/switchroom/hooks/`
- `src/agents/scaffold.ts` — `switchroomPreToolUse` entry, matcher `^Bash$`, `run-hook.sh`-wrapped, timeout 10

## Test plan

- `tests/foreground-hog-pretool.test.ts` — full decision matrix (14 tests, spawn-driven against source like the hindsight hook test): every gated pattern → deny with the instructive message; near-misses (`tail -n 100`, `sleep 5`, `sleep 30`, `watchman`, `git log --follow`, `gh pr checks` without `--watch`, `docker logs --tail`, loop-without-sleep, `2>&1` fd redirects) → allow; shell-backgrounded segments → allow; `run_in_background: true` → allow; kill-switch off values → allow / junk values stay ON; non-Bash tools and malformed stdin → allow.
- `tests/scaffold.test.ts` — settings.json wiring (matcher, bundled path, run-hook.sh wrap, timeout).
- `tests/reconcile-hooks-drift.test.ts` — registration parity assertion.

Local: 14/14 new hook tests green; `tests/scaffold.test.ts` + `tests/reconcile-hooks-drift.test.ts` 224/224 green; `npm run lint` clean; `npm run build` bundles and the dist artifact was smoke-tested under `node` (deny + allow paths).

## Risk

Low. Fail-open on protocol errors; deny only on the explicit shape list; kill-switch requires no rebuild. Hook only fires on Bash (scoped matcher). Live in agents only after image rebuild + restart.

## Job spec

`reference/jobs/steer-or-queue-mid-flight.md` — a turn wedged behind an unbounded foreground command can't be steered, and the user can't tell why the agent went dark; also serves `feel-like-a-colleague`'s "no foreground watches over 30s" protocol by making it deterministic instead of prompt-dependent (dev protocol: deterministic mechanisms over model-dependent behavior).

### DIFF
--- a/docker/Dockerfile.agent
+++ b/docker/Dockerfile.agent
@@ -244,6 +244,17 @@ RUN chmod 0755 /opt/switchroom/hooks/skill-validate-pretool.mjs
 COPY dist/cli/hindsight-mental-model-pretool.mjs /opt/switchroom/hooks/hindsight-mental-model-pretool.mjs
 RUN chmod 0755 /opt/switchroom/hooks/hindsight-mental-model-pretool.mjs
 
+# Foreground turn-hog gate. A PreToolUse hook scoped to `^Bash$` that DENIES
+# effectively-unbounded FOREGROUND command shapes (tail -f, watch, sleep > 30s,
+# sleep-loops, gh --watch, docker/kubectl/journalctl followers) with an
+# instructive "re-run with run_in_background: true" message, so the turn is
+# never wedged behind an open-ended command. Backgrounded calls
+# (run_in_background: true) always pass. Not in the security-plugin: it's
+# turn-hygiene (responsiveness), not a load-bearing safety control. Kill
+# switch SWITCHROOM_FOREGROUND_HOG_GATE=0.
+COPY dist/cli/foreground-hog-pretool.mjs /opt/switchroom/hooks/foreground-hog-pretool.mjs
+RUN chmod 0755 /opt/switchroom/hooks/foreground-hog-pretool.mjs
+
 # Agent self-improvement Stop gate (RFC reference/rfcs/agent-self-improvement.md
 # slice 1). A cheap deterministic per-turn triage; on a learning signal
 # it injects a forked review turn via the gateway socket (inject_inbound,
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -308,6 +308,24 @@ if (mmHookEscape.changed) {
   console.log(`[build] ASCII-escaped ${mmHookEscape.nonAsciiCount} non-ASCII code units in dist/cli/hindsight-mental-model-pretool.mjs`);
 }
 
+// Bundle the foreground-hog-pretool hook — deterministic gate that denies
+// effectively-unbounded FOREGROUND Bash shapes (tail -f, watch, long sleeps,
+// sleep-loops, gh --watch, docker/kubectl/journalctl followers) with an
+// instructive "re-run with run_in_background: true" message. Same pattern
+// as hindsight-mental-model-pretool: no src/ imports (a pure tool_input
+// gate), bundled for consistency so it ships to /opt/switchroom/hooks/.
+console.log("[build] bundling src/cli/foreground-hog-pretool.ts -> dist/cli/foreground-hog-pretool.mjs");
+execSync(
+  `bun build ${JSON.stringify(resolve(root, "src/cli/foreground-hog-pretool.ts"))} --outfile ${JSON.stringify(resolve(outDir, "foreground-hog-pretool.mjs"))} --target node`,
+  { stdio: "inherit", cwd: root }
+);
+const fgHogHookOutFile = resolve(outDir, "foreground-hog-pretool.mjs");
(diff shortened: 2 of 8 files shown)


mekenthompson:
Review fixes pushed in caf3f5ab — all four FP classes were one root cause (quote/heredoc-blind segmentation) and are fixed by a single `blankQuotedAndHeredocs()` pre-pass:

- Length-preserving blanking of `'…'` / `"…"` spans (including embedded newlines, so multi-line commit bodies can't create phantom segments) and heredoc bodies (`<<TAG` / `<<'TAG'` / `<<"TAG"` / `<<-TAG`, terminator-line aware, multiple heredocs per line; `<<<` here-strings skipped) runs BEFORE segmentation and LOOP_SLEEP.
- gh matcher now keys on unquoted positional tokens only (`positionals[0]==='pr' && positionals[1]==='checks' && --watch`, `run watch`); flag values are data.
- Fail-open preserved and extended: unterminated quote (prose apostrophe), unterminated heredoc, or un-parseable heredoc tag → ALLOW. Known accepted false negatives documented in tests (e.g. `gh --repo o/r run watch`).
- Every review repro pinned as an allow-case (FP-1..FP-4 + fail-open cases); true-positive corpus re-pinned around quotes/heredocs; original 14 matrix cases untouched and green (20/20).
- Review LOW addressed: `tests/bundled-hooks-parity.test.ts` derives the bundled-hook list from `buildSettingsHooksBlock` and asserts build.mjs + Dockerfile.agent parity for each (pre-existing gap, all 8 current bundled hooks covered).

Local: 20/20 hook matrix, 3/3 parity, 224/224 scaffold+drift, lint clean, dist artifact smoke-tested under node (FP-1/FP-2 allow, tail -f deny).

PR #3169, opened 2026-07-12
fix(scaffold): agent dotfiles written root-owned on rollout — permission-card storm on every update

mekenthompson:
Fixes #3168.

## Summary

Every update rollout left every agent's `.claude/settings.json` root:root mode 0600 — unreadable by the agent's own container uid — so Claude Code silently loaded NO permission allowlist and every tool call threw an operator approval card. 2026-07-12 incident: all 12 fleet agents wedged simultaneously on the v0.18.13 → v0.18.14 roll, operator Telegram flooded with cards for 40+ minutes, manual host-side `chown` as stopgap. Recurs on every update.

## Root cause

- hostd spawns `switchroom agent restart <a> --wait --force` **as root** for every agent on every roll (`src/cli/rollout.ts:681`; root requirement at `rollout.ts:42`; hostd is `USER 0:0`, `docker/Dockerfile.hostd:124`).
- Restart reconciles first (`reconcileAndRestartAgent` → `reconcileAgent`, `src/cli/agent.ts:614`).
- `reconcileAgent` rewrites settings.json via `atomicWriteFileSync(settingsPath, after, 0o600)` (`src/agents/scaffold.ts:6576` pre-fix) — tmp + fsync + rename(2). **The rename replaces the inode, so the file's owner becomes the writer's euid: root.** Same for `.mcp.json` (`scaffold.ts:6790`).
- The apply path has always been protected by the `alignAgentUid` chown sweep (`src/cli/apply.ts:1501`); the reconcile path — the one every rollout takes — had no ownership alignment at all. In-place `writeFileSync` sites (start.sh, CLAUDE.md) truncate the existing inode and preserve ownership, which is why only the atomically-rewritten files flipped.

## Same-class audit (whole scaffold/reconcile write surface)

| Path | Writer | Pre-fix state after root reconcile | Symptom |
|---|---|---|---|
| `.claude/settings.json` | atomic rewrite (new inode) | root:root 0600 | **allowlist gone → card storm** |
| `.mcp.json` | atomic rewrite (new inode) | root:root 0600 | MCP servers silently dropped |
| `.claude-cron/.mcp.json` | first-write 0600 | root:root 0600 | cron session loses MCP |
| `.claude/agents/<sa>.md` (new subagent) | first-write | root:root 0644 | read OK, agent can't update |
| `.resume-mode-migration-warned`, workspace migration files | first-write | root:root 0644 | agent can't rewrite |
| hindsight plugin dir | rm + recopy (new inodes) | root:root 0644/0755 | read OK, write-blocked |
| `start.sh`, `CLAUDE.md` | in-place truncate | owner preserved | none (why the incident was settings-only) |

All classes are now covered by one mechanism rather than per-call-site patches.

## Fix shape (deterministic, not model/discipline-dependent)

- **`src/agents/agent-owned-tree.ts`** (new): `alignAgentTreeOwnershipIfRoot` — end-of-reconcile ownership sweep, `chown -R <uid>:<uid>` over the agent tree whenever `reconcileAgent` runs as root (the reconcile twin of apply's `alignAgentUid`; same symlink-safe `chown -R` semantics, so bundled-skills pool symlinks are never re-owned). Throws loudly if the chown fails.
- **Post-sweep assertion** (`findAgentUnreadablePaths`, wired into `reconcileAgent`): after the sweep, every file the reconcile touched (plus settings.json/.mcp.json/start.sh always) must be readable by the agent uid, or the reconcile **fails loudly** — a future write-after-sweep or broken sweep fails the rollout instead of silently storming the operator. This is the deterministic guard for the recurrence.
- **`src/agents/agent-uid.ts`** (new): `allocateAgentUid` extracted from compose.ts (re-exported there for the existing broker/kernel/apply/doctor importers) so scaffold.ts can derive the uid without a compose ↔ scaffold import cycle.
- **Guidance**: CLAUDE.md dev-flow rule (root-written agent-home files MUST land agent-owned + regression-test pinned; tmp+rename re-owns the inode) and a one-line ownership-invariant note in `docs/architecture.md`. Audited docs/, skills/, scripts/, reference/ for guidance instructing root writes into agent homes without chown — none found; existing auth/vault docs already document chown-to-agent-uid correctly.

hostd's container retains CAP_CHOWN/DAC_OVERRIDE (compose cap set) and is Debian-based, so the sweep works exactly where rollout reconciles run. In-container reconciles (agent-config bridge) run as the agent uid → sweep no-ops. Operator `sudo` reconciles → covered.

## Test plan

`tests/reconcile.agent-ownership.test.ts` (13 tests, all passing; would fail pre-fix):
- root reconcile chowns the tree to `allocateAgentUid(name)` (recorder seam; fails pre-fix — no sweep existed)
- unprivileged reconcile (in-container agent uid) never chowns
- failed root chown fails the reconcile with an actionable message
- **pre-fix incident shape** (root writer, ownership never restored, 0600 settings.json) now makes `reconcileAgent` throw `agent-unreadable files` instead of succeeding
- root-gated real-outcome pin (`it.runIf(getuid()===0)`): after a root scaffold (settings.json verifiably root-owned) + reconcile, `statSync` uid/gid of settings.json, .mcp.json, start.sh, .claude/ equal the agent uid — executed for real in the dev sandbox (root), self-skips on non-root CI runners
- unit coverage for `findAgentUnreadablePaths` (0600 foreign file flagged, 0644 accepted, agent-owned 0600 accepted, 0700 dir flagged, symlinks/missing skipped) and the `allocateAgentUid` extraction parity

Local: `tsc --noEmit` + full `npm run lint` clean; scoped vitest green — new suite (13), scaffold/reconcile suites (54), compose-generator + workspace.git (208), apply/agent-reconcile/restart-dirty (46), docker doctor/isolation (19). `npm run build` bundles clean.

## Risk

- The sweep runs `chown -R` on every root reconcile — idempotent and identical in scope to what `alignAgentUid` already does on every apply.
- Fail-loud assertion could abort a reconcile that previously "succeeded" — that reconcile was producing a wedged agent; aborting before restart is strictly better and names the offending paths.

Job spec: `reference/jobs/survive-reboots-and-real-life.md` — agents must come back cleanly after restarts/rollouts; a rollout that wedges all 12 agents on approval cards is that job failing.


mekenthompson:
Review findings addressed in c652673c:

**F1 (error-path ownership backstop) — FIXED.** `reconcileAgent` is now a thin wrapper around `reconcileAgentInner`; on ANY throw it re-runs `alignAgentTreeOwnershipIfRoot` best-effort (backstop's own failure logged + swallowed, so the original error is never masked). Fail-loud readability assertion remains success-path only. New tests force a throw AFTER the atomic rewrites (file planted at `workspace/` → `mkdirSync` EEXIST) and assert: sweep runs on the error path with the right uid/tree, original error surfaces, and — root-gated real outcome — settings.json/.mcp.json are agent-owned on disk after the failed reconcile. A "backstop itself fails" test pins the no-masking property.

**F2 (busybox chown dereference) — FIXED.** `chownTree` now passes `-h` (`--no-dereference`); comment corrected (GNU -R defaults to -P, busybox has no -H/-L/-P and dereferences without -h). Empirically re-verified on GNU coreutils 9.7: `-h -R` re-owns files and link inodes, leaves the link target outside the tree untouched. Same `-h` applied to `alignAgentUid`'s two pre-existing shell-outs (plain + sudo) which had the identical exposure, and to its error-message hint.

Local: `tsc --noEmit` + full `npm run lint` clean; ownership suite 16/16 (root sandbox executes the real-chown pins), scoped scaffold/reconcile/apply/compose suites 280 passed.

PR #3173, opened 2026-07-12
fix(telegram): wire draft-stream edits through the send gate's per-message edit floor

mekenthompson:
## Summary

`stream-controller.ts` drove rapid same-message `editMessageText` through its own 400 ms DM / 1000 ms group throttle and called `robustApiCall` with only `{ threadId, chat_id }` — omitting `messageId` / `editPayload` / `priorityClass` — so the send gate treated the draft/answer stream as ordinary sends and its per-message edit floor (>=1.5 s), last-write-wins coalescing, no-op skip, and cosmetic shedding never applied. part3-design §4 names rapid same-message edits as the #1 documented flood-ban trigger, and this exact bypass produced a production 429 with `retry_after=21397s` (~6 h) on 2026-07-12: draft edits 472 ms apart, consistent with the 400 ms DM throttle (evidence on #3110).

Fixes #3110 (carried follow-up from #3084 PR 2 review L3).

## What changed

- **`stream-controller.ts`** — `RetryPolicy` opts widened to `SendGateOpts & { threadId?: number }` (structural superset, assignable to `RetryCallOpts`; type-only import). All four edit call sites (anchor edit, anchor plain parse-fallback, tail edit, tail plain parse-fallback) now pass `messageId` + `editPayload` (the exact wire payload, rich wrapper included) + `priorityClass`. The local throttle stays as a cheap pre-filter; the gate is the authority.
- **Priority classes** — intermediate draft edits are `cosmetic` (part3-design §2 lists "stream updates" there): shed under pressure / an open flood window; the next flush carries full state. The finalize flush is `critical`, mirroring the reply path's preview-finalize convention (`gateway.ts` `editPreview`): never shed, waits out a short window, fails fast with a structured `FLOOD_WAIT_ACTIVE` on a long one. Sends stay untagged (gate admits untagged non-edit sends as `critical`) — a shed send resolves `undefined` and would break message-id capture.
- **Shed honesty (reworked per review F1–F3)** — the gate's shed paths now resolve a distinguishable exported `SEND_GATE_SHED` sentinel (`Symbol.for`-keyed) instead of `undefined`, which was overloaded (gate no-op drop and robustApiCall's swallowed benign 400s also resolved `undefined`). stream-controller keys shed detection off `isSendGateShed()` exactly; benign `undefined` resolutions are treated as delivered. The anchor's shed outcome is decided AFTER the tail-upsert loop (a benign or shed anchor never starves tails), and ANY shed piece marks the flush not-delivered. draft-stream recognizes the property-tagged `DraftEditShedError`, preserves the shed snapshot in `shedText` (not `pendingText` — avoids a flushLoop busy-spin against an open window), and `finalize()` — with or without `finalText` — re-flushes it as the stream's final critical state, so argument-less cleanup finalizes (disconnect-flush, turn-end, PTY previews) no longer lose content.
- **`draft-stream.ts`** — `finalize(finalText?)`: an optional final snapshot that supersedes any pending draft (last-write-wins) and flushes with `final` already set. Needed because `stream_reply done=true` used to flush the final text inside `update()` — i.e. before the stream was final — which would have classified the completed answer as a sheddable draft.
- **`stream-reply-handler.ts`** — `done=true` routes the text through `stream.finalize(effectiveText)` instead of `update()` + `finalize()`. Side effect: the final flush no longer waits out the local throttle window.
- **`tests/bot-api.harness.ts`** — mock `editMessageText` now resolves `true` (faithful to grammy's `Message | true`; `undefined` from the production retry stack now specifically means gate-shed/skip).

## Test plan

New `telegram-plugin/tests/stream-controller-send-gate.test.ts` — `createStreamController` driven through a REAL `createSendGate` on the same fake clock as `send-gate.test.ts` (vitest fake timers restricted to `setTimeout`/`clearTimeout` for the local throttle; the gate clock advanced explicitly):

- draft edits inside the floor never hit the API; intermediates collapse; latest snapshot lands floor-paced (wire spacing 0/1500/3000 ms)
- the floor is per-message (stream B not delayed by stream A's floor)
- gate no-op skip engages for stream edits (second surface repeating the on-screen payload → zero API calls)
- open flood window: drafts shed with ZERO API calls; full state lands after close
- a shed draft is NOT recorded as delivered — a later finalize of the SAME text still lands
- finalize is critical: not shed; waits out a short (<= fail-fast ceiling) window then lands; fails fast (structured, logged, `failedFast` counted) on a long window with zero API calls and no hang
- regression pins: every edit passes `messageId`/`editPayload`/`priorityClass` through the retry policy (anchor, finalize, and the plain parse-fallback), sends stay untagged

Plus draft-stream unit pins (`finalize(text)` flushes final-tagged, supersedes pending, still dedupes text that actually landed) and a handler-level pin (`done=true` edit reaches `robustApiCall` as `critical`, intermediates `cosmetic`).

**Sabotage-verified**: with the gate fields stripped from stream-controller, 8/9 of the new integration tests fail.

Local: 302 tests across the 20 files touching send-gate/stream/draft/handler modules pass under vitest; `bun test telegram-plugin/tests/draft-stream.test.ts` (dual-run file) passes; `npm run lint` fully green (tsc + all 8 guards).

## Risk

Medium-low, flag-gated like the rest of #3084: with `SWITCHROOM_TELEGRAM_SEND_GATE=0` the gate is a pure passthrough and the only behavior deltas are (a) `done=true` flushes via `finalize(text)` (fewer edits, no throttle wait — same final state) and (b) the harness mock return value. With the gate on, draft edits now pace at >=1.5 s/message instead of 400 ms/1000 ms — slightly less "live" draft streaming, which is exactly the design's intent (the 400 ms cadence is the documented ban trigger).

Known missed benefit (review F5, documented in-code): draft-stream serializes flushes, so the gate's last-write-wins coalescing engages only for CONCURRENT writers to one message; a stale draft sleeping on the floor lands before the final, and finalize can trail by up to ~2x floor. Correctness unaffected. Known limitation (pre-existing, unchanged): draft-stream swallows edit failures, so a `stream_reply done=true` during a >60 s ban still returns `finalized` while the fail-fast is only logged — surfacing that as a structured MCP error is a candidate follow-up. Also deferred: the post-flood cooldown multiplier suggested in the #3110 evidence comment.

Job spec: `reference/jobs/talk-to-agents-from-anywhere.md` — advances the **always-available** outcome; outbound edit pacing is what keeps an agent from flood-banning itself off Telegram for hours.

### DIFF
--- a/telegram-plugin/draft-stream.ts
+++ b/telegram-plugin/draft-stream.ts
@@ -26,6 +26,40 @@
 
 const TELEGRAM_MAX_CHARS = 32768
 
+/**
+ * Error a transport layer (stream-controller.ts) throws from its edit
+ * callback when the send gate SHED the edit — it did NOT land, and the
+ * stream must not record the snapshot as on-screen (#3110). Marked with a
+ * property (not a message pattern) so the classification is exact.
+ *
+ * `flush()` recognizes this and PRESERVES the snapshot in `shedText` so a
+ * later `finalize()` — including the argument-less finalize the gateway's
+ * cleanup paths use — re-flushes it as the stream's final state instead of
+ * silently losing the content (review F2). It deliberately does NOT restore
+ * `pendingText`: `flushLoop` drains while `pendingText != null`, and a
+ * restore there would busy-spin the loop against an open flood window.
+ */
+export interface DraftEditShedError extends Error {
+  draftEditShed: true
+}
+
+/** Build the marker error a transport throws for a gate-shed edit. */
+export function makeDraftEditShedError(messageId: number | null): DraftEditShedError {
+  return Object.assign(
+    new Error(
+      `draft edit shed by send gate (id=${messageId ?? 'unknown'}); snapshot preserved for re-flush`,
+    ),
+    { draftEditShed: true as const },
+  )
+}
+
+/** True when `err` is the shed marker thrown by a stream transport. */
+export function isDraftEditShedError(err: unknown): err is DraftEditShedError {
+  return (
+    err instanceof Error && (err as Partial<DraftEditShedError>).draftEditShed === true
+  )
+}
+
 // Throttle defaults for the in-place engine.
 //   DM chats: 400 ms — slightly more responsive than groups while staying
 //     well under Telegram's practical ~1 edit/sec/message ceiling. This
@@ -172,6 +206,14 @@ export function createDraftStream(
   let messageId: number | null = config.initialMessageId ?? null
   let pendingText: string | null = null
   let lastSentText: string | null = null
+  /**
+   * Snapshot of the newest text the transport reported as SHED (thrown
+   * `DraftEditShedError`) — content that never landed. Cleared on any
+   * successful flush (a newer snapshot superseded it) and consumed by
+   * `finalize()` so the stream's last state is re-delivered once the
+   * pressure clears instead of being lost (review F2).
+   */
+  let shedText: string | null = null
   let lastSentAt = 0
   let inFlight: Promise<void> | null = null
   // Observability — per-stream fire counters for the stream-end trace.
@@ -225,11 +267,21 @@ export function createDraftStream(
       await sendViaMessage(textToSend)
       lastSentText = textToSend
       lastSentAt = Date.now()
+      shedText = null // a newer snapshot landed — drop any older shed one
     } catch (err) {
       const msg = (err as Error).message ?? String(err)
-      if (/\bmessage is not modified\b/i.test(msg)) {
+      if (isDraftEditShedError(err)) {
+        // #3110 review F2: the send gate shed this edit — it did NOT land.
+        // Preserve the snapshot for finalize()'s re-flush (as the stream's
+        // final state, sent critical) instead of silently losing it. Do NOT
+        // restore pendingText: flushLoop drains while pendingText != null
+        // and would busy-spin against an open flood window.
+        shedText = textToSend
+        log?.(`stream → shed by send gate (id: ${messageId}); snapshot preserved for re-flush`)
+      } else if (/\bmessage is not modified\b/i.test(msg)) {
         lastSentText = textToSend
         lastSentAt = Date.now()
+        shedText = null // on-screen text == this snapshot; nothing to recover
         log?.(`stream → not modified (id: ${messageId})`)
       } else if (
         /\bmessage to edit not found\b/i.test(msg)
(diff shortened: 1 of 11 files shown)


mekenthompson:
## Review fixes — all 5 findings addressed in 063e8ba5

Root cause accepted and fixed as suggested: `undefined` at the robustApiCall seam was overloaded (gate shed / gate no-op drop at send-gate.ts / retry-api-call's swallowed "message is not modified"). The gate's two shed paths now resolve an exported **`SEND_GATE_SHED`** sentinel (`Symbol.for`-keyed so duplicated module instances agree on identity); shed detection keys off `isSendGateShed()` exactly. Blast radius audited: every production consumer of a cosmetic call's result discards it (reactions seam, typing, boot/issues-card edits, pending-progress via swallowingApiCall, answer-stream, worker-feed); three gate tests asserting shed-resolves-undefined updated to the sentinel.

Per finding:

- **F1 (HIGH)** — fixed twice over: benign `undefined` resolutions are now treated as delivered (onEdit fires, no shed line), and the anchor's shed outcome is decided AFTER the tail-upsert loop so even a genuinely shed anchor can't starve the tails. Your probe is pinned as a test: `('a_b_c_d_e ').repeat(3000)` → 8 plain pieces, tail-only growth across three flushes exercising the not-modified swallow (flush 2) and the gate no-op drop (flush 3) — tails land on every flush, zero shed/failure log lines, and a same-text finalize dedupes with zero further API calls.
- **F2 (MED)** — draft-stream recognizes the property-tagged `DraftEditShedError` (factory + guard exported from draft-stream.ts) and preserves the shed snapshot in a new `shedText` field — deliberately NOT `pendingText`, which would busy-spin `flushLoop` against an open flood window. `finalize()` (argument-less included — the disconnect-flush/turn-end/PTY cleanup path) re-flushes it as the final critical state. A newer landed flush clears it, so a stale shed snapshot is never resurrected (unit-pinned both ways in draft-stream.test.ts).
- **F3 (MED)** — `upsertTail` reports shed per piece; ANY shed piece (anchor or tail) throws the marker so draft-stream never records the body as delivered — the PR-body claim you falsified is corrected there too. Recovery flush costs nothing for landed pieces (gate no-op skip + `tailLastText` short-circuit). The previously-untested tail call sites now have integration coverage: the F1 repro above plus a scoped `msg-edit:` window shedding ONLY the last tail, recovered by an argument-less `finalize()`.
- **F4 (LOW)** — the no-op-skip test now asserts the benign mechanism: 1 API call total, `dropped=1`, `onEdit` fired, zero shed/failure logs, and a third identical update dedupes inside draft-stream without consulting the gate (`dropped` stays 1).
- **F5 (LOW)** — documented in the send-gate wiring comment block in stream-controller.ts: draft-stream serializes flushes, so gate coalescing engages only for concurrent writers to one message; a stale draft on the floor lands before the final and finalize can trail by up to ~2x floor. Correctness unaffected; noted in the PR body as well.

Verification: 306 tests across the 20 send-gate/stream/draft/handler files pass; bun dual-run draft-stream suite passes; `npm run lint` fully green (including the plugin-strict `check-plugin-references` gate). Sabotage checks: re-overloading `undefined` into `isSendGateShed` → 13 failures; removing the `shedText` preservation → 13 failures.

PR #3191, opened 2026-07-12
feat(telegram): premium-model tier-downgrade failover when walled fleet-wide

mekenthompson:
## Summary

Adds a **second, lower-priority recovery tier** to the fleet-fallback machinery. When a user's selected premium model (e.g. `/model fable`) is overloaded / throttled mid-turn (HTTP 529 / 429 / `overloaded_error` / 503) **and** account-level failover finds no other account still serving that model, the live turn no longer stalls: the agent **downgrades to its configured default model** (fleet default `opus`) and resumes so work continues.

Account-level failover (swap OAuth account, same model) already exists and is untouched — this adds a strictly lower-priority tier that only fires when account-swap is exhausted.

Cites job spec: `reference/jobs/steer-or-queue-mid-flight.md` (via `reference/rfcs/session-model-stickiness.md` §0.1 rev 4), advancing the **always-available** outcome.

## What actually happens (no auto-return to the premium model)

A `/model fable` override is **session-scoped and in-memory only** — `recordTypedModelSwitch` writes **no** `.session-model` carrier, so the override dies on the downgrade SIGTERM. The downgrade writes a consume-once carrier for the **configured default** (`opus`), so:

- the resume boot runs `opus`, replays the interrupted turn, and keeps working;
- every subsequent restart also boots `opus`;
- **the premium model is never restored automatically.** The user must re-issue `/model fable` once it frees up.

The user-facing notice says exactly that:

> ⤵️ **Downgrading model to keep going** on agent **klanker**
> `fable` is overloaded across every account right now, so this turn is resuming on the default `opus` to keep working. Re-issue `/model fable` once it frees up — it won't switch back on its own.

(An earlier revision of this PR claimed the agent "reverts to `fable` on the next restart" — that was false and impossible under the session-scoped `/model` rule; it has been corrected in code, comments, and this description. `renderTierDowngradeNotice` is a pure function pinned by a test that fails on any revert-to-premium wording.)

## Design

- **Precedence A — account-swap first.** The downgrade is consulted *only* from the `all-blocked` branch of `doFireFleetAutoFallback`, i.e. after `runFleetAutoFallback` returned no eligible account. A `switched` outcome never reaches it, so a model merely throttled on **one** account is always recovered by account-swap, never a downgrade.
- **Consume-once carrier for the configured default.** `maybeTierDowngrade` writes a consume-once `.session-model` carrier for the configured default; start.sh applies + deletes it on the resume boot. General by construction — **any** premium override → the agent's default, not a hardcoded `fable→opus`.
- **Effort is native.** No `.session-effort` carrier is written, so the self-restart **sheds** any live `/effort` override and the downgraded default boots at the configured `thinking_effort` (the fleet `low` pin, #1978 / `src/config/thinking-effort-risk.ts`).
- **Loop bound is the natural on-default guard.** After the downgrade boot the session runs the configured default (the override is gone), so a re-entry returns `skip` ('on-default') and never re-downgrades — even if the default is itself walled fleet-wide (then the normal all-blocked card fires; no loop). The ceremonial restart-surviving counter from the first revision was **removed**: it was unreachable in the normal single-turn flow and advertised a "name-as-lost after 2nd failure" UX that never fired. Pacing reuses the `fleetFallbackResumeGate` single-flight + 3h staleness.
- **Concurrent-turn race is silent, not contradictory.** If a first turn already armed a resume restart (a downgrade or an account-swap), a second turn that also hits `all-blocked` does **not** broadcast a give-up card — the armed restart replays the interrupted turn. `planTierDowngrade` maps a `skip-inflight` resume-gate verdict to `suppress`.
- **Latch armed only after the carrier write.** The resume gate is split into `peek()` / `arm()`: `maybeTierDowngrade` peeks (no arm), writes the carrier, and arms only on write success — a throwing write can no longer leave an armed latch with no pending restart (which would suppress a legitimate later account-swap resume for the single-flight window).

## Test plan

Scoped vitest only (no full suite run):

- `telegram-plugin/tests/tier-downgrade.test.ts` — `decideTierDowngrade` (skip on-default / unresolved / downgrade-to-default / general target); `renderTierDowngradeNotice` asserts resume-on-default + manual re-issue and **no** revert-to-premium promise; `planTierDowngrade` routing incl. the LOW-9a `suppress` on `skip-inflight`.
- `telegram-plugin/tests/fleet-fallback-resume.test.ts` — new `peek()`/`arm()` seams: `peek` never arms, `arm` commits the single-flight window, `decide` == peek+arm-on-resume.
- `tests/scaffold.session-model.test.ts` — rendered start.sh outcomes: the consume-once carrier applies + reverts to the configured default; a downgrade boot with no effort carrier resolves `--effort low`.

`npm run lint` — clean (tsc + all 8 guards).

## Risk

Behavior change is confined to the `all-blocked` fleet-fallback branch (previously a give-up card). Deterministic and loop-bounded by the natural on-default guard; no new raw `bot.api` calls (notices use `sendRichMessage`). Does not touch account-level failover.

---

## Follow-up (commit `81c6cf18`) — "premium model recovered" ping

The downgrade above strands the user on the default with no signal for when
their premium selection frees up. This follow-up adds a one-tap heads-up.

- **Durable marker.** `maybeTierDowngrade` persists a consume-once
  `.premium-recovery` marker (shape-gated, parity with the `.session-model` /
  `.session-effort` carriers) recording the **dropped premium token** + the
  chats to notify. `start.sh` never consumes this name, so it survives the
  downgrade self-restart.
- **Deterministic recovery signal — no new poller, no broker change.** The
  signal is the auth-broker's **own per-account eligibility**, read off the
  `list-state` the gateway's existing `runQuotaWatch` tick already polls
  (~15 min). Recovery = at least one account **neither `exhausted` nor
  `premium_walled`** — the exact complement of the `all-blocked` condition that
  fired the downgrade. Both fields are the broker's live-authoritative verdicts
  (`isAccountExhausted` / `isAccountPremiumWalled` → `isModelTierWalled`, a pure
  `premium_walled_until > now` timestamp compare) — **P0 deterministic, never
  model-decided.**
- **Exactly one ping, one button, one apply.** On recovery, fire ONE ping
  (`✅ \`fable\` is available again — tap to switch back…`) with an inline
  **"Switch to fable"** button. Its `callback_data` is `mdl:alias:<premium>`,
  which routes through the **same session-scoped `/model` apply path** as the
  model menu (`handleModelMenuCallback` → `recordModelMenuSideEffects`, PR #3184)
  — it does **not** bypass it. At-most-once: fleet-wide `claim-notification`
  dedup + marker cleared **before** the send (never-storm).
- **No stale ping.** The marker is cleared the instant the user **manually**
  re-issues `/model <premium>` (both `/model` record chokepoints).

### Review close-out (@overlord adversarial review)

- **[MEDIUM] "reverts to fable" misinform** — resolved in `5d9bbf44` (honest
  `renderTierDowngradeNotice`). Confirmed on tip.
- **[LOW] attempts-counter pollution** — the counter was removed entirely in
  `5d9bbf44`; moot (no counter remains).
- **[LOW] `maybeTierDowngrade` glue untested** — **fixed in `81c6cf18`.**
  Extracted the order-sensitive glue into `runTierDowngrade`
  (`tier-downgrade-wiring.ts`, injected-deps runner); the new
  `tier-downgrade-wiring.test.ts` **pins the broadcast notice** (asserts
  resume-on-default + manual re-issue, fails on any revert-to-premium phrasing)
  and covers the `skip-inflight` suppress path + carrier-before-arm ordering.

### Added tests

- `telegram-plugin/tests/premium-recovery.test.ts` — `decidePremiumRecovery`
  (deterministic predicate), `renderPremiumRecoveryPing` (honest wording),
  `premiumRecoveryClaimKey`, `.premium-recovery` carrier round-trip + shape gate.
- `telegram-plugin/tests/tier-downgrade-wiring.test.ts` — the gateway-glue
  integration test (notice-text pin + suppress path + ordering invariant).

Scoped vitest + `npm run lint` (tsc + 8 guards) clean.

### DIFF
--- a/telegram-plugin/fleet-fallback-resume.ts
+++ b/telegram-plugin/fleet-fallback-resume.ts
@@ -81,12 +81,24 @@ export interface FleetFallbackResumeGate {
   /**
    * Decide whether to resume the dead turn after a successful swap. Call ONLY
    * when the swap outcome was `'switched'`. Records the arm time on a 'resume'
-   * verdict so a follow-on swap within `singleFlightMs` is suppressed.
+   * verdict so a follow-on swap within `singleFlightMs` is suppressed. Equivalent
+   * to `peek()` followed by `arm()` on a 'resume' verdict.
    *
    * @param failedTurnStartedAtMs epoch-ms the failed turn began, or null when
    *        unknown (then the staleness guard is deferred to the boot path).
    */
   decide(failedTurnStartedAtMs: number | null): ResumeDecision;
+  /**
+   * Evaluate the resume verdict WITHOUT arming the single-flight latch. Use when
+   * the caller must perform fallible work (e.g. write a consume-once carrier)
+   * BEFORE committing to a restart: peek → do the writes → `arm()` only on
+   * success, so a throwing write can never leave an armed latch with no pending
+   * restart (which would suppress a legitimate later resume for `singleFlightMs`).
+   */
+  peek(failedTurnStartedAtMs: number | null): ResumeDecision;
+  /** Commit the single-flight arm (records "now" as the arm time). Pair with a
+   *  prior `peek()` that returned 'resume'. */
+  arm(): void;
   /** Test seam — reset to fresh state. Production code should not call this. */
   reset(): void;
   /** Test/debug — current internal state. */
@@ -105,7 +117,7 @@ export function createFleetFallbackResumeGate(
   // -Infinity = never resumed. A concrete value arms the single-flight window.
   let lastResumedAtMs = Number.NEGATIVE_INFINITY;
 
-  function decide(failedTurnStartedAtMs: number | null): ResumeDecision {
+  function peek(failedTurnStartedAtMs: number | null): ResumeDecision {
     const now = nowFn();
     // Guard 2 — single-flight. A second swap within the window does not
     // re-arm; the in-flight restart owns the resume.
@@ -115,12 +127,23 @@ export function createFleetFallbackResumeGate(
     if (failedTurnStartedAtMs != null && now - failedTurnStartedAtMs > maxAgeMs) {
       return 'skip-stale';
     }
-    lastResumedAtMs = now;
     return 'resume';
+  }
+
+  function arm(): void {
+    lastResumedAtMs = nowFn();
+  }
+
+  function decide(failedTurnStartedAtMs: number | null): ResumeDecision {
+    const verdict = peek(failedTurnStartedAtMs);
+    if (verdict === 'resume') arm();
+    return verdict;
   }
 
   return {
     decide,
+    peek,
+    arm,
     reset() {
       lastResumedAtMs = Number.NEGATIVE_INFINITY;
     },
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -468,8 +468,16 @@ import {
   readConfiguredDefaultModel,
   writeSessionEffortFile,
   clearSessionEffortFile,
+  writePremiumRecoveryFile,
+  readPremiumRecoveryFile,
+  clearPremiumRecoveryFile,
 } from './session-model-file.js'
-import { decideTierDowngrade, planTierDowngrade } from '../tier-downgrade.js'
+import { runTierDowngrade } from './tier-downgrade-wiring.js'
+import {
+  decidePremiumRecovery,
+  renderPremiumRecoveryPing,
+  premiumRecoveryClaimKey,
+} from '../premium-recovery.js'
 import { discoverModels, selectModel } from '../../src/agents/model-picker.js'
 import { resolveMainModel, SWITCHROOM_DEFAULT_THINKING_EFFORT } from '../../src/agents/scaffold.js'
 import {
@@ -21940,6 +21948,10 @@ function recordTypedModelSwitch(
   }
   if (!reply.selectedModel) return ''
   sessionModelSource.setOverride(reply.selectedModel)
+  // A manual /model apply to the dropped premium clears any pending recovery
+  // marker so the "available again" ping can't still fire after the user has
+  // already switched back themselves.
+  clearPremiumRecoveryOnManualSwitch(reply.selectedModel)
   return ''
 }
 
@@ -21968,6 +21980,11 @@ function recordModelMenuSideEffects(
   // (recommended)" selection clears any leftover carrier.
   if (outcome.selectedModel) {
     sessionModelSource.setOverride(outcome.selectedModel)
+    // Clear a pending premium-recovery marker when the tap re-selects the
+    // dropped premium (menu OR the recovery ping's own switch-back button):
+    // no stale "available again" ping once we're back on it. Idempotent — the
+    // ping-send path already consumed the marker, so this is a no-op there.
+    clearPremiumRecoveryOnManualSwitch(outcome.selectedModel)
   }
   if (outcome.clearedDefault) {
     const smDir = resolveAgentDirFromEnv()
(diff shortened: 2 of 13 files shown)


mekenthompson:
## Adversarial review (independent, commissioned via @overlord) — REQUEST-CHANGES (minor)

Mechanism verified **safe** on every high-risk axis:

- **429 cross-wire: safe.** litellm-local + generic-transient 429s early-return and never reach `fireFleetAutoFallback`; only account-scoped-beyond-threshold / `quota_exhausted` reach fleet fallback; 529/overload excluded from `isAutoKind`; tier-downgrade is consulted only inside `all-blocked`, strictly after account-swap. (`gateway.ts:7709,7757,7763,7822,7890,7927,23761`)
- **Builds on #3187, no reimplementation.** Consumes the existing `all-blocked` verdict + `sessionModelSource.getOverride()` / `readConfiguredDefaultModel()`.
- **Routing invariants intact (I1/I3/I5).** Writes only `.session-model` / `.tier-downgrade-attempts` + self-restart; never touches `credentials.json`, never injects `ANTHROPIC_AUTH_TOKEN`, never re-auths. Only the model token changes; litellm traversal unaffected.
- **No wedge.** on-default short-circuit + per-chain counter cap + 10-min staleness; downgrade-target-also-walled falls through to the give-up card, no loop.

### Findings to fix before merge

**[MEDIUM] Downgrade notice misinforms the user → silent quality demotion.** The notice says the session "reverts to `fable` on the next restart." Under rev-4 session-scoped `/model` (merged today), the next restart reverts to the **configured default (opus)**, not the user's premium selection — the fable override lived only in in-memory `sessionModelSource` and is destroyed at restart; nothing re-selects premium after boot. This contradicts the module docstring (`tier-downgrade.ts:22-29`) and the scaffold tests. A user reads "reverts to fable," never re-issues `/model fable`, and stays silently on opus indefinitely — the exact silent-demotion this tier was meant to avoid, surfacing through messaging. Fix: reword to say the premium selection was dropped and to re-issue `/model <model>` (or add genuine premium-intent persistence + re-apply, which does not currently exist).

**[LOW] Attempts-counter pollution on carrier-write failure.** `writeTierDowngradeAttempts` persists BEFORE `writeSessionModelFile`; if the carrier write throws, `count=1` is left while the session is still on the premium model → a later independent overload within the 10-min staleness window hits `decideTierDowngrade` with premium-active + `count>=1` → `'exhausted'` on its first real attempt, wrongly declaring the turn lost without ever downgrading. Roll the counter back (or write it after the carrier) on the abort path.

**[LOW] `maybeTierDowngrade` integration path untested.** The pure `decideTierDowngrade` + FS helpers + start.sh block have real-outcome tests, but the gateway glue (notice text, resume-gate pacing, `exhausted` broadcast, clear-on-outbound) has none — which is why the MEDIUM slipped through. Add one integration test that pins the notice text.

Everything else clean (broadcast posture matches existing fleet-fallback; counter is deliberate belt-and-suspenders; no dead code). Fix the MEDIUM + two LOWs, add the notice-pinning test, then re-review → merge.

mekenthompson:
## Review close-out (verified against current tip `81c6cf18`)

Per-finding, against the branch tip (not the original commit the review read):

- **[MEDIUM] "reverts to fable" misinforms → silent demotion** — **resolved in `5d9bbf44`.** `renderTierDowngradeNotice` (`tier-downgrade.ts:134`) is now pure + unit-pinned and states the turn resumes on the **default** and the user must re-issue `/model <premium>` themselves: _"Re-issue `/model fable` once it frees up — it won't switch back on its own."_ No revert-to-premium promise anywhere. Confirmed on tip; no further action beyond this note.

- **[LOW] attempts-counter pollution on carrier-write failure** — **moot.** The `.tier-downgrade-attempts` counter (and its `exhausted` decision variant + the persist-before-carrier ordering the finding hinged on) was **removed entirely** in `5d9bbf44` (option b — it was unreachable behind the natural on-default guard). `grep` for `tier-downgrade-attempts` / `writeTierDowngradeAttempts` / the `exhausted` variant in the tier-downgrade source returns nothing on tip. The carrier-before-arm ordering that remains is now pinned by a test (below).

- **[LOW] `maybeTierDowngrade` integration path untested** — **newly addressed in `81c6cf18`.** Extracted the order-sensitive downgrade glue into `runTierDowngrade` (`telegram-plugin/gateway/tier-downgrade-wiring.ts`, injected-deps runner mirroring `throttle-tier-wiring.ts`); `maybeTierDowngrade` is now a thin adapter wiring the real seams. New `telegram-plugin/tests/tier-downgrade-wiring.test.ts` exercises the **actual glue** and **pins the broadcast notice** — asserts the resume-on-default + manual-re-issue wording and **fails on any revert-to-premium phrasing** (the exact regression that slipped through). It also covers the resume-gate `skip-inflight` **suppress** path (no notice, no restart) and the carrier-**before**-arm ordering invariant (a throwing carrier write leaves no armed latch).

Also shipped in `81c6cf18` (Ken-approved): the **"premium model recovered" ping** — on recovery the user gets one Telegram ping with a one-tap **"Switch to fable"** button, off the broker's **existing** deterministic per-account eligibility (`list-state` `exhausted` / `premium_walled`, polled by the gateway's existing `runQuotaWatch` — no new poller, no broker change). At-most-once (fleet-claim dedup + consume-once marker); the button routes through the same session-scoped `/model` apply path (PR #3184). Details in the updated PR description.

Scoped vitest (premium-recovery + tier-downgrade-wiring + tier-downgrade + fleet-fallback-resume + session-model-file + scaffold.session-model) and `npm run lint` (tsc + 8 guards) green.

PR #3219, opened 2026-07-13
fix(gateway): close idle-clear check-to-send race with a write-time precondition (#3116)

mekenthompson:
Closes #3116.

## Problem
`maybeIdleClear()` evaluated the idle gate, then dispatched `/clear` into the agent's tmux asynchronously. An inbound arriving between the gate check and the tmux send landed `/clear` in claude's prompt buffer, where it ran at the *next* idle prompt — against a session that was **no longer idle**, destroying live context. Observed 2026-07-11 (overlord stray injected text). Narrowed but not closed by #3113.

## Fix
The inject primitive is shared with proactive-compact, so the guard lives in the primitive:

- `injectSlashCommand`/`injectSlashCommandWith` take an optional `precondition: () => boolean`, re-evaluated **immediately before** the first `send-keys` write (after the pane lock is held and the session confirmed). False → nothing is typed; returns the new `outcome: 'skipped'` (`errorCode: 'precondition_failed'`). Every interactive/operator caller omits it → unchanged behaviour.
- `maybeIdleClear` passes `stillIdle`: `!turnInFlightForGate() && Date.now() - max(lastActivityAt, lastTurnEndedAt) >= idleClearMs`. An inbound in the gap calls `markIdleActivity()` (bumps the clock, resets `idleAutoCleared`), so the precondition reads false and the gate re-arms cleanly.

## Test (fails on main)
`src/agents/inject.test.ts` — precondition false → `runner.send` **never** called + `outcome==='skipped'`; precondition true → sent. The "send never called" assertion **fails on current main** (no precondition → the write always fires). Verified via mutation check (source stashed → 2 RED).

Scoped tests green (99/99 in `inject.test.ts`); `npm run lint` clean.

## Stacking
First of the idle-clear cluster (#3116 → #3117 → #3114 → #3115). Based on main; independent of the others at the file level except a shared touch of `maybeIdleClear`. Later PRs in the stack branch off this one.

### DIFF
--- a/src/agents/inject.test.ts
+++ b/src/agents/inject.test.ts
@@ -919,3 +919,98 @@ describe("#2566 — /model and /memory allowlist treatment", () => {
     expect(result.outcome).toBe("ok");
   });
 });
+
+// ─── #3116 — check-to-send race guard (precondition re-eval at write) ──────
+describe("injectSlashCommandWith — precondition (#3116)", () => {
+  it("precondition false → command is NEVER typed into the pane (outcome=skipped)", async () => {
+    // The race: idle-clear decided /clear, but an inbound arrived before the
+    // tmux write. The precondition re-reads the idle gate at write time and
+    // returns false, so nothing is sent.
+    const sent: string[][] = [];
+    const runner = makeFake({
+      hasSession: true,
+      captures: ["before\n"],
+      onSend: (args) => sent.push(args),
+    });
+
+    const r = await injectSlashCommandWith(runner, {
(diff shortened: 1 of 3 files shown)


mekenthompson:
Superseded — Ken's #3217 already merged the #3116 fix, and overlord is rebuilding the whole idle-clear cluster (#3114-3117) against live main with a stronger approach (full idle-decision re-run at write time + buffer-cancel follow-up). Closing to avoid duplicate work.

PR #3221, opened 2026-07-13
fix(gateway): TTL-bounded background-work suppressor for idle-clear (#3117)

mekenthompson:
Closes #3117. **Stacked on #3219 (#3116)** — review/merge that first; this branch contains its commit.

## Problem
`turnInFlightForGate()` sees the main turn only. #3113 fixed in-flight background subagents *as activity* (a worker that EMITS keeps the window warm), but a worker alive yet **silent** for a full idle window can still be `/clear`ed out from under itself.

## Fix — TTL-bounded suppressor (self-healing)
An unbounded `backgroundWorkInFlight` boolean was deliberately rejected in #3113: a stuck pending flag would disable idle-clear forever. The TTL fixes that.

- `pending-work-progress.ts`: `anyPendingAsyncDispatch()` — keyless scan (single-session gateway).
- `idle-clear.ts`: pure `decideBackgroundSuppression({ backgroundInFlight, suppressingSince, now, ttlMs })`. Clock starts at the would-be-clear point; suppresses until `now - since >= ttlMs`; then self-heals. Default 30m (`SWITCHROOM_IDLE_BG_SUPPRESS_TTL_MS`).
- `maybeIdleClear`: base decider first; if it would not clear, reset the clock; if it would, apply the suppressor — so grace is measured from the moment the clear would otherwise fire (not from when the worker appeared).

## Tests (both directions, RED on main)
`idle-clear.test.ts` — silent worker in flight past the window but within TTL → **NOT cleared**; flag stuck past window+TTL → clear **still fires once** (not permanently disabled). `pending-work-progress.test.ts` — `anyPendingAsyncDispatch` sees work under any key. Both files import symbols that don't exist on main → RED.

Scoped tests green (52/52); `npm run lint` clean.

### DIFF
--- a/src/agents/inject.test.ts
+++ b/src/agents/inject.test.ts
@@ -919,3 +919,98 @@ describe("#2566 — /model and /memory allowlist treatment", () => {
     expect(result.outcome).toBe("ok");
   });
 });
+
+// ─── #3116 — check-to-send race guard (precondition re-eval at write) ──────
+describe("injectSlashCommandWith — precondition (#3116)", () => {
+  it("precondition false → command is NEVER typed into the pane (outcome=skipped)", async () => {
+    // The race: idle-clear decided /clear, but an inbound arrived before the
+    // tmux write. The precondition re-reads the idle gate at write time and
+    // returns false, so nothing is sent.
+    const sent: string[][] = [];
+    const runner = makeFake({
+      hasSession: true,
+      captures: ["before\n"],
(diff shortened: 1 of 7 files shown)


mekenthompson:
Superseded — overlord is rebuilding #3117 against live main with dispatchedAt-stamped TTL wired into the idle decision (not the fleet inbound gate). Closing this and its duplicate #3222.

PR #3222, opened 2026-07-13
fix(idle): gate idle-clear on in-flight background sub-agents, TTL-bounded (#3117)

mekenthompson:
Closes #3117. Second staged PR of the idle-clear cluster, on top of #3116 (`c3a665fa`).

## Problem
The idle auto-/clear gate consulted only main-turn state (`turnInFlightForGate`, gateway.ts ~:2721). A detached background sub-agent — `Agent`/`Task` dispatched, main turn ended before the worker returned — was invisible to it. `hasPendingAsyncDispatch` existed but was wired only to silence-poke, never to the idle decision. So a worker alive-but-silent for a full idle window (one long tool call, no stream events to re-stamp the activity clock) could be `/clear`'ed out from under the context its handback needs.

## Fix
1. **Timestamp first.** `noteAsyncDispatch` (pending-work-progress.ts) set `pending=true` with no timestamp, so no TTL could be built. Added a `dispatchedAt` epoch (stamped/re-stamped on dispatch, cleared whenever `pending` goes false via `startTurn`/`clearPending`).
2. **TTL-bounded suppressor.** New `asyncDispatchAgeMs(key)` + agent-wide `anyPendingAsyncDispatchWithin(ttlMs)`, and `BACKGROUND_WORK_SUPPRESS_TTL_MS` (= `MAX_LIFETIME_MS` = 30m). Idle-clear is suppressed only while a pending dispatch is within the TTL; past it the suppression lapses so a leaked/never-cleared flag can never disable idle-clear forever (self-healing).
3. **Wired into the decision, not the gate.** New optional `backgroundWorkInFlight` input on `decideIdleClear`; suppress the clear while true. Deliberately NOT folded into `turnInFlightForGate` — that gate also feeds fleet-wide inbound buffering/receipt gates and must stay main-turn-only.

## Composes with #3116
#3116's precondition re-runs the FULL `decideIdleClear` at `/clear` write time. Because the suppressor lives inside `decideIdleClear` (and is fed at both the decision site and the write-time re-eval), it's enforced at decision time AND write time for free — a sub-agent dispatched in the check-to-send gap suppresses the buffered `/clear`.

## TTL bound + rationale
30 min, tied to `MAX_LIFETIME_MS`. Both express the same "how long do we believe a background worker is still legitimately running" budget, so keeping them equal makes the idle-suppression window and the ambient-liveness window agree by construction. Shorter risks clobbering a legit ~25-min worker; longer weakens self-healing.

## Test
4 new assertions in `idle-clear.test.ts`, verified to FAIL on post-3116 main (no background input to the decider today) and pass here:
- pending dispatch inside TTL → `decide` returns clear=false
- flag stuck past TTL → clear self-heals (feature not permanently disabled)
- dispatch cleared (worker returned) → clear re-allowed before TTL
- fresh dispatch re-arms the TTL

The end-to-end block drives the REAL pending-work-progress state (via an injected clock) into `decideIdleClear`, so the wiring is exercised as an outcome, not mirrored.

## Adversarial review
- TTL boundary tested (`< ttl` at ttl-1 true, ttl+1 false).
- No path sets `pending=true` without `dispatchedAt`; a defensive stale entry (rolling upgrade) is treated as no-suppression → safe.
- `turnInFlightForGate` and the 6 other IdleClearState-shaped callsites untouched — `decideIdleClear` is only called in `maybeIdleClear`.
- Kill switch (`SWITCHROOM_DISABLE_PENDING_PROGRESS=1`) → no dispatch tracking → pre-3117 behaviour, acceptable.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -5073,6 +5073,14 @@ function maybeIdleClear(): void {
       idleClearMs,
       alreadyCleared: idleAutoCleared,
       turnInFlight: turnInFlightForGate(),
+      // #3117 — TTL-bounded background-work suppressor. A detached sub-agent
+      // (Agent/Task dispatched, main turn ended before it returned) is invisible
+      // to turnInFlightForGate(); consult the pending-dispatch flag directly so a
+      // silent long-running worker isn't /clear'ed out from under its handback.
+      // The TTL keeps a leaked flag from disabling idle-clear forever.
+      backgroundWorkInFlight: pendingProgress.anyPendingAsyncDispatchWithin(
+        pendingProgress.BACKGROUND_WORK_SUPPRESS_TTL_MS,
+      ),
     },
     Date.now(),
   );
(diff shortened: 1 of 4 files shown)


mekenthompson:
Superseded — duplicate of #3221, both replaced by overlord's live-main #3117 rebuild. Closing.

PR #3239, opened 2026-07-14
fix(worker-feed): absolute row-lifetime cap for immortal-but-updating pinned card

mekenthompson:
## The bug (evidenced on live agent Carrie, v0.18.23)

A coalesced worker-feed card gets pinned when a sub-agent registers. When the sub-agent finishes, `subagent-watcher` logs `cleaned up terminal agent <id>`, but the worker-feed **row** never transitions to finished — it stays `state=running, finished=false`. Meanwhile something keeps calling the feed row's `update()` every ~6s. The `#3226` TTL reaper in `heartbeatTick` is keyed off `row.lastUpdateAt`, so **each update resets the staleness clock and the silence sweep NEVER fires**.

Observed: a pinned zombie card lived **5 hours**, re-edited **3000+ times**, only cleared by the restart-time dm-pin-sweep.

`#3226` shipped two mitigations that both fail to bite Carrie's row:
1. `onTerminalCleanup` → `workerActivityFeed.terminate()` wiring (gateway.ts:30343 → subagent-watcher.ts:2258) — the terminal signal is wired, but the row keeps being re-driven, so it is either never actually terminated or immediately re-created/kept alive.
2. the `heartbeatTick` silence reaper — both its branches (`finished` GC and `lastUpdateAt >= staleWorkerTtlMs` reap) require either `finished=true` or silence. Carrie's row is **neither** finished **nor** silent, so neither can ever match.

> Note on the dispatch premise: the requested "Part 1" (drive `terminate` from the watcher's terminal-cleanup path) was **already shipped in #3226** and is present in v0.18.23 — verified via `git log -S`. The genuinely-missing durable guarantee is the absolute cap below.

## The fix — absolute row-lifetime cap

Add an **absolute** row-lifetime cap, anchored to a new immutable `createdAtMs` stamped once at row creation and **never** re-stamped by later updates. The heartbeat sweep now reaps any row past `absoluteRowLifetimeCapMs` **regardless of `lastUpdateAt` or finished state** — force-terminating + unpinning it. Once reaped, the existing `finalized` gate blocks a late/spurious cue from resurrecting the row, so the card is dead for good.

The gateway derives the cap from the watcher's effective in-flight terminal cap (`resolveInflightTerminalCapMs() * 8` → 6h at the 45-min default), matching the existing `staleWorkerTtlMs` derivation pattern — it tracks operator overrides and is never a bare magic number. It sits far above any legitimate single worker's lifetime, so it can only bite a genuine leak; a real long worker that ever hit it collapses its cosmetic card to `incomplete` and unpins — its actual result still reaches the user via the separate handback reply.

## Tests (assert outcomes, RED without the fix)

- A row that **keeps updating** past the absolute cap (resetting `lastUpdateAt` every tick — the exact Carrie failure mode) is still reaped + unpinned.
- A row past the cap is reaped even when the silence TTL would never fire.

Both assert outcomes (`feed.size === 0`, pin `messageId === null`), not that a function was called. Verified they go RED with the source change stashed.

## Files changed
- `telegram-plugin/worker-activity-feed.ts` — `createdAtMs` field + `absoluteRowLifetimeCapMs` option + absolute-age branch in `heartbeatTick`.
- `telegram-plugin/gateway/gateway.ts` — derive + pass `absoluteRowLifetimeCapMs`.
- `telegram-plugin/tests/worker-feed-coalesce.test.ts` — 2 new regression tests.

## Local checks
- `vitest run` worker-feed-coalesce (26→28 tests) + worker-feed-terminal-cleanup: **28 passed**.
- `npm run lint` (tsc --noEmit + repo checks): **clean**.

Completes #3207.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -2043,13 +2043,21 @@ const WORKER_FEED_STALE_TTL_MARGIN_MS = 5 * 60_000
  * row's immutable creation time, so it reaps an immortal card even when the row
  * keeps receiving `update()` cues that reset `lastUpdateAt` every heartbeat
  * (the Carrie 5h zombie-pin leak, re-edited 3000+ times, that the silence sweep
- * could never match). At the 45-min default cap this yields a 6-hour absolute
- * ceiling — far above any legitimate single worker's lifetime, so it can only
- * ever bite a genuine leak, while still bounding a ghost card to hours not days.
- * Derived from the same base as `staleWorkerTtlMs` (never a bare magic number)
- * so it tracks any operator override of the terminal cap.
+ * could never match). At the 45-min default cap this yields a 3-hour absolute
+ * ceiling — comfortably above any legitimate single worker's lifetime, so it
+ * can only ever bite a genuine leak, while bounding a ghost card well UNDER the
+ * ~5h symptom Ken hit (the initial 8x/6h was too loose). Derived from the same
+ * base as `staleWorkerTtlMs` — the watcher's effective in-flight terminal cap
+ * (`resolveInflightTerminalCapMs()`: the `SWITCHROOM_SUBAGENT_INFLIGHT_TERMINAL_
+ * CAP_MS` env override, else the 45-min default) — never a bare magic number, so
+ * it tracks the env-level terminal-cap override in lockstep with the silence
+ * backstop. (Both this and `staleWorkerTtlMs` call `resolveInflightTerminalCapMs()`
+ * with NO arg, matching the watcher's OWN effective value here: the gateway does
+ * not pass a config-file `inflightTerminalCapMs` to `startSubagentWatcher`, so
+ * there is no config-file override to thread through — env + default is the
+ * complete override surface on this path.)
  */
-const WORKER_FEED_ABSOLUTE_ROW_LIFETIME_CAP_MULTIPLE = 8
+const WORKER_FEED_ABSOLUTE_ROW_LIFETIME_CAP_MULTIPLE = 4
 const workerFeedOwnerDmFallbackLogged = new Set<string>()
 
 /**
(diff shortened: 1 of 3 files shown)


mekenthompson:
## Follow-up per Ken's review (pushed a5f9db9b)

**1. Boot/reconnect purge (the key ask).** Added `workerActivityFeed.purgeAllOnBoot()` — finalizes+removes every row and drives `reconcilePin(messageId: null)` for every group, unconditionally. Wired into the bridge-connect path **before** the old feed is stopped/replaced. This closes a real leak I found while investigating: on a bare **bridge reconnect** (no full process restart, so no boot pin-sweep), `workerActivityFeed?.stop()` only cleared the heartbeat — the outgoing feed's `wk:group:` pins were orphaned because the replacement feed is empty and never knew those groups. Tests assert a restored running row + pinned card is emptied + unpinned (and a late cue can't resurrect it), plus multi-chat + idempotency.

Note on scope: worker rows are **in-memory only** — a full restart destroys them and the persisted `wk:group:` pin is already swept at boot by `statusPinBootCleanup`. For a **DM** the actual Telegram unpin is necessarily deferred to first-inbound (positive DM chat ids return `400 chat not found` until the user messages — a Bot API limit, see the #3026 comment), which is why the Carrie pin cleared on first inbound. `purgeAllOnBoot` fixes the in-process reconnect leak, which needs no user message.

**2. agentId mismatch — investigated, NOT reproducible from source (reported, not force-fit).** Registration (`registerAgent`), `onProgress`→`update`, and `cleanupTerminalAgent`→`onTerminalCleanup`→`terminate` all use the **identical** `entry.agentId` (the `agent-<uuid>.jsonl` stem, `subagent-watcher.ts:2822`, never transformed). Both gateway `update` sites (`gateway.ts:30704`, `:30855`) and the `terminate` site (`:30345`) pass the same destructured `agentId`. So `terminate(id)` **does** match the row when cleanup fires for it — no key divergence exists. The real immortal-updating leak is a **missed** terminal signal (cleanup never fires for that row — e.g. a sibling in a coalesced group whose JSONL neither vanished nor hit the cap), which the absolute cap (#3) + boot/reconnect purge (#1) durably cover, as you authorized.

**3. Tightened the absolute cap** 8× → **4×** the in-flight terminal cap → **~3h** at the 45-min default (was 6h), comfortably under the ~5h symptom. Still derived, no bare magic number.

**4. Doc-comment fix.** Corrected the cap comment: it tracks the env override (`SWITCHROOM_SUBAGENT_INFLIGHT_TERMINAL_CAP_MS`) + default — the **complete** override surface here, since the gateway does not pass a config-file `inflightTerminalCapMs` to `startSubagentWatcher`, so `resolveInflightTerminalCapMs()` with no arg matches the watcher's own effective value. No overclaim of general "operator overrides".

**Local checks:** worker-feed-coalesce (28) + worker-feed-terminal-cleanup (2) = **30 passed**; `npm run lint` **clean**.

PR #3246, opened 2026-07-15
fix(hindsight): durable retain watermark + boot reconciliation (#3244)

mekenthompson:
## Summary

PR1 of 2 for #3244 — stop Hindsight silently dropping work from abruptly-ended
sessions (SIGKILL / OOM / watchdog kill / gateway SIGTERM). **Forward fix only;
the one-time backfill is PR2 and is gated (below).** Builds the revised design
`design-20260715.md` §1–§1.5, §4, §5.

## Why

The `.jsonl` transcript survives every kind of session death (Claude Code writes
it independent of any hook), but retain durability depended on a graceful
SessionEnd that an abrupt kill skips, a per-turn Stop hook that silently
swallowed its own POST failures, and a boot path that only drained a queue —
never reconciled the transcript. When those holes lined up, an entire session's
work vanished with a clean recall (the clerk 2026-07-11 loss).

Job spec: this advances "always available" / memory-survives-restart
(`reference/jobs`), the same outcome the `jtbd-memory-survives-restart` UAT
guards.

## Three holes closed

- **A2 (silent drop).** `retain.py:main()` now ENQUEUES the failure payload it
  already builds (mirrors `session_end.py`) instead of discarding it.
- **B1 + C (no jsonl-tail reconciliation).** New `reconcile_tail.py` runs at
  SessionStart *after* the drain: it diffs a durable per-session watermark
  against the on-disk transcript tail and retains any un-committed human turns
  before the new session starts. This is the load-bearing guarantee; A2/B1 are
  defense-in-depth.
- **Duplicate-namespace id.** The live chunked-`n>1` path drops the wall-clock
  `{session_id}-{time.time()*1000}` id for the content-derived
  `{session_id}-r{start_uuid}-{end_uuid}` (FULL uuids) that reconcile and the
  PR2 backfill also compute — one id namespace, so any two writes of the same
  turn-slice upsert instead of double-storing.

## Also in scope

- `lib/watermark.py` — atomic (tmp+fsync+rename) + flock, advances ONLY on
  confirmed persistence, never regresses, explicit uuid-not-found (compaction)
  branch.
- `build_retain_payload()` — network-free, mission-write-free seam reused by
  reconcile (and PR2 dry-run).
- Shared `retain-inflight.lock` pacing — non-blocking for the live path so turn
  latency never regresses; blocking for boot reconcile.
- `client.retain(async_processing=False)` commit-before-ack on durability
  paths, so a bare async-200 (ack-of-receipt) can never falsely advance the
  watermark.
- Honest bounds — 48h lookback / 200-turn cap / 4s budget each ENQUEUE the
  remainder rather than truncate. No silent loss of any turn written to disk
  (modulo the documented ≤1-turn torn-tail residual, design §3).

## Merge preconditions still open (do NOT merge until resolved)

1. **Empirical daemon upsert-on-`document_id` probe** (throwaway bank: POST same
   id twice with different content, assert ONE doc reflecting the second write).
   The whole single-id dedup — including the live throttle-edge re-fire — relies
   on it.
2. **`async=false` commit-before-ack probe** (assert the doc is queryable the
   instant the sync POST returns).
3. This is **PR1 of 2**. The backfill (PR2) is hard-gated on #1, on this PR's
   single-id redesign, and on a Ken-reviewed fleet dry-run before `--commit`.

## Test plan

`vendor/hindsight-memory/scripts/tests/test_reconcile_durability.py` — fake
in-process daemon, no network, asserts outcomes:
SIGKILL→boot-reconcile recovers all 5 turns; failed Stop retain enqueued (not
dropped); live-Stop + reconcile of the same window ⇒ ONE document (single-id
fix); commit-before-ack gates the watermark (200-not-persisted doesn't advance);
turn-cap ENQUEUES the remainder; `build_retain_payload` is write-free; watermark
monotonicity + stale-anchor.

Local (scoped): `python3 -m unittest discover tests/` from
`vendor/hindsight-memory/scripts` → **256 pass** (CI-authoritative path). The
vendor-root `tests/` set → **229 pass** (one assertion updated to the new
commit-before-ack contract). No Python linter is configured in-repo.

## Risk

Blast radius of the id change is contained to `retain.py:260-261` (the only
wall-clock-id site; the full-session/`n==1` stable-id path is unchanged). Live
throughput unaffected except the durability POST is now synchronous on the
async-registered Stop hook (off the user's critical path). Notable deviation:
the `memory.retain.reconcile_on_start` YAML surface in `scaffold.ts` was
deferred — the knob is fully functional and defaults ON via config /
`HINDSIGHT_RECONCILE_ON_START`, and wiring it through the byte-identical
settings.json reconcile invariant risked destabilising the TS suite for no
behavioural gain in PR1.

### DIFF
--- a/vendor/hindsight-memory/scripts/drain_pending.py
+++ b/vendor/hindsight-memory/scripts/drain_pending.py
@@ -72,7 +72,18 @@ def _budget_seconds() -> float:
 
 
 def _retry_one(entry: dict, timeout: int) -> None:
-    """POST a single queued retain. Raises on failure."""
+    """POST a single queued retain. Raises on failure.
+
+    Posts ``async_processing=False`` (commit-before-ack, switchroom #3244 §1.1):
+    the drain is a DURABILITY path — it deletes the pending entry on a 200, so
+    the 200 must prove durable persistence, not merely ack-of-receipt. A bare
+    async 200 followed by a dropped extraction would delete the queue entry
+    while the content never lands, and (for boot-reconcile remainders whose
+    watermark already advanced) there is no reconcile backstop — silent loss
+    (the #3244 bug). All drained entries — Stop-hook A2 failures, SessionEnd
+    failures, and reconcile-remainder deferrals — are durability retries, so
+    sync is correct for every one.
+    """
     client = HindsightClient(entry["api_url"], entry.get("api_token"))
     client.retain(
         bank_id=entry["bank_id"],
@@ -82,6 +93,7 @@ def _retry_one(entry: dict, timeout: int) -> None:
         metadata=entry.get("metadata") or {},
         tags=entry.get("tags"),
         timeout=timeout,
+        async_processing=False,
     )
 
 
--- a/vendor/hindsight-memory/scripts/lib/client.py
+++ b/vendor/hindsight-memory/scripts/lib/client.py
@@ -173,12 +173,22 @@ class HindsightClient:
         metadata: Optional[dict] = None,
         tags: Optional[list] = None,
         timeout: int = 15,
+        async_processing: bool = True,
     ) -> dict:
         """Retain content into a bank's memory.
 
-        Posts with async=true so the server processes in the background.
-        The context field helps Hindsight cluster memories by provenance
-        (e.g. "claude-code" vs manual retains).
+        By default posts with ``async=true`` so the server processes extraction
+        in the background (a 200 is an ack-of-receipt, not proof of durable
+        persistence). The context field helps Hindsight cluster memories by
+        provenance (e.g. "claude-code" vs manual retains).
+
+        ``async_processing=False`` (switchroom #3244 §1.1) posts ``async=false``
+        so the 200 is returned only after the daemon has durably committed the
+        item — commit-before-ack. Durability paths whose success advances the
+        retain watermark (the Stop-hook durability retain and boot
+        reconciliation) MUST use this so a bare async 200 can never falsely mark
+        unpersisted work as committed. (Merge precondition: the daemon honours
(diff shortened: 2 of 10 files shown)


mekenthompson:
Addressed the diff-review blockers in `dd92af43`:

- **F1 (HIGH, reproduced silent loss)** — closed two ways: (a) `drain_pending._retry_one` now POSTs `async_processing=False` (commit-before-ack), so delete-on-200 is safe for every durability retry; (b) `reconcile_tail` no longer advances the watermark on a turn-cap split — the older remainder precedes `recent`, so advancing to the transcript tail would jump past un-confirmed content. On a split the recent slice is upserted for recall but the anchor stays put; the next boot re-derives and idempotently re-upserts until the tail is confirmed end-to-end.
- **F2 (HIGH, advance-on-enqueue-fail)** — subsumed by F1's "never advance on a split": the watermark never moves past an unconfirmed remainder whether the enqueue succeeded or not.
- **F3 (test gap)** — test 10 now sets `drop_async=True` on the drain path and asserts (a) the watermark did NOT advance to the transcript end after the capped reconcile, and (b) all turns land durably after the drain. Verified it FAILS against pre-fix code (`AssertionError` on the watermark check) and PASSES after.
- **uuid precondition** — verified on real on-disk transcripts that Claude Code stamps a stable outer `uuid` on every user/assistant entry (only non-message `queue-operation` lines lack it, and those aren't surfaced), so `last_uuid` is non-None in production and the watermark persists normally.

`cd vendor/hindsight-memory/scripts && python3 -m unittest discover tests/` → **256 pass** (CI-authoritative). Vendor-root `tests/` → 229 pass.

PR #3249, opened 2026-07-15
feat(hindsight): one-time transcript backfill to recover pre-fix misses (#3244)

mekenthompson:
One-time operator-run recovery tool for #3244 — scans on-disk transcripts and restores work missing from agents' Hindsight banks (e.g. clerk's 2026-07-11 total-loss sessions). Rebased onto main after PR #3246 (durable retain) merged; was PR #3248 (auto-closed when its base branch was deleted on PR1 merge).

## Safety (validated: adversarial review + re-review, all findings fixed)
- **Dry-run is the default; zero writes.** Requires explicit `--commit` to write, and `--commit` requires an explicit `--agent` (no fleet-wide blind commit).
- **No duplication:** total-loss classification gate + watermark advance (NOT id-convergence — ids intentionally do not match the live path; the gate is load-bearing and tested as such).
- **Never storms:** strictly serial, async=false commit-before-ack, shared inflight lock, 1.5s pace, backoff + circuit-breaker.
- **Wrong-bank safe:** dynamic-bank agents refused (zero writes); resolved bank surfaced per agent/session in the dry-run report.
- **Active-session safe:** `--min-idle-s` (default 3600s) skips in-flight sessions.
- Not wired to any hook/cron — operator-invoked only.

## Depends on
PR #3246 (durable retain seams: build_retain_payload, slice_document_id, watermark, pacing) — now merged to main.

## Run flow
1. `python3 backfill_transcripts.py` (dry-run, whole fleet, zero writes) → review per-agent report.
2. `python3 backfill_transcripts.py --commit --agent clerk` (static-bank incident target).

267 hindsight tests pass. Do NOT run against real banks without operator go.

### DIFF
--- a/vendor/hindsight-memory/scripts/backfill_transcripts.py
+++ b/vendor/hindsight-memory/scripts/backfill_transcripts.py
@@ -1,0 +1,613 @@
+#!/usr/bin/env python3
+"""One-time backfill of transcript work already missing from Hindsight banks.
+
+switchroom #3244 Part 2 (design-20260715.md). Recovers work an abrupt session
+death silently dropped BEFORE the forward fix (PR1) shipped — e.g. clerk's
+2026-07-11 planner sessions whose Stop-hook retains were discarded and never
+reconciled. This is a standalone, operator-run CLI, NOT runtime code: it scans
+every agent's on-disk ``.jsonl`` transcripts (which survive every kind of
+death) and re-POSTs anything missing from the bank, paced and resumable.
+
+**Reuse, not fork (design Part 2).** It imports PR1's seams verbatim — the
+deterministic content-derived ``document_id`` (``retain.slice_document_id`` via
+``build_retain_payload``), the durable watermark, the shared inflight pacing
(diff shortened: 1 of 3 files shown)


mekenthompson:
⚠️ RUN-SAFETY CAVEAT (from a fleet-wide dry-run 2026-07-15): Do NOT run `--commit` against pre-deployment history. The total-loss classification ('no watermark = lost') is only valid AFTER PR #3246 (durable retain) is deployed and has been writing watermarks — before that, ALL historical sessions look total-loss (dry-run showed 3659 sessions / 17279 turns), including work already retained by the existing per-3-turn Stop hook. Because backfill document_ids do not converge with the live path's ids, re-posting already-saved history would DUPLICATE it, not upsert. Also: run only in an environment with correct per-agent bank resolution — a run from a single agent's container resolves every agent to that agent's bank (cross-contamination). Recommended: deploy #3246 first; treat historical incidents (e.g. clerk 2026-07-11) as narrow targeted restores, not a fleet sweep. The tool + dry-run are safe; this caveat governs --commit.

PR #3251, opened 2026-07-15
feat(hindsight): --from-logs slice-level recovery for backfill (#3244)

mekenthompson:
## Summary

Adds a SAFE, log/registry-driven `--from-logs` recovery mode to the Hindsight backfill tool (`vendor/hindsight-memory/scripts/backfill_transcripts.py`), the #3244 follow-up. It uses the per-agent turns registry to NARROW candidate sessions (bounding cost), then makes **session-level, id-scheme-agnostic bank-membership** the real safety gate: it restores only sessions the agent's true bank has **zero** documents for (true total loss). It reuses the merged tool's pacing / dry-run / commit-gating / deterministic-id / watermark machinery unchanged.

Cites job: memory-survives-restart / durable-memory recovery (operator-run repair tool, not runtime).

## Why

The merged backfill is a total-loss gate driven by the durable watermark, which did not exist for pre-fix history. `--from-logs` re-derives the total-loss signal from the durable turn-lifecycle log (`registry.db`) instead, and verifies against the real bank before any write.

## The 3 design red-team must-fixes folded in

1. **BROAD candidate classifier (no dead in-flight predicate).** The design's original narrowing predicate `restart AND tool_call_count>0 AND last_assistant_done=0` is dead on real data — both columns are NULL on 100% of deployed `restart` rows. The classifier is broad: any row that is not a clean, closed `stop` (`ended_via IS NULL OR ended_via <> 'stop' OR ended_at IS NULL`) is a candidate — `timeout`, `sigterm`, null-open, and ALL `restart` rows. Cost stays bounded because membership is one cheap server-side-filtered GET.

2. **Membership as the real gate (session-level, scheme-agnostic).** One server-side query (`GET .../documents?q=<session_id>`) returns that session's docs. Presence is decided by the `{session_id}` id PREFIX, which BOTH the legacy `{session}-{epoch_ms}` scheme (what production banks hold today) and the new `{session}-r{uuid}` scheme share. ANY doc ⇒ present ⇒ skip; ZERO ⇒ total loss ⇒ restore. **Fail-closed** on query error (treat present ⇒ skip, never duplicate).

3. **Event-span CONTAINMENT join (not mtime-window).** Deployed registries have no `session_id` column, so each turn's transcript is resolved by span containment: the transcript whose `[first_event_ts, last_event_ts]` CONTAINS the turn's `started_at`, chat-filtered. 0 or >1 matches ⇒ refused (reported, not guessed). Migrated registries with a stamped `session_id` take the direct join.

Should-fixes folded: membership GETs paced through the shared inflight lock; server-side `q=` filter avoids a read storm; the false "metadata is {}" justification dropped.

## BLOCKER-1 fix (adversarial review of the first push)

The first version used **slice-level exact `-r{uuid}` id membership**. That scheme is **not deployed** — production banks are keyed with the legacy `{session}-{epoch_ms}` scheme, so every computed `-r` slice id would miss the legacy docs, classify every slice "missing", and re-post — **duplicating memory on any populated bank, even intact sessions**. Fixed by making membership **session-level and scheme-agnostic** (id-prefix presence). Consequence, stated plainly: the mode recovers **only zero-document (total-loss) sessions**. Partial-tail recovery within a *populated* session is **OUT OF SCOPE** on legacy epoch-ms banks (a legacy id doesn't encode which turns it covers, so the gap can't be determined safely) — it can be added once the `-r{uuid}` scheme (`retain.py`'s not-yet-live `slice_document_id`) is deployed and banks are re-keyed. Documented in the tool docstring, CLI `--help`, and the report footer.

## Safety argument

- Restore fires only for a session with zero bank docs → the backfill is the sole writer for it; its own deterministic ids converge, so re-runs upsert (idempotent, no dup).
- A session with any doc (legacy or new scheme) is skipped wholesale → a populated/intact session can never be duplicated.
- Fail-closed membership: any daemon error ⇒ present ⇒ skip.
- Correct-bank targeting unchanged: bank = `switchroom.yaml memory.collection ?? name`, cross-checked against `.claude/settings.json` X-Bank-Id, REFUSE on disagree / dynamic / missing row. Never ambient env / stale plugin default.
- Dry-run default, genuinely zero-write; `--commit` still requires an explicit `--agent`.

## Recovers clerk's 2026-07-11 loss

clerk's bank is empty (`total:0`), so that session is a true total loss. The incident row (`sigterm`, 01:39:35) is a candidate under the broad classifier; its transcript is resolved by span containment (mtime-window would miss it); membership returns zero docs → restore all its slices. This is exactly the case the session-level gate is safe for.

## Safe RUN envelope (do NOT run in this PR)

1. `--from-logs` dry-run fleet-wide (zero writes) → review the per-agent candidate report (per session: join kind, `has_documents` vs `total_loss` vs `ambiguous`/`unmatched`, resolved bank).
2. Then per agent: `--from-logs --commit --agent clerk`.

`--commit` restores **only zero-document (total-loss) sessions** and is therefore safe against populated legacy banks (klanker/overlord/carrie/gymbro — they have docs → every candidate session is skipped as `has_documents`). Fleet-wide commit stays refused; dynamic/mismatched/missing bank stays refused.

## Test plan

`cd vendor/hindsight-memory/scripts && python3 -m unittest discover tests/` → **278 passed** (11 in `tests/test_backfill_from_logs.py`, deterministic, fake daemon + fake registry/transcripts, no network/no real bank):
- **legacy-scheme docs present → zero restores** (BLOCKER-1 regression; verified to FAIL on the pre-fix code — it posts 5 duplicate `-r` docs — and PASS after)
- genuinely-empty (total-loss) `restart` with NULL in-flight cols → restored (proves broad classifier)
- mixed fleet: legacy-present session skipped + empty session restored, one run
- span-containment picks the right transcript where naive mtime mismatches; ambiguous → refused
- membership error → fail-closed → not restored
- wrong/mismatched/missing bank → refused, zero writes
- dry-run → zero writes (no POST, no watermark, no progress)
- pacing: membership GET + restore POSTs serialised + spaced via fake clock
- direct session_id join on a migrated registry

## Risk

Additive mode; the existing watermark path and all merged guarantees are untouched. Operator-run tool, not runtime code. No fleet run performed.

### DIFF
--- a/vendor/hindsight-memory/scripts/backfill_transcripts.py
+++ b/vendor/hindsight-memory/scripts/backfill_transcripts.py
@@ -98,6 +98,12 @@ from lib.config import debug_log, load_config
 from lib.content import _is_tool_result_only_user_message
 from lib.daemon import get_api_url
 from lib.pacing import inflight_lock
+from lib.turnlog import (
+    read_candidate_turns,
+    resolve_transcript_for_turn,
+    resolve_true_bank,
+    transcript_span,
+)
 from retain import build_retain_payload, read_transcript
 
 # Injectable sleep so tests assert pacing via a fake clock, not wall time.
@@ -144,6 +150,36 @@ def _min_idle_s(override: Optional[int] = None) -> float:
         return max(0.0, float(os.environ.get("HINDSIGHT_BACKFILL_MIN_IDLE_S", "3600")))
     except ValueError:
         return 3600.0
+
+
+def _window_slack_s(override: Optional[int] = None) -> float:
+    """Slack (seconds) added to each side of a transcript's event span when
+    resolving the historical span-containment join (turnlog #3). Default 120s;
+    override with ``--window-slack-s`` / ``HINDSIGHT_BACKFILL_WINDOW_SLACK_S``."""
+    if override is not None:
+        return max(0, override)
+    try:
+        return max(0.0, float(os.environ.get("HINDSIGHT_BACKFILL_WINDOW_SLACK_S", "120")))
+    except ValueError:
+        return 120.0
+
+
+def _switchroom_yaml_path(agents_root: str) -> str:
+    """Path to switchroom.yaml. Override with ``HINDSIGHT_SWITCHROOM_YAML``;
+    otherwise the sibling of the agents root (``~/.switchroom/switchroom.yaml``)."""
+    override = os.environ.get("HINDSIGHT_SWITCHROOM_YAML")
+    if override:
+        return override
+    return os.path.join(os.path.dirname(os.path.normpath(agents_root)), "switchroom.yaml")
+
+
+def _registry_tmpl(agents_root: str) -> str:
+    """Template (with ``{agent}``) for a per-agent turns registry. Override with
+    ``HINDSIGHT_REGISTRY_TMPL``; default ``<agents>/{agent}/telegram/registry.db``."""
+    override = os.environ.get("HINDSIGHT_REGISTRY_TMPL")
+    if override:
+        return override
+    return os.path.join(agents_root, "{agent}", "telegram", "registry.db")
 
 
 def _is_dynamic_bank(config: dict) -> bool:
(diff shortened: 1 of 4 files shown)


mekenthompson:
Addressed review before merge:
- **[LOW]** Fixed the stale test-module docstring (`test_backfill_from_logs.py`) to accurately describe SESSION-LEVEL membership (superseded slice-level design) — commit 5ca4bc40. Tests still green (11/11).
- **[MED]** Filed the partial-POST-failure under-recovery gap as tracked follow-up #3291 (not fixed here, per review).

PR #3279, opened 2026-07-16
fix(gateway): deterministic turn-flush delivery via one send primitive (#3276)

mekenthompson:
Closes #3276. Follow-up #3278 (read-back verification) deferred.

## Problem

On a turn where the model composes a final answer as plain assistant text but does NOT call the `reply` tool, the turn-flush backstop is supposed to deliver it. Instead it delivered by **editing the ephemeral progress card** (`editMessageText` onto the taken-over card), pushed the **card message id** into `sentIds`, logged `status=ok` / recorded `complete` — and the card was garbage-collected ~60-90s later. Net: nothing durable reached the chat while the turn record said `complete`. Observed live on klanker 2026-07-16 (6 silent drops; operator lost trust in delivery). Delivered turns were exactly the reply-tool turns; undelivered turns were exactly the card-coupled backstop turns.

## Fix — one delivery primitive

New `deliverAnswer()` routes the backstop through `sendReplyChunks` (the same battle-tested send core `executeReply` uses: THREAD_NOT_FOUND fallback, length re-split, parse-reject plaintext fallback). It posts a **fresh chat message** and returns the **real fresh message ids**. It never edits the card; the card is unpinned/collapsed as a cosmetic follow-up.

### The 8 guards (each in the diff)

1. `deliverAnswer` stays **downstream** of the guards — called only inside the `flush` branch, after the `replyCalled` short-circuit and the 2s `getRecentOutboundCount` suppression (both kept, unhoisted).
2. **Synchronous latch before async send** — `turn.answerDelivered = true` and `backstopDeliveryLedger.claim(turnId)` set at flush-fire, before any `await`.
3. **Real `sentIds` into supersede** — `flushedTurnSupersede.record({turnId, messageIds: sentIds})` now gets genuine chat ids, so a late reply corrects in place instead of second-bubbling.
4. **Card collapsed to a NON-answer state** — answer flows only through `deliverAnswer`; the card is just unpinned. A card-collapse-fail and a fresh-send-fail can't leave both an answer-card and an answer-bubble visible.
5. **Once-per-turn delivery latch** (`BackstopDeliveryLedger`, keyed on `turnId`) is the backstop-vs-reply arbiter and the double-fire guard (a second fire for the same turn is a no-op).
6. **Per-chunk sent-ledger + pre-send pending marker** — a retry after a partial send resumes at the first unsent chunk; chunk 0 is never re-sent.
7. **Receipt gate = fresh non-card id** — `backstopSendOutcomeGated` / `finalizeBackstopSendGated` count only fresh chat ids (excluding `backstopCardMessageId`); a card-only "success" ⇒ `send_failed`, not `complete`.
8. **Deterministic narration strip** — `isNarrationBlock` broadened to catch non-opener progress narration ("Checking now…", "Pulling the numbers:") that `NARRATION_OPENER` missed, WITHOUT re-gating on `FLUSH_SUBSTANTIVE_MIN_CHARS` (short answers like "yes, done" still deliver).

### New / changed modules

- `telegram-plugin/gateway/backstop-delivery.ts` (new) — pure latch + per-chunk idempotency ledger + receipt gate.
- `telegram-plugin/gateway/turn-record-status.ts` — `backstopSendOutcomeGated` / `finalizeBackstopSendGated` (card-aware receipt gate).
- `telegram-plugin/turn-flush-safety.ts` — guard-8 narration heuristic.
- `telegram-plugin/gateway/gateway.ts` — `deliverAnswer` + backstop rewiring.

## Test matrix

- `backstop-delivery.test.ts` (12): receipt gate (fresh id ≠ card id; card-present ⇒ delivered id ≠ card id), status honesty (`complete` IFF a real non-card id; card-only ⇒ `send_failed`; partial/throw/empty ⇒ `failed`), once-per-turn latch, per-chunk idempotency (partial send + retry ⇒ chunk 0 not duplicated; index-ordered ids).
- `turn-flush-safety.test.ts`: narration strip ("Checking now…\n\n<answer>" ⇒ answer only; colon-terminated; short terminal answer still delivered).

Local: scoped vitest **167 passed** (touched + adjacent flush/supersede/answer-ready suites); full `npm run lint` (tsc + all check-*.mjs + bot-api-wrapping) green.

## Deferred (out of scope)

- #3278 — read-back/echo verification for API-ok-but-silently-dropped sends.
- The ~150s liveness/orphaned-reply floor for crash/watchdog turn-ends that never emit `turn_end` is unchanged.

### DIFF
--- a/telegram-plugin/gateway/backstop-delivery.ts
+++ b/telegram-plugin/gateway/backstop-delivery.ts
@@ -10,26 +10,31 @@
  * This module owns the pure, side-effect-free arbitration + accounting the
  * backstop needs so it can be unit-tested without the 30k-line gateway:
  *
- *  - a once-per-turn delivery LATCH keyed on `turnId` (guard 5) — the real
- *    arbiter of backstop-vs-reply, replacing the blind 2s recent-outbound
- *    window as the primary decision;
  *  - a per-chunk sent LEDGER with a pre-send pending marker (guard 6) so a
  *    retry after a partial send or a lost ack resumes at the first unsent
  *    chunk and NEVER re-sends chunk 0;
+ *  - a bounded in-turn RETRY orchestrator (`runBackstopDelivery`) that is the
+ *    live caller of the ledger's resume machinery — the real fix for the
+ *    `send_failed` silent-drop: it retries mid-chunk before giving up, and
+ *    reports whether the answer was actually delivered so the caller can leave
+ *    the delivery obligation OPEN on terminal failure;
  *  - a RECEIPT GATE (guard 7) that counts only fresh, non-card chat message ids
- *    — a card message id can never satisfy the delivery obligation.
+ *    — a card message id can never satisfy the delivery obligation;
+ *  - a once-per-turn double-fire LATCH (guard 5) so a turn that already fired a
+ *    backstop (answer-ready quiescence, then the turn-end backstop) does not
+ *    deliver twice.
  *
  * Every function here is pure over its arguments (the ledger is an explicit
  * per-turn store the caller owns); nothing reads gateway module state.
  */
 
 /**
- * Per-turn delivery latch + per-chunk idempotency ledger for the turn-flush
+ * Per-turn double-fire latch + per-chunk idempotency ledger for the turn-flush
  * backstop. One instance is held at gateway module scope; entries are keyed by
  * the per-turn `turnId` nonce so distinct turns never collide.
  */
 export class BackstopDeliveryLedger {
-  /** turnIds that have claimed the delivery latch (guard 5). */
+  /** turnIds that have claimed the backstop double-fire latch (guard 5). */
   private latched = new Set<string>()
   /** turnId -> (chunkIndex -> landed message ids for that chunk). */
   private chunks = new Map<string, Map<number, number[]>>()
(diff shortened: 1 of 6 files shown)


mekenthompson:
## Adversarial-review findings addressed — commit `48694a08`

All 7 findings closed; re-review welcome.

**1 (MED) — `send_failed` silently dropped the answer.** Added a bounded in-turn retry `runBackstopDelivery` (default 3 attempts, `SWITCHROOM_BACKSTOP_DELIVERY_MAX_ATTEMPTS`). Each attempt **resumes at the first unsent chunk** via the per-chunk ledger — chunk 0 is never re-sent. Only after retries are exhausted do we record `send_failed`, and in that terminal-fail case the delivery obligation is left **OPEN** so the ~150s liveness floor re-presents it. This required deferring the obligation close out of `endCurrentTurnAtomic` (new `deferObligationClose` option, mirroring `deferRecord`) and resolving it in the async finally from the REAL send outcome: `delivered` ⇒ `close`, terminal-fail ⇒ `finalAnswerDelivered=false` + `noteTurnEnded` (left open). The suppressed-reply branch closes the (now-deferred) obligation explicitly since the reply delivered.

**2 — guard 6 now has a live caller.** `runBackstopDelivery` consults `unsentIndices`/`hasChunk` every attempt; the ledger `clear(turnId)` moved to run only after success OR exhaustion, never before a retry can read prior progress.

**3 — guard 5 de-overclaimed.** The ledger latch arbitrates **backstop-vs-backstop double-fire only**; late-reply arbitration remains `flushedTurnSupersede` + `turn.answerDelivered` (unchanged from main). Deleted the dead `isLatched`; corrected the comments and the two synchronous-arm comments to name the two distinct arbiters.

**4 (LOW) — post-delivery throw.** Terminal status is now gated on the receipt-gated `delivered` flag returned by the orchestrator, not a blanket outer-catch. A throw in the post-delivery bookkeeping (dedup/supersede record) can no longer demote a genuinely delivered turn to `send_failed`.

**5 (LOW) — recordOutbound id/text misalignment.** History `texts` are zipped from `ledger.entries()` (source text repeated per landed id), so a length-resplit chunk that lands >1 id stays aligned. Test asserts `ids.length === texts.length` on a resplit.

**6 — integration oracles.** `runBackstopDelivery` is a pure, injectable orchestrator (gateway.ts isn't importable — Bun.listen at import). New tests drive the real retry/ledger/receipt code: fresh non-card id recorded in history (id ≠ card), **mid-chunk retry proving chunk 0 is sent exactly once on attempt 2**, terminal failure ⇒ `delivered=false`/`exhausted` (the exact input the gateway uses to leave the obligation open), and text/id alignment on resplit.

**7 — guard 8 low.** Added tests that a colon-terminated line that IS the whole answer is never dropped, both standalone and as the terminal block after narration (the narration heuristic only ever strips PRECEDING blocks).

**Local results:** `tsc --noEmit` clean, full `npm run lint` chain green, scoped vitest green — `backstop-delivery` (18), `turn-flush-safety` (57), `turn-record-status` (13), `obligation-turn-end` (6), plus adjacent flush/supersede/answer-ready/obligation-determinism suites (obligation determinism: 62 across 3000-schedule fuzz).

PR #3280, opened 2026-07-16
fix(gateway): deterministic delivery for the turn-flush backstop (#3276)

mekenthompson:
Fixes #3276.

## Root cause (confirmed against real source at HEAD `a1872cc1`)

Two send paths in `gateway.ts`:

- **Reply-tool path** (`executeReply`, gateway.ts:13321) → `sendReplyChunks` (`outbound-send-path.ts:239`) → emits `[outbound] path=reply msg_id=…` per chunk + `reply: finalized … messageIds=[…]`, records a durable message_id. **Delivered.**
- **Turn-flush backstop** (gateway.ts:~18923 IIFE) → a hand-rolled loop sending raw `bot.api.sendRichMessage(backstopChatId, …)` with **no** `path=reply`/`messageIds` receipt, then `finalizeBackstopSend` (gateway.ts:19143 → `turn-record-status.ts:89`) stamped `delivered` purely from `!threw && sentCount >= chunkCount`. **No independent receipt** — a non-throwing `status=ok` that produced no real durable message was recorded `complete` while the operator got nothing.

### Contradiction with the issue's stated mechanism (reported, not force-fit)

The issue's decisive 2026-07-16 comment attributes the drop to the backstop *painting the ephemeral status card* via `editMessageText`/`sendRichMessage`. At HEAD `a1872cc1` (and the deployed klanker `v0.18.28`, built 04:58Z — at/after this HEAD) that card-takeover branch is **dead code**: `progressDriver === null` (gateway.ts:9392) forces `backstopCardMessageId === null`, so the `editMessageText` branch never executes. The `editMessageText status=ok` in the incident logs comes from a *different* subsystem (the activity/progress-card painter), not the backstop. This matches the maintainer's own triage comment. The **real surviving gap** is the receipt-less send + non-throw⇒delivered inference — which this PR fixes deterministically regardless of the card theory.

## Changes

1. **Route the backstop through `sendReplyChunks`** — the same primitive real replies use. A flushed answer now lands as a real, durable chat message, emits the identical `[outbound] path=reply msg_id=…` receipts + a `reply: finalized … messageIds=[…] via=turn-flush` line, and collects real Telegram message_ids. Same THREAD_NOT_FOUND / oversize re-split / parse-reject fallback ladder — no robustness lost. Dead card-takeover `editMessageText` branch removed. (`gateway.ts`)
2. **Independent-receipt requirement** — `backstopSendOutcome`/`finalizeBackstopSend` now take `receiptIds` and return `delivered` only when every chunk yielded a positive message_id. A no-throw send with zero positive receipts is `send_failed`, never `complete`. (`turn-record-status.ts`)
3. **Suppression oracle** no longer relies solely on `history.db` (`getRecentOutboundCount`, blinded by a stale history.db that returned 0). The in-turn `turn.replyCalled` latch is the primary signal; history.db is a best-effort corroborator. (`gateway.ts`)

## Tests

Regression asserts the OUTCOME (`turn-record-status.test.ts`): a no-throw backstop send with **zero recorded message ids → `send_failed`** (fails on the old non-throw⇒delivered behavior); with a real message id → `complete`. Non-positive ids (0/negative) don't count.

- Scoped vitest: `turn-record-status`, `outbound-send-chunks`, `outbound-send-path`, `turn-flush-safety`, `turn-end-gate`, `answer-ready-flush`, `turn-flush-dedup-controller`, `gateway-no-reply-single-emit` — **165 passed**.
- `npm run lint` — clean (strict tsc + all guard scripts).

## Consolidated release

Ships alongside #3277 (deterministic model switch) in the same release.

## Open risks (honest)

- Suppression on `turn.replyCalled === true` treats "reply tool invoked in-window" as delivered; a reply tool that was invoked but *failed* to deliver would suppress the flush. Reply path has its own robust retries and `computeTurnStatus` still gates `suppressed → complete` on `finalAnswerDelivered`, so the exposure is narrow — but noted.
- The gateway IIFE itself isn't unit-testable (gateway.ts isn't importable); the deterministic guarantee is enforced at the pure `backstopSendOutcome`/`finalizeBackstopSend` seam that the IIFE drives verbatim.
- Secondary finding from the issue (history.db stopped recording at msg 18855) is not fixed here — it's a separate defect; this PR removes the backstop's *dependence* on it but does not repair history.db recording.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -18922,12 +18922,30 @@ function handleSessionEvent(ev: SessionEvent): void {
 
         void (async () => {
           await new Promise<void>(resolve => setTimeout(resolve, 500))
-          if (HISTORY_ENABLED) {
+          // #3276 SUPPRESSION-ORACLE FIX. The primary "did a reply already
+          // land?" signal is the IN-TURN `turn.replyCalled` flag — flipped
+          // synchronously the moment the reply tool is invoked (gateway
+          // reply-tool observer), so a reply that lands in this post-fire
+          // window is caught even when history.db is stale. The issue found
+          // the old sole check `getRecentOutboundCount` was blinded by a
+          // history.db that stopped recording (always returned 0), defeating
+          // suppression. We now check the in-memory latch FIRST and keep the
+          // history.db count only as a secondary/best-effort corroborator.
+          const replyLatched = turn.replyCalled === true
+          if (HISTORY_ENABLED || replyLatched) {
             try {
-              const { getRecentOutboundCount } = await import('../history.js')
-              const recentCount = getRecentOutboundCount(backstopChatId, 2)
-              if (recentCount > 0) {
-                process.stderr.write(`telegram gateway: turn-flush suppressed — reply tool sent ${recentCount} message(s) within 2s\n`)
+              let recentCount = 0
+              if (HISTORY_ENABLED) {
+                try {
+                  const { getRecentOutboundCount } = await import('../history.js')
+                  recentCount = getRecentOutboundCount(backstopChatId, 2)
+                } catch {}
+              }
+              if (replyLatched || recentCount > 0) {
+                const why = replyLatched
+                  ? `reply tool invoked in-turn (replyCalled=true)${recentCount > 0 ? `; history corroborates ${recentCount}` : ''}`
+                  : `reply tool sent ${recentCount} message(s) within 2s`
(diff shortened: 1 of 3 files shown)


mekenthompson:
Superseded by #3279, which merged to main (14b156fe) and carries the identical guarantees: receipt-bearing backstop send via deliverAnswer/sendReplyChunks, delivered-status gated on every chunk landing a fresh non-card message id, and no finalAnswerEverDelivered self-satisfaction on the flush branch. Verified equivalent by adversarial review. Closing as duplicate.

PR #3306, opened 2026-07-17
fix(telegram): line-leading issue refs (#123) are not headings

mekenthompson:
## Problem

An outbound message whose line starts with an issue reference — e.g. `#3304 is merged. The regression suite is green.` — renders in Telegram as a huge heading spanning the whole paragraph.

**Before (Telegram):** the entire first paragraph in giant heading text.
**After:** plain body text, literal `#3304` preserved.

## Root cause

- The parser is spec-correct: micromark reads `#3304` as a plain paragraph (GFM/CommonMark: `#` opens an ATX heading only when followed by a space, another `#`, or EOL). No bug there.
- The renderer (`telegram-plugin/render/render.ts`, `renderBlock` paragraph case) re-emits the text via `escapeMarkdown`, which deliberately never escapes `#` (format.ts:53). So the wire body still begins with a literal line-leading `#`.
- Telegram's Bot API 10.1 server-side rich-markdown parser is **non-spec**: it promotes any line-leading `#` run to a heading, space or not. That's the promotion step.

## Fix

New `escapeLineLeadingHash` helper applied only to rendered **paragraph** output: backslash-escapes the first `#` of each line (`/^([ \t]{0,3})#/gm`). Telegram honours backslash escapes (same mechanism as the existing link-href `)` escape), so the reader sees a literal `#3304`. Heading IR nodes render untouched, so genuine `# Title` headings keep working; mid-line `#` is untouched; code blocks/tables/blockquote markers unaffected. This is the single shared render path (`renderOutboundChunks` → stream-controller), so every agent's outbound gets the fix automatically.

## Tests

New describe block in `tests/render/render.test.ts`:
- multi-line message starting `#3304 is merged.` → stays paragraphs, output starts `\#3304`, no heading on re-parse, literal `#3304` round-trips
- `#123` at line start mid-message → escaped, no heading
- genuine `# Heading` still renders as a heading
- mid-line `#` left alone

Results: `tests/render/` 10 files / 232 tests pass; normalizer suites (paragraph-normalizer, paragraph-spacer-golden, telegram-format, format-consistency, formatting-parse-regression, card-format) 221 tests pass — no block-spacing/bullet regressions. `npm run lint` clean.

### DIFF
--- a/telegram-plugin/render/render.ts
+++ b/telegram-plugin/render/render.ts
@@ -203,10 +203,34 @@ function renderTable(node: TableNode): string {
   return [headerLine, sepLine, ...bodyLines].join("\n");
 }
 
+/**
+ * Escape a line-leading `#` in rendered PROSE so Telegram cannot promote the
+ * line to a heading (#3304 regression).
+ *
+ * Per CommonMark/GFM, `#` opens an ATX heading only when followed by a space,
+ * another `#`, or end-of-line — so micromark correctly parses `#3304 is
+ * merged.` as a plain paragraph, and this renderer re-emits it verbatim
+ * (`escapeMarkdown` deliberately leaves `#` alone). But Telegram's Bot API
+ * 10.1 server-side rich-markdown parser is NON-spec here: any line-leading
+ * `#` run is read as a heading, so an issue reference opening a line renders
+ * the whole paragraph as huge heading text. Backslash-escaping the first `#`
+ * of each such line keeps it literal on Telegram while remaining a no-op
+ * visually (Telegram honours backslash escapes, same mechanism the link-href
+ * `)` escape relies on).
+ *
+ * Applied ONLY to paragraph output (including paragraphs nested in
+ * blockquotes / list items via their renderBlock recursion), never to
+ * heading, code-block, or table nodes — so a genuine `# Title` heading node
+ * still renders as `# Title`.
+ */
+function escapeLineLeadingHash(text: string): string {
+  return text.replace(/^([ \t]{0,3})#/gm, "$1\\#");
+}
+
 function renderBlock(node: Block): string {
   switch (node.type) {
     case "paragraph":
-      return renderInlineChildren(node.children);
+      return escapeLineLeadingHash(renderInlineChildren(node.children));
     case "heading":
       return `${"#".repeat(node.level)} ${renderInlineChildren(node.children)}`;
     case "blockquote":
--- a/telegram-plugin/tests/render/render.test.ts
+++ b/telegram-plugin/tests/render/render.test.ts
@@ -143,6 +143,48 @@ describe("render: inline palette", () => {
   });
 });
 
+describe("render: line-leading issue refs are not headings (#3304 regression)", () => {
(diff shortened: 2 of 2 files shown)


mekenthompson:
## Adversarial review (overlord/reviewer, 2026-07-18) — static PASS, live check PENDING

**Verdict: DO-NOT-MERGE yet** — solely because the load-bearing assumption (Telegram Bot API 10.1 honours a line-leading `\#` escape and renders a literal `#`) is still **unverified live**. The vault grant for `telegram-test-bot-token` timed out with the operator away; the live probe (escaped `\#3304…` vs unescaped control into the Test Harness supergroup `-1003982767197` via `sendRichMessage`) is queued for when the grant is re-approved.

### Static review — what's good
- `escapeLineLeadingHash` (`telegram-plugin/render/render.ts:226`) is scoped to paragraph output only; heading/code/table nodes untouched. A paragraph line can never legitimately begin `# ` (ATX headings interrupt paragraphs in CommonMark, so parse.ts would have split it) — so the escape cannot eat an intended heading.
- Idempotent: after escaping, the line starts `\`, which no longer matches `/^([ \t]{0,3})#/gm`; the `richMessage()` guards (`guardAccidentalBlockConstructs` etc.) don't touch `#` or the inserted `\`, so no double-escape on the composed pipeline.
- Chunking safe: `renderOutboundChunks` (rich-render.ts:130+) re-splits the RAW source and re-renders each piece, so a chunk whose first line newly starts with `#` gets escaped per-piece. Blockquote/list nesting gets the escape before the `> ` / `- ` prefix is added — harmless either way.
- Tests assert outcomes (escaped output + round-trip reparse to paragraph, heading preserved, mid-line `#` untouched); scoped run green (232/232 in tests/render), CI fully green.

### Findings to fix (follow-up or in-PR)
1. **Coverage gap — the `richMessage()` seam is still exposed.** Only stream-controller answers flow through the renderer; the ~93 `richMessage()` callsites in `gateway.ts` (approval cards, banners, previews) send dynamic content that can open a line with `#3304`-style refs and will still be heading-promoted. The unambiguous sub-case `^#(?=\S)` (no space — never an intended heading) belongs in `guardAccidentalBlockConstructs` (`render/line-start-guard.ts`), which would cover every wire send including this PR's case. Suggest follow-up issue.
2. **Stale contradicting doc.** `render/line-start-guard.ts` header still claims "`#1` / `#foo` (NO space) is not a GFM heading at all → Telegram renders it literally → no guard needed" — this PR's whole premise (and #3304's observed behaviour) is that Telegram's 10.1 parser is non-spec here. Update that comment (and `guard-linestart-note.md`) when the live check confirms.
3. **`SWITCHROOM_RICH_RENDER=0` path uncovered** — raw passthrough skips the renderer entirely. Acceptable (documented escape hatch), noting for the record.

Live-check protocol on resume: send `\#3304 is merged.` + unescaped control via the UAT bot (`meken_switchroom_test_bot`) `sendRichMessage` to the UAT supergroup, compare returned message entity JSON; merge only if the escaped variant shows no heading entity AND no literal `\` in the visible text.

PR #3309, opened 2026-07-17
fix(gateway): key PendingUserNoticeGate by turn/topic (#3294)

mekenthompson:
## Summary

Keys the deferred non-operator failure-notice gate (`telegram-plugin/pending-user-notice.ts`, from #3293) by **turn/topic** instead of per-agent, so under the PR-4e keyed-liveness flag (concurrent per-topic turns) one topic's turn end no longer resolves another topic's pending notice.

Closes #3294.

## Why

The user-facing "couldn't complete that — it's on our side" notice is never sent at error time; it is *scheduled* and resolved at the single turn-end funnel (`endCurrentTurnAtomic`): a turn that recovered (delivered a reply) drops it, a turn that died reply-less sends it. The gate collapsed **per-agent** and `resolveTurnEnd` cleared *all* pending notices on *any* turn end.

Under PR-4e keyed liveness (concurrent per-topic turns) that races: turn B's reply-less end flushed turn A's pending notice while A might still recover — a **false failure claim** in that interleaving. Single-turn gateways (the current norm — only one topic ever) were unaffected, which is why this lands *before* the keyed-turn flag ships broadly.

## What changed

- `PendingUserNotice` gains an optional `key` (topic key = `statusKey(chatId, threadId)`).
- `schedule` collapses per `(agent, key)` — concurrent topics each keep their own pending notice; a burst within one turn still collapses to one.
- `resolveTurnEnd(turnKey, delivered, now)` resolves **only** notices matching the ending turn's topic (or unattributed `undefined`-key ones — the legacy agent-wide fallback that keeps single-turn byte-identical). Other topics' notices stay pending until their own turn end or TTL.
- Gateway wiring: the schedule site attributes the notice to the live turn's topic; `endCurrentTurnAtomic` passes its own topic `key` into the flush.

### User-visible behavior change

Under keyed liveness, a concurrent topic ending reply-less (or recovering) no longer triggers/drops another topic's "couldn't complete" notice. Each user's failure notice now fires IFF *their own* turn genuinely died. Single-turn gateways: no change.

## Honest limitation

The `api_error` operator event is agent-level (its wire `chatId` is always empty), so the notice is attributed to the **most-recently-set live turn** — best-effort under the sequential-CLI invariant, and strictly better than the pre-existing agent-wide flush. Perfect per-turn attribution would require plumbing a topic onto the operator-event IPC, which is out of scope for this issue. `undefined` key (no live turn: error between turns / after a silence poke) keeps the legacy agent-wide resolution so a genuine death still notifies.

## Test plan

`telegram-plugin/tests/litellm-proxy-auth-misconfig.test.ts` — updated to the keyed contract + 4 new keyed-liveness cases that assert **outcomes** (a notice gated under the wrong topic key fails):
- cross-topic reply-less end returns `[]` and leaves A pending; A's own end flushes it
- cross-topic recover does not drop A's notice
- per-topic notices do not cross-collapse (each topic's end flushes only its own)
- an unattributed (undefined-key) notice still resolves on any turn end (legacy)

`./node_modules/.bin/vitest run telegram-plugin/tests/litellm-proxy-auth-misconfig.test.ts` → 27 passed. Adjacent `operator-events` + `llm-error-present` → 108 passed. `npm run lint` clean (tsc, `check-bot-api-wrapping` — all gateway raw calls use shift-stable inline markers — and the gateway line-ratchet: +20 lines within the 50-line slack).

## Risk

Low. Single-concern, additive keying; single-turn path byte-identical (undefined-key + one-topic collapse reproduce the old per-agent behavior). No new imports, no send-path change, recipients unchanged. Disjoint from the #2996 P0a registration-extraction region (~L23484–29013) — edits are at L5384, L8655, L8689.

Job spec: `reference/jobs/survive-reboots-and-real-life.md` — "The user is told when something went wrong, never left in silence." This fix guards the inverse failure: never told a turn *failed* that didn't.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -5191,8 +5191,9 @@ function endCurrentTurnAtomic(
   // gate drops the notice; reply-less → the turn genuinely died, the notice is
   // sent now. replyCalled covers the short-answer/#2624 shape where
   // finalAnswerDelivered stays false despite an explicit reply. No-op when
-  // nothing is pending (the overwhelmingly common path).
-  flushPendingUserFailureNotices(turn.finalAnswerDelivered || turn.replyCalled)
+  // nothing is pending (the overwhelmingly common path). #3294 — `key` scopes
+  // resolution to THIS turn's topic under keyed liveness.
+  flushPendingUserFailureNotices(turn.finalAnswerDelivered || turn.replyCalled, key)
   return turnEndedAt
 }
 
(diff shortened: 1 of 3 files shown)


mekenthompson:
Rebased onto main @ ab34bb7b (P0a/P0b/P0c/P0e + P1 cleanup all landed; anchors re-validated — the gate wiring sites survived unchanged, now at endCurrentTurnAtomic ~L5198 and the operator-event schedule site ~L8469). Two additional fixes in the force-push:

1. **Purged 551 stray `.tmp/` cache files** accidentally swept into the original commit by `git add -A` (a local TMPDIR misconfiguration). The PR is now the intended 3 files.
2. **Trimmed inline comments in gateway.ts** to satisfy the P0.5 anti-inflation ratchet, which tightened to 31905 when the P0/P1 extractions merged: net gateway.ts addition is now +11 lines (31948 vs ceiling 31955). The full rationale lives in `pending-user-notice.ts` JSDoc, which the gateway comments now reference.

**Prior bun-test failure was a runner-side hang, not a test failure:** the job was killed at its 8-minute `timeout-minutes` cap with zero assertion failures — output stopped after `subagent-watcher.test.ts` passed, and the runner terminated an orphaned `bun` process. No bun-test suite imports `pending-user-notice.ts` (only the vitest suite does), and the wedge-prone sqlite suites (`subagent-tracker-hooks`, `history-reaper`) reproduce identical lock/timeout behavior on a pristine `origin/main` worktree under load — branch-independent. Local verification on the rebased branch: scoped vitest 27/27, full `npm run lint` clean.

PR #3311, opened 2026-07-18
fix(scaffold): harden Hindsight boot gate — real MCP initialize probe, 120s bound, logged outcome

mekenthompson:
## Why

On 2026-07-18 the **clerk** agent booted toolless ("1 MCP server failed") despite the existing 30s pre-launch gate in `profiles/_base/start.sh.hbs`. Root cause: the gate polled a **bare GET** on `/mcp/`, which can return HTTP 200 mid-restart while a real MCP `initialize` still fails — and 30s undershoots a Hindsight restart window.

## What changed

- **Real readiness probe**: POST a JSON-RPC `initialize` request to `/mcp/` with `Content-Type: application/json` and `Accept: application/json, text/event-stream`, and require a JSON-RPC `"result"` in the response body (dependency-free: curl + grep; works for both plain-JSON and SSE-framed responses). Verified functionally against a mock MCP HTTP server.
- **Bound extended 30s → 120s**; on timeout the agent still boots (degraded beats never-boots) with a clear `WARNING ... TIMED OUT` line.
- **Gate outcome logged** to the start.sh stderr stream either way (`initialize OK after Ns` / timeout warning) so future incidents are diagnosable.
- **`export MCP_TIMEOUT=${MCP_TIMEOUT:-30000}`** so claude's own MCP startup connect is patient too. Env var name verified present in the shipped claude binary (`bin/claude.exe`, v2.1.205); operator env overrides win via the default-expansion form.
- **Tests**: extended the existing scaffold render test to assert the rendered start.sh contains the initialize probe, the 120s bound, and the timeout warning; the omit-when-memory-disabled test still covers the negative path.

This is the only occurrence of the gate block — no other profile/template duplicates it (`grep -rn HINDSIGHT_WAIT profiles/` → `_base` only; `docker/hindsight-entrypoint.sh`'s `SWITCHROOM_HINDSIGHT_WAIT_S` is the container's own wait, untouched).

## Test results

- `vitest run tests/scaffold.test.ts` — 212 passed (one later re-run showed an unrelated 5s-timeout flake in the `resume_mode` test on a slow-IO sandbox; passed on the clean run).
- `npm run lint` — tsc + all guard scripts clean.
- Rendered gate block passes `bash -n` and `sh -n`; probe verified end-to-end against a mock MCP initialize endpoint.

### DIFF
--- a/profiles/_base/start.sh.hbs
+++ b/profiles/_base/start.sh.hbs
@@ -613,22 +613,37 @@ export MCP_TIMEOUT="${MCP_TIMEOUT:-30000}"
 # a 30s bare-GET gate — the endpoint returned HTTP 200 mid-restart while a
 # real MCP initialize still failed). So: probe with an actual JSON-RPC
 # initialize POST and require a JSON-RPC result, not just HTTP 200.
-# Bounded at 120s; on timeout we log a warning and boot anyway (degraded
+# Bounded at 120s of real wall-clock time (each probe costs up to ~3s of
+# curl timeout + 1s sleep, so an iteration count would overshoot — measure
+# elapsed seconds against the deadline instead; HINDSIGHT_WAIT holds real
+# elapsed seconds). On timeout we log a warning and boot anyway (degraded
 # beats never-boots — the plugin reconnects once Hindsight recovers).
 HINDSIGHT_WAIT=0
 HINDSIGHT_READY=0
+_hs_start=$(date +%s)
 while [ $HINDSIGHT_WAIT -lt 120 ]; do
-  if curl -sf --max-time 3 \
+  _hs_resp=$(curl -sf --max-time 3 -i \
        -X POST {{{hindsightApiBaseUrlQ}}}/mcp/ \
        -H "Content-Type: application/json" \
        -H "Accept: application/json, text/event-stream" \
        -d '{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"switchroom-boot-gate","version":"1.0"}}}' \
-       2>/dev/null | grep -q '"result"'; then
+       2>/dev/null)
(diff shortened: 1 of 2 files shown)


mekenthompson (commented):
Adversarial review — verdict: MERGEABLE. CI fully green (26 passing, rest skipped). Verified: the curl -d payload in start.sh.hbs contains no `{{` sequences, extracts as valid JSON (checked with json.loads), and the tests assert against the *rendered* start.sh, so render survival is covered; the gate sits inside `{{#if hindsightEnabled}}` (existing 'omits wait loop when memory disabled' test still covers the off path — no 120s penalty for memory-off agents); MCP_TIMEOUT is the documented Claude Code MCP startup env in ms, and `${MCP_TIMEOUT:-30000}` + userEnvQuoted exporting *after* this block means an operator-set value wins either way. Minor findings, none blocking:

1. **'120s bound' is 120 iterations, not seconds.** Each failed probe costs up to 3s (curl --max-time) + 1s sleep, so worst-case wall time is ~8 min when hindsight accepts TCP but stalls, and the log line 'after ${HINDSIGHT_WAIT}s' underreports. Pre-existing pattern (old loop had it too), but the log claims seconds. Fix: count with $SECONDS or note iterations.

2. **`grep -q '"result"'` false-positive surface.** curl -f filters HTTP errors, but a 200 JSON-RPC error body whose message contains 'result' would pass. Contrived; tightening to '"serverInfo"' or '"protocolVersion"' would make it airtight. (SSE-framed responses are fine — the data: line still matches.)

3. **Orphan MCP session per boot** — the probe's initialize opens a streamable-HTTP session that's never DELETEd. One per boot; negligible, noting for the record.

4. **Test depth**: assertions are substring-presence on rendered output. They'd catch a handlebars swallow of the asserted substrings but not a mangled `{}` elsewhere in the payload. A one-line extract-the--d-payload-and-JSON.parse assertion would close that.

Nothing in the diff outside stated scope (2 files only).

PR #3313, opened 2026-07-18
feat(memory): lazy-connect stdio MCP shim for hindsight — registration can never fail at session start

mekenthompson:
## Why

Claude Code retries a failed MCP handshake ~3x at session start and then marks the server **failed for the entire session** — manual `/mcp` is the only recovery (mid-session HTTP outages recover per-request, startup ones do not). Hindsight (vectorize-io/hindsight 0.8.4) offers HTTP MCP only and officially leaves client resilience to clients. So the current `type: "http"` entry couples every agent's entire memory tool surface (`mcp__hindsight__*`) to the hindsight container being up at the exact instant the agent session boots. A hindsight restart racing an agent restart = a whole session with no memory tools.

## What

New hidden CLI verb **`switchroom hindsight-mcp-shim`** (`src/cli/hindsight-mcp-shim.ts`), wired as the default `hindsight` `.mcp.json` entry by `generateHindsightMcpConfig` (`src/memory/hindsight.ts`):

- **Handshake never fails.** The shim answers stdio `initialize` itself, immediately, without touching the backend — echoing the client's protocol version when supported (2025-06-18 / 2025-03-26 / 2024-11-05), else offering our newest. `ping` is answered locally too. The shim's client-side and backend-side version negotiations are independent; only version-stable methods are forwarded.
- **tools/list**: live fetch with a 3s timeout; on success the manifest is persisted (atomic tmp+rename) to `$HINDSIGHT_SHIM_CACHE_DIR/hindsight-tools-list.json`; on failure the cache is served, or on first boot a built-in static manifest (all 29 tools, permissive schemas) pinned by test against `tests/fixtures/hindsight-tools-list.snapshot.json`.
- **tools/call**: lazy per-call Streamable-HTTP session (initialize → notifications/initialized, `mcp-session-id` + `X-Bank-Id` headers, JSON and SSE-framed responses both parsed), bounded timeout (120s) + exactly one retry with a fresh session (covers cold backends AND expired upstream sessions/404). Backend down ⇒ a proper `isError: true` tool result telling the agent memory is *temporarily* unavailable — never a shim crash, never a session-wide failure. Next call after recovery goes straight through (tested).
- **Everything else** forwarded transparently when up; JSON-RPC error `-32001` when down. Malformed stdin lines and handler throws are absorbed (last-resort catch responds `-32603`).

### Wiring / config

- Default entry becomes `{type:"stdio", command:"/usr/local/bin/switchroom", args:["hindsight-mcp-shim"], env:{HINDSIGHT_MCP_URL, HINDSIGHT_BANK_ID, HINDSIGHT_SHIM_CACHE_DIR, HOME}}` — env threaded explicitly because Claude Code spawns MCP commands with a sanitized env (same pattern as the gdrive launcher).
- **Escape hatch**: `memory.config.mcp_transport: "http"` (new schema enum) restores the legacy direct HTTP entry verbatim.
- `HINDSIGHT_SHIM_CLI_PATH` / `HINDSIGHT_SHIM_AGENT_HOME` are deliberate local copies of `DOCKER_SWITCHROOM_CLI_PATH` / `DOCKER_AGENT_HOME` (importing scaffold.ts from memory/hindsight.ts would be a cycle); `tests/memory.test.ts` pins them in lockstep.
- Docs: `docs/configuration.md` hindsight bullet updated.

### Design notes

- The tool-name surface (`mcp__hindsight__<tool>`) is unchanged — the server key stays `hindsight` — so the enumerated permission allow-list and all guidance keep working.
- `run()` resolves only after stdin ends AND every in-flight handler has flushed — a live CLI smoke caught the exit-on-stdin-end race dropping the final response; there's a regression test for it.
- Fixed a real drift while writing the fallback table: it is generated from (and test-pinned to) the golden snapshot, mirroring the repo's existing `EXPECTED_HINDSIGHT_TOOLS` convention.

## Test evidence

Scoped runs in the worktree (bun install frozen):

- `vitest run tests/hindsight-mcp-shim.test.ts` — **13 passed**: handshake with backend down; version negotiation; local ping; tools/list from cache when down; static fallback when no cache; tools/call `isError` when down; pass-through against a mock Streamable-HTTP backend (lazy upstream handshake order + `X-Bank-Id` + session header asserted); SSE-framed response parsing; live tools/list refreshes a stale disk cache; per-call recovery (isError while down → real result after backend comes up on the same port, same shim instance); stdio drain regression; fallback-vs-snapshot lockstep; fallback schema validity.
- `vitest run tests/memory.test.ts tests/setup.test.ts` — 66 passed (entry-shape pins updated: stdio default, url-override threading, http escape hatch, docker-constant lockstep).
- `vitest run tests/scaffold.test.ts` — 212/212 passed (one vitest-worker "Timeout calling onTaskUpdate" RPC flake on the slow test host, not a test failure; suite is 2–3 min of FS-heavy work here).
- `vitest run tests/cheap-cron-trim-resources.test.ts tests/scaffold.mcp-json-user-declared.test.ts tests/memory.hindsight-contract.fixture.test.ts src/memory/scaffold-integration.test.ts tests/config.test.ts` — 49 passed.
- `npm run lint` — clean (tsc + all 10 guard scripts).
- Live CLI smoke (`bun run dev -- hindsight-mcp-shim` with a dead backend URL): initialize answered instantly, tools/list served the fallback manifest, tools/call returned the isError result.

CI is the full-suite authority.

## Deviations from spec

- "Persist under the agent state dir": the cache lives at `/state/agent/home/.hindsight-shim/` (the agent's writable HOME bind mount) via `HINDSIGHT_SHIM_CACHE_DIR`, falling back to `$HOME/.hindsight-shim` outside docker.
- Language: TypeScript inside the switchroom CLI (matching the existing `drive-mcp-launcher` / `notion-mcp-launcher` stdio-MCP launcher convention) rather than a separate bun package — the entry must resolve at `/usr/local/bin/switchroom` inside agent containers with zero new image plumbing.


mekenthompson (commented):
Adversarial review (posted as COMMENT — GitHub blocks request-changes on the PR author's own account; treat as **REQUEST CHANGES / DO-NOT-MERGE** until findings 1–2 are fixed).

Verified the up-path against the **live** hindsight backend (127.0.0.1:18888/mcp/): initialize answers over SSE with **no `mcp-session-id` header** (stateless), `notifications/initialized` returns 202, and `tools/list` with the shim's exact header pattern (`mcp-protocol-version`, `X-Bank-Id`, no session id) succeeds — the mock backend is faithful where it matters and the shim's null-session handling covers the real server. Design is sound; tests assert outcomes (down-path, cache, fallback, recovery, SSE, drain-on-stdin-close).

### Must fix

**1. Blind one-retry can double-execute non-idempotent tool calls** — `src/cli/hindsight-mcp-shim.ts`, `UpstreamClient.request()`: the catch wraps the *entire* attempt, including failures that occur **after the POST was delivered** (120s abort while the backend is still processing, connection reset mid-response, SSE parse failure). Scenario: `tools/call retain` is delivered, backend is slow, abort fires → catch → `reset()` → full re-send → **duplicate retain** (or duplicate `create_directive` etc.). The legacy `type: http` path never auto-retried — a behavioral regression hidden inside the resilience feature. Fix: retry only pre-delivery failures (connect refused / `ensureSession` failure / 404-session-expired), never timeouts or mid-response errors.

**2. Head-of-line blocking: one slow call freezes the whole shim** — `run()`: `chain = chain.then(async () => { await this.handle(msg); ... })` serializes **handling**, not just writes. One `tools/call` against a hung-but-listening backend holds the chain up to 240s (120s × 2), during which queued `ping`, `tools/list`, and parallel tool calls from the same turn get zero progress — client-visible symptom: the whole hindsight server unresponsive, the exact failure mode this PR exists to prevent. Also makes `notifications/cancelled` structurally useless: the cancel queues *behind* the call it targets. Fix: run `handle(msg)` concurrently, serialize only `output.write`, track in-flight promises for the drain-on-close guarantee.

### Should fix / follow-up

**3. `listChanged: true` advertised but never emitted** — the shim never sends `notifications/tools/list_changed` (not on fallback→live manifest upgrade, and it can't relay the backend's — no GET/SSE listen stream). Degradation is safe (permissive fallback schemas; unknown tools surface as isError) but the capability is decorative and stale schemas persist all session. Emit after the first live refresh, or stop advertising.

**4. Stale comment**: `FALLBACK_TOOL_TABLE` doc points at `tests/hindsight-mcp-shim.fallback.test.ts`; the drift guard actually lives in `tests/hindsight-mcp-shim.test.ts`.

**5. Capabilities note**: live backend also advertises `prompts`/`resources`; shim advertises only `tools`, so a client will never call `prompts/list` despite `forward()` supporting it. Fine today — worth a comment.

**6. Security note (no regression)**: `HINDSIGHT_BANK_ID` env and the agent-HOME cache dir are inside the agent's own trust domain — the legacy http entry's `X-Bank-Id` header was equally agent-rewritable. Poisoned cache = tool-description injection into the agent's *own* context only; no cross-agent escalation.

**Verified good**: fixture `tests/fixtures/hindsight-tools-list.snapshot.json` is on the branch and the drift guard genuinely pins the table; escape hatch renders `type: http` (tested); CLI-path/HOME lockstep pins exist; atomic tmp+rename cache write; `HOME === "/"` guard; per-call AbortController bounds each fetch; scaffold/reconcile/setup tests updated coherently; no diff scope creep. CI: all checks green (lint, 4 vitest shards, bun-test, uat-gate, images-ok).

mekenthompson:
Review findings addressed in 90f4be23 (all six, including lows):

1. **Double-execution guard** — `request()` split into pre/post-delivery phases; retry now fires only on `ensureSession` failure, connect-level errors (`isPreDeliveryError`: ECONNREFUSED/ENOTFOUND/EAI_AGAIN/UND_ERR_CONNECT_TIMEOUT via the undici cause chain, incl. AggregateError), or a 404 session rejection. Timeouts/aborts/mid-response failures after delivery are never re-sent. Regression test counts backend POSTs against a hang-on-tools/call mock: exactly 1.
2. **Head-of-line blocking** — `run()` handles messages concurrently (in-flight Set, drained on input close); stdout stays safe via single-chunk atomic line writes. Regression test: ping answered in <1s while a 1.5s-hung tools/call is pending, and the drain still flushes the hung call's isError result.
3. **listChanged made honest** — after serving a cached/fallback manifest, the first successful upstream call emits `notifications/tools/list_changed` (one-shot), wired through a `notificationSink` that run() binds to stdout. Tested: emitted exactly once on recovery, not while down, not on subsequent calls.
4. Stale comment pointer fixed (drift guard lives in `tests/hindsight-mcp-shim.test.ts`).
5. Tools-only capabilities now documented in `initializeResult()` (no offline fallback exists for prompts/resources; `forward()` still proxies them).
6. No change (noted as fine).

Scoped: `vitest run tests/hindsight-mcp-shim.test.ts tests/memory.test.ts tests/setup.test.ts` — 83 passed (17 shim tests). `npm run lint` clean. Ready for re-review; not merging.

mekenthompson:
Re-review of 90f4be23 — **all six findings verified fixed, MERGEABLE**.

- **Must-fix 1 (double execution)**: retry now three-phased; post-delivery failures never re-sent. `isPreDeliveryError()` cannot misclassify — its code set (ECONNREFUSED/ENOTFOUND/EAI_AGAIN/UND_ERR_CONNECT_TIMEOUT) contains only connect-phase codes; AbortError has no code → false; ECONNRESET/parse failures → false; AggregateError requires *every* sub-error pre-delivery (conservative). Regression test is genuine: hung mock backend + 300ms timeout asserts the backend saw the tools/call **exactly once** — would fail on the old blind retry.
- **Must-fix 2 (head-of-line blocking)**: concurrent handling with an in-flight Set; each response/notification is a single synchronous `write()` of one complete line, so stdout cannot interleave. Drain-on-close preserved: input close stops new tasks, `while (inflight.size) await Promise.allSettled(...)` terminates (finally-deletion and allSettled continuations are both microtasks, so each iteration makes progress). Ping-while-hung test asserts ordering AND latency (<1s).
- **Finding 3**: `listChanged` now honest — one-shot `notifications/tools/list_changed` on stale-manifest → backend-recovery, cleared when a live list is served; once-only asserted.
- **Findings 4–6**: stale test pointer fixed; tools-only capabilities decision documented with rationale; security posture documented.

Residual (non-blocking, noted for the record): concurrent first calls can race duplicate `ensureSession()` initializes — benign against the stateless backend, no side effects.

CI: all checks green. Merge state: CLEAN. Squash-merging.

PR #3314, opened 2026-07-18
feat(routing): deterministic + observable agent→LiteLLM routing (fable demotion split by cause, .routing-mode, boot-card row)

mekenthompson:
## Incident context

2026-07-17 boot race: a slow LiteLLM co-boot left klanker demoted fable→opus on the /anthropic passthrough while overlord landed fable on the router root — **both silent divergence**. klanker has since converged via restart. This PR makes the landing deterministic where possible and observable everywhere.

### 1. Fable demotion split by cause (`profiles/_base/start.sh.hbs`)
- **Missing key** (vault key absent → routing stripped): keeps the opus demotion — fable is a retired codename direct-to-Anthropic and every call 4xxs. Alert wording strengthened (names the cause and the recovery: `switchroom apply` + restart).
- **Unreachable** (key present, probe failed → routing left in place): demotion **removed**. Boots the declared fable model pointed at the LiteLLM **router root** (the sr-*/fable repoint now also fires on the unreachable path), loud-degraded — same policy as sr-* configured defaults. Requests fail until the proxy heals, then work with zero divergence, no restart needed.
- Inner probe budget 120s → **240s**, env-tunable via `SWITCHROOM_LITELLM_BOOT_BUDGET`. The **outer** (gateway-pass) probe stays at 120s — extending it would double gateway boot delay.

### 2. `.routing-mode` state file + stderr line
One overwrite-per-boot line written just before `exec claude`:
`mode=<router-root|passthrough|direct-oauth> base=<url> model=<effective> litellm_ok=<0|1> declared=<mode> ts=<iso>`, echoed to stderr prefixed `routing:`. `_DECLARED_ROUTING` is baked at render time (`src/agents/scaffold.ts`, both start.sh contexts) with the same `isClaudeModel` split `compose.ts` uses for `ANTHROPIC_BASE_URL`; it resolves to `direct-oauth` when LiteLLM was never configured (captured via `SWITCHROOM_LITELLM_DECLARED` before any fail-open strip), so non-LiteLLM fleets never false-alert.

### 3. Divergence alert, deduped
Landed ≠ declared appends to `.session-model-alert` (existing consume-once gateway relay) — only when the landed mode differs from the **previous boot's** `.routing-mode` (read before overwrite), so a persistently degraded agent alerts once, not on every restart.

### 4. Boot-card routing row (`telegram-plugin/gateway/boot-card.ts`)
New `readRoutingMode` / `renderRoutingRow` + a `Routing` row wired into the post-settle render. Tolerant of an absent file (row omitted) and of the render-timing hazard: a `.routing-mode` whose mtime predates this gateway boot renders suffixed **"(prev boot)"** — stale data is never shown as current. Divergence gets the 🟡 dot.

## Deliberately NOT in this PR (follow-up)

**LiteLLM `pass_through_endpoints` change** — live-config op handled operator-side. Security constraint for that follow-up: scope strictly to `/api/oauth/usage` (+ profile), forward the client `Authorization` header unreplaced, and it must **not** create an unauthenticated relay.

## Tests

- `tests/scaffold.routing-mode.test.ts` (11 tests, all outcome-asserting against the RENDERED start.sh): unreachable-fable no longer rewrites to opus and lands on the router root (a rendered script that would demote fails); missing-key still demotes + alerts; `.routing-mode` line shape + `routing:` stderr echo; divergence alert first-boot-only dedupe + re-alert on mode change; converged/never-configured boots stay silent; 240s inner budget + `SWITCHROOM_LITELLM_BOOT_BUDGET` rendered, outer 120s preserved; declared-routing bake per model class.
- `telegram-plugin/tests/boot-card-routing.test.ts` (8 tests): parse, absent file → null, garbage → null, stale-mtime → prevBoot; row render converged/diverged/stale; renderBootCard row present/omitted.
- Existing suites green locally: `scaffold.test.ts` (212), boot-card suites (83), `reconcile.default-model` / `cheap-cron-session-render` / `claude-cli-contract` / `reconcile.skip-profile-templates`. `npm run lint` clean.
- `tests/scaffold.session-model.test.ts`: 7 failures in the fake-`switchroom`-binary "live configured-default resolution" describe are **identical on pristine origin/main** in this environment (verified via a detached main worktree; CLAUDE.md's known fake-binary-harness env class) — CI is the authority there.

### DIFF
--- a/profiles/_base/start.sh.hbs
+++ b/profiles/_base/start.sh.hbs
@@ -96,6 +96,13 @@ if [ "$SWITCHROOM_RUNTIME" = "docker" ] && [ -z "$SWITCHROOM_DOCKER_TMUX_INNER" 
   #
   # SWITCHROOM_AGENT_NAME is injected by compose env (compose.ts:1816)
   # so it is available here, before the inner-pass export at line ~320.
+  #
+  # Capture whether LiteLLM routing was configured at compose time BEFORE the
+  # missing-key fail-open below can strip SWITCHROOM_LITELLM. The routing-mode
+  # observability block (inner pass, pre-exec) reads this to distinguish
+  # "litellm never configured" (declared intent IS direct-oauth — no
+  # divergence) from "configured but stripped at boot" (real divergence).
+  export SWITCHROOM_LITELLM_DECLARED="${SWITCHROOM_LITELLM_DECLARED:-${SWITCHROOM_LITELLM:-}}"
   if [ -n "${SWITCHROOM_LITELLM:-}" ] && [ -z "$ANTHROPIC_CUSTOM_HEADERS" ] && command -v switchroom >/dev/null 2>&1; then
     sr_ll_key="$(switchroom vault get "litellm/$SWITCHROOM_AGENT_NAME/api-key" 2>/dev/null || true)"
     sr_ll_ok=""
@@ -1096,7 +1103,20 @@ fi
 # unreachable at boot, or claude would 4xx an unknown model against Anthropic.
 # Empty by default (proxy not confirmed live); set to "1" only on a successful
 # liveliness probe. Deliberately NOT unset with the sr_ll_* scratch vars.
+#
+# _LITELLM_UNREACHABLE distinguishes WHY _LITELLM_OK is empty (routing-cause
+# split, 2026-07-17 boot-race incident): "1" means the key was present and the
+# probe failed — routing was LEFT IN PLACE pointing at the proxy, so a
+# proxy-only model (fable/sr-*) can boot loud-degraded and self-heal per
+# request. Empty means either the key is MISSING (routing stripped — a
+# proxy-only model genuinely cannot work and must demote) or LiteLLM is not
+# configured at all (same conclusion). Deliberately NOT unset either.
 _LITELLM_OK=""
+_LITELLM_UNREACHABLE=""
+# Idempotent mirror of the outer-pass capture (first setter wins) — on the
+# non-docker path only this inner block runs, so capture here too, before the
+# missing-key fail-open below can strip SWITCHROOM_LITELLM.
+export SWITCHROOM_LITELLM_DECLARED="${SWITCHROOM_LITELLM_DECLARED:-${SWITCHROOM_LITELLM:-}}"
 if [ -n "${SWITCHROOM_LITELLM:-}" ] && command -v switchroom >/dev/null 2>&1; then
   sr_ll_key="$(switchroom vault get "litellm/$SWITCHROOM_AGENT_NAME/api-key" 2>/dev/null || true)"
   sr_ll_ok=""
@@ -1112,10 +1132,14 @@ if [ -n "${SWITCHROOM_LITELLM:-}" ] && command -v switchroom >/dev/null 2>&1; th
     # co-boots, the litellm proxy's heavy Python app is often not yet healthy
     # in the first few seconds. A single one-shot probe fails-open and takes
     # this agent dark (untracked) for the WHOLE session. Poll the liveliness
-    # endpoint every ~3s for up to 120s (hard budget) and only fall open if
+    # endpoint every ~3s for up to SWITCHROOM_LITELLM_BOOT_BUDGET seconds
+    # (default 240 — the INNER budget only; the outer/gateway-pass probe stays
+    # at 120s so gateway boot latency is not doubled) and only fall open if
     # it is STILL unreachable after that whole window.
     sr_ll_url="${SWITCHROOM_LITELLM_BASE:-${ANTHROPIC_BASE_URL%/anthropic}}/health/liveliness"
-    sr_ll_deadline=$(( $(date +%s) + 120 ))
+    sr_ll_budget="${SWITCHROOM_LITELLM_BOOT_BUDGET:-240}"
+    case "$sr_ll_budget" in *[!0-9]*|"") sr_ll_budget=240 ;; esac
+    sr_ll_deadline=$(( $(date +%s) + sr_ll_budget ))
     sr_ll_up=""
     while :; do
       if curl -fsS -m 5 -o /dev/null "$sr_ll_url" 2>/dev/null; then sr_ll_up="1"; break; fi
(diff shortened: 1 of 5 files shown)


mekenthompson (commented):
## Adversarial review — REQUEST-CHANGES

**Verdict: DO-NOT-MERGE** — one critical false-positive that defeats the PR's own headline scenario, plus two lows. Everything else verified sound (details below).

### CRITICAL-1: healthy configured-fable boots read as permanently diverged

`declaredRoutingQ` is baked with `isClaudeModel(resolveMainModel(model)) ? "passthrough" : "router-root"` (src/agents/scaffold.ts:3250 and 6471). `isClaudeModel('fable')` is **true** (`fable` is in `MODEL_ALIASES`, telegram-plugin/gateway/model-command.ts:51,81-85), so a configured-fable agent gets `_DECLARED_ROUTING='passthrough'`. But start.sh's own repoint case (`sr-*|fable|claude-fable-5`, start.sh.hbs:1530-1535) deliberately lands every healthy fable boot on the **router root**.

Reproduced against the rendered script (scaffold `model: fable`, `_LITELLM_OK=1`, healthy path):

```
.routing-mode: mode=router-root base=http://litellm:4010 model=fable litellm_ok=1 declared=passthrough
.session-model-alert: Routing divergence: this agent landed on `router-root` … but its declared routing is `passthrough`. …
```

So the exact fleet this PR exists for (pinned `model: fable`, LiteLLM healthy) fires a spurious "Routing divergence" operator alert on first boot and renders a permanent 🟡 `router-root ≠ declared passthrough` boot-card row forever after. The observability surface cries wolf on the healthy case, which trains operators to ignore the row that's supposed to catch the next real incident.

Fix: declared intent must mirror the template's repoint split, not compose's env split — fable/claude-fable-5 (and sr-*) declare `router-root`; only non-fable Claude models declare `passthrough`. (compose.ts:2295 putting fable's *env* on the passthrough is fine — start.sh repoints — but "declared" is compared against the *landed* mode, so it must encode the post-repoint intent.)

Test gap that let this through: `"fable + LiteLLM up boots fable via the router root"` asserts EFFECTIVE/BASEURL but never `alertText()`; the `.routing-mode` write test asserts only `declared=[a-z-]+`. Add: healthy-fable boot asserts `declared=router-root` and no divergence alert.

### LOW-1: "(prev boot)" mislabels a gateway-only restart

`routingBootStartMs = Date.now() - process.uptime()*1000` (boot-card.ts, startBootCard). If the gateway process is respawned by the supervisor without a container restart, the still-current `.routing-mode` (claude session unchanged) has mtime < new gateway start → labeled "(prev boot)". Conservative direction (understates freshness, never overstates), so acceptable if documented — but the doc comment claims mtime-older ⇒ "belongs to the PREVIOUS boot", which is not true in this order. At minimum fix the comment; ideally key staleness on container boot time (e.g. /proc/1 start) rather than gateway process start.

### LOW-2: divergence dedup keys only on landed mode, not declared

start.sh.hbs:1568-1570: alert fires when `landed≠declared && landed≠prev`. If the operator changes the configured model class (declared flips) while the agent stays stuck on the same wrong landed mode, no new alert — the mode didn't change. Edge case; consider including declared in the dedup key (e.g. store `mode+declared` in the compare).

### Verified sound (evidence)

- **Flag interplay / no-write paths**: single writer — only the inner pass runs the routing block; the outer docker pass `exec tmux`s at line 391, well before it. All four `exec claude` branches (1577-1585) sit after the `.routing-mode` write (1546-1570). Shebang is bash, no `set -u`; all new vars initialized before use (`_LITELLM_UNREACHABLE` at 1077, `SWITCHROOM_LITELLM_DECLARED` captured before the strip on both docker (line 105) and non-docker (1081) paths, first-setter-wins so tmux env inheritance is safe). `litellm_ok=${_LITELLM_OK:-0}` renders 0 correctly.
- **Never-configured fleets**: `SWITCHROOM_LITELLM_DECLARED` empty → declared forced `direct-oauth`, no false alert — covered by a real outcome test.
- **Repoint-on-unreachable**: header + `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY` + `ANTHROPIC_DEFAULT_FABLE_MODEL` are all exported on the `sr_ll_unreachable` branch too (1131-1163), so fable actually resolves and auths once the proxy heals; the sr-* precedent comment matches the code. One honest caveat: I could not black-box-verify the claude CLI's behavior against a refused router root at process start (discovery soft-fail is asserted by template comments, not tested here) — worth one manual degraded-boot smoke before rollout.
- **Dedup ordering**: `_prev_routing_mode` is read (1566) before the overwrite (1568); no outer-pass writer exists, so no race and no self-suppression on the first degraded boot.
- **Budget**: inner default 240s, digit-validated fallback (`case *[!0-9]*|""`), outer probe verified still hard-coded 120s (line 120).
- **Both render contexts patched identically** (scaffold.ts 3250 / 6471); `declaredRoutingQ` has exactly these two producers and one template consumer; all profiles extend `_base`, no non-claude profile renders the template.
- **Tests are outcome tests, not theater**: they execute the rendered block in bash and assert `EFFECTIVE=fable` / `BASEURL=<router root>` / file+alert contents — a regression back to unconditional demotion fails them. Ran locally: 11/11 scaffold + 8/8 boot-card pass on the PR head.
- **Pre-existing failures claim verified**: `tests/scaffold.session-model.test.ts` fails 7/44 on the PR head **and** 7/44 with origin/main's start.sh.hbs+scaffold.ts+test swapped in — genuinely pre-existing, not introduced.
- **Scope**: 5 files, all on-topic; no creep. **CI: green** (all checks pass, none pending/failed).

Fix CRITICAL-1 (+ the two test assertions), then this is mergeable; the shell mechanics are otherwise solid.

mekenthompson:
Addressed the adversarial review in `d9a3b643` (rebased on fresh main):

- **CRITICAL-1**: added `declaredRoutingMode()` (scaffold.ts) — declared intent now mirrors start.sh's post-repoint landing (`fable`/`claude-fable-5`/`sr-*` → `router-root`; other Claude models → `passthrough`). A healthy configured-fable boot now records `declared=router-root` and fires no divergence alert. Both render sites updated.
- **LOW-1**: boot-card staleness keys on container (PID 1) boot time via new `containerBootStartMs()`, stable across gateway-only respawns, falling back to gateway process start when /proc is unavailable. Comments corrected.
- **LOW-2**: divergence dedup now keys on the `(landed, declared)` pair, so a declared-class flip re-alerts even when the landed mode stays stuck.

Tests added for all three (scaffold.routing-mode + boot-card-routing, 24 pass); `npm run lint` clean.

PR #3364, opened 2026-07-18
feat(rollout): gate scoped sweep on canary via per-invocation env (#3333 stage 4)

mekenthompson:
## Stage 4 of #3333 — scoped agent-home ownership sweep

Stacked on #3363 (stage 3). Wires the scoped sweep to the rollout canary so it can be observed on one agent before the default flips ON.

### Amendment C — per-invocation env, not global
The reconcile sweep runs **inside** the `switchroom agent restart` process the rollout spawns (`deps.run(restartArgs)`). So injecting `SWITCHROOM_SCOPED_OWNERSHIP_SWEEP=1` into **only** the `isCanary` spawn's env flips exactly one agent per roll — the first-restarted canary — and the sweep reads it there. A global / container-boot env would flip the whole fleet at once and defeat the canary gate, so it is deliberately not that.

- `RolloutDeps.run` gains an optional `{ env }` that layers onto the spawned process env; production impl merges over `process.env`
- the `restart-agent` step passes the flag env only when `isCanary`

### Verify on the roll (before stage 5)
The canary's restart-time collapse (`scope=scoped` in its `[ownership-sweep] #3333` log from stage 1) with no approval-card storm.

### Test
`src/cli/rollout.test.ts` — the canary (first restart) spawn carries `{ SWITCHROOM_SCOPED_OWNERSHIP_SWEEP: "1" }`, every subsequent agent's spawn and all non-restart steps (apply/webd/hostd) do not.

### Not in this PR (follow-up issues)
- **Stage 5** (flip default ON + one-release escape hatch) — waits for a real canary-roll observation
- **Stage 6** (apply-path twin `alignAgentUid`, amendment D) — after bake

Refs #3333 (does not close — closes when stage 5 default-ON ships).

### DIFF
--- a/src/cli/rollout.test.ts
+++ b/src/cli/rollout.test.ts
@@ -34,6 +34,7 @@ import {
   type RolloutResult,
   type RolloutStep,
 } from "./rollout.js";
+import { SCOPED_SWEEP_ENV } from "../agents/agent-owned-tree.js";
 
 describe("normalizeVersion", () => {
   it("strips a leading v and trims so config pin == in-container version", () => {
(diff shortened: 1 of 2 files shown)


mekenthompson:
**HOLD — do not merge yet (deliberate).**

Adversarial review verdict on this PR is APPROVE, but the reviewer confirmed a rollout-behavioural consequence: the moment this merges, **every subsequent roll runs the scoped ownership sweep on that roll's canary agent** (the per-invocation env is injected into the first restart spawn unconditionally — there is no separate operator switch). That is exactly the designed stage-4 behaviour, but it means merging IS starting the canary observation.

So this PR lands only when we deliberately begin the canary observation window: merge it, run a roll, and watch the canary's `[ownership-sweep] #3333 ... scope=scoped` line for the restart-time collapse and confirm no approval-card storm. That observation is the gate for stage 5 (#3365, default ON — which closes #3333).

Branch is rebased on the current stack (stages 2+3) and CI-green; nothing else blocks it.

PR #3370, opened 2026-07-18
perf(reconcile): scoped ownership sweep behind flag, default OFF (#3333 stage 2)

mekenthompson:
## Stage 2 of #3333 — scoped agent-home ownership sweep

Stacked on #3360 (stage 1). Adds `chownScopedTree` behind the `SWITCHROOM_SCOPED_OWNERSHIP_SWEEP` env flag, **default OFF** — zero behaviour change until the canary flip (stage 4).

### Amendment A shape
- **SHALLOW** chown (no `-R`) of the agent dir + every direct child (files, symlinks, subdir inodes) — captures `start.sh`, `CLAUDE.md*`, `cron-session.sh`, `.mcp.json`, `.resume-mode-migration-warned`, `SOUL.md` regardless of name
- recursive `chown -h -R` **only** into the scoped state subdirs that exist: `SCOPED_RECURSIVE_SUBDIRS = {.claude/, .claude-cron/, telegram/}`
- `home/` `workspace/` `tmp/` are never traversed (home/ is ~96% of inodes)
- the walk is **never hand-rolled**: `readdirSync` reads ONE level for the shallow pass, recursion is `chown -h -R`; `-h` symlink semantics preserved throughout

Folds in finding-1's leak list: `cron-session.sh` and `.resume-mode-migration-warned` are top-level files captured by the shallow pass; `workspace/CLAUDE.custom.md` stays excluded (per amendment A, workspace/ is not swept — it is world-readable today, and the widened assert in stage 3 catches it if it ever becomes critical).

`SCOPED_RECURSIVE_SUBDIRS` is the single source of truth for scope-of-sweep; stage 3's widened assert derives its candidate set from the same constant (scope-of-sweep == scope-of-assert).

### Tests
`tests/reconcile.scoped-sweep.test.ts`:
- call shape — shallow path list + recursion only into existing scoped subdirs, excluded caches never a recursion root
- flag gating — OFF = full sweep (unchanged), ON = scoped
- env parsing (default OFF)
- root-gated real outcomes — planted `.claude` 0600 re-owned, planted `home/` 0600 left, `-h` symlink target not dereferenced
- non-root `-h` pin (review fix 2) — real `chownShallow` over a dangling symlink to our own uid succeeds only because of `-h`; fails on `-h` removal on the non-root CI runner

### Residual risk (documented per review fix 3)
A future root writer that creates a NEW top-level **subdirectory** not in `SCOPED_RECURSIVE_SUBDIRS` gets its dir inode re-owned by the shallow pass, but its **contents** are neither swept nor asserted unless the write site pushes the paths into `changes` (which the union assert then covers). `SCOPED_RECURSIVE_SUBDIRS` is therefore a **maintained invariant**: anyone adding a new root-written state subdir under the agent home must add it to the constant (which automatically widens both sweep and assert, since both derive from it). Same discipline as the agent-owned-tree.ts "rule for future scaffold/reconcile authors" doc block.

Refs #3333.

### DIFF
--- a/src/agents/agent-owned-tree.ts
+++ b/src/agents/agent-owned-tree.ts
@@ -57,8 +57,38 @@
  */
 
 import { execFileSync } from "node:child_process";
-import { existsSync, lstatSync, statSync } from "node:fs";
+import { existsSync, lstatSync, readdirSync, statSync } from "node:fs";
+import { join } from "node:path";
 import { allocateAgentUid } from "./agent-uid.js";
+
+/**
+ * The agent-dir SUBDIRECTORIES the scoped sweep recurses into (#3333
+ * amendment A). Root reconcile writers land inside these state dirs
+ * (`.claude/` — settings.json + subagents + backups; `.claude-cron/` —
+ * the cron .mcp.json; `telegram/` — the inbox). Everything else at the top
+ * level is a single file (start.sh, CLAUDE.md*, cron-session.sh,
+ * .mcp.json, .resume-mode-migration-warned, the SOUL.md symlink) captured
+ * by the top-level SHALLOW chown; the giant caches that make the full
+ * sweep slow (home/ ~96% of inodes, workspace/, tmp/) are deliberately
+ * NOT recursed into.
+ *
+ * This is the single source of truth for scope-of-sweep; the widened
+ * post-sweep assert (#3333 stage 3) derives its candidate set from the
+ * same constant so scope-of-sweep == scope-of-assert.
+ */
+export const SCOPED_RECURSIVE_SUBDIRS = [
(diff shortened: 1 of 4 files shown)


mekenthompson:
Supersedes #3362 — that PR was auto-closed by GitHub when its stacked base branch (`fix/3333-instrument-sweep`, PR #3360) was deleted on merge, and cannot be reopened after the review-fix force-push. Same branch, same content plus the review fixes: non-root `-h` pin test (fix 2) and residual-risk documentation (fix 3, in the body above). Reviewer verdict on #3362 was APPROVE-WITH-FIXES; both fixes are in.

PR #3435, opened 2026-07-19
recall: strip <channel> envelope from query + breach telemetry (A1, #3431)

mekenthompson:
PR 1 of the hindsight leverage programme (epic #3430; workstream A child #3431). Single concern: **recall query hygiene + observability**. Implements A1 (envelope strip on both query paths) plus the A3 stage-1 telemetry fields (moved forward from PR 3 so a fresh pre-parallelism breach baseline can start accruing ≥3 days before PR 3 merges).

## Problem

- The recall query on the single-turn path re-used the **raw prompt** (`recall.py` `else: query = prompt`), so ~100–200 chars of `chat_id`/`ts`/`user`/`message_id` XML from the `<channel …>` transport envelope entered the embedding and consumed the `recallMaxQueryChars` (800) cap.
- `compose_recall_query` did **not** fix this: it stripped context messages only (`content.py:96-98`) and returned/appended the latest query **raw** (`:83-84`, `:110-115`). So the bug lived on **both** the single-turn and multi-turn paths.
- `recall_log.jsonl` carried no per-bank latency or timeout signal, so there was no way to establish the fresh pre-A3 breach baseline the parallelism change (PR 3) measures against. The oft-quoted 17–26% figure is the stale 2026-05-24 pre-fix audit, not a valid before-number.

## Change

- **`recall.py`** — strip the envelope **once** (`strip_channel_envelope`) before both branches; feed the stripped text to `compose_recall_query`, the single-turn assignment, and `truncate_recall_query`'s `latest_query` arg. The raw `prompt` is left untouched for the ack/nudge gates and the cache-hit log-length field.
- **`content.py`** — `compose_recall_query` strips `latest_query` internally too, so the helper is envelope-safe for any future caller (defence in depth, e.g. the A2 `contextTurns` flip).
- **Telemetry** — each non-cache recall log now carries per-bank `{elapsed_ms, timed_out}`, `directives_elapsed_ms`, `total_elapsed_ms`, and a derived `deadline_hit` (any bank hit its hard per-request timeout). A new `_is_timeout_error` classifier distinguishes a deadline hit from a fast 5xx/refused failure. The stripped query text is recorded in the log so the query-hygiene acceptance check can assert directly.
- Cache-hit log rows carry the same telemetry keys (nulls / empty) for a uniformly queryable schema.

## Acceptance evidence

- **New `tests/test_recall_envelope_strip_telemetry.py` (15 tests):**
  - single-turn wrapped prompt produces a query identical to the bare prompt; no `<channel`/`chat_id`/`ts=` substrings;
  - **multi-turn composed query** (regression guard for the future `contextTurns` flip) has no `<channel` in the `Prior context:` lines or the trailing latest-query segment — both via the helper directly and end-to-end through `main()` with an on-disk transcript;
  - logged query has no envelope (`e["query"] == bare`);
  - telemetry fields present; per-additional-bank timing record; `deadline_hit` **true** on a bank `socket.timeout`, **false** on a `RuntimeError("HTTP 503")`;
  - `_is_timeout_error` unit coverage (socket.timeout / TimeoutError / URLError-wrapped / message-sniff / non-timeout).
- **Scoped suite green:** `python3 -m unittest discover tests/` → **293 passed**.
- **Lint:** `check-no-pii-secrets`, `check-plugin-references`, `check-stale-tool-descriptions`, `check-vault-test-hermeticity` all clean (synthetic `chat_id` used, no operator PII). `tsc`/bun-dependent guards not run locally (Python-only change, no TS touched, deps not installed offline) — CI is the authority there.

## Deviations from the design

- The A1 acceptance requires "the query recorded in `recall_log.jsonl` contains no `<channel`", but the log previously stored only `query_chars` (length). Added an explicit `query` field holding the envelope-stripped, truncated query text so the assertion (and D2's later query↔hit pairing) is possible. Query is already capped at `recallMaxQueryChars` (≤800).
- `deadline_hit` is defined here as "any per-bank hard-timeout" (there is no shared deadline until A3 stage-2); documented inline. Also added `directives_elapsed_ms` since the directives fetch is on the same critical path A3 will parallelise.

Do **not** merge until adversarial review + CI green.

### DIFF
--- a/vendor/hindsight-memory/scripts/lib/content.py
+++ b/vendor/hindsight-memory/scripts/lib/content.py
@@ -79,7 +79,14 @@ def compose_recall_query(
 
         <latest query>
     """
-    latest = latest_query.strip()
+    # Switchroom A1 (hindsight-leverage PR 1) — strip the <channel> envelope
+    # from the latest query INSIDE the helper as well, so any caller (present
+    # or future) gets an envelope-free composed query. The recall.py caller
+    # already strips before calling, but keeping the strip here is defence in
+    # depth: it guarantees the trailing latest-query segment appended below
+    # (and returned on the turns<=1 short-circuit) never carries the raw
+    # chat_id/ts/user XML noise into the embedding or the char cap.
+    latest = strip_channel_envelope(latest_query).strip()
     if recall_context_turns <= 1 or not isinstance(messages, list) or not messages:
         return latest
 
--- a/vendor/hindsight-memory/scripts/lib/content.py
+++ b/vendor/hindsight-memory/scripts/lib/content.py
@@ -33,11 +33,17 @@ def strip_channel_envelope(content: str) -> str:
     This is the Claude Code equivalent of Openclaw's stripMetadataEnvelopes().
     Extracts the inner text, preserving the actual user message while removing
     transport metadata that Hindsight doesn't need.
+
+    A single prompt may carry MORE THAN ONE envelope (e.g. a coalesced
+    burst where several inbound messages were concatenated). Since this now
+    sits on the live recall-query path (hindsight-leverage PR 1, review
+    finding 6), coalesce EVERY envelope's inner text rather than keeping only
+    the first and silently dropping everything after the first ``</channel>``.
     """
-    # Match <channel ...>content</channel> — extract inner text
-    match = re.search(r"<channel\b[^>]*>([\s\S]*?)</channel>", content)
-    if match:
-        return match.group(1).strip()
+    # Match every <channel ...>content</channel> — extract & join inner texts.
+    matches = re.findall(r"<channel\b[^>]*>([\s\S]*?)</channel>", content)
+    if matches:
+        return "\n".join(m.strip() for m in matches if m.strip()).strip()
     return content
 
 
--- a/vendor/hindsight-memory/scripts/recall.py
+++ b/vendor/hindsight-memory/scripts/recall.py
@@ -176,6 +176,15 @@ def _summarise_source_topics(results: list) -> dict:
 # `switchroom memory recall-log <agent>`.
 RECALL_LOG_FILE = "recall_log.jsonl"
 RECALL_LOG_MAX_LINES = 5000
+# Honest per-line upper-bound estimate for the size-gated trim (hindsight-
+# leverage PR 1, review finding 5). The A3-telemetry row (bounded query
+# excerpt ≤200 chars + bank_timings + memory_ids) runs ~700-900 bytes worst
+# case; 1024 is a safe upper bound. The size gate only reads+trims the file
+# when it plausibly exceeds the line cap, so it must OVER-estimate row size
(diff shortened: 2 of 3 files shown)


mekenthompson:
## Review round 1 addressed — commit `652c2d5c`

All six findings fixed in the same branch (repo protocol: lows fixed too).

**1 (MAJOR) — total-failure turns now log.** The `_write_recall_log` call was hoisted **above** the `not directives_block and not memories_block` early-return, so a turn where every bank times out and directives fail now records `bank_timings` / `deadline_hit` / `total_elapsed_ms`. New test `test_own_bank_timeout_no_directives_still_writes_row` asserts the row exists with `deadline_hit: true` and no injected block.

**2 (MAJOR) — ceiling-kill denominator documented.** A hook the UserPromptSubmit ceiling kills mid-flight (sequential worst case ~18s > 12s) never reaches the writer, so a true ceiling breach is a **missing row**, not a logged one. Documented inline at the telemetry write and stated here for PR 3's baseline method:

> **PR-3 baseline is two-signal.** (a) **Ceiling-breach (total-drop) rate** = recall-log **rows-missing vs the UserPromptSubmit turn count** over the window. (b) **`deadline_hit`** = per-bank hard-timeout **among the hooks that survived long enough to log**. Neither number alone is the breach rate; the before/after must cite both.

**3 (MINOR) — cache-hit schema honesty.** Cache-hit rows now carry `deadline_hit: null` (not `false`) alongside the None/[] siblings: `false` = "banks ran, none timed out"; `null` = "no banks ran". Added `test_cache_hit_row_has_none_telemetry`.

**4 (MINOR) — HTTP 502/504 no longer misclassified.** `_is_timeout_error` now rules out `HTTPError` (raised directly, via `__cause__`, and via lib.client's `HTTP <code> from …` wrapper) **before** the `"timed out"` message sniff, so a proxy body saying "upstream timed out" is not counted as a client deadline hit. Added three classifier tests.

**5 (MINOR) — trim thrash fixed.** The stored `query` is now a **≤200-char excerpt** (`query_chars` keeps the full length), and the size-gated trim uses an honest per-line **upper bound** (`RECALL_LOG_BYTES_PER_LINE_EST = 1024`) so it no longer does a full-file read on every hook once rows grow past the old 250 B/line assumption. Added `test_query_field_is_bounded_excerpt`.

**6 (MINOR) — multi-envelope strip.** `strip_channel_envelope` (now load-bearing on the live query path) coalesces **every** envelope's inner text instead of keeping only the first and dropping everything after the first `</channel>`. Added `MultiEnvelopeStrip` (single/bare unchanged, 2- and 3-envelope coalesced).

**Tests:** full plugin suite `python3 -m unittest discover tests/` → **303 passed** (was 293; +10 new). Lint guards (`check-no-pii-secrets`, `check-plugin-references`, `check-stale-tool-descriptions`, `check-vault-test-hermeticity`) clean. Holding for full CI green before any merge; not merging.

PR #3437, opened 2026-07-19
fix(model): rev5 carrier auto-remediation + test hardening (#3427)

mekenthompson:
Implements all four items of #3427 (follow-ups to #3424, which shipped detection + honesty only).

## 1. Auto-remediation (primary) — DONE
`switchroom doctor --fix` heals rev5 `.session-model` carrier drift instead of only reporting it:

- New module `src/cli/doctor-fix-session-model.ts`: for each agent, run the #3424 tripwire (`checkStartShSessionModelCarrier`); on `fail`/`warn` regenerate `start.sh` via the standard per-agent **reconcile** (`reconcileAgent(…, { preserveClaudeMd: true })` — the same code path `switchroom apply` uses), then **re-run the check** and report the post-fix status. The heal claim is always a re-run check, never an assumed success.
- **Deterministic + idempotent**: healthy agents are never reconciled (proven by test — second run performs zero reconcile calls, content + mtime unchanged). `start.sh` is purely template-driven ("no user-merged sections, safe to overwrite" — `scaffold.ts`), so full regeneration IS the sanctioned mechanism; no operator content exists in it to clobber. `preserveClaudeMd` keeps the fix scoped away from CLAUDE.md.
- Why not option (a) (boot-time self-heal): the `profiles/_base/` template tree is **not shipped into the agent image** (see `ReconcileOptions.skipProfileTemplates` in `scaffold.ts`) — an agent cannot re-render its own start.sh in-container. The host-side doctor is the earliest place the drift signal and the template coexist.
- The `--fix` remediation section runs before the standard sections, so the Agents section reports post-fix state; check fix strings and `docs/cli-reference.md` now advertise `switchroom doctor --fix`.

## 2. Behavioral tests over source-string assertions — DONE
- The card-vs-suppress decision moved out of the gateway boot IIFE into pure `resolveModelSwitchBootNotice` (model-command.ts) — line-ratchet friendly and unit-testable.
- `gateway-session-model-relaunch.test.ts`'s `toContain`-on-source tests are replaced by behavioral tests that run the real `classifyModelSwitchConfirmation → formatModelRelaunchDiagLog / resolveModelSwitchBootNotice` pipeline and assert the **emitted** outcomes: `NOT-APPLIED` / `outcome=applied` / `outcome=default` log lines, the green applied/default cards, the ⚠️ silent-revert card (with the back-ticked re-issue hint), and the LOW-2 suppress arm (including "applied + alert still confirms"). One thin wiring pin remains for the un-bootable gateway IIFE (documented as such).

## 3. Fix the failing `scaffold.session-model.test.ts` subset — DONE (root-caused, not skipped)
- Root cause: the fake `switchroom` CLI was written under `os.tmpdir()`; `/tmp` is mounted **noexec** in the dev container, so every live read exited 126 (stderr suppressed by the launcher) and fell back to the bake — failing the 7 tests that expect the live value. Repo already has a convention for this (`/var/tmp`, see `turn-pacing-hook.test.ts`, `host-control/server.test.ts`); the fake bin now lives there. **45/45 green** (was 38/45), and runtime dropped 33s → 8s (no more retry timeouts).

## 4. Requested-vs-served divergence window — DONE (in this PR, right-sized)
Two deterministic surfaces replace the silent self-heal:
- **Immediate (command time)**: the `/model` ack for a free-text full `claude-*` id carries a caveat that the id cannot be pre-validated (Claude-native constraint — no raw API probe) and that `--fallback-model` may substitute. Aliases / sr-* / menu-picker tokens (vouched by their own gates) are unaffected.
- **First-output tripwire**: `createSessionModelSource` now verifies the FIRST post-override transcript line against the requested token via `servedModelMatchesRequested` (conservative comparator — alias↔family, full-id↔date-stamped-descendant, never accuses non-comparable pairs like sr-*/labels). On mismatch the gateway logs greppable `gw /model served-model DIVERGENCE …`, **drops the bogus override** (so /status + menu stop claiming it), and sends a one-shot ⚠️ card to the initiating chat.
- RFC `session-model-stickiness.md` "Determinism scope" updated to the new loud contract.
- True pre-launch validation of arbitrary ids would need an API probe, which the Claude-native pillar forbids — this is the tightest deterministic closure available; noted in code comments.

## Tests & lint
- New: `tests/doctor-fix-session-model.test.ts` (6) — real scaffold → mangle to live pre-rev5 legacy shape → real `reconcileAgent` heal → tripwire passes; idempotency (no reconcile calls, no rewrite); missing-start.sh heal; reconcile-throw honesty; still-failing-after-write honesty; skip passthrough.
- New behavioral suites in `model-command.test.ts` (+10), `session-model-source.test.ts` (+7), rewritten `gateway-session-model-relaunch.test.ts`.
- Scoped run: **11 files, 288 passed | 2 skipped** (`doctor.test.ts`, `doctor-fix-session-model.test.ts`, `scaffold.session-model.test.ts`, `check-gateway-line-ratchet.test.ts`, `gateway-session-model-relaunch.test.ts`, `session-model-source.test.ts`, `model-command.test.ts`, `session-model-file.test.ts`, `bot-commands-model-effort.test.ts`, `narrative-lane-golden.test.ts`, `stream-render-golden.test.ts`).
- `npm run lint` exit 0 (tsc + all 12 guard scripts; gateway line ratchet clean at 25099 ≤ 25053+50).
- `npm run build` clean.

Closes #3427

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,9 +32,13 @@
   (idempotent; healthy agents untouched; takes effect on the agent's next
   restart). The requested-vs-served `--fallback-model` masking window is now
   loud: the `/model` ack warns immediately for unvalidatable free-text
-  `claude-*` ids, and the first assistant line after an apply-boot is verified
-  against the requested token — a mismatch logs `gw /model served-model
-  DIVERGENCE`, drops the bogus override, and warns the initiating chat. The
+  `claude-*` ids, and the first LIVE assistant line after an apply-boot is
+  verified against the requested token — a mismatch logs `gw /model
+  served-model DIVERGENCE` and warns the initiating chat, naming both
+  candidate causes (invalid id or transient unavailability). Verification
+  arms only at the boot-rehydration site and skips first-attach replay lines
+  from the pre-relaunch session, so a valid mid-turn switch can never be
+  false-accused. The
   #3424 source-string (`toContain`) tests are converted to behavioral tests of
   the classify → log/card pipeline, and the `scaffold.session-model.test.ts`
   live-resolution subset red on `main` is fixed (the fake `switchroom` CLI now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,20 @@
   process writes (and sub-agent / Bash-CLI writes with no main-transcript
   tool_use) are at most TTL stale. Saves the 2s-timeout `list_directives`
   round-trip on cache-hit turns. (hindsight-leverage A4)
+- **`/model` rev5 carrier: auto-remediation + test hardening** (#3427, follow-up
+  to #3424) — `switchroom doctor --fix` now HEALS rev5 `.session-model` carrier
+  drift instead of only detecting it: each drifted/missing `start.sh` is
+  regenerated from the current template via the standard per-agent reconcile
+  (idempotent; healthy agents untouched; takes effect on the agent's next
+  restart). The requested-vs-served `--fallback-model` masking window is now
+  loud: the `/model` ack warns immediately for unvalidatable free-text
+  `claude-*` ids, and the first assistant line after an apply-boot is verified
+  against the requested token — a mismatch logs `gw /model served-model
+  DIVERGENCE`, drops the bogus override, and warns the initiating chat. The
+  #3424 source-string (`toContain`) tests are converted to behavioral tests of
+  the classify → log/card pipeline, and the `scaffold.session-model.test.ts`
+  live-resolution subset red on `main` is fixed (the fake `switchroom` CLI now
+  lives on an exec-capable filesystem — `/tmp` is noexec in the dev container).
 
 ## v0.19.2 — Hindsight recovery, External spend on /usage, auth-broker hardening, /model carrier doctor
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,10 @@
   `claude-*` ids, and the first LIVE assistant line after an apply-boot is
   verified against the requested token — a mismatch logs `gw /model
   served-model DIVERGENCE` and warns the initiating chat, naming both
-  candidate causes (invalid id or transient unavailability). Verification
+  candidate causes (invalid id or transient unavailability). The override is
+  KEPT on divergence (M2): a transient substitution self-corrects and /status
+  freshness already surfaces the served model, so the card only advises
+  re-issuing `/model <valid id>` (or `/model default`) if it persists. Verification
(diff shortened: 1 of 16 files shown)


mekenthompson:
Review findings addressed in b39c35e8 — all four fixed in place (item 4 not severed):

**H1 (blocker)** — verification arming is now an explicit opt-in: `setOverride(model, { verify: true })`, passed ONLY by the boot-rehydration site. The command-time `scheduleModelRelaunch` record and both rollback restores stay unarmed (a plain set also *clears* any pending arm — newest write wins, unarmed). Behavioral tests: "H1: a command-time setOverride (no verify) NEVER arms" (old-model line in the pre-restart window after a valid switch → zero fires, override intact) and "H1: a verify-arm is CLEARED by a later plain set". Gateway call-site pins added.

**H2 (blocker)** — session-tail now records the first-attach replay window (`replayUntilByFile`, set when `computeFirstAttachCursor` returns an offset below size-at-attach) and marks model events read from it `replayed: true` (chunk-granular, conservative: over-marking a boundary batch only delays verification to the next live line — it can never accuse). `stream-render` passes the flag; the source records replayed lines for /status freshness exactly as before but excludes them from verification. Covered by two integration tests against the REAL tailer in `session-tail-first-attach.test.ts` (mid-turn-restart replay → flagged old-model event, later live line unflagged; completed-turn attach → nothing flagged) plus source-level tests asserting NO fire on the valid-switch boot-replay flow and that a genuinely bogus id still fires on the first live line.

**M2** — log + card now name both candidate causes (invalid/unknown id OR transient unavailability), card says a transient substitution self-corrects and to re-issue only "if it persists"; the handler no longer calls `setOverride(null)` — the override record survives (freshness already points /status at the served model). RFC + changelog updated to the corrected contract.

**M1** — `deps.reconcile` now returns `ReconcileResult.changes` and the heal row discloses them verbatim ("Full per-agent reconcile applied — rewrote N managed file(s): …", capped listing), and the `--fix` help text states it is the FULL per-agent reconcile with staged yaml edits landing. Integration test asserts the disclosure names start.sh among the changes.

**Nits** — L2: commander wiring test (doctor registers `--fix` with the full-reconcile disclosure). L3: `servedModelMatchesRequested` accepts legacy shapes (`claude-3-opus-20240229` for alias `opus`) via a dash-segment fallback that only widens toward "match". afterEach `rmSync` guards added.

Scoped run: 16 files, **404 passed | 2 skipped** (incl. all session-tail suites + goldens). `npm run lint` exit 0 (gateway ratchet clean at 25103 = ceiling+slack exactly). `npm run build` clean.

PR #3472, opened 2026-07-20
feat(gateway): per-tool-call wall-clock deadline (Stage A hang prevention)

mekenthompson:
## What & why

There is no working hang-watchdog for the "carrie" mode: an in-flight tool call that never returns. While any `tool_use` is open, `isLegitimatelyWorking()` returns true and the silence-poke 300s framework fallback is **deferred to its 15-min hard ceiling** — so a hung tool pins the conversation for 15 minutes before any net even considers acting, and even then the fallback only resets gateway state, never the child (why carrie needed a manual restart).

This PR is **Stage A: prevention** — a deterministic per-tool-call wall-clock deadline. It ships the mechanism with **enforcement opt-in / default-off** so blast radius is low. Stage B (progress-based restart keyed on turn-active-marker staleness) is a separate PR.

## Design

New pure module `telegram-plugin/gateway/tool-call-deadline.ts` bounds every in-flight tool with a ceiling measured from its own `startedAt`. Three classes:

| class | tools | ceiling |
|---|---|---|
| `standard` | ordinary foreground tools (Read, Grep, most MCP) | ~120s |
| `human` | `ask_user` / approval-card waits | **excluded** — owned by the existing #2724 permission-TTL sweep (`permission-ttl-sweep.ts`); a 120s clock would deny operators mid-decision |
| `background` | `Task`/`Agent`/`run_in_background` Bash | **excluded** — health judged by marker staleness (Stage B); keying a kill on "tool open N seconds" is the anti-pattern that discriminator exists to avoid |

On a breach the sweep force-fails the tool: logs the synthetic `tool_result` error and drops the masked flight entry from `ToolFlightTracker` + the silence-poke in-flight map, so `isMidToolCall` stops masking the hang and the net can act ~13 min sooner than the 15-min defer ceiling.

**Critical per-`toolUseId` guard:** after a force-fail the id is remembered in a `reaped` set; a real `tool_result` arriving *after* the synthetic error returns `false` from `noteResult` and is dropped — never double-processed.

### Honest scope note (source contradicts the naive premise)

A built-in tool executing *inside the unmodified claude process* has **no injection channel** — auto-allowed tools short-circuit in the bridge and never reach the gateway. The permission-gated waits that *do* have a channel (`dispatchPermissionVerdict`) are exactly the `human` class, already reaped by #2724. So Stage A's real production effect for a standard hang is **un-masking** the flight state to let the Stage B net act; the "synthetic tool_result injection" is realized at the module boundary (tested via the `sweepToolCallDeadlines` deps) and logged in the gateway. This is why enforcement is opt-in and the real kill lives in Stage B.

## Tests (fail on current head)

`telegram-plugin/tests/tool-call-deadline.test.ts` (26) + a `drop()` case in `interrupt-defer.test.ts`. They fail on `main` because `tool-call-deadline.ts` does not exist there. Key assertions:
- a never-returning standard tool breaches at exactly its 120s ceiling; the turn proceeds (flight cleared);
- the per-`toolUseId` guard drops a real result that arrives after the synthetic error;
- `human` + `background` classes are never force-failed on the short ceiling;
- observe-only vs enforce behaviour of the sweep wrapper.

## Safety
- Enforcement default-OFF (`SWITCHROOM_TOOL_DEADLINE_ENFORCE=1` to enable); observation-only otherwise.
- Gateway growth stays under the anti-inflation ratchet (decision+logging in the module).
- `tsc --noEmit` + full `npm run lint` clean; scoped vitest green. Full suite is CI's authority.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -243,6 +243,14 @@ import {
   distinctRequestIds,
 } from './permission-rearm.js'
 import { sweepPermissionTtl } from './permission-ttl-sweep.js'
+import {
+  ToolCallDeadlineTracker,
+  resolveDeadlineConfig,
+  observeToolUse,
+  toolCallDeadlineEnabled,
+  toolCallDeadlineEnforced,
+  sweepToolCallDeadlines,
+} from './tool-call-deadline.js'
 import { createMissedApprovalsStore, type MissedApproval } from './missed-approvals-store.js'
 import {
   createAlwaysAllowPersistQueue,
@@ -4429,6 +4437,13 @@ function postBusyAck(chatId: string, threadId: number | undefined, text: string)
 // stashed inbound, the original deadline is preserved (bounded wait).
 const toolFlightTracker = new ToolFlightTracker()
 
+// Per-tool-call wall-clock deadline (Stage A prevention) — see
+// tool-call-deadline.ts. `standard` tools get a ~120s ceiling; `human` waits
+// (owned by the #2724 permission-TTL sweep) and `background` tools (Task/Agent/
+// long Bash) are excluded. Enforcement is opt-in (SWITCHROOM_TOOL_DEADLINE_ENFORCE=1).
+const toolCallDeadline = new ToolCallDeadlineTracker()
+const TOOL_DEADLINE_CONFIG = resolveDeadlineConfig()
+
 interface PendingDeferredInterrupt {
   agentName: string
   inboundMsg: InboundMessage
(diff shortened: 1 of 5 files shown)


mekenthompson:
Closing per adversarial review. Stage A is near-inert and not required substrate:

- **No injection channel for the actual hang.** A built-in tool executing inside the unmodified `claude` process has no channel for the gateway to deliver a synthetic `tool_result` — `onForceFail` only drops the flight entry + logs, it never makes the model continue. The permission-gated waits that *do* have a channel (`dispatchPermissionVerdict`) are the `human` class, already reaped by the existing #2724 permission-TTL sweep. So this does **not** fix the carrie hang mode.
- **Inert by default** (enforcement opt-in) and even enforced it can't recover a built-in hang.
- **Stage B (#3473) does not import or depend on this** — it stands alone as the real fix (live SIGTERM restart keyed on marker staleness + ask-first resume).
- Plus two red CI gates (gateway line-ratchet, the brittle #3020 `/stop` source-slice test).

Superseded by #3473. The tool-class taxonomy in `tool-call-deadline.ts` (`background` vs `standard` vs `human`) is worth reusing — it's being ported into Stage B to protect known-long foreground tools from a false-positive restart (review Finding A). Not salvaging this PR now.

PR #4067, opened 2026-07-31
fix(gateway): drain parked turn-starts on silence-fallback

mekenthompson:
## Summary

The gateway carries **two independent "turn in flight" states**, and the 300 s silence-fallback only resets one of them.

The park gate in `stream-render.ts` (`case 'enqueue'`) treats the session as busy if **either** a live `currentTurn` exists **or** `parkedTurnStarts` is non-empty. The framework-fallback in `liveness-wiring.ts` (`onFrameworkFallback`) that recovers a hung turn nulls `currentTurn` and redelivers `pendingInboundBuffer` — but **never touched `parkedTurnStarts`**. That store's only real drains are the CLI's own `dequeue` / `remove` stream events, so when the `claude` REPL hangs mid-turn those events never arrive: a leftover parked envelope latches the busy gate permanently, and every later inbound parks unseen until a manual container restart.

`#3927` introduced the parked store but never wired it into fallback recovery. In the gymbro incident, msgs **4502→4504** sat parked behind a dead lock until a SIGTERM — while the `drained_buffered=0/0` log read honest-but-misleading (nothing was in `pendingInboundBuffer`; the stranded messages were in `parkedTurnStarts`).

## The fix

- **`stream-render.ts`** — export `drainParkedTurnStartsForChat(deps, chatId, threadId)`. It keeps the module-private `parkedTurnStarts` encapsulated: pops the entries **scoped to the fallback's chat/thread**, re-begins each via the same `beginTurn` path a real `dequeue` uses (**oldest-first**, so the newest survives as the live turn with clean supersession), and returns the drained envelopes.
- **`liveness-wiring.ts`** — `onFrameworkFallback` calls it **after** the `currentTurn` teardown, scoped to `fbChatId`/`fbThreadId`, and folds the count into the log line as `drained_parked=N` so the log stops lying. Guarded optional getter → degrades to prior behaviour when absent.
- **`gateway.ts`** — wire the dep to the real store via `gatewayStreamRenderDeps()`. Kept net-zero on the gateway line-ratchet (single-line import, trimmed an adjacent comment) so the one-way ratchet stays clean at 24917.

### Fix B deferred

The optional idle-timer park-gate sweep backstop is **deferred to a follow-up issue** to keep this a single-concern PR (see linked issue). Fix A closes the actual wedge; Fix B is defence-in-depth.

## Test plan

New `telegram-plugin/tests/framework-fallback-drains-parked.test.ts` drives the **real** `handleSessionEvent` and the **real** `onFrameworkFallback` (via `buildSilencePokeOptions`) in **one shared world** — the fallback's `drainParkedTurnStarts` dep routes to the real store + `beginTurn` over the harness deps. It asserts the **outcome**, not a code path:

- **before (origin/main):** RED — the parked entry is never drained (`expected 1 to be +0`).
- **after:** GREEN — `parkedTurnStarts` emptied **and** the parked message handed back into turn processing (its turn begun, its card opened), not silently dropped.

Plus a chat-scoping test (a fallback for chat A leaves chat B parked) and an idempotency test (a late CLI `dequeue` after the drain does not double-start).

```
Test Files  1 passed (1)
     Tests  3 passed (3)
telegram gateway: silence-poke framework-fallback ended wedged turn chat=1001 ... drained_buffered=0/0 drained_parked=1 extra_keys_purged=1
```

`tsc --noEmit` clean; gateway-line-ratchet / test-runner-coverage / no-pii / agent-state-dir-hermeticity / bun-test-imports / auth-test-hermeticity guards all clean.

## Risk

Low and contained. The drain is chat/thread-scoped (no cross-chat leak), idempotent against a late `dequeue` (empty-store no-op), and routes through the same `beginTurn` a real dequeue uses. The dep is an optional guarded getter, so older wiring degrades to today's behaviour.

### DIFF
--- a/telegram-plugin/gateway/gateway.ts
+++ b/telegram-plugin/gateway/gateway.ts
@@ -457,7 +457,7 @@ import {
   type SendReplyGatewayDeps,
   type DeliverCapturedProseDeps,
 } from './outbound-send-path.js'
-import { handleSessionEvent as handleSessionEventCore } from './stream-render.js'
+import { handleSessionEvent as handleSessionEventCore, drainParkedTurnStartsForChat } from './stream-render.js'
 import { createNarrativeLane } from './narrative-lane.js'
 import {
   parseAgentCallback,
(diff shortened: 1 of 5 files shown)


mekenthompson:
Fix B (idle-timer park-gate sweep backstop) deferred to #4068.

PR #4492, opened 2026-08-07
fix(memory): gate resolved-secret emission behind --resolve-secrets, pin --recreate vault wiring (#4486, #4487)

mekenthompson:
## Summary

Both follow-ups from the review of #4471 (`fix(hindsight): resolve vault: LLM api_key refs on the recreate/rollout launch path`), fixed together since both live in `src/cli/memory.ts`.

### #4487 — `memory docker-compose` no longer prints live secrets to stdout by default

#4471 wired `vault:` resolution into the compose-snippet emit site (`src/cli/memory.ts:1058` at the time) — functionally required (a compose file with a literal `vault:` ref is broken) and consistent with what `cp_access_key` already did, so not a careless regression. But it changed what the command emits to stdout: a live `sk-…` key, from an operator-invoked command whose output routinely gets pasted into terminals, files, and chat.

**Implemented the design from the issue** (did not substitute my own):
- Default: emit `vault:` refs unresolved, plus a note that the snippet needs resolving before use.
- New `--resolve-secrets` opt-in flag restores the resolved-key behaviour.
- `cp_access_key` follows the same rule (previously always resolved) — one consistent default for the command instead of two.

I considered the alternatives the issue raises (emit to a `0600` file; keep resolution on but redact on a TTY):
- **0600 file** is a bigger behavioural change (a `docker-compose` verb that writes to disk instead of stdout) for a command whose whole contract today is "print a snippet to paste"; it also doesn't help the common case of piping into a file the operator names themselves (`> compose.yml`), which would still be an unencrypted live-key file.
- **Redact-on-TTY, resolve-when-piped** is TTY-detection-dependent (model-dependent-ish, and every agent/CI/hook invocation is non-TTY, i.e. *always* piped) — it would resolve secrets by default in exactly the automated contexts where an accidental `docker-compose > some-log-capturing-place.txt` is most likely, which is the opposite of what we want.

`--resolve-secrets` is the simplest mechanism that is deterministic regardless of TTY/pipe context and matches the issue's own proposal, so I went with it as specified.

### #4486 — wiring regression test for `memory setup --recreate`

No test file referenced `registerMemoryCommand` at all, so there was no harness to hang a direct assertion on the `--recreate` launch path (`src/cli/memory.ts:991` at review time) — the path that actually caused the 2026-08-06 outage (reached from `src/cli/rollout.ts:1412` via hostd's refresh step). The existing regression test (`tests/setup-verification.test.ts:578`) only covers `stepMemoryBackend`, a different first-run-setup code path.

Added `tests/cli.memory-secrets-wiring.test.ts`: the first `registerMemoryCommand` harness (`new Command()` + `registerMemoryCommand(program)` + `parseAsync`, mirroring the shape of `tests/cli.agent-start-exit-code.test.ts`). It mocks only the docker-touching surface of `src/setup/hindsight.js` (container probes, port allocation, pull/stop) and the vault broker client, keeping every resolver (`resolveHindsightLlmSecrets`, `resolveHindsightCpAccessKey`, `hindsightGpuDecision`, `generateHindsightComposeSnippet`, ...) real. On top of that harness:

- a wiring assertion for `--recreate`, in the same shape as the existing `stepMemoryBackend` test (spies `startHindsight`, asserts the resolved `sk-…` key reaches the 4th positional arg and no `vault:` literal does)
- coverage for the new `--resolve-secrets` flag (default unresolved + note, `--resolve-secrets` resolved, `cp_access_key` follows the same rule)

Both new guards were verified to FAIL when the fix they pin is reverted (a live `sk-…` key reaching stdout without `--resolve-secrets`, and the `--recreate` launch path handed the raw `vault:` config instead of the resolved one) — pasted into the PR description's test plan below.

## Test plan

- [x] `npx vitest run tests/cli.memory-secrets-wiring.test.ts tests/setup-verification.test.ts` — 56/56 pass
- [x] Confirmed the #4486 guard fails when `--recreate`'s `startHindsight` call is reverted to pass the raw (unresolved) `hindsight.llm` config
- [x] Confirmed the #4487 guard fails when `--resolve-secrets` is forced on unconditionally (simulating the pre-fix always-resolve behaviour)
- [x] `npx tsc --noEmit` — clean
- [x] `npm run lint` — clean (full check suite; no full test suite run locally per instructions)
- [x] CHANGELOG `## Unreleased` entry added under a new `### Hindsight` section (rebased cleanly, no conflict)

Not run locally (CI is the full-suite authority): the full `vitest` suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,11 @@ now an anomaly worth investigating, not the norm.
   trivial rename of this header instead of reconstructing the whole section
   from `git log` at release time (as v0.20.12 had to).
 
+### Memory CLI secret hygiene
+
+- **`memory docker-compose` no longer prints live secrets to stdout by default (#4487)** — the compose snippet now emits `vault:` refs unresolved, plus a note, unless the new `--resolve-secrets` flag is passed; `cp_access_key` follows the same rule. Closes the follow-up from #4471, which correctly wired `vault:` resolution into the snippet emit site but changed the command to print a live `sk-…` key on an operator-invoked command whose stdout is routinely pasted around.
+- **`memory setup --recreate` vault-ref wiring now has a direct test (#4486)** — adds the first `registerMemoryCommand` test harness and a wiring assertion pinning that the `--recreate` launch path (the one that caused the 2026-08-06 outage) resolves `hindsight.llm` `vault:` refs before they reach `startHindsight`.
+
 ### Hindsight keyword recall moves to ParadeDB pg_search (BM25)
 
 - **New installs default to the `pg_search` BM25 backend (#4470)** — the keyword
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ now an anomaly worth investigating, not the norm.
 
 - **`memory docker-compose` no longer prints live secrets to stdout by default (#4487)** — the compose snippet now emits `vault:` refs unresolved, plus a note, unless the new `--resolve-secrets` flag is passed; `cp_access_key` follows the same rule. Closes the follow-up from #4471, which correctly wired `vault:` resolution into the snippet emit site but changed the command to print a live `sk-…` key on an operator-invoked command whose stdout is routinely pasted around.
 - **`memory setup --recreate` vault-ref wiring now has a direct test (#4486)** — adds the first `registerMemoryCommand` test harness and a wiring assertion pinning that the `--recreate` launch path (the one that caused the 2026-08-06 outage) resolves `hindsight.llm` `vault:` refs before they reach `startHindsight`.
+- **The cleartext-secrets warning now fires on every path that actually emits one** — it is decided ONCE, from the values about to be printed rather than from the flag, so `--resolve-secrets` (the only path that emits a live `sk-…` key) can never warn LESS than the safe default does; a PLAINTEXT api_key written straight into `switchroom.yaml`, which no `vault:` gate ever applied to, is warned about too; and a snippet carrying only unresolved `vault:` refs stays quiet, since a `vault:…` string is not a secret and crying wolf there would devalue the warning where it matters.
 
 ### Hindsight keyword recall moves to ParadeDB pg_search (BM25)
 
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -62,6 +62,7 @@ import {
   HINDSIGHT_CONTAINER_NAME,
   REPAIR_FAILED_EXIT,
 } from "../memory/hindsight-repair.js";
+import { isVaultReference } from "../vault/resolver.js";
 import { spawn } from "node:child_process";
 import { existsSync, readFileSync, statSync } from "node:fs";
 import { writeConfigFileSync } from "../util/atomic.js";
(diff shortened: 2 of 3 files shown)


mekenthompson (commented):
## ❌ REQUEST CHANGES — fresh-process review (CLAUDE.md Dev flow §5)

*(Posted as a review comment: GitHub forbids `--approve` / `--request-changes` on one's own PR, and every PR in this batch is authored by the same account the reviewer holds. Treat this as a **blocking DO-NOT-MERGE**.)*

One **major**, and I have **disarmed auto-merge** on this PR — it was `CLEAN` + CI-green + armed, and the repo ruleset (`main branch protection`, id 16470166) requires **no approving review**, only status checks. It was one queue tick from landing with this defect.

## Major — `--resolve-secrets`, the path that actually emits secrets, lost its runtime warning

On `origin/main`, `src/cli/memory.ts:1112` unconditionally printed to stderr before emitting the snippet:

> `# ! This snippet embeds resolved secrets in cleartext (LLM api_key and cp_access_key). Treat the output as sensitive — do not paste it into shared channels or commit it unredacted.`

This PR deletes that `console.error` and **does not re-add it to the `if (resolveSecrets)` branch** (`src/cli/memory.ts:1128-1142`). Net effect:

```
switchroom memory docker-compose --resolve-secrets | tee compose.yml
```

now prints live `sk-…` LLM keys and the `cp_access_key` to stdout with **zero runtime warning**. Before this PR, that same invocation warned. In a PR whose entire purpose is secret-emission hygiene, the one path that emits secrets is strictly *less* safe than it was.

The asymmetry is the tell: the **default** branch (`:1148-1156`) — which emits no secrets at all — carries a five-line NOTE including *"If you do, treat this command's output as a secret: don't paste it into chat, logs, or an unencrypted file."* So the warning was clearly judged necessary, then placed in the branch where it isn't needed and omitted from the branch where it is. The `--help` text carries a similar caution, but `--help` is not shown at emit time and is not where an operator piping output is looking.

**Suggested fix (small):** restore the `console.error` inside the `if (resolveSecrets)` branch. Keep it on stderr with the `#` prefix — a redirected stdout stays a valid compose file, and it stays a YAML comment even under `2>&1`.

## Major (same root cause) — a plaintext, non-`vault:` secret is now emitted with no warning at all

`hasUnresolvedHindsightVaultRef` (`:162-183`) gates the NOTE on the presence of a `vault:` **reference**. An operator whose `switchroom.yaml` holds a literal `cp_access_key: "somesecret"` or a literal `llm.api_key` (both schema-legal) takes the `else` branch at `:1146-1147`, which passes those literals straight to `generateHindsightComposeSnippet`. Result: the default, no-flag invocation prints a real cleartext secret and emits **neither** the NOTE (no ref → predicate false) **nor** the old blanket warning (deleted). `main` warned here. Not hypothetical — a plaintext key in config is exactly the pre-vault state an older fleet is in.

One fix covers both: warn whenever the snippet actually carries a secret *value* — i.e. `resolveSecrets === true` **or** a non-ref secret is present — rather than keying the warning off whether a `vault:` ref exists.

## The tests do not cover either regression

`tests/cli.memory-secrets-wiring.test.ts` is otherwise good work — it asserts outcomes (what actually printed) rather than call graphs, which is the right instinct, and the `cp_access_key` parity test at `:266-300` is a nice touch. But nothing asserts the sensitive-output warning survives on `--resolve-secrets`, so nothing went red when it was removed. The `--resolve-secrets` test at `:227-236` asserts only that `REAL_KEY` appears and `VAULT_REF` does not.

Please pin both cases. Note the harness collects `logs` from `console.log` while the warning is on `console.error` — a new test must capture stderr or it will assert nothing.

## What I verified as sound

- CI rollup 22/22 SUCCESS; `mergeStateStatus: CLEAN`.
- The core #4487 guarantee holds: default emits the `vault:` ref literally, never resolves, and **never touches the broker** (`expect(mocks.getViaBrokerStructured).not.toHaveBeenCalled()`) — the right shape for that assertion.
- `--resolve-secrets` genuinely reaches both `resolveHindsightCpAccessKey` and `resolveHindsightLlmSecrets`.
- **Disjoint from #4498**, confirmed at hunk level: #4498 touches `resolveLiteLLMForHindsight` (~:161-189) and the `--recreate` launch site (~:1012); this PR touches the `docker-compose` action (~:1085-1160). They do not undo each other. They **will** conflict textually on the import block — #4498 removes the `getViaBrokerStructured, readVaultTokenFile` import from `memory.ts` which this PR still uses, so whichever lands second must keep that import alive.

Happy to re-review promptly once the warning is restored and pinned.

Issue #4477, opened 2026-08-06
P3 — ANN over-fetch + reranker candidate-cap tuning (measured tradeoff curve)


mekenthompson:
Part of **#4474**. **Phase P3 — cheap CPU/latency levers that are currently unexamined.** Depends on **#4475** (P1 harness) and **#4476** (P2 residency).

## Why this phase exists

The original lever list assumed `ef_search` needed *raising* to 200 and that filter-aware ANN was missing. Reading the source contradicts both:

- **`ef_search=200` is already in force.** `hindsight-api-slim/hindsight_api/engine/search/retrieval.py:132-135`: *"semantic arms over-fetch by 5x (min 100) and trim to limit in Python to compensate. `ef_search=200` is set globally on pool connections at init time."* The two pgvector profiles are defined at `_vector_index.py:67` (`hnsw.ef_search=60`, low-latency) and `:70` (`=200`, high-recall) — the fleet runs the **high-recall** one, globally, on every connection.
- **ANN is already filter-partitioned.** Per-(bank, fact_type) partial HNSW indexes exist (`idx_mu_emb_worl_*`/`_obsv_`/`_expr_`, Alembic `d5e6f7a8b9c0_add_bank_internal_id_and_per_bank_hnsw.py`). There is no `bank_id` post-filter on the semantic arm to remove.

So the real, unexamined levers point the *other* way, and they are cheap:

1. **`ef_search` 200 → 60** (or an intermediate), i.e. the documented low-latency profile, traded against recall@k.
2. **The 5x over-fetch** on the semantic arms — over-fetching 5x (min 100) per fact_type, then trimming in Python, multiplies both HNSW work and rows shipped, across **3 fact_type arms per query**.
3. **`HINDSIGHT_API_RERANKER_MAX_CANDIDATES` 150 → 75-100.** Verified live at 150. The cross-encoder rerank is the GPU/CPU-heaviest stage and its cost is roughly linear in candidate count; it is also the stage that competes most directly with the retain/consolidation LLM load that P2 isolates.

Each of these trades quality for latency, so **none may be changed without a quality measurement** — that is why P6 exists and why this phase reports quality alongside latency.

## Scope

A measured sweep, then a durable config change for whichever settings win, landed switchroom-side (env / `HINDSIGHT_PERF_*` defaults), not as engine edits.

Sweep matrix, at minimum:
- `hnsw.ef_search` ∈ {60, 100, 200}
[...]

PR #4495, opened 2026-08-07
feat(hindsight): recall latency bench harness + baseline (P1, #4475)

mekenthompson:
## Summary

Adds `switchroom hindsight-bench`: a **model-free** harness that measures Hindsight recall latency as a function of bank size and concurrency, and reduces two runs to a PASS/FAIL verdict.

- `src/hindsight-bench/` — stats, DB probe, recall driver, contention generator, sweep runner, report/CSV, SVG plot, bank selection
- `src/cli/hindsight-bench.ts` — the verb, registered in `src/cli/index.ts`
- `docs/hindsight-bench.md`, `docs/hindsight-bench-baseline.md`, `docs/baselines/hindsight-recall-2026-08-07/` — how to run it, blast radius, and the captured baseline

Closes the P1 half of #4475.

## Why

#4474's goal — *"consistent latency across any bank size and good quality/performance including when we have contention on the box for many queries"* — was unfalsifiable. There was no number to move, so every later phase (#4476 index/cache work, #4477, #4478) would have shipped on assertion. The instrument is the deliverable.

Serves `reference/jobs/remember-across-sessions.md`: recall is the read half of it, and a 20-second recall is a recall the user gave up waiting for.

## Design points worth reviewing

- **Percentiles, not means** (nearest-rank p50/p95/p99, plus variance, sample count, error count). Errors are counted separately and never folded into percentiles — a timeout is not a fast call.
- **The harness never mutates the bank.** Every psql session is clamped read-only via `PGOPTIONS`, and the clamp is *verified* by attempting a real `CREATE TEMP TABLE` and requiring Postgres to reject it. `SHOW transaction_read_only` would pass against a session that could still write. `pg_stat_reset()` is opt-in only (`--reset-stats`), never a side effect.
- **True concurrency, not batching.** A pull-based worker pool keeps N recalls genuinely in flight; batching would have measured batch-max latency, a different and less useful number.
- **The contention generator is honest about its ceiling.** It reproduces buffer-pool eviction and WAL/checkpoint pressure — the box-level shape of a retain storm — but not the LLM half, because driving real model calls to manufacture load would burn subscription quota (`CLAUDE.md` § subscription-honest). Results under it are a documented lower bound. Defaults are conservative and every load knob is a flag; orphaned load backends are bounded by both an in-SQL deadline and an `application_name`-keyed terminate sweep wired to `SIGINT`/`SIGTERM`.
- **Cells record result counts, not just latency.** Without `meanResults`/`zeroResultCalls` a latency table cannot distinguish "recall got faster" from "recall stopped finding anything".
- **Exit-code contract:** `0` pass · `1` verdict failed · `2` could not run.

## Test plan

- `npx vitest run src/hindsight-bench` — 123 scoped tests
- `npm run lint` — clean (tsc + guard scripts)
- Verdict modes exercised end-to-end against synthetic result files: compare PASS `0`, compare FAIL `1`, contention PASS `0`, contention no-movement FAIL `1`, duplicate-cell file `2`
- Ran live against the fleet's Hindsight instance for the committed baseline

## Two bugs the harness found in itself, and what they cost

Both were found by using it for real, not by reading it.

**1. The contention generator was a silent no-op.** `contentionSql` emitted
`FROM memory_units TABLESAMPLE SYSTEM (25) m` — the alias belongs *before*
`TABLESAMPLE`, so PostgreSQL rejected it and every load backend died on
connect. Worker stderr went to `/dev/null`, so the harness cheerfully printed
"contention running with 8 backend(s)" while measuring a completely idle box.
The first contended sweep was void. The generator now counts its own backends
in `pg_stat_activity` and throws if none attached; a test pins the exact string
that shipped.

**2. Both verdicts graded whichever cells survived.** The Hindsight container
was recreated by unrelated fleet work partway through a contended sweep. Two of
ten cells lost their data, `compareContention` skipped them silently, and the
eight survivors still cleared the threshold — a confident PASS over a run that
had partly failed. One damaged cell kept 6 samples of 40 and reported the
*lowest* p95 in the run, because a percentile over the calls that were fast
enough not to time out is not a fast cell. Cells are now classified ungradeable
(no samples, or >10 % error rate), named in the report, and block the pass.

## What the baseline actually says

**AC1 (reproducibility) FAILED and is reported as failed.** Worst p95 delta
82.7 %, median 30.5 %, 3/20 cells within ±10 % at n=40. The diagnosis is in the
baseline: at low concurrency the p95 estimator's own bootstrap interval (median
half-width 15.8 %, up to 145 %) is *wider than the gate*, so ±10 % is
unattainable at these sample counts — roughly n≈2700/cell would be needed. At
high concurrency the noise floor is tight (3.8–11 %) and the cells still
disagree by 20–36 % in a consistent direction, which is real fleet drift. p50 at
n=100 reproduced at 6.4 %.

The tolerance stays at 10 %. Every cell reports its own noise floor beside the
delta so a reader can tell "the system moved" from "the estimator moved", and
the report says outright when the gate is unattainable — but the gate is not
widened to fit, because that is how a measurement stops being one.

**AC4 (contention moves the tail) PASSED**, measured A/B/A so drift cannot be
mistaken for effect: idle p95 2881 → contended **7919** (+174.9 %) → idle 2551,
with 4/4 load backends confirmed attached.

**Unplanned finding: bank size is not the dominant axis in the measured range.**
The 230,020-row bank has a *lower* single-threaded p95 than the 12-row one,
while going from c=1 to c=16 multiplies p95 by 9–13× on every bank. #4474 is
framed around size; the data says queueing.

## Risk

`--contention` deliberately degrades the live box while it runs; that is documented with its blast radius, defaults are conservative, and it is opt-in. Everything else is read-only. Zero model tokens. No behaviour change to any existing code path — this is additive: one new CLI verb and one registration line.

## Known gap

No synthetic bank above production's largest (#4475 item 3): creating one means writing hundreds of thousands of embedded rows into the live database, which the harness's own safety contract forbids, and no restored clone is available. Documented in the baseline; needs a clone first.


mekenthompson (commented):
## ❌ DO NOT MERGE AS-IS — fresh-process review (CLAUDE.md Dev flow §5)

*(Posted as a review comment: GitHub forbids `--request-changes` on one's own PR. Treat this as blocking.)*

**The harness code is good and I'd approve it on its own. The committed baseline data must not land in this repo in its current form.** Auto-merge is deliberately **not** armed.

## 🚨 Blocker — the baselines publish the operator's private fleet roster to a public repo

`switchroom/switchroom` is `"visibility": "PUBLIC"` (verified via the API, not assumed). Every committed baseline file carries the live fleet's **bank names with exact per-bank row counts**:

```
"bank": "assistant" · "carrie" · "finn" · "gymbro" · "kdogg" · "ken-profile"
"bank": "klanker"   · "lawgpt" · "lisa-profile" · "marko" · "overlord"
"bank": "reggie"    · "switchroom-dev" · "ziggy"
```

Present in **all 13** files under `docs/baselines/hindsight-recall-2026-08-07/` — `idle-a.json` (21 hits), `idle-b.json` (21), `arms.json` (30), `contended2.json` (15), `flank-idle1.json` (15), the `aba3-*.json` (8 each), the four `.csv` files (6–12 each), and `README.md:18-20`.

Two of these are named natural persons' profile banks — **`ken-profile`** and **`lisa-profile`** — the latter being a third party who has not consented to appearing in a public repo. Combined with row counts, this discloses the private fleet's membership, and each member's relative memory volume and activity, permanently and in git history. Force-pushing after a merge does not reliably remove it once GitHub has the objects.

Two things make this worse rather than incidental:
- `check-no-pii-secrets` passes clean on this diff, so the existing lint guard has **no coverage for this class of leak**. Nothing would have caught it.
- Rewriting public git history is materially harder than fixing it now, and this is a **purely mechanical** fix.

**Fix — pick either:** anonymise the bank identifiers across all 13 files (`bank-1`, `bank-2`, … — the harness only needs *a* size distribution, not real names), **or** strip the `bankRows` / `largestIndexes` / per-bank identity blocks from the committed baselines and keep only the aggregate latency cells. Update `README.md:18-20` to match.

**Worth doing as a follow-up:** extend `check-no-pii-secrets` (or add a `check-no-fleet-identifiers`) to fail on agent/bank names appearing under `docs/baselines/`. This leak reached a green CI and a review queue; a deterministic guard is the durable fix, per the repo's own "deterministic mechanisms over model-dependent behaviour" rule. Right now the only thing standing between a fleet name and the public internet is a reviewer noticing.

**Good news on scope:** I checked for leaked *content* too, and there is none. `"querySet": "generic-v1"` is the only query field — no real recall text, no memory bodies, no credentials. The exposure is names + counts, nothing more.

## What I verified as sound (the harness itself)

Genuinely strong work; every claim in the body held up.

- **Tests: 123/123 pass**, exactly as claimed (7 files, ~6s). `npm run lint` clean. CI rollup 22/22 SUCCESS.
- **Read-only clamp is real and stronger than advertised.** `src/hindsight-bench/db.ts:104-141` sets `PGOPTIONS='-c default_transaction_read_only=on'` per session and *proves* it by attempting a real `CREATE TEMP TABLE` in a `DO/EXCEPTION` block, reading back both the declared GUC and whether the write actually landed. `db.test.ts` pins the case where the flag claims `on` but the write lands anyway. No SQL path escapes the gate: `resetStats()` (`db.ts:172`) and the `write` contention profile are the only writable exceptions and are reachable only via explicit `--reset-stats` / `--allow-writes`.
- **The contention generator never touches bank data** — its write arm targets only the harness-owned `hindsight_bench_scratch`, never `memory_units` / `reflections`, and the unit tests assert this by scanning every write statement in the generated SQL.
- **Both self-found bugs are real regression tests, not decorative ones** — I read what they assert. `contention.test.ts`'s "puts the table alias BEFORE the TABLESAMPLE clause" and `report.test.ts`'s ungradeable-cell tests construct the exact original failure shapes (misordered alias; a 6-of-40-survivors biased cell) and would fail against the buggy behaviour. The generator counting its own backends in `pg_stat_activity` and throwing when none attach is the right durable fix for a class of silent no-op, not a patch.
- **Exit-code contract implemented as documented**: `fail()` → `exit(2)` (`hindsight-bench.ts:113`); `--compare` / `--contention-compare` set `exitCode = 1` on `!rep.pass` (`:237`, `:244`); measure/plot fall through to 0.
- **CLI registration is inert** — `src/cli/index.ts` is a two-line import + register, same pattern as existing verbs, no top-level side effects.
- CHANGELOG entry present under `### Benchmarking` and accurate against the code.

## Low (disclosed, not misrepresented)

`src/hindsight-bench/db.ts:60` + `contention.ts` `stop()` — on **SIGKILL** (not SIGINT/SIGTERM), or if `pg_terminate_backend` itself throws, a `write`-profile load backend can keep hammering the scratch table for up to `--contention-max-seconds` (clamped at 3600s). It never touches bank tables even then, and the code comments disclose this honestly. Noting it for the record rather than asking for a change.

---

Fix the baseline identifiers and I'll re-review and arm this immediately — the harness half is ready.

PR #4496, opened 2026-08-07
feat(hindsight): configurable container memory cap + shared_buffers safety check

mekenthompson:
## The bug

`HINDSIGHT_DEFAULT_MEM_LIMIT = "16g"` was hard-coded at `src/setup/hindsight.ts:882` with no config path, and it reached docker on both launch paths as a constant — `--memory=${HINDSIGHT_DEFAULT_MEM_LIMIT}` (`hindsight.ts:2278`) and `mem_limit: ${HINDSIGHT_DEFAULT_MEM_LIMIT}` (`hindsight.ts:2875`).

Meanwhile `shared_buffers` — the thing that cap has to hold — **has** been operator-configurable since the pg0 work: `SWITCHROOM_HINDSIGHT_PG_SHARED_BUFFERS` is a managed key (`hindsight-pg-defaults.ts:229`, in `HINDSIGHT_PG_ENV_KEYS` at `:309`), resolved by `resolveHindsightPgOverrides()` (`:326`) from the operator's `hindsight.env` block.

**That asymmetry is the defect.** An operator raised the live container to 24 GiB by hand so a 12 GiB buffer pool would fit. The next `switchroom memory setup --recreate` put the cap silently back to 16 GiB and left the 12 GiB `shared_buffers` standing inside it. pg0 runs Postgres with `shared_memory_type=mmap` (`hindsight-pg-defaults.ts:110-117`), so the buffer pool is charged to the cgroup as **unreclaimable** anonymous memory — the container came back pinning three quarters of its own ceiling, with no warning anywhere. It already happened once this week, and nothing in the codebase would have stopped it happening again.

## Plumbing-path decision — first-class field, NOT the `HINDSIGHT_PG_ENV_KEYS` channel

**Requested justification.** I did not put it on the `hindsight.env` channel, and the reason is mechanical rather than stylistic.

`hindsight.env` is the map of variables switchroom **emits into the container**. Every managed key travels one route and only one: `hindsightPgEnvPairs()` (`hindsight.ts:1914` pre-change) → `.flatMap(([k, v]) => ["-e", ...])` in the docker-run argv (`:2013`) and `.map(([k, v]) => "      - K=V")` in the compose `environment:` block (`:2722`). Those pairs are read by `docker/hindsight-entrypoint.sh` inside the container.

The memory cap is a docker `HostConfig.Memory` flag. The daemon consumes it; no process inside the container ever sees it. Routing it through that channel gives two bad options — emit `-e SWITCHROOM_HINDSIGHT_MEM_LIMIT=24g` into a container where nothing reads it (a lie in `docker inspect`), or put a key in the `env` map that is deliberately *not* emitted (breaking that map's one invariant, so `docker inspect`'s env no longer matches the config). The task said not to force-fit it; I didn't.

The repo already decided this category once. `hindsight.gpu` (`schema.ts:2782`) is the other non-env container-shape knob — `--gpus all` / the compose `deploy.resources.reservations.devices` stanza — and it is a **first-class field on `HindsightConfigSchema`**, not an `env` key. `hindsight.mem_limit` follows that precedent exactly.

**Threading:** on the existing trailing `HindsightPerfOptions` object, not a new positional. That object's own doc (`hindsight.ts:2004-2009`) states the rationale — "the positional-parameter list is already long enough to make a silent argument-order mismatch between them plausible" — and a 9th positional across two generators with *different* parameter orders is precisely that hazard. Both paths resolve through one `resolveHindsightMemLimit()`, so they cannot drift; a test asserts the outcome of both generators for the same inputs rather than that both call it.

## The safety check — warn, and why not fail

`HINDSIGHT_PG_SHARED_BUFFERS_BUDGET_MIB` is a **compile-time** budget derived from a literal 16 GiB, pinned by a test. It protects the pair of numbers in this repo and nothing else. Once both sides are operator-configurable they move independently, at different times, by different mechanisms — which is the observed failure.

So: every launch now compares the *resolved* cap against the *resolved* `shared_buffers` and warns when the cap does not clear the pool by `HINDSIGHT_PG_APP_ANON_MIB + HINDSIGHT_PG_PAGE_CACHE_FLOOR_MIB` (4608 MiB) — the **same arithmetic** the shipped budget subtracts, exported as one constant so the two cannot disagree about what "comfortably larger" means.

```
! hindsight: container memory cap 16g (16 GiB) leaves only 4 GiB above shared_buffers
  12288MB (12 GiB) — Postgres pins that buffer pool as unreclaimable shared memory, so
  the container needs at least 4608 MiB beyond it (2560 MiB app working set + 2 GiB
  page-cache floor). Raise `hindsight.mem_limit` to at least 17 GiB, or lower
  `hindsight.env.SWITCHROOM_HINDSIGHT_PG_SHARED_BUFFERS`.
```

It fires from **`startHindsight()` itself**, not only the CLI wrapper, because `switchroom memory setup --recreate` is the thing that did the reverting — a CLI-only warning would be skipped by every other caller.

**Warn, not fail** — argued, per the request:

1. The dangerous state is reached by a **default re-asserting itself**, not by a bad operator edit. Hard-failing there converts "the memory container is badly sized" into "the fleet has no memory container at all", which is strictly worse for every agent on the host.
2. The headroom figure is a measured heuristic (an anon working set plus a page-cache floor), not a hard kernel limit. A heuristic must not be able to brick `switchroom setup` on a host whose workload it was never measured against.

Two things **are** hard rejections, because they are unambiguous rather than heuristic: a malformed size string (schema regex at `apply`, and again at launch rather than silently substituting the default — silently defaulting would reproduce the exact "reverted to 16g" behaviour this removes), and a cap **below the container's 4g `--memory-reservation`**. That last one is an adversarial finding on my own diff: the reservation is emitted unconditionally, docker refuses to create a container whose limit is under its reservation, and the constant made that unreachable while this knob makes it reachable. Docker's own rejection is late and opaque; this one names both numbers.

## Tests — `src/setup/hindsight-mem-limit.test.ts` (23)

Each would go red on the original code, not just exercise a path:

- **(a) explicit value reaches both emitters.** `--memory=24g` in the docker-run argv (asserted as the *exact* set of `--memory=` flags, so a duplicate would also fail) and `mem_limit: 24g` in the compose snippet. On `main` both are `16g` regardless. Plus a parity loop asserting run ⇄ compose agree for `undefined` / `24g` / `32768m`, and that the pre-existing `--memory-swap` invariant still holds alongside a configured cap.
- **(b) default unchanged.** `HINDSIGHT_DEFAULT_MEM_LIMIT` is still `16g`; an unconfigured `startHindsight()` / `generateHindsightComposeSnippet()` emits exactly that on both paths; blank/whitespace/absent all resolve to the default rather than erroring.
- **(c) the check fires on the dangerous combination and stays quiet on safe ones.** Red on cap `16g` + `shared_buffers` `12288MB` — asserting the message names both numbers *and* both keys, since a warning that names neither was the original defect. Quiet on the shipped defaults, quiet on the operator's actual fix (24 GiB + 12 GiB). Boundary asserted exactly (cap == buffers + headroom passes; one MiB under does not). The end-to-end shape too: `hindsightMemBudgetWarningFor({ env: { SWITCHROOM_HINDSIGHT_PG_SHARED_BUFFERS: "12288MB" } })` — the default cap meeting an operator-raised pool, i.e. the recurrence path — warns, and goes quiet once `mem_limit` is also raised. And that `startHindsight()` itself warns on that config and does not on a stock launch.
- Size parsing is asserted in both grammars, because they differ and getting it wrong would misfire the check by a factor of 8192: docker's bare number is **bytes**, Postgres' bare number is **8 kB blocks**, plus the `off` sentinel resolving to pg0's own 256MB.

Unparseable input on either side returns `null` (say nothing rather than assert a wrong number) — asserted.

## Also

- Wired on **first-run `switchroom setup`** too (`src/cli/setup.ts`), which previously passed `perf: undefined`. A knob honoured on `memory setup` but ignored on `setup` would be the same silent divergence again. That call site now also forwards `hindsight.env` — necessary, not incidental: the check cannot honestly compare against `shared_buffers` it cannot see.
- Docs: new `### Memory cap for the Hindsight container (hindsight.mem_limit)` section in `docs/configuration.md`, alongside the other `hindsight.*` keys.
- `CHANGELOG.md` `## Unreleased` (the one at the top).

## Verification

- `npm run lint` — clean (tsc + all 23 guards).
- `vitest run src/setup/ src/config/` — 41 files, 1002 passed.
- `src/cli/apply.test.ts` fails 10 identically on unmodified `origin/main` in this environment (needs a built `dist/`); pre-existing, not from this change.
- **No container was touched.** Code only.

## Considered and not done

A recreate-time *drop guard* for the cap, mirroring the GPU drop guard (`memory setup --recreate` refusing to silently shrink a running container's cap). The warning covers the dangerous end state, and the guard is a separate concern with its own inspect-path surface — deliberately out of scope for a single-concern PR rather than forgotten.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -191,6 +191,10 @@ now an anomaly worth investigating, not the norm.
   changes to expect: the **ultraplan** feature is gone, and `/review` is now an
   alias for `/code-review`.
 
+### Hindsight
+
+- **`hindsight.mem_limit` — the container's memory cap is configurable, and a cap that cannot hold `shared_buffers` now says so** — the cap was a hard-coded `16g` with no config path, while `shared_buffers` (a managed `hindsight.env` key) was configurable. An operator who raised the live container to 24 GiB by hand to fit a 12 GiB buffer pool had it silently reverted to 16 GiB by the next `switchroom memory setup --recreate`, with the 12 GiB `shared_buffers` left standing inside it — Postgres pinning most of its own cgroup as unreclaimable shared memory, warned about nowhere. `hindsight.mem_limit` (a docker size string; absent → the unchanged `16g` default) now reaches docker `--memory` and the compose `mem_limit:` from one resolver, so the two launch paths cannot disagree, and it is honoured on first-run `switchroom setup` as well as on `memory setup`. It is a first-class field rather than an `env:` key because it is a docker `HostConfig.Memory` flag, not a variable emitted into the container. On top of the knob, every launch now checks the resolved cap against the resolved `shared_buffers` and warns — naming both numbers and both keys — when the cap does not clear the buffer pool by the app working set plus the page-cache floor (4608 MiB). Warning, not refusal: the dangerous state is a default re-asserting itself, and hard-failing there would leave the fleet with no memory container at all. A malformed size string is still rejected outright, at `switchroom apply` by the schema and at launch rather than silently falling back to the default.
+
 ## v0.20.12 — native WebSearch fleet-wide, faster sub-agent cards, Hindsight recall speedups, and token-aware LiteLLM pacing
 
 ### Web tooling
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -756,6 +756,35 @@ hindsight:
 
 If you run your own Hindsight container outside `switchroom memory --start` (e.g. you point `memory.config.url` at an external server), switchroom doesn't manage that container's env — set the cap on your own image.
 
+### Memory cap for the Hindsight container (`hindsight.mem_limit`)
+
+The Hindsight container is created with a hard memory cap — docker `--memory` on the `switchroom memory setup` path, `mem_limit:` in the `switchroom memory docker-compose` snippet. The default is **16g**, and until this key existed that was also the *only* value: there was no config path at all.
+
+```yaml
+hindsight:
+  mem_limit: 24g          # a docker size string: 24g, 16384m, …
+  env:
+    SWITCHROOM_HINDSIGHT_PG_SHARED_BUFFERS: 12288MB
+```
+
+It is a first-class field rather than a `hindsight.env` key on purpose: `env` is the map of variables switchroom **emits into** the container, and the memory cap is a docker `HostConfig.Memory` flag consumed by the daemon — no process inside the container ever sees it.
+
+**Move it together with `shared_buffers`.** They are one decision, not two knobs. Hindsight's embedded PostgreSQL runs with `shared_memory_type=mmap`, so `shared_buffers` is charged to the container's cgroup as **unreclaimable** anonymous memory: whatever you give the buffer pool is subtracted from the cap before anything else in the container gets a byte. The container still has to hold its ~2.5 GiB app working set (API process, embedder, cross-encoder, next-server) plus a page-cache floor on top of that.
+
+The failure this key removes is the asymmetry, not the number. `shared_buffers` was configurable (a managed `hindsight.env` key); the cap was not. So an operator who raised the live container's cap by hand to fit a larger buffer pool had it silently reverted to 16g by the **next `switchroom memory setup --recreate`**, while their `shared_buffers` stayed exactly where they had put it — leaving Postgres pinning most of its own cgroup, with no warning anywhere. Setting `hindsight.mem_limit` makes the cap survive a recreate like every other configured value.
+
+Switchroom now also **warns** when the resolved cap does not clear the resolved `shared_buffers` by at least the app working set plus the page-cache floor (4608 MiB — the same arithmetic the shipped defaults are derived from). The warning names both numbers and both keys, and fires on the launch path itself, so a `--recreate` that puts the container into that state says so:
+
+```
+! hindsight: container memory cap 16g (16 GiB) leaves only 4 GiB above shared_buffers
+  12288MB (12 GiB) — … Raise `hindsight.mem_limit` to at least 17 GiB, or lower
+  `hindsight.env.SWITCHROOM_HINDSIGHT_PG_SHARED_BUFFERS`.
+```
+
+It is a **warning, not a refusal**, deliberately: the dangerous state is reached by a default re-asserting itself, so hard-failing there would turn "the memory container is badly sized" into "the fleet has no memory container at all". Two things *are* rejected outright: a malformed size string (by the schema at `switchroom apply`, and again at launch rather than silently falling back to the default), and a cap below the container's 4g memory *reservation* — docker will not create such a container, and its own rejection is late and opaque.
+
+Both launch paths resolve the cap from one place, so the `docker run` and `--compose` outputs cannot disagree. Changing it takes effect on the next container recreate (`switchroom memory setup --recreate`), not live.
+
 ### Logging in to the Hindsight dashboard (`hindsight.cp_access_key`)
 
 The Hindsight container also serves a **control-plane dashboard** (Next.js, host port 9999) alongside the memory API. That dashboard's login is armed by exactly one thing — upstream's `HINDSIGHT_CP_ACCESS_KEY`. When the variable is unset the access-key middleware short-circuits and lets every request through: the dashboard is not weakly protected, it has **no authentication at all**.
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -30,6 +30,7 @@ import {
   getHindsightContainerEnv,
   getHindsightImageEnv,
   hindsightContainerEnvPairs,
+  hindsightMemBudgetWarningFor,
   hindsightResolvedCapabilities,
   preflightHindsightPorts,
   getHindsightDeviceRequests,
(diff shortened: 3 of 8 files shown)


mekenthompson (commented):
## ✅ APPROVE — fresh-process review (CLAUDE.md Dev flow §5)

*(Posted as a review comment: GitHub forbids `--approve` on one's own PR and this batch is all one author. Treat this as the §5 approval.)*

No blockers, no majors. Reviewed adversarially against the PR body's claims rather than accepting them.

### The rebase question — both sides preserved, but it has re-conflicted

The body's "just rebased to fix a CHANGELOG add/add conflict" **no longer holds**: this PR is `DIRTY` again, because **#4497 merged at 01:51Z after the rebase**, adding its own `### Hindsight …` sections to `## Unreleased` at the same anchor.

I inspected the actual three-way result rather than assuming. The conflict is a clean, purely **additive** add/add: `origin/main` contributes `### Hindsight keyword recall moves to ParadeDB pg_search (BM25)`, `### Hindsight vault: secrets outside the apply path`, `### Recall quality is now a measurable, gating number`, `### Test isolation and CI`, `### Dependencies`; this PR contributes its single `### Hindsight` block. **Nothing on either side is dropped or overwritten** — resolution is keep-both, with this PR's section appended at the end of `## Unreleased` (the add/add-safe anchor the repo already uses). Confirm the merged file has both blocks after you rebase.

One **low** while you're in there: after resolution the generic heading `### Hindsight` sits alongside main's more specific `### Hindsight keyword recall moves to…` and `### Hindsight vault: secrets…`. Consider retitling to something specific like `### Hindsight container memory cap` for consistency.

### Verified — the substance

- **Tests:** `vitest run src/setup/hindsight-mem-limit.test.ts` → 23/23. `vitest run src/setup/` → 552/552 across 22 files. The tests pin **artefacts, not call graphs** — real `--memory=…` argv and real `mem_limit:` compose text. "keeps the two launch paths on the same cap for the same input" iterates `[undefined, "24g", "32768m"]` and asserts `fromCompose === fromRun`, so a divergence bug goes red. The "sits exactly on the boundary, not near it" test would catch an off-by-one in the 4608 MiB arithmetic.
- **Lint:** CI `lint-run` passes (1m5s) and the rollup is **23/23 SUCCESS**. A local `npm run lint` failure in a scratch worktree turned out to be missing `telegram-plugin/` type deps in that environment — the diff touches zero `telegram-plugin/` files, and CI is the authority here.
- **One resolver, both paths (the central claim) — holds.** `resolveHindsightMemLimit()` at `src/setup/hindsight.ts:2093` is called from `startHindsight` (`:2380`, `--memory=${…}`) **and** `generateHindsightComposeSnippet` (`:2991`, `mem_limit: ${…}`). Same function, same `perf` shape — no input makes them disagree.
- **Size parsing** (`parseDockerSizeToMib`, `src/setup/hindsight-pg-defaults.ts:239`) and the schema regex (`src/config/schema.ts:2799`) were checked to **agree with each other** across 16 adversarial cases (`16G/16g/16gb/16GiB/16/""/"  16g  "/1.5g/16t/0/-1/"16 g"/16_000_000/huge`), and cross-checked against real dockerd 27.3.1 via `docker create --memory=…`. Both the schema-level and launch-time rejections claimed in the body genuinely exist.
- **The safety check arithmetic is right:** `HINDSIGHT_PG_APP_ANON_MIB` (2560) + `HINDSIGHT_PG_PAGE_CACHE_FLOOR_MIB` (2048) = 4608, matching the claimed floor (`hindsight-pg-defaults.ts:106,117,238`). `parsePgSizeToMib` correctly treats a bare Postgres number as **8 kB blocks** rather than bytes, and `off` as the 256 MiB fallback — a different grammar from docker's, correctly kept separate. It **warns rather than refuses**, as claimed, and names both numbers and both keys (`:342-350`).
- **`shared_buffers` is read from the resolved env, not raw config** — `hindsightPgEnvPairs(perf)` (`hindsight.ts:1888-1891`) uses `hindsightPgEnv(resolveHindsightPgOverrides(perf?.env, perf?.processEnv))`, the exact resolution the container env pairs use (`:1987`). The warning therefore cannot see a different `shared_buffers` than the container gets. This was the failure mode I most expected to find and it is closed.
- **Backward compatibility proven, not assumed:** on `origin/main` both launch paths hard-coded `HINDSIGHT_DEFAULT_MEM_LIMIT` with no config lookup (`git show origin/main:src/setup/hindsight.ts:2278,2875`); `resolveHindsightMemLimit(perf)` returns exactly that constant when `mem_limit` is absent or blank. An existing fleet gets byte-identical behaviour.
- **Docs** (`docs/configuration.md`) accurately state the field, the `16g` default, the units, and warning-not-refusal. The worked example's arithmetic checks out: 16g − 12288 MB = 4 GiB headroom < 4608 MiB floor → warns, and the suggested raise matches `ceil((12288+4608)/1024)*1024 = 17408` MiB = 17 GiB.

### Lows (non-blocking; fix if convenient)

1. `src/config/schema.ts:2799` — the regex is a strict **subset** of docker's real `--memory` grammar (rejects `16GiB`, `16_000_000`, which dockerd accepts). Fails safe, so not a bug, but it will surprise someone.
2. `src/setup/hindsight.ts:2093` — a googol-scale `mem_limit` passes switchroom's validation and is then rejected by dockerd with a misleading, unrelated error instead of switchroom's clear one.
3. `src/setup/hindsight.ts:2107` — `mem_limit: "0"` is rejected as "below reservation", but dockerd actually treats `0` as *unlimited* and accepts it even with a reservation set. The comment overstates docker's behaviour for this one input. Harmless in practice.
4. **`src/config/schema.test.ts` has no direct test of the `mem_limit` regex**, despite schema-level rejection being an advertised feature of this PR. I hand-verified 16/16 cases, but nothing in CI would catch future regex drift. Worth a few lines — this is the one low I'd actually spend time on.

None of these change behaviour for any realistic config. Rebase for the CHANGELOG and this is good to land.

PR #4498, opened 2026-08-07
fix(hindsight): one derivation for LiteLLM routing env, never a silent key drop

mekenthompson:
## Summary

Hindsight's LiteLLM routing env (`ANTHROPIC_BASE_URL` + `ANTHROPIC_CUSTOM_HEADERS`) is what makes its LLM traffic transit the proxy and land in `LiteLLM_SpendLogs` as `end_user=hindsight`. Two defects, both in that env's production:

1. **The credential-bearing literal existed twice.** `hindsightContainerEnvPairs` (`src/setup/hindsight.ts:2201`, the docker-run / `memory setup --recreate` path) and `generateHindsightComposeSnippet` (`:2837`) each built the header inline. Two copies of the same string, one of which is the only one the fleet actually *runs* and the other the only one an operator ever *reads* — the shape that drifts. Both now come from one `hindsightLiteLlmEnvPairs()`.
2. **A dropped key was indistinguishable from "LiteLLM is off".** `resolveLiteLLMForHindsight` returned a bare `undefined` for LiteLLM-disabled, broker-denied, and broker-down alike — and existed as a *verbatim copy* in both `src/cli/memory.ts:166` and `src/cli/setup.ts:1030`. So a denied grant launched hindsight with no routing env, no warning, and unmetered spend. The container comes up healthy; reflect/consolidation just quietly stop appearing under `end_user=hindsight`.

## Why — the live evidence

Observed on the dev fleet 2026-08-07: `docker inspect switchroom-hindsight` had zero `ANTHROPIC_*` entries after a `memory setup --recreate`, and `vault-audit.log` shows the cause:

```
{"op":"get","key":"litellm/hindsight/api-key","caller":"agent:overlord",
 "result":"denied:agent 'overlord' has no ... standing grant declaring
 'litellm/hindsight/api-key'; nothing is broker-accessible"}
```

`litellm.enabled: true` and `base_url` were both set, so this was branch 2 exactly: the broker denied, the `.catch(() => null)` swallowed it, routing vanished. The only operator-visible trace was the ENV DRIFT report naming the two vars — which reads as "these were set imperatively", i.e. operator drift. It has been misdiagnosed as exactly that. The drift report structurally cannot know the real cause; the resolver can, and now says so.

The two sibling secrets on the same launch path already fail loudly (`HINDSIGHT_CP_NO_ACCESS_KEY_WARNING`, `hindsightLlmDroppedKeyWarning`). This is the third, now held to the same standard.

## What changed

- **`hindsightLiteLlmEnvPairs(litellm, llmModel)`** (`src/setup/hindsight.ts`) — THE single construction of both vars. Returns `[]` unless a base URL *and* a non-blank key are present, so a blank key can no longer bake `x-litellm-api-key: Bearer ` into the container (rejected at LiteLLM auth — a lane that *looks* routed is worse than an obviously-unrouted one). Both emitters call it; compose escapes the newlines for its YAML scalar exactly as before, so its output is byte-identical.
- **`resolveHindsightLiteLlm(config, deps)`** — one resolver for all launch paths, returning `{ litellm?, droppedRef? }`. `droppedRef` is set only in the enabled-but-unresolved case and carries the `vault:` **reference**, never the secret. `memory.ts` / `setup.ts` are now thin delegators over it, and `setup.ts`'s test seam (`deps.getViaBrokerStructured`) now reaches this resolution too — previously it did not, so first-run-setup tests hit the real broker for this one key.
- **`hindsightLiteLlmDroppedKeyWarning(ref)`** — printed by both launch paths. Names the reference, both vars, the metering consequence, and that this is *not* env drift.
- Deliberately **not** falling back to `litellm.admin_key` (`vault:litellm/master-key`): that is the proxy master credential and must never enter a service container. A test pins the ref against `HINDSIGHT_KEY_VAULT_REF` (the doctor check's copy) and asserts it contains no `master`.

## Test plan

New `src/setup/hindsight-litellm-env.test.ts` (20 tests) asserts artefacts, not call graphs: the real `docker run` argv `startHindsight` issues, and the real compose text.

- [x] `vitest run src/setup/hindsight-litellm-env.test.ts` — 20/20
- [x] **Fails without the fix**: reverting the three source files takes it to **14 failed / 6 passed**. The behavioural failure on `main` is `omits both rather than emitting an empty Bearer  when the key is blank` (`expected true to be false` — main emits the broken header); the remaining 13 fail because the fail-loud resolver/warning do not exist there. The 6 that pass on `main` are the ones pinning that the recreate path *does* inject when the key resolves — kept as the regression guard.
- [x] `vitest run src/setup/ tests/setup-verification.test.ts` — 602/602
- [x] `vitest run tests/hindsight-env-keys-are-reachable.test.ts tests/memory-key-drift.test.ts tests/cli.doctor.memory.test.ts src/litellm/` — 227/227
- [x] `npm run lint` — clean (incl. `check-no-pii-secrets`, `check-litellm-config-guard`)
- CHANGELOG entry appended at the END of `## Unreleased` (the add/add-conflict-safe anchor).

Not run locally, CI is the authority: the full vitest/bun suites.

## Risk

Low, and the one behaviour change is a tightening: a `LiteLLMHindsightConfig` with a blank `apiKey` now emits nothing instead of a malformed header. No production caller can construct that — `resolveHindsightLiteLlm` never returns a blank key.

## Follow-up (deliberately out of scope)

`switchroom memory docker-compose` still passes `litellm: undefined` to the snippet generator (`src/cli/memory.ts`, "litellm: omitted ⇒ same default the docker-run path uses" — which is no longer true, if it ever was), so the emitted compose file carries no LiteLLM routing at all. Wiring it would mean resolving and printing a live `sk-` key to stdout, which is precisely what **#4492** is in flight to stop. That change belongs on top of #4492's `--resolve-secrets` gate, not underneath it. This PR is otherwise disjoint from #4492: it touches `resolveLiteLLMForHindsight` and the `--recreate` launch site, not the `docker-compose` action.

## Design contract

`reference/jobs/` — memory that survives a restart, and `reference/invariants.md` § "Operator-controlled gateway carve-out": the carve-out is only honest if its metering is intact, and a metering lane that fails silently is not operator-controlled.


mekenthompson (commented):
## ✅ APPROVE — fresh-process review (CLAUDE.md Dev flow §5)

*(Posted as a review comment: GitHub forbids `--approve` on one's own PR and this batch is all one author. Treat this as the §5 approval; the repo ruleset requires status checks, not an approving review.)*

I verified the load-bearing claims by **execution**, not reading, because several claims in this batch have not survived checking.

### (a) Compose byte-identity — proven, not accepted

I did not take the `\n` → `\\n` claim on trust. I generated `generateHindsightComposeSnippet` output over an 11-case input matrix (bare / trailing-slash / multi-slash base URLs; Claude vs non-Claude vs absent model; blank and whitespace-only key; blank and whitespace-only base URL; whitespace-padded key + URL) on this branch and on pristine `origin/main`, and diffed the dumps.

**10 of 11 cases are byte-identical.** The single divergence is the whitespace-padded case, where `main` emits

```
- ANTHROPIC_BASE_URL=  http://pad:4010  /anthropic
- ANTHROPIC_CUSTOM_HEADERS=x-litellm-api-key: Bearer   sk-…  \n…
```

and this branch emits the correctly-trimmed value. So the escaping mechanism is genuinely unchanged and the only delta is a strict improvement — `main` was emitting a broken value there.

Minor precision note for the record: the body's flat *"byte-identical"* is very slightly overstated. It is "byte-identical except where main was broken." Not a defect, but I'd rather the claim be exact, since the next reader will lean on it.

### (b) No `vault:` literal and no resolved `sk-` into env, stdout, logs, or the warning

- `resolveHindsightVaultString` (`src/setup/hindsight.ts:1544`) returns `undefined` — never the literal — for denied / missing / non-string / blank / broker-down, and trims. So `hindsightLiteLlmEnvPairs` can only ever receive an already-resolved key.
- `hindsightLiteLlmEnvPairs` (`:2186`) hard-gates on **both** a non-blank base URL and a non-blank key, so `Bearer ` can no longer be baked in. On `main` it could — see (d).
- The warning takes only the `vault:` **reference** (`hindsightLiteLlmDroppedKeyWarning(ref)`); the test asserts it contains neither the key nor `sk-`.
- Compose stdout: the only production caller of `generateHindsightComposeSnippet` is `src/cli/memory.ts:1118`, which passes `litellm: undefined`. There is no path by which this refactor puts a LiteLLM key on stdout.

### (c) No master/admin-key fallback

`HINDSIGHT_LITELLM_KEY_VAULT_REF = "litellm/hindsight/api-key"` (`:1587`) byte-matches `HINDSIGHT_KEY_VAULT_REF` in `src/litellm/key-allowlist-check.ts:64`, and the test pins that equality plus the absence of `master`. Grepping `admin_key` / `litellm/master-key` in `hindsight.ts` finds only the two doc comments explaining why it is *not* used. `parseVaultReference("vault:litellm/hindsight/api-key")` yields exactly the raw key string the old code passed positionally, so the broker request is byte-for-byte unchanged.

### (d) The test evidence reproduces exactly

Reverting the three source files to `origin/main` gives **14 failed / 6 passed** — the stated number. More important, I checked the failures are not all missing-symbol `TypeError`s. The first is a genuine behavioural assertion:

```
FAIL  omits both rather than emitting an empty `Bearer ` when the key is blank
AssertionError: expected true to be false
```

`main` really does bake the malformed header. That is a real regression guard, not a shape test. Restored afterwards and confirmed the worktree matches HEAD.

### Also run

- `vitest run src/setup/ tests/setup-verification.test.ts tests/hindsight-env-keys-are-reachable.test.ts tests/memory-key-drift.test.ts tests/cli.doctor.memory.test.ts src/litellm/` → **829/829 pass**, 34 files.
- `npm run lint` → clean, including `check-no-pii-secrets`, `check-litellm-config-guard`, `check-secret-pattern-parity`.
- CI rollup 22/22 SUCCESS.

### Disjointness from #4492 — confirmed

Hunk-level: #4492 touches the `docker-compose` action (~:1085-1160) and adds `hasUnresolvedHindsightVaultRef`; this PR touches `resolveLiteLLMForHindsight` (~:161-189) and the `--recreate` launch site (~:1012). Semantically disjoint, as claimed. They **will** conflict textually on the import block — this PR removes the `getViaBrokerStructured, readVaultTokenFile` import from `memory.ts`, which #4492 still uses. Whichever lands second must keep that import alive. Worth care at rebase; not a review blocker.

### One low, deferred — and I agree with the author's call

`src/cli/memory.ts:1120`'s comment `// litellm: omitted ⇒ same default the docker-run path uses.` is now actively misleading: the docker-run path derives routing and this one does not. The body acknowledges it and correctly defers, since fixing it properly means resolving a live `sk-` key onto stdout, which must land **on top of** #4492's `--resolve-secrets` gate rather than underneath it. Please make it a tracked follow-up issue rather than an informal note — a stale comment at a secret-emitting call site is exactly what the next reader trusts.

### Merge note

Currently `DIRTY`: CHANGELOG add/add conflict created when **#4497 landed after this branch was cut** (not a pre-existing problem with the branch). Resolution is purely additive — keep both sides, nothing is dropped. Rebase and this is good to go.

PR #4724, opened 2026-08-15
fix(worktree): count nested-cwd holders as live; add report-only reap pass

mekenthompson:
## Summary

Two changes to the worktree reclaim subsystem, one correctness fix and one
evidence-gathering mode.

**1. The liveness probe missed nested holders (`src/worktree/reaper.ts:134`
on `main`).** `probePathInUse` answered with `fuser <path>` then
`lsof -t <path>`. Both match the path **exactly**, so a process whose cwd is a
nested subdirectory of a worktree — i.e. an agent working inside a checkout —
was reported `free`. Reproduced on `main`:

```
$ mkdir -p /tmp/tree/src/deep && (cd /tmp/tree/src/deep && sleep 60) &
$ fuser /tmp/tree          ; echo $?     # (no output)   1   ← "free"
$ fuser /tmp/tree/src/deep ; echo $?     # 262751c       0   ← in-use
```

`free` is exactly what clears the reaper's third fail-safe ("an in-use probe
can DEFINITIVELY report the path as free"), so any automation built on this
probe would `git worktree remove --force` live work. The same probe is the
idle guard for the task-tree sweep (`gc.ts:834`).

The probe now walks procfs (`src/worktree/proc-liveness.ts`): a process is a
holder when its **cwd is the tree root or any directory beneath it**, or when
any **open fd** resolves inside the tree. Two subtleties handled explicitly:

- **Container/host boundary.** These verbs run host-side, but the processes
  holding agent task trees run inside agent containers where the same
  directory has a different path, so `readlink /proc/<pid>/cwd` from the host
  returns a path that shares no prefix with the host path. We therefore also
  walk `/proc/<pid>/cwd`, `/proc/<pid>/cwd/..`, … comparing `(dev, ino)`
  against the tree root — the kernel resolves those in *our* mount namespace,
  so a container-side holder is found regardless of its internal path.
- **A partial sweep is not a negative.** Reading another uid's
  `/proc/<pid>/cwd` needs ptrace-read permission. If any userspace process
  could not be inspected, the probe returns `unavailable` (⇒ treat as live,
  keep) rather than `free`. Kernel threads are excluded from that count.
  `lsof` is now invoked with `+D` (recursive); `fuser` is kept as a
  positive-only signal since it has no recursive mode.

**2. `switchroom worktree reap-report` — scheduled, report-only.** Runs both
classifiers (the claim reaper and the per-agent task-tree sweep) and prints
what they *would* reclaim: grouped per agent against a size budget
(`--budget-gb`, default 5 GB, or `SWITCHROOM_AGENT_TREE_BUDGET_GB`), ordered
**oldest-first**, which is the eviction order a budget-driven reaper would
use. Only trees that already clear every safety guard are ever marked
`over-budget`; when the guards keep an agent over budget the report says how
much stays over rather than widening eligibility.

## Why

RFC `reference/rfcs/agent-home-lifecycle.md` §2 — the deterministic task-tree
reaper and its per-agent size budget. §4 (gc coverage) has landed; §2's
periodic pass had not. The point of shipping the *report* first is that
nothing in the fleet has ever measured whether these classifiers are right,
and the first time anyone would find out is the first time something
automatic deletes a checkout. This builds a week of evidence first.

**Where the safety default lives:** it is structural, not a flag.
`src/worktree/reap-report.ts` composes only plan-only predicates (`planReaper`,
`planGc`) and imports no removal primitive — not `applyGc`, not `runReaper`,
not `removeCheckout`. `planGc` is always called with `escapeHatch: false`.
The CLI verb has no `--yes`. There is no scheduled deletion anywhere in this
PR; deleting stays the separate, explicit `worktree gc --yes` / `worktree
reap`.

## Test plan

- `tests/worktree.proc-liveness.test.ts` — the regression test spawns a real
  process with its cwd in a nested subdirectory and asserts the tree is
  classified `in-use`. **It fails against the unfixed probe**: run on
  `origin/main`'s `reaper.ts` it reports `expected 'free' to be 'in-use'`;
  it passes on this branch. Plus fd holders, the cross-namespace inode walk
  (synthetic procfs built from real symlinks), the partial-sweep →
  `unavailable` rule, kernel-thread exclusion, and a vanished tree.
- `tests/worktree.reap-report.test.ts` — oldest-first ordering, unknown-age
  entries sorted last, the budget marking exactly the oldest prefix needed,
  the budget never promoting a guard-blocked tree (`stillOverBytes` instead),
  unattributed trees labelled not dropped, and an end-to-end outcome test
  that a report pass over a would-reap claim + a would-reap task tree leaves
  every file and registry record on disk.
- `./node_modules/.bin/vitest run tests/worktree.{reaper,gc,proc-liveness,reap-report}.test.ts`
  → 111 passed, 1 skipped (the skip is the non-root EACCES accounting case;
  its decision rule is pinned unconditionally).
- `bun run lint` clean end to end (`tsc --noEmit` included).

## Risk

- **Reapability on a host where the probe cannot see every process.** Run as
  a uid that can't read the holder's `/proc/<pid>/cwd`, trees now resolve
  `unavailable` and are KEPT where `fuser` previously said `free`. That is
  the module's stated contract ("can't prove it's idle" must never license a
  force-remove) and it fails toward preservation; the report surfaces the
  count (`could not be proven idle`) so the exposure is measurable. Documented
  in `docs/operators/worktree-gc.md`.
- **Probe cost.** One `readlink` + one `readdir` per pid, short-circuiting on
  the first holder; the inode walk only runs when the path test misses.
- `TaskTreeAction` gains `newestMtimeMs` (the signal `idle` was already
  derived from) so the report can order oldest-first. Additive.

*Job spec:* `reference/jobs/give-each-agent-its-own-workspace.md` — "Orphan
cleanup / reaping: abandoned trees are reaped, not accumulated", under its
Durability clause that a dirty or live tree is never reset on the agent's
behalf. This PR is the guard half of that: it stops the classifier from
calling a live tree dead, and it measures the classifier before anything acts
on it.

### DIFF
--- a/changelog.d/4724-worktree-liveness-and-reap-report.fix.md
+++ b/changelog.d/4724-worktree-liveness-and-reap-report.fix.md
@@ -1,0 +1,27 @@
+- **worktree reaper: a process working inside a checkout counted as "free" (#4724)** —
+  The in-use probe answered with `fuser <path>` / `lsof -t <path>`, both of
+  which match the path **exactly**. A process whose cwd was a nested
+  subdirectory of a worktree — i.e. an agent actually working in the checkout —
+  was invisible to it, so the probe returned `free` for a live tree. `free` is
+  precisely what clears the reaper's third fail-safe ("an in-use probe can
+  DEFINITIVELY report the path as free"), so anything automating this would have
+  `git worktree remove --force`d live work; the same probe is the idle guard for
+  the task-tree sweep. The probe now walks procfs: a process holds the tree when
+  its **cwd is the root or any directory beneath it**, or when an open **fd**
+  resolves inside it. Because these verbs run host-side while the holders are
+  usually processes inside agent containers (same directory, different path),
+  the cwd test also walks `/proc/<pid>/cwd/..` upward comparing `(dev, ino)`,
+  which the kernel resolves in *our* namespace. A sweep that could not inspect
+  every process now reports `unavailable` (⇒ keep) rather than `free`; `lsof` is
+  invoked with `+D` and `fuser` retained as a positive-only signal.
+- **New `switchroom worktree reap-report` — report-only, deletes nothing (#4724)** —
+  Runs both classifiers on a schedule and prints what they *would* reclaim,
+  grouped per agent against a size budget (`--budget-gb`, default 5 GB, or
+  `SWITCHROOM_AGENT_TREE_BUDGET_GB`) and ordered oldest-first — the eviction
+  order a budget-driven reaper would use (RFC `agent-home-lifecycle.md` §2).
+  Only trees that already clear every safety guard are ever marked over-budget;
+  when the guards keep an agent over, the report says how much stays over rather
+  than widening eligibility. The safety default is structural, not a flag: the
+  module composes only plan-only predicates and imports no removal primitive,
+  and the verb has no `--yes`. Deleting remains the separate, explicit
+  `worktree gc --yes` / `worktree reap`.
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -208,6 +208,7 @@ switchroom worktree claim <repo> [--task <name>] [--agent <name>] [--json]
 switchroom worktree list [--json]
 switchroom worktree release <id> [--json]
 switchroom worktree reap [--dry-run] [--json]
+switchroom worktree reap-report [--json] [--append <file>] [--budget-gb <n>]
 ```
 
 - **`claim <repo>`** claims an isolated checkout for a repo (alias or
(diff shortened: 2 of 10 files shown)


mekenthompson:
Probe cost, measured rather than asserted: a full-sweep MISS (the worst case — no holder, so no short-circuit) is **16ms over 64 pids** in a container here, i.e. roughly 100–150ms per tree on a host with ~500 pids. A 53-tree `gc` sweep pays a few seconds on top of a pass that is already dominated by per-tree `git`/`gh` probes. The inode walk only runs on pids whose cwd path did not match, and terminates at the filesystem-root fixed point (~10 hops), not at `maxDepth`.

Lint is green end to end including `check-changelog-entry` now that the fragment is committed.

PR #4727, opened 2026-08-15
feat(worktree): capture a checkout to a verified bundle before deleting it

mekenthompson:
## Why

`worktree gc --reclaim-dirty` quarantines by **moving** a directory — an undo only until someone empties the quarantine. There was no primitive producing a durable, restorable artefact: `git bundle` appears nowhere in `src/` on `main`.

When a fleet host hit 85% root-disk full and 41 stale checkouts had to go, the deletion was only safe because of a bash script written by hand that morning and thrown away afterwards. This PR makes that a real, tested verb.

## What

```sh
switchroom worktree capture <dir> [--dest <dir> | --out <file>] [--delete] [--json]
```

It rescues the four things a naive `git bundle create --all` silently drops:

1. uncommitted working-tree changes
2. untracked files
3. stash entries below `stash@{0}` — see the correction below
4. unreachable commits from local amends/rebases — **5,703** in one measured fleet checkout, ~12k across four, and invisible both to `git status` (clean) and to "is HEAD pushed?" (yes)

## Three things worth reviewing

**1. The source repo is never written to.** Rescuing an unreachable commit means putting a *ref* on it (bundle tips must be nameable) and snapshotting the tree means writing blobs. Doing that in the source would mean a `capture` verb that mutates 12,000 refs into a repo the operator may well keep — fine in a script that runs seconds before `rm -rf`, not fine in a verb whose default mode does **not** delete. So every write goes into a disposable bare **staging** repo that borrows the source's object DB via `objects/info/alternates`, and `git add -A` runs with `--git-dir` at staging. Source refs, index, HEAD and work tree are provably untouched — a test byte-compares the index before and after.

(This is strictly stronger than the `GIT_INDEX_FILE` approach: that relocates the index but still writes rescue refs into the source.)

**2. `git bundle verify` is NOT a sufficient deletion gate.** Measured on git 2.47.3: a bundle truncated to its first 200 bytes, and a bundle with one flipped byte inside the pack, **both** report `is okay` and exit 0 — while a readback dies with `index-pack died` / `pack has bad object at offset 1864`. Gating deletion on `verify` alone would have deleted a checkout against an unreadable bundle. The gate here is `bundle verify` **AND** a real `git fetch 'refs/*:refs/*'` into a fresh bare repo with ref-set equality asserted. `--delete` is off by default and is unreachable unless both pass; on any failure the tree is left exactly where it was. There is a test pinning that `verify` alone would not have caught either corruption.

**3. Destination reuses #4723's bulk volume, but not `/scratch`.** Writing rescue bundles onto the disk you are trying to reclaim is self-defeating, so `scratch.volume` (default `/mnt/bulkdata`) is the right default — but the per-agent `/scratch` subdir is not: it is documented as relocated package **caches**, purgeable by design and chowned to one agent's container uid. Wrong lifetime, wrong owner for the only surviving copy of someone's work. Bundles get a sibling `switchroom/rescue/` directory. `SWITCHROOM_RESCUE_DIR` / `--dest` / `--out` override; with no volume mounted it falls back to `~/.switchroom/rescue` and the CLI says so loudly.

## Restore

```sh
# working copy, checked out on the rescued snapshot:
git clone /mnt/bulkdata/switchroom/rescue/<name>.bundle <dir>

# EVERYTHING, refs verbatim (refs/rescued/unreachable/*, refs/rescued/stash/*):
git clone --mirror /mnt/bulkdata/switchroom/rescue/<name>.bundle <dir>.git
```

A plain clone only fetches `refs/heads/*` + `refs/tags/*`, which is why the work-tree snapshot is filed as a **branch** (`rescued-wip`) and staging's HEAD points at it. A `<bundle>.json` sidecar records source path, HEAD, ref counts and timestamp.

## Two corrections to the received wisdom

- **`refs/stash` IS carried by `--all`** on modern git — verified with `git bundle list-heads`. What actually dies is the stash *reflog*: every entry below `stash@{0}`. The implementation walks `rev-list -g refs/stash` and files each entry as `refs/rescued/stash/<n>`.
- **`git fsck --unreachable` alone is not enough.** Reflogs are fsck roots, so a freshly amended commit is reported as *reachable* and skipped — and then dies anyway, because bundles carry no reflogs. Verified: plain `--unreachable` reports 0 for that commit, `--no-reflogs` reports it. `--connectivity-only` is passed too; the pack's own integrity is proved by the readback, not by fsck.

## Edge shapes covered by tests

Non-git directory (refused outright, prints the `tar` alternative — no half-handling), missing directory, repo with **no commits** (unborn HEAD), detached HEAD, bare repo (no snapshot attempted), linked git worktree (captured, and `worktree prune` run after deletion), a source branch already named `rescued-wip` (not overwritten), a destination that cannot be created, a bundle path *inside* the checkout being deleted (refused — a capture that deletes itself), and submodules (gitlinks only, warned about explicitly).

## Tests

24/24 in `tests/worktree.capture.test.ts`, all outcome assertions — the headline test clones from the bundle and reads content back out, and a control test proves a plain `--all` bundle loses the same hazards.

`npm run lint` passes clean.


mekenthompson:
Follow-up commit `ee7e995` — CI caught a real bug that a dev box hides.

`git commit-tree` refuses to run without a committer identity, and the hosts this verb is aimed at are exactly the ones least likely to have one: a CI runner, a rescue shell, a fresh container with no `~/.gitconfig`. Without an identity the working-tree snapshot died with *"Committer identity unknown"*, the failure was swallowed into `warnings[]`, and a repo whose only content is uncommitted work reported `nothing-to-capture` — i.e. hazards 1 and 2 lost silently, the exact failure mode this PR exists to prevent.

The snapshot now gets a fixed synthetic identity (`switchroom worktree capture <capture@switchroom.invalid>`), passed as env rather than `-c user.*` so an inherited `GIT_COMMITTER_*` cannot override it. The regression test scrubs every identity source from the environment the capture child inherits and asserts the uncommitted + untracked content still comes back out of a clone of the bundle; it fails against the previous commit.

25/25 local, all CI green.

PR #4739, opened 2026-08-16
memory: pin the hindsight shim's REST route contract to /openapi.json

mekenthompson:
## Summary

Implements the piece design-v2.md §2.5 named as owed and never built ("a
grep of the shipped shim finds no `/openapi.json` reference, so the pin is
not built"): the engine version pin plus doctor contract probe for the
hindsight-mcp-shim's five **synthesized** tools.

- `deactivate_directive`, `reactivate_directive`, `search_knowledge_pages`,
  `get_knowledge_page`, `get_knowledge_tree` are hand-rolled REST calls
  (`DirectiveAdmin` / `KnowledgeAdmin`), not MCP tools — the engine's own
  `/mcp` never registers them (32 backend tools, 0 knowledge tools). The
  existing `classifyToolContract` diff (`tools/list` vs
  `EXPECTED_HINDSIGHT_TOOLS`) can therefore never notice one of their REST
  routes moving or disappearing — it never looks at REST at all.
- `src/memory/hindsight-shim-contract.ts` is new: `SYNTHESIZED_TOOL_ROUTES`
  pins the exact `{path, method}` pairs each synthesized tool depends on
  (confirmed live against a running 0.9.0 engine's `/openapi.json`), and
  `ShimContractPin` fetches + memoizes the spec and answers a per-tool
  `preflight()`.
- Wired into `HindsightShim.synthesizedCall` (`src/cli/hindsight-mcp-shim.ts`):
  when `/openapi.json` PROVES a route is gone, the call is rejected loudly
  — naming the missing route — **before** the doomed REST call is even
  attempted. This extends the shim's existing anti-silent-drop guard
  (`synthesizedCall`'s unknown-arg rejection) to route drift.
- `src/cli/doctor-hindsight-shim-contract.ts` is new, wired into
  `switchroom doctor` right after the existing `hindsight contract` /
  `hindsight version` rows: a `fail` row per synthesized tool with a
  missing route, and a `hindsight shim contract: version` row comparing
  `/openapi.json`'s `info.version` against `HINDSIGHT_MIN_API_VERSION`.

### Behaviour on version/contract mismatch — explicit choice

**No hard startup crash.** The shim's foundational invariant is that
`initialize` always succeeds regardless of backend state — that is the
entire reason the shim exists (see its module header). This preflight
follows the same rule: it is lazily fetched on the first synthesized call,
never during `initialize()`, and an unreachable or malformed
`/openapi.json` degrades to "proceed" (the REST layer remains the
backstop, exactly as it already was) rather than blocking every
synthesized tool on a probe that might itself be flaky. Only a
**confirmed**-missing route — `/openapi.json` fetched successfully and the
route is provably absent — is rejected, and the rejection is scoped to
that one tool, not the whole shim. This is the "least surprising
behaviour" call the task asked me to make explicit where §2.5 was silent
on it.

A failed fetch is **not** cached as failure — the next synthesized call
retries, so a backend that was briefly unreachable doesn't wedge every
later call into permanent "unknown" for the session.

### What I could not implement as written from §2.5

§2.5 (and §5, migration step 2) describes "a doctor contract probe
**against a throwaway bank**". I implemented a static `/openapi.json`
schema-presence check instead — non-mutating, no bank created/exercised/
torn down. This matches the task's own literal wording for the doctor
requirement ("the 5 synthesized ones have their routes present") and
avoids `doctor` (documented elsewhere in this codebase as a read-only
diagnostic — see `hindsight-repair.ts`'s header: "`doctor` is a read-only
diagnostic and must stay that way") acquiring a live-bank side effect. I
could not find an elaboration of "throwaway bank" beyond that one phrase —
the design doc cites "E-01's method" but E-01 itself lives in a separate
evidence-ledger file this repo does not contain. Flagging this rather than
guessing at what a throwaway-bank exercise should create/assert/clean up.

## Test plan

- [x] `vitest run tests/hindsight-shim-contract.test.ts
  tests/hindsight-mcp-shim-contract-preflight.test.ts
  tests/doctor-hindsight-shim-contract.test.ts tests/hindsight-mcp-shim.test.ts
  tests/hindsight-directive-admin.test.ts tests/hindsight-knowledge-admin.test.ts
  tests/memory.hindsight-contract.fixture.test.ts
  tests/memory.hindsight-iserror-guard.test.ts` — 190 passed
- [x] `vitest run tests/doctor.test.ts tests/cli.doctor.hindsight-consumer.test.ts
  tests/cli.doctor.memory-health.test.ts tests/cli.doctor.memory.test.ts` —
  146 passed, 2 skipped (no regression)
- [x] `tsc --noEmit` clean
- [x] `npm run lint` — full lint suite clean, including
  `check-changelog-entry` (fragment staged) and `check-mutation-coverage`
- [x] The load-bearing guard test
  (`hindsight-mcp-shim-contract-preflight.test.ts`, "THE GUARD") drives a
  real `tools/call` for `search_knowledge_pages` against a mock REST server
  whose `/openapi.json` has the search route removed, and asserts BOTH
  `isError:true` naming the missing route AND that the REST search endpoint
  was **never hit** — proving this is a refusal at the contract layer, not
  an incidental REST failure the shim happened to also report

### DIFF
--- a/changelog.d/pin-the-hindsight-shim-s-rest-route-contract-to.feat.md
+++ b/changelog.d/pin-the-hindsight-shim-s-rest-route-contract-to.feat.md
@@ -1,0 +1,13 @@
+- **memory: pin the hindsight shim's REST route contract to /openapi.json** —
+  design-v2.md §2.5's engine version pin, previously "not built": the
+  hindsight-mcp-shim's five synthesized tools (`deactivate_directive`,
+  `reactivate_directive`, `search_knowledge_pages`, `get_knowledge_page`,
+  `get_knowledge_tree`) are REST calls that `tools/list`'s existing contract
+  check can never see move. The shim now fetches `/openapi.json` and, when a
+  route is CONFIRMED missing, rejects that tool's call loudly before ever
+  attempting the doomed REST request — consistent with the shim's existing
+  anti-silent-drop guard. An unreachable/malformed spec degrades to
+  "proceed" rather than blocking every synthesized tool on a flaky probe.
+  `switchroom doctor` gains a matching contract probe: fail rows when a
+  synthesized tool's route is missing, and a row comparing the live
+  engine's declared version against the pin.
--- a/changelog.d/pin-the-hindsight-shim-s-rest-route-contract-to.feat.md
+++ b/changelog.d/pin-the-hindsight-shim-s-rest-route-contract-to.feat.md
@@ -10,4 +10,8 @@
   "proceed" rather than blocking every synthesized tool on a flaky probe.
   `switchroom doctor` gains a matching contract probe: fail rows when a
   synthesized tool's route is missing, and a row comparing the live
-  engine's declared version against the pin.
+  engine's declared version against the pin. Adversarial review: doctor now
+  emits a `warn` row (not silence) when the engine is reachable but
+  `/openapi.json` is not, and `ShimContractPin` gained a short negative/
+  positive cache TTL so a permanently-missing (or later-restored) route no
+  longer re-pays the fetch timeout on every synthesized call forever.
--- a/src/cli/doctor-hindsight-shim-contract.ts
+++ b/src/cli/doctor-hindsight-shim-contract.ts
@@ -1,0 +1,174 @@
+/**
+ * `switchroom doctor` section for the hindsight-mcp-shim's REST contract —
+ * design-v2.md §2.5's "engine version pin ... and a doctor contract probe",
+ * called out as "not built" ("a grep of the shipped shim finds no
+ * `/openapi.json` reference"). This file is that probe.
+ *
+ * Three things a healthy shim needs; here is doctor's existing coverage of
+ * each, and what this file adds:
+ *
+ *   1. **The 32 backend MCP tools resolve.** Already covered by
+ *      `classifyToolContract` (`doctor-memory.ts`), driven off a live
+ *      `tools/list` and wired into `doctor.ts` as the "hindsight contract:
+ *      <tool>" rows. NOT duplicated here.
+ *   2. **The five SHIM-SYNTHESIZED tools' REST routes are present.** These
+ *      tools (`deactivate_directive`, `reactivate_directive`,
+ *      `search_knowledge_pages`, `get_knowledge_page`, `get_knowledge_tree`)
+ *      are REST, not MCP — `tools/list` can never see one of their routes
+ *      move. This was the gap: nothing in doctor previously read
+ *      `/openapi.json`, so a route rename/removal underneath one of these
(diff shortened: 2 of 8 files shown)


mekenthompson:
Fixed both minors from the adversarial review — thanks for the catch, both were real.

**Finding 1** (doctor-hindsight-shim-contract.ts:166-174): confirmed — by the time `doctor.ts` calls `runHindsightShimContractCheck`, its own TCP-reach + "speaking MCP" probe has already passed (both early-return the whole `checkHindsight*` pass otherwise), so a null `/openapi.json` spec at that call site can *only* mean the route-drift guard itself is unable to run, not a duplicate of a total-outage row. Changed `[]` to a `warn` row naming the actual consequence: the guard (and `ShimContractPin.preflight`'s loud call-time check) is inactive, so route drift degrades silently to "unknown, proceed."

**Finding 2** (hindsight-shim-contract.ts:186-190): confirmed — no cache on a failed `/openapi.json` fetch meant every synthesized call against an engine without it re-paid up to `OPENAPI_FETCH_TIMEOUT_MS` (3s), forever. Added an opt-in `negativeCacheMs` TTL to `ShimContractPin`, set to **30s** in production (`hindsight-mcp-shim.ts`) — bounds the tax to at most one timeout per 30s of wall-clock regardless of call volume, while still clearing a transient outage within a session (no restart needed). It's opt-in (0/unset = old always-retry behavior) specifically so `hindsight-shim-contract.test.ts:280`'s "does NOT cache a FAILED fetch" test keeps passing unmodified — negative caching is off by default and production turns it on explicitly at the call site.

Also took the related third case since it fell out of the same TTL mechanism cleanly: a *successful* fetch was memoized for the process lifetime, so a route restored by a mid-session engine upgrade stayed rejected — with the error text falsely claiming "the shim confirmed the route is gone" — until shim restart. Added `positiveCacheMs` (production: **5 min**) using the identical age-check pattern, so a restored route recovers without a restart well inside a typical session.

New tests (all proven to fail against the pre-fix code by reverting locally and re-running):
- `runHindsightShimContractCheck > THE GUARD, FINDING 1: a warn row — not silence — when /openapi.json is unreachable`
- `ShimContractPin.preflight > negativeCacheMs > skips the network fetch entirely while within the negative-cache TTL`
- `ShimContractPin.preflight > negativeCacheMs > retries once the negative-cache TTL has expired`
- `ShimContractPin.preflight > positiveCacheMs > re-fetches after the positive-cache TTL and un-rejects a route the engine restored`

All 33 shim tests + 51 mcp-shim tests + doctor suite (146 tests across doctor.test.ts / cli.doctor.memory*.test.ts / cli.doctor.hindsight-consumer.test.ts) stay green. `tsc --noEmit` clean.

PR #4760, opened 2026-08-17
memory: Memory v2 M2 — directive-triage card generator + apply-batch executor

mekenthompson:
**REVIEW-ONLY — do NOT merge.** Dispatched as a worker to implement M2 (directive triage) per `carve-M2.md` / `redteam-M2.md`, TDD-first, in an isolated worktree off `origin/main` (`11361acb4`, which already includes M1 #4758). No live agent bank was triaged by this PR.

## The three binding decisions, as implemented

1. **windows-boxes fix = path (A), create-first.** `reconcileDirectiveSuperset` (`src/memory/directive-triage-executor.ts`) creates the new superset copy FIRST, then deactivates the old copy SECOND — a create failure leaves the original directive completely untouched (never a zero-copy gap). `hindsight-directive-admin.ts`'s `buildDirectivePatchBody` field whitelist (`is_active`/`tags` only — no `content`) is **not touched**; the shim's refusal to PATCH directive content stays exactly as shipped.

   **Deviation/finding:** the carve's literal plan (create a same-named copy, then call `DirectiveAdmin.deactivate({name})`) doesn't work as written — once two directives share a name, `resolve()` deliberately throws "ambiguous" on >1 hit. Fixed by capturing the old directive's `id` via `list()` *before* the create, then PATCHing the old copy **by id**, bypassing `resolve()`'s name lookup entirely (reuses the same whitelisted `buildDirectivePatchBody`, so the content-immutability guarantee can't drift between call sites). See the module docblock and the `reconcileDirectiveSuperset` tests.

2. **Actor model = Shape α (per-agent, bank-pinned).** No central cross-bank directive admin CLI was built. `applyDirectiveTriageBatch` and `reconcileDirectiveSuperset` both operate on a single `DirectiveAdmin` instance the caller constructs with its own bank — no caller-supplied cross-bank id anywhere in the new surface.

3. **M2 leaves `rules-block`-category directives ACTIVE, code-enforced.** `applyDirectiveTriageBatch` refuses to deactivate any row categorised `rules-block`, independent of what the card generator marked (`RulesBlockDeactivationRefusedError`, or a non-throwing skip mode that still lets the rest of the batch land). `classifyDirective` (`src/memory/directive-triage.ts`) also can never itself produce `action: "retire"` for a `rules-block` row (defense in depth). Both guards are mutation-tested: temporarily disabling the executor's check made exactly the 3 rules-block-refusal tests fail with specific diffs, then restored.

## Scope delivered (per the carve's T1..Tn)

- **(a) Triage-card generator** — `src/memory/directive-triage.ts`: `classifyDirective`/`buildDirectiveTriageRows`/`renderDirectiveTriageCard`. Sorts into E-45's five categories (`rules-block`/`reflect-directive`/`disposition`/`retire`/`retain-as-memory`), defaults every row to KEEP, shows per-row deterministic signal + priority, and visually separates KEEP/staged (rendered first) from retirement candidates (rendered second) — never interleaved.
- **(b) Residue-measurement harness** — `src/memory/directive-residue.ts`: `measureDirectiveResidue` (counts `rules-block` + `reflect-directive` rows only — the categories that stay always-injected post-M3) and `writeDirectiveResidueArtifact` (per-agent markdown table, header written once). Demo run against a throwaway 24-directive fixture (not committed, not a live bank):
  ```
  { agent: "klanker-fixture-throwaway", residueDirectiveCount: 20, totalDirectiveCount: 24, residueBytes: 4699, residueTokensEstimate: 1270 }
  ```
  (via `estimateTokens` from `src/cli/debug.ts`, the fleet-standard ~3.7 chars/token estimator.) This is the number-shape M3 needs to set its rules budget — real per-agent numbers still need a live pass.
- **(c) `mental-model-curator` skill extension** — rewrote its `## Directive merge/retire pass` section into `## Directive triage pass (Memory v2 M2)`: E-45 taxonomy table, deterministic-signal-only retirement rule, the hard rules-block-never-deactivated instruction (restating decision 3 for the interactive execution path — since a skill only calls MCP tools one at a time, it can't inherit the code-level guard, so this prose IS the enforcement mechanism there), the two-section card shape, an explicit "stop for operator go-ahead, never same-turn" step, sequential (never parallel) execution, and a rewritten anti-patterns list. Added `deactivate_directive`/`reactivate_directive` to `allowed-tools` (both are fleet-pre-approved, no per-call card); deliberately did NOT add `create_directive` — the windows-boxes-class reconcile stays a rare, out-of-band script/operator action, not a skill-driven one.
- **(d) Code-enforced rules-block refusal** — covered under decision 3 above.

## Known residual gap (flagging honestly, not hiding it)

The code-level `RulesBlockDeactivationRefusedError` guard only fires when something calls `applyDirectiveTriageBatch` directly. The actual interactive path (an agent running the `mental-model-curator` skill, calling `deactivate_directive` one row at a time via MCP) does **not** route through that function — it's enforced by SKILL.md prose only for that path. I judged building a new CLI wrapper or synthesized MCP tool to close this gap out of proportion to Ken's approved scope (new mutation surface beyond what was asked), so I left it as documented prose enforcement and flag it here as a candidate follow-up rather than silently declaring the problem solved.

## Files changed

- `src/memory/directive-triage.ts` (new)
- `src/memory/directive-triage-executor.ts` (new)
- `src/memory/directive-residue.ts` (new)
- `tests/memory.directive-triage.test.ts` (new, 11 tests)
- `tests/memory.directive-triage-executor.test.ts` (new, 8 tests)
- `tests/memory.directive-residue.test.ts` (new, 4 tests)
- `skills/mental-model-curator/SKILL.md` (rewrote the directive pass section)
- `changelog.d/4900-memory-v2-m2-directive-triage.feat.md` (new)

`hindsight-directive-admin.ts` and `hindsight-mcp-shim.ts` are **untouched**.

## Tests

- 23 new tests across the 3 new files, all passing (`npx vitest run`), including anti-vacuity guards (full row-count preservation, all-retired-set genuinely-zero residue, throw-on-missing-lookup) and hermetic in-process `node:http` mock REST servers — no live bank touched.
- Existing `tests/hindsight-directive-admin.test.ts`: 17/17 still passing, unmodified.
- `tsc --noEmit`: new files typecheck clean; the 14 `nostr-tools`-module errors in `src/buzz-gateway/*`/`src/buzz-provision.ts` are pre-existing baseline noise (confirmed identical on unmodified `origin/main`, not introduced by this PR).
- Scoped lint checks run clean: `check-test-runner-coverage`, `check-hindsight-bank-hermeticity`, `check-hindsight-write-redaction`, `check-mutation-coverage`, `check-stale-tool-descriptions`. `check-changelog-entry` SKIPs locally (commonmark devDependency not installed in this worktree's symlinked `node_modules`; CI installs devDependencies before lint, per its own message).
- Mutation-tested the rules-block refusal by hand: disabling the executor's category check made exactly the 3 refusal-specific tests fail with clear assertion diffs, then restored from backup.

## Hard rules honored

No live agent bank was triaged, no live `switchroom.yaml` edited, no agent restarted. All test writes went to throwaway in-process mock servers, torn down in `afterEach`.


mekenthompson:
Review response pushed (41c889a6e). All 2 blockers + 4 majors addressed, each with a red-then-green TDD repro (verified failing against the pre-fix code via `git stash` before the fix landed, per your instruction).

**B1** — `classifyDirective` ran the `superseded-by:` tag scan before checking rules-block, so a directive with both signals misclassified as `retire`. Fixed: rules-block (marker tag, then override) now checked FIRST, unconditionally. Guarded by `tests/memory.directive-triage.test.ts` — "rules-block wins over the superseded-by tag scan" (both override and persisted-marker cases).

**B2** — the refusal only lived in `applyDirectiveTriageBatch`, which nothing outside tests calls; the live path (`mental-model-curator` skill → `hindsight-mcp-shim.ts`'s `deactivate_directive` → `DirectiveAdmin.deactivate`) bypassed it. Moved the refusal into `DirectiveAdmin` itself: `deactivate`/`deactivateById`/`deactivateByIdWithTag` all call `refuseIfRulesBlock`, which reads a persisted `RULES_BLOCK_MARKER_TAG` off the directive — independent of any caller's classification. Guarded by `tests/memory.directive-triage-executor.test.ts` — "the admin-level chokepoint refuses even when the ROW LIES" (row.category is NOT rules-block but the directive carries the marker anyway). Also restored per-retirement human confirmation in the skill: `deactivate_directive` is pre-approved fleet-wide (no per-call Telegram card), so a single batch-level prose go-ahead had silently replaced the per-retirement tap `delete_directive`'s approval card gave on main. The skill now stops after presenting each retirement candidate and requires an explicit reply before the next — never a blanket "yes" for the whole list.

**M1** — batch deactivation resolved by name, which throws "ambiguous" once a reconcile leaves two same-named directives mid-batch. Added `DirectiveAdmin.deactivateById`/`resolveById`; executor and residue harness resolve everywhere by id now, and skip `isActive:false` rows. Guarded by "survives a same-name-duplicate bank state that would make `resolve()` throw 'ambiguous'" + "never deactivates a row whose isActive is already false".

**M2** — `reconcileDirectiveSuperset`'s "reconciled on retry" docstring claim was false; a retry always created a new copy. Made idempotent: reuse an existing active same-name copy whose `content` matches before creating, and deactivate every remaining stale active by id (plural now — `deactivatedOldIds`). Guarded by the new "is idempotent/recoverable across a deactivate-step failure retry" test — the one you flagged as missing.

**M3** — `ReconcileDirectiveSupersetArgs` accepted its own `apiBaseUrl`/`bankId`, so Shape α was comment-only. Removed those fields entirely; added `DirectiveAdmin.create()` + a `bankId` getter, so `reconcileDirectiveSuperset` derives everything from the `admin` instance passed in — cross-bank reach is now a type-level impossibility, not caller discipline. (Chose to fold `create()` onto `DirectiveAdmin` rather than folding the whole reconcile function in, since the by-id-only stale-copy retirement (`deactivateByIdWithTag`) still needed to stay outside the class to avoid coupling reconcile's specific two-step ordering into the admin API.)

**M4** — no real entry point existed for the windows-boxes-class fix. Added `switchroom memory directive reconcile <agent> <name> <content...>` (`src/cli/memory-directive.ts`), wired into `src/cli/memory.ts`. Preflight/wiring covered by `tests/cli.memory.directive.test.ts` (agent-not-found, missing-content, empty-content, --json error envelope) — note: this file shells out to the built `dist/cli/switchroom.js` like the existing `memory demote` CLI tests, and this sandbox's worktree is missing the `nostr-tools` dependency needed for `npm run build` (confirmed pre-existing on the unmodified main checkout too, unrelated to this change) so I could not execute it locally; `tsc --noEmit` is clean on it and CI's full `npm run build` + test run will exercise it.

**Lows, all fixed**: residue harness now counts `isActive` rows only; the new superset copy is tagged `supersedes:<name>` for provenance symmetry; overrides/residue maps are keyed by directive `id`, not `name` (a name is not unique mid-reconcile).

**Untouched, as instructed**: the content-PATCH refusal in `hindsight-directive-admin.ts` and create-first ordering are unchanged.

Scoped verification: `tests/memory.directive-triage.test.ts` (15), `tests/memory.directive-residue.test.ts` (5), `tests/memory.directive-triage-executor.test.ts` (12, including new M1/M2/B2 tests), `tests/hindsight-directive-admin.test.ts` (17, no regressions), `tests/profile-template-skill-references.test.ts` (3) — all green (52 total). `tsc --noEmit` clean. `check-test-runner-coverage`, `check-hindsight-bank-hermeticity`, `check-hindsight-write-redaction`, `check-mutation-coverage`, `check-stale-tool-descriptions` all pass.

Not merging — leaving for another review pass.

